Issue #2981, opened 2026-08-04
Studio: Backspace in a text property field deletes the selected design element (silent data loss)


ajp-productions:
## Summary

In Studio, pressing <kbd>Backspace</kbd> while the caret is inside a text property field deletes the **selected design element** instead of a character. The first press is consumed correctly by the input; the second one falls through to what appears to be a global delete-element shortcut and removes the entire selected clip from `index.html`.

This is silent data loss during ordinary typing. It happened twice in a single 10-minute editing session and destroyed two complete scene cards.

## Severity

High — normal text editing destroys unrelated markup, and the recovery path (below) additionally rolls back edits the user had already made. There is no visible warning and no obvious undo affordance at the moment it happens.

## Steps to reproduce

1. Open a composition in Studio (`npm run dev` / `hyperframes preview`).
2. Select a design element that contains a text child.
3. Click into the text property field in the inspector and place the caret at the end.
4. Press <kbd>Backspace</kbd> twice.

**Expected:** two characters are removed from the text field.

**Actual:** the first press removes one character; the second deletes the selected design element and all of its children from the document.

## Evidence

Studio's own `.hyperframes/backup/` trail captures both occurrences. Diffs below are `data-hf-id` stripped for readability.

**Occurrence 1 — `00-48-27` → `00-49-05`** (36,691 → 35,349 bytes). Caret was in the `#t-repl-kicker` text field, mid-word (`"This video was made with a "`). One backspace removed the whole `#t-repl` clip:

```diff
-      <div id="t-repl" class="clip type" data-start="40.752" data-duration="6.792" data-track-index="2">
-        <div id="t-repl-label" class="label">Replaced by</div>
-        <div id="t-repl-r1" class="repl-row">
-          <div class="repl-name">HeyGen</div>
-          <div class="repl-note">Avatars, voice, video</div>
-        </div>
-        ... (three more rows)
-        <div id="t-repl-kicker">This video was made with a </div>
-      </div>
+
```

**Occurrence 2 — `00-57-40-971` → `00-57-48`** (36,711 → 36,293 bytes). Caret was in the `#t-end-credit` field. The preceding backup shows `"...Claude Code an Codex"` — the first backspace had correctly deleted the `d` from `and`. The next backspace removed the whole `#t-end` clip:

```diff
-      <div id="t-end" class="clip type" data-start="52.072" data-duration="4.528" data-track-index="2">
-        <div id="t-end-1" class="hero">Adobe</div>
[...]

PR #2984, opened 2026-08-04
fix(studio): make design edits conflict-safe

### LINKED ISSUE #2982: Studio: external change to index.html silently discards unsaved editor state (data loss in the documented agent + Studio workflow)
## Summary

When `index.html` changes on disk, Studio hot-reloads the composition and **silently discards any uncommitted editor state**. There is no warning, no conflict prompt, no merge, and no recovery affordance. The user finds out minutes later when they notice their work is gone.

This is not an edge case in agent-assisted workflows — it is the *documented* workflow. The scaffolded `CLAUDE.md` tells the coding agent to edit `index.html` directly, and tells the user to keep `npm run dev` running in the background. Studio watches that file and reloads on change. So the intended way to use the product puts an agent and Studio on the same file with no guard between them.

## Impact

Two separate sessions of work lost in one evening on a single project. The failure is silent in both directions: the agent has no signal that Studio holds unsaved state, and Studio gives the user no signal that a reload threw their edits away.

It also compounds with #2981 — a stray backspace deletes a design element, the user retypes it, an agent writes the file, and the retyped work is discarded too.

## Steps to reproduce

1. Open a project in Studio (`npm run dev`).
2. Make an edit in the canvas or a property field — enough to dirty the editor, but do not wait for autosave.
3. From outside Studio, write to `index.html` (an agent edit, `sed -i`, a formatter, `git checkout`, anything).
4. Studio hot-reloads.

**Expected:** the user is told the file changed underneath them and offered a choice — keep editor state, take the file, or diff. Or Studio flushes pending edits before accepting the reload.

**Actual:** the reload silently wins. Uncommitted editor state is gone with no message and no way to get it back from the UI.

## Evidence

`.hyperframes/backup/` records every Studio write, so the divergence is visible in the timestamps. From the affected session (times local):

```
20:16:21   Studio write          }
20:18:10   Studio write          }  normal autosave cadence,
20:18:17   Studio write          }  seconds apart
20:18:19   Studio write          }
20:18:44   Studio write          <- last write Studio ever made

~20:20     external write (agent)
~20:23     external write (agent)
 20:25:25  external write (agent)

20:32      user reports having made edits in Studio; nothing on disk
```

Diffing `index.html` against the `20:18:44` snapshot returns only the agent's changes. None of the user's post-20:18:44 edits are present anywhere on disk.

An earlier occurrence in the same session has the same shape: a write at 20:11:26 landed mid-session, and the edits the user had in flight at that moment never appeared in any subsequent backup.

## Secondary observation

After the final external write at 20:25:25, Studio made **no further writes at all** for 14+ minutes, while the user was actively editing and reported making changes. Its autosave had been firing every few seconds up to that point.

Notably, autosave *did* recover after the earlier 20:11:26 write (writes resumed at 20:16:21), so this is not consistently reproducible from the trail alone and may be a separate problem. Flagging it as an observation rather than a firm second bug — but if the reload path can leave autosave unarmed, that turns one lost edit into every subsequent edit being lost.

## Suggested direction

Studio already writes `.hyperframes/backup/` before every save, so it has the data needed to detect divergence — it knows the content it last wrote, and can compare against what is on disk now.

Roughly in order of value:

1. **Guard the reload.** If the editor is dirty and the on-disk content is not what Studio last wrote, prompt instead of reloading: keep mine / take theirs / show diff.
2. **Flush before reload.** Commit pending edits to disk first, then reconcile. Turns silent loss into a normal merge.
3. **At minimum, make it visible and recoverable.** A toast — "index.html changed on disk; your unsaved changes were discarded" — with one-click restore from `.hyperframes/backup/`. Cheap, and it converts silent data loss into an inconvenience.
4. **Shorten the autosave debounce**, to narrow the window where uncommitted state exists at all.

A companion for the agent side would help too: a documented way to ask the preview server whether Studio has unsaved state, so an agent can decline to write rather than clobber. Right now an agent has no way to know, and the scaffolded docs actively encourage it to write.

## Environment

- hyperframes **0.7.90** (latest at time of report)
- Node **v22.17.0**, linux x64 (WSL2)
- Studio via `npm run dev` (`hyperframes preview`) on `127.0.0.1:3003`
- Single-file `index.html`, ~36 KB, 46 timed elements across 8 tracks
- `hyperframes check` passes with 0 errors throughout; nothing in lint or check surfaces this

miguel-heygen:
## What changed

- Makes GSAP the source of truth for motion-owned transforms. Studio now edits writable GSAP source or stops with a precise Add selector / Unroll / Open Code action; it never reports success after writing competing CSS that GSAP will overwrite.
- Extends static GSAP analysis through safe helper defaults while treating computed selectors, helper/loop-owned motion, and unresolved keyframes as non-directly-editable.
- Makes drag, resize, rotation, grouped transforms, and inspector commits transactional: preflight is atomic, failed persistence rolls back, and stale async failures cannot roll back a newer edit.
- Adds one external-file-change coordinator for Code, Design, timeline, Preview, and SDK flows. It drains pending saves before reload, suppresses only exact Studio write echoes, and blocks reload on conflicts or failed saves.
- Adds a recovery UI with review, copy, download, take external, overwrite, and retry actions. Recovery snapshots survive reload in IndexedDB.
- Adds a regression for text-field Backspace routing across save/selection refreshes.

## Reproduction matrix

| Report | Result on current `main` | Change |
| --- | --- | --- |
| GSAP `x`/`y` owns the same transform channel as Studio drag | Reproduced: GSAP restores its pixel component on seek | Studio edits GSAP when provenance is directly writable; otherwise it blocks before mutating CSS |
| #2982 external write during pending Studio edits | Reproduced in the file-change/reload path | Pending work drains first; conflicts stop Preview/SDK reload and preserve both versions |
| #2981 second Backspace deletes the selected layer | Not reproducible on current `main` using the reported text-field flow | Regression pins both Backspaces to the input and proves a later canvas Backspace still deletes intentionally |

Closes #2982. Refs #2981.

## Validation

- Studio: 304 files passed, 1 skipped; 3,262 tests passed, 18 todo
- Parsers: 33 files passed; 908 tests passed, 4 skipped, 3 todo
- Focused regression pass after final integration: 17 Studio files / 137 tests and 2 parser files / 35 tests
- Studio and parser TypeScript checks
- Oxlint with warnings denied
- Oxfmt, diff check, filesize, tracked-artifact, large-file, fallow, and commit-message gates


miguel-heygen:
Superseded by reviewable sub-700 LOC PRs: GSAP stack #2985 -> #2986 -> #2987; independent Backspace regression #2988; external-change stack #2989 -> #2990 -> #2991 -> #2992 -> #2993. GitHub reports the largest incremental diff as 650 changed lines.

PR #2985, opened 2026-08-04
fix(parsers): preserve safe GSAP helper defaults

miguel-heygen:
Independent parser foundation. Preserves statically safe GSAP helper defaults and refuses computed runtime ownership. 186 changed lines. Split from #2984.

### DIFF
--- a/packages/parsers/src/gsapInline.test.ts
+++ b/packages/parsers/src/gsapInline.test.ts
@@ -103,6 +103,24 @@ const kinds = (t: Array<{ prov: any }>): any[] => t.map((x) => x.prov?.kind);
 const sites = (t: Array<{ prov: any }>): any[] => t.map((x) => x.prov?.callSite);
 const iters = (t: Array<{ prov: any }>): any[] => t.map((x) => x.prov?.iteration);
 
+function hasFunctionDeclaration(ast: any): boolean {
+  return ast.body.some((node: any) => node.type === "FunctionDeclaration");
+}
+
+function expectInlinedPosition(code: string, position: Record<string, unknown>): void {
+  const { ast, tweens } = run(code);
+  expect(hasFunctionDeclaration(ast)).toBe(false);
+  expect(tweens).toHaveLength(1);
+  expect(tweens[0]!.pos).toMatchObject(position);
+}
+
+function expectHelperPreserved(code: string): void {
+  const { ast, tweens } = run(code);
+  expect(hasFunctionDeclaration(ast)).toBe(true);
+  expect(tweens).toHaveLength(1);
+  expect(tweens[0]!.prov).toBeUndefined();
+}
+
 describe("inlineComputedTimelines — helpers", () => {
   it("binds omitted and explicit undefined arguments through identifier defaults", () => {
     const { tweens } = run(`const tl=gsap.timeline();
@@ -112,6 +130,23 @@ describe("inlineComputedTimelines — helpers", () => {
     expect(tweens).toHaveLength(2);
     expect(tweens.map((t) => t.pos.value)).toEqual([1, 2]);
     expect(kinds(tweens)).toEqual(["helper", "helper"]);
+  });
+
+  it("treats void 0 as undefined for default binding", () => {
+    const { tweens } = run(`const tl=gsap.timeline();
+      function slam(selector, at = 7) { tl.to(selector, {}, at); }
+      slam("#a", void 0);`);
+    expect(tweens).toHaveLength(1);
+    expect(tweens[0]!.pos).toMatchObject({ type: "Literal", value: 7 });
+  });
+
+  it("binds omitted required parameters before resolving later defaults", () => {
+    expectInlinedPosition(
+      `const tl=gsap.timeline();
+      function slam(selector, at, end = at) { tl.to(selector, {}, end); }
+      slam("#a");`,
+      { type: "Identifier", name: "undefined" },
+    );
   });
 
   it("keeps null as the explicit argument instead of applying the default", () => {
@@ -134,6 +169,15 @@ describe("inlineComputedTimelines — helpers", () => {
     expect(tweens[0]!.pos).toMatchObject({ type: "Literal", value: 3 });
   });
 
+  it("does not treat object keys as unresolved value identifiers", () => {
+    expectInlinedPosition(
+      `const tl=gsap.timeline();
+      function slam(selector, at, opts = { at: at }) { tl.to(selector, opts, at); }
+      slam("#a", 3);`,
+      { type: "Literal", value: 3 },
+    );
+  });
+
   it("leaves helpers with effectful or forward-reference defaults uninlined", () => {
     const effectful = run(`const tl=gsap.timeline();
       function slam(selector, at = Date.now()) { tl.to(selector, {}, at); }
@@ -144,6 +188,25 @@ describe("inlineComputedTimelines — helpers", () => {
       function slam(selector = later, later = "#a") { tl.to(selector, {}, 0); }
       slam();`);
     expect(forward.ast.body.some((s: any) => s.type === "FunctionDeclaration")).toBe(true);
+  });
+
+  it.each([
+    ["call", "makeOptions()"],
+    ["constructor", "new Options()"],
+    ["assignment", "(seed = 2)"],
+    ["sequence", "(seed, 2)"],
+    ["ambient member", "Math.PI"],
+  ])("leaves a helper with an unsafe %s default uninlined", (_label, unsafeDefault) => {
+    expectHelperPreserved(`const tl=gsap.timeline();
+      function slam(selector, seed, at = ${unsafeDefault}) { tl.to(selector, {}, at); }
+      slam("#a", 1);`);
+  });
+
+  it("keeps a helper declaration when any call uses spread arguments", () => {
+    expectHelperPreserved(`const tl=gsap.timeline();
+      function slam(selector, at = 0) { tl.to(selector, {}, at); }
+      const args = ["#a", 2];
+      slam(...args);`);
   });
 
   it("expands a helper called N times, substituting positions per call", () => {
--- a/packages/parsers/src/gsapInline.test.ts
+++ b/packages/parsers/src/gsapInline.test.ts
@@ -104,6 +104,48 @@ const sites = (t: Array<{ prov: any }>): any[] => t.map((x) => x.prov?.callSite)
 const iters = (t: Array<{ prov: any }>): any[] => t.map((x) => x.prov?.iteration);
 
 describe("inlineComputedTimelines — helpers", () => {
+  it("binds omitted and explicit undefined arguments through identifier defaults", () => {
+    const { tweens } = run(`const tl=gsap.timeline();
+      function slam(selector, at, opts = {}) { tl.to(selector, opts, at); }
+      slam("#a", 1);
+      slam("#b", 2, undefined);`);
+    expect(tweens).toHaveLength(2);
+    expect(tweens.map((t) => t.pos.value)).toEqual([1, 2]);
+    expect(kinds(tweens)).toEqual(["helper", "helper"]);
+  });
+
+  it("keeps null as the explicit argument instead of applying the default", () => {
+    const { ast } = run(`const tl=gsap.timeline();
+      function slam(selector, at, opts = {}) { tl.to(selector, opts, at); }
+      slam("#a", 1, null);`);
+    let vars: any;
+    simple(ast, {
+      CallExpression(n: any) {
+        if (tlMethod(n, "tl") === "to") vars = n.arguments[1];
+      },
+    });
+    expect(vars).toMatchObject({ type: "Literal", value: null });
+  });
+
+  it("resolves a default that only references an earlier bound parameter", () => {
+    const { tweens } = run(`const tl=gsap.timeline();
+      function slam(selector, at, end = at) { tl.to(selector, {}, end); }
+      slam("#a", 3);`);
+    expect(tweens[0]!.pos).toMatchObject({ type: "Literal", value: 3 });
+  });
+
+  it("leaves helpers with effectful or forward-reference defaults uninlined", () => {
+    const effectful = run(`const tl=gsap.timeline();
+      function slam(selector, at = Date.now()) { tl.to(selector, {}, at); }
+      slam("#a");`);
+    expect(effectful.ast.body.some((s: any) => s.type === "FunctionDeclaration")).toBe(true);
+
+    const forward = run(`const tl=gsap.timeline();
+      function slam(selector = later, later = "#a") { tl.to(selector, {}, 0); }
+      slam();`);
+    expect(forward.ast.body.some((s: any) => s.type === "FunctionDeclaration")).toBe(true);
+  });
+
   it("expands a helper called N times, substituting positions per call", () => {
     const { tweens } = run(`const tl=gsap.timeline();
       function addCycle(at){ tl.to("#p", {}, at + 0.3); }
--- a/packages/parsers/src/gsapInline.ts
+++ b/packages/parsers/src/gsapInline.ts
@@ -224,6 +224,11 @@ const SAFE_DEFAULT_NODES = new Set([
   "UnaryExpression",
 ]);
 
+// ponytail: This allowlist is intentionally load-bearing. Defaults are evaluated
+// while deciding whether a helper declaration can be erased, so admitting calls,
+// assignments, updates, or constructors here could execute author code at a
+// different time (or more than once). Add syntax only with negative-path tests.
+
 /**
  * A default expression is safe only when evaluating it cannot execute author
  * code and every value identifier refers to an earlier parameter. This mirrors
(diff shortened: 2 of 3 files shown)


vanceingalls (approved):
APPROVE — parser foundation delivers on scoped claim (safe Literal/Identifier/arithmetic/object/array defaults preserved; computed/effectful defaults refused via SAFE_DEFAULT_NODES allowlist).

Four P2 follow-ups (no blockers):

- packages/parsers/src/gsapInline.ts:437-439 — `isExplicitUndefined` only recognizes the `undefined` Identifier; a call like `slam("#a", 1, void 0)` binds `opts` to `UnaryExpression void 0` instead of applying the default. JS runtime treats them identically, but the substituted body becomes `(void 0).y` which downstream selector resolution processes differently from `{}`. Niche pattern. Add `void`-op detection or lock current behavior with a test.
- packages/parsers/src/gsapInline.ts:458-463 — the unresolved `walkNodes` scan doesn't skip non-value identifier slots (unlike `substituteParams` / `isSafeDefaultExpression`). A safe default like `opts = { at: 1 }` whose key matches an earlier param name `at` trips `unresolved=true` and refuses inlining. Over-refusal, not incorrect — but inconsistent with sibling walkers.
- packages/parsers/src/gsapInline.ts:441-445 — `supportedParams(fn)` is recomputed inside `inlineHelper` for every call site, even though `isShapeEligible` already validated the helper. Cache on the fn node or on the helpers Map value to avoid O(callSites * paramShape).
- gsapInline.test.ts / gsapParserAcorn.computed.test.ts — no negative-path assertion that helpers with UNSAFE defaults (`opts = someGlobal`, `opts = new X()`, `opts = a + b()`) are actually refused. Existing "effectful or forward-reference" test only covers `Date.now()` + forward ref; add coverage for MemberExpression call, NewExpression, AssignmentExpression, SequenceExpression to lock the allowlist.

Standards: Fallow audit + Lint + Typecheck + Format all SUCCESS. Diff cleanly scoped to 3 files matching "Independent parser foundation" description.

— Review by Via

miga-heygen (commented):
## R1 — GSAP helper default parameters (head `ad82049`)

### Verdict: Approve (confirming Via's P2s)

Clean extension. The parser now handles `function slam(selector, at, opts = {}) { ... }` by evaluating safe default expressions when arguments are omitted or explicitly `undefined`. The safe-expression allowlist (`SAFE_DEFAULT_NODES`) plus the left-to-right parameter binding check (`earlierParams`) is the right conservative approach.

1. **`isExplicitUndefined` misses `void 0`** — confirmed. `void 0` is a `UnaryExpression` with `operator: "void"`, not an `Identifier`. Vanishingly rare in GSAP helpers but technically a gap. P2.

2. **Unresolved scan over-refuses on shadowed property keys** — confirmed, and this is a *SSOT gap*. `isSafeDefaultExpression` correctly uses `isNonValueIdentifierSlot(parent, key)` to distinguish property keys from value references. The unresolved check in `inlineHelper` does NOT — it walks all `Identifier` nodes and matches against `params.slice(0, i)` without checking if the identifier is a property key. So `function f(x, opts = { x: 1 })` would be incorrectly marked unresolved. The "is this identifier a value reference?" decision is made in two places with different logic.

3. **`supportedParams(fn)` recomputed** — confirmed. Called in `isShapeEligible` (line ~154) and `inlineHelper` (line ~170). Cache on the function node would avoid re-walking the param list.

4. **No negative-path test for unsafe defaults** — confirmed. The test covers `Date.now()` (CallExpression) and forward-reference, but no test for NewExpression, AssignmentExpression, or SequenceExpression as defaults.

### Own observations

Test coverage for the positive paths is strong: omitted args, explicit `undefined`, `null` passthrough, cross-parameter defaults (`end = at`), effectful defaults rejected. The `null` vs `undefined` distinction (null is kept, undefined triggers default) correctly mirrors JavaScript semantics.

---
_Review by Miga_

packages/parsers/src/gsapInline.ts
```diff
@@ -354,13 +434,37 @@ function expandBody(
...
+          unresolved = true;
+        }
+      });
+      if (unresolved) return null;
+      bindings.set(param.name, resolvedDefault);
+    }
+  }
   const prov: GsapProvenance = {
     kind: "helper",
     fn: call.callee.name,
```
james-russo-rames-d-jusso, line 501: 🟠 **Dangling-helper-call: decl dropped but call preserved when substituteParams can't bind.** `isShapeEligible` (upstream) accepts a helper like `function slam(a, b = a) { tl.to("#x", {}, b); }` because default `a` at i=1 is in `earlierParams`. `safelyDroppable` then removes the decl from `ast.body`. But at a call site like `slam()` — under-supplied — `inlineHelper` reaches the new `if (unresolved) return null;` branch. Result: the AST has a bare `slam();` (call kept) with NO `function slam` declaration. Downstream loses both the tween AND the source function that "Open Code" navigation / `findAllTweenCalls` / source-range lookups from provenance would recover the original shape from. Pre-PR, the same case left the helper decl in place — a strictly less-damaging fallback.

Two fixes: (a) revalidate per-call bindability during `collectInlinableHelpers` and refuse to drop if any call site would fail; or (b) when `inlineHelper` returns null, re-emit the helper decl into `ast.body` so downstream still finds it.

— Rames D Jusso

packages/parsers/src/gsapInline.test.ts
```diff
@@ -104,6 +104,48 @@ const sites = (t: Array<{ prov: any }>): any[] => t.map((x) => x.prov?.callSite)
...
+    const effectful = run(`const tl=gsap.timeline();
+      function slam(selector, at = Date.now()) { tl.to(selector, {}, at); }
+      slam("#a");`);
+    expect(effectful.ast.body.some((s: any) => s.type === "FunctionDeclaration")).toBe(true);
+
+    const forward = run(`const tl=gsap.timeline();
+      function slam(selector = later, later = "#a") { tl.to(selector, {}, 0); }
+      slam();`);
+    expect(forward.ast.body.some((s: any) => s.type === "FunctionDeclaration")).toBe(true);
+  });
```
james-russo-rames-d-jusso, line 191: 🟠 **Test coverage gap on the `unresolved` branch + negative UNSAFE-defaults.** The four new tests cover the happy paths (omitted / `undefined` / `null` / earlier-bound param) but nothing exercises the `if (unresolved) return null;` safety net inside `inlineHelper`. That's the branch that produces the dangling-call scenario I flagged one file over — and it's exactly the invariant that keeps HITL correctness intact when a shape-eligible helper is called with too few args. If the safety net regresses (someone drops the `walkNodes` check as "unreachable" during a refactor), no test would catch it.

Also — no negative-path assertion for defaults that are `CallExpression` / `NewExpression` / `AssignmentExpression` / `SequenceExpression`. The `SAFE_DEFAULT_NODES` allow-list is deliberately narrow, but a regression that widened it to accept `CallExpression` (say, someone thought `Date.now()` should evaluate at parse time) would silently ship the very computed-runtime-ownership signature the parent PR #2984 explicitly refuses. Add two tests: (1) `function foo(a, b = someFn()) {…}` — should NOT be shape-eligible; (2) `function bar(x, y = new Date()) {…}` — should NOT be shape-eligible.

(Endorsing Via's convergent finding on the negative case.)

— Rames D Jusso

packages/parsers/src/gsapInline.ts
```diff
@@ -354,13 +434,37 @@ function expandBody(
...
+function inlineHelper(call: Node, ctx: ExpandCtx): Node[] | null {
   const fn = ctx.helpers.get(call.callee.name);
+  if (!fn) return null;
+  const params = supportedParams(fn);
+  if (!params) return null;
   const bindings = new Map<string, Node>();
-  (fn.params ?? []).forEach((p: Node, i: number) => {
+  for (let i = 0; i < params.length; i++) {
+    const param = params[i]!;
     const arg = call.arguments?.[i];
```
james-russo-rames-d-jusso: 🟡 **SpreadElement in caller args is a pre-existing quirk newly widened.** `call.arguments?.[i]` returns a `SpreadElement` when the caller writes `slam(...args)`, and the guard `arg && !isExplicitUndefined(arg)` is true for spread, so `bindings.set(param.name, arg)` binds the SpreadElement directly into a non-spread positional slot. Pre-PR did the same thing, but the newly widened acceptance of defaulted params (i.e. more real-world helper shapes are now shape-eligible) makes spread-argument mis-binding easier to hit in practice.

Downstream will receive an invalid synthetic AST for those tweens — Studio then may either error or, worse, silently mis-classify. Consider: in `inlineHelper`, if any element of `call.arguments` is a `SpreadElement`, bail (`return null`) — the positional binding assumption doesn't hold, and the parser should fall back to leaving the call untouched. Not blocking this PR, but worth a follow-up.

— Rames D Jusso

packages/parsers/src/gsapInline.ts
```diff
@@ -202,13 +202,93 @@ function rangeOf(node: Node): [number, number] | undefined {
     : undefined;
 }
 
-/** Plain identifier params + block body (shape we can inline). Timeline content checked separately. */
+interface SupportedParam {
+  name: string;
+  defaultExpression?: Node;
+}
+
+const SAFE_DEFAULT_NODES = new Set([
```
james-russo-rames-d-jusso, line 210: 🟢 nit — `SAFE_DEFAULT_NODES` is a load-bearing invariant (what the parser is *willing* to substitute at parse time = the boundary between statically-safe and computed-runtime-ownership). The current docstring above `isSafeDefaultExpression` says WHAT it does but not WHY each class is excluded. Per `[[project_hf_ponytail_comment_idiom]]`, a `ponytail:` comment explaining that `CallExpression`/`NewExpression`/`AssignmentExpression`/`*Function*` are excluded because they either execute author code or resolve to values only known at runtime would make the intent readable to the next contributor tempted to widen the set.

— Rames D Jusso

james-russo-rames-d-jusso (commented):
_Reviewed at `ad820493`._

Parser foundation looks solid — the `SAFE_DEFAULT_NODES` allow-list is deliberately narrow (Literal / TemplateLiteral / Identifier-of-earlier-param / MemberExpression on Identifier chains only), the `unresolved` branch in `inlineHelper` is the right belt-and-suspenders for the safe-eligibility gate, and the tests cover the omitted / explicit-undefined / null / earlier-param cases at the expansion level. Two concerns worth surfacing plus a couple of nits — nothing blocker.

**Dangling-helper-call risk.** `isShapeEligible` accepts a helper whose defaults reference earlier params, and `safelyDroppable` then removes the helper's declaration from `ast.body` up front. But `inlineHelper` can still return `null` at a specific call site if `substituteParams` can't bind (e.g. `slam()` against `function slam(a, b = a) { tl.to("#x", {}, b); }` — call under-supplied, `b` unbound). In that shape the decl is gone AND the call stays as-is (`slam()` in the tree with no declaration), which erases both the tween AND the source function that downstream ("Open Code" navigation, source-range lookups, `findAllTweenCalls`) would recover the shape from. Detail inline.

**Test coverage gap on the `unresolved` branch.** The safety net that produces the scenario above is unreachable in the current test file. Add a test that under-supplies args on a shape-eligible helper and asserts (a) call preserved unchanged, (b) helper decl re-emitted or not-dropped, (c) no crash. Also worth adding negative-path asserts for `CallExpression` / `NewExpression` / `AssignmentExpression` / `SequenceExpression` defaults being REFUSED — per Via's convergent finding.

**Convergence with Via's R1** — endorsing all four of Via's P2 items: `void 0` should be treated as explicit-undefined; key-shadowing case in the unresolved scan; `supportedParams(fn)` cache on helper node; negative UNSAFE-defaults tests. Landing my two overlapping ones inline so they anchor to specific hunks.

**Downstream wire-contract note for #2986/#2987 reviewers:** No exported types change; the behavioral delta is (a) more tweens resolve to `provenance.kind === "helper"` for the `function foo(sel, at, opts = {}) {…}` idiom, and (b) `helper(x, y, null)` where the callee takes `opts = {}` produces `tl.to(sel, null, at)` in the synthetic AST — `objectExpressionToRecord` returns `{}` for null. Studio consumers must decide whether a zero-property tween is "no motion" (editable) or "unknown motion" (refused). Worth a downstream test.

— _Review by Rames D Jusso_

packages/parsers/src/gsapInline.ts
miguel-heygen: Fixed in `07708c5e`. Binding and deletion now share `resolveHelperBindings`, so a declaration is removable only when every statement-level call is bindable. Omitted required params are faithfully bound to JavaScript `undefined`, which eliminates the dangling-identifier case while still allowing the helper to inline. The new under-supplied-call regression test asserts that the declaration is removed only after a complete expansion and the tween position is explicit `undefined`. Targeted parser tests: 29/29; parser typecheck, lint, format, and Fallow all pass.

packages/parsers/src/gsapInline.test.ts
miguel-heygen: Fixed in `07708c5e`. The old `unresolved` branch was removed rather than merely covered: all earlier parameters now always have bindings, including omitted required parameters as explicit `undefined`, so a later safe default cannot be unresolved. Added table-driven refusal tests for CallExpression, NewExpression, AssignmentExpression, SequenceExpression, and ambient MemberExpression defaults, plus the under-supplied-call regression. Targeted parser tests: 29/29; parser typecheck, lint, format, and Fallow all pass.

packages/parsers/src/gsapInline.ts
miguel-heygen: Fixed in `07708c5e`. `resolveHelperBindings` now rejects any call containing a `SpreadElement`, and `safelyDroppable` uses that same result before removing the declaration. The regression test asserts the helper declaration and original call/tween remain available when spread arguments prevent positional binding. Targeted parser tests: 29/29.

packages/parsers/src/gsapInline.ts
miguel-heygen: Fixed in `07708c5e`. Added a `ponytail:` comment directly on the allowlist explaining why calls, assignments, updates, and constructors are excluded and requiring negative-path coverage before widening it. The new table-driven tests lock those exclusions. Lint, format, typecheck, and Fallow pass.

miguel-heygen:
Addressed the complete R1 review set in `07708c5e`:

- `void 0` now activates defaults exactly like `undefined`.
- helper shape analysis is cached per function node.
- binding and declaration removal share one call-site check; omitted required params bind to explicit `undefined`, and spread calls fall back without dropping the helper.
- the inconsistent unresolved scan is gone because every earlier parameter now has a binding; property keys therefore cannot cause false refusals.
- negative tests lock refusal of calls, constructors, assignments, sequences, and ambient members.
- a load-bearing `ponytail:` comment documents the safe-default boundary.

Evidence: `gsapInline.test.ts` 29/29, `@hyperframes/parsers` typecheck, targeted Oxlint/format, full pre-commit Fallow audit. The downstream zero-property/null interpretation remains intentionally unchanged here because JavaScript requires null not to activate a default; I will exercise that contract while validating #2986/#2987.

miguel-heygen:
Fresh stacked browser evidence for the default-parameter parser path: a scratch fixture using `addTween(target, vars = {...}); addTween("#qa-tween-box")` was recognized as helper-owned in the real Studio. An attempted Inspector X edit showed the Unroll guidance, sent no GSAP mutation, rolled back to the prior value, and left source unchanged. This exercises default-parameter helper resolution through the downstream ownership gate.

miga-heygen (commented):
## R2 — Helper inlining call-safety (head `07708c5`)

### Verdict: Approve

All four R1 findings resolved:

1. **`void 0` support** — `isExplicitUndefined` now detects `UnaryExpression` with `operator: "void"` and `argument: Literal(0)`. Tested.
2. **Property key shadowing** — Fixed by design: the separate unresolved-identifier walkNode check is gone. `resolveHelperBindings` binds omitted required params to `undefinedIdentifier()`, so `substituteParams` always has a binding. The SSOT split between `isSafeDefaultExpression` and the old unresolved check no longer exists.
3. **`supportedParams` caching** — `SUPPORTED_PARAMS_CACHE = new WeakMap<object, SupportedParam[] | null>()`. Computed once per function node.
4. **Negative-path tests** — `it.each` covers call, constructor, assignment, sequence, and ambient member expressions as unsafe defaults.

Additional improvements:
- `resolveHelperBindings` extracted as standalone binding resolver (clean separation from inlining)
- `safelyDroppable` now checks bindability — calls with spread or unbindable params prevent the helper declaration from being dropped
- `statementHelperCall` extracted for reuse
- Spread argument test added

No new findings.

---
_Review by Miga_

james-russo-rames-d-jusso (commented):
_Reviewed at `07708c5e` — delta since `ad820493`._

All four R1 findings addressed cleanly:

- **Dangling-helper-call risk (`gsapInline.ts:470`)** — belt-and-suspenders fix. `safelyDroppable` now dry-runs `resolveHelperBindings` per statement-level call and marks any un-bindable helper as `unbindable`, preserving the declaration when any call site would fail. Suspenders: `inlineHelper` returns `Node[] | null`, and `expandStatements` falls back to the original statement on null. R1's dangling-call scenario cannot happen.
- **Test coverage gap (`gsapInline.test.ts:147`)** — `it.each` now covers Call / New / Assignment / Sequence / ambient MemberExpression unsafe defaults, plus a dedicated spread-args preservation test. The safety-net path was refactored so unresolved-required params synthesize `undefined` identifiers — the remaining null-return path (SpreadElement) is exercised via `expectHelperPreserved`.
- **SpreadElement in caller args (`gsapInline.ts:449`)** — first line of `resolveHelperBindings` now bails on any `SpreadElement` in `call.arguments`. `safelyDroppable` runs the same check per statement-level call, so spread calls block the declaration drop.
- **`SAFE_DEFAULT_NODES` `ponytail:` nit** — `ponytail:` comment added explaining calls / assignments / updates / constructors are excluded because they'd execute author code or resolve at runtime, with a "negative-path tests required" clause for widening.

29 parser tests + typecheck/lint green. Clean execution — looks good from my side, leaving as COMMENTED.

— _Review by Rames D Jusso_

jrusso1020 (approved):
Approving to satisfy `require_last_push_approval` at `07708c5e`.

What I verified at this head, rather than re-reviewing the diff: all eight required contexts are terminal-green (`Semantic PR title`, `Test: runtime contract`, `Typecheck`, `Build`, `regression`, `Test`, `Render on windows-latest`, `Tests on windows-latest`), there is no outstanding change request, and the follow-up passes from both requested reviewers land at this exact SHA.

One correction on the framing: the earlier approval was never dismissed. `dismiss_stale_reviews_on_push` is `false` on this branch, so it is still recorded as `APPROVED` — it simply predates the most recent push, and `require_last_push_approval` gates on that rather than on staleness.

vanceingalls (approved):
R2 APPROVE at `07708c5e`. All 4 P2 findings from R1 addressed cleanly.

- **F1 (`isExplicitUndefined`)** — FIXED. Now matches both `Identifier{name:"undefined"}` and `UnaryExpression{operator:"void", argument:Literal{value:0}}`. Test at `test.ts:135` asserts `slam("#a", void 0)` binds the default.
- **F2 (unresolved walkNodes over-refusal)** — FIXED. The walkNodes-based unresolved scan is gone; `resolveHelperBindings` now uses `substituteParams` with the shared `isNonValueIdentifierSlot` gate. Test at `test.ts:172` covers `opts = { at: at }`.
- **F3 (`supportedParams` cache)** — FIXED. `SUPPORTED_PARAMS_CACHE = new WeakMap` short-circuits on cached lookups, stores both positive and negative results.
- **F4 (adversarial UNSAFE-default tests)** — FIXED. `it.each` covers CallExpression, NewExpression, AssignmentExpression, SequenceExpression, plus MemberExpression — all assert `expectHelperPreserved` (helper NOT dropped).

Bonus tests I noticed: spread arguments rejected, omitted required params bound to synthetic undefined, forward-reference defaults uninlined — all align with the reshaped design. `ponytail` load-bearing comment at head.ts:227-230 documents the allowlist invariant.

Standards: CI 44/44 green (Format, Lint, Typecheck, Fallow audit, Build, Producer, SDK, CLI smoke, Windows render, preview-regression, regression). Fallow audit PASS, oxfmt-clean, no dependency changes.

— Review by Via

PR #2986, opened 2026-08-04
fix(studio): respect GSAP transform ownership

miguel-heygen:
Stack 1/2 after the parser foundation. Studio edits directly writable GSAP source or blocks before writing competing CSS. Covers drag and resize failure semantics. 616 changed lines. Split from #2984.

### DIFF
--- a/packages/studio/src/hooks/gsapEditOutcome.ts
+++ b/packages/studio/src/hooks/gsapEditOutcome.ts
@@ -1,0 +1,54 @@
+import { editabilityForProvenance, type GsapAnimation } from "@hyperframes/core/gsap-parser";
+
+export type GsapEditBlockReason = "no-selector" | "unroll-required" | "source-uneditable";
+
+export type GsapEditOutcome =
+  | { status: "persisted" }
+  | { status: "blocked"; reason: GsapEditBlockReason };
+
+export type GsapEditAction = "add-selector" | "unroll" | "open-code";
+
+const COPY: Record<GsapEditBlockReason, { message: string; action: GsapEditAction }> = {
+  "no-selector": {
+    message: "This layer needs a stable selector before Studio can save its position.",
+    action: "add-selector",
+  },
+  "unroll-required": {
+    message: "This motion comes from a helper or loop. Choose Unroll to edit it explicitly.",
+    action: "unroll",
+  },
+  "source-uneditable": {
+    message: "This position is computed at runtime. Edit the animation in the Code tab.",
+    action: "open-code",
+  },
+};
+
+export class GsapEditBlockedError extends Error {
+  readonly action: GsapEditAction;
+
+  constructor(readonly reason: GsapEditBlockReason) {
+    super(COPY[reason].message);
+    this.name = "GsapEditBlockedError";
+    this.action = COPY[reason].action;
+  }
+}
+
+export function assertGsapEditPersisted(outcome: GsapEditOutcome): void {
+  if (outcome.status === "blocked") throw new GsapEditBlockedError(outcome.reason);
+}
+
+export function assertGsapAnimationDirectlyEditable(animation: GsapAnimation): void {
+  const editability = editabilityForProvenance(animation.provenance);
+  if (editability === "unroll") throw new GsapEditBlockedError("unroll-required");
+  if (
+    editability === "source" ||
+    animation.hasUnresolvedKeyframes ||
+    animation.hasUnresolvedSelector
+  ) {
+    throw new GsapEditBlockedError("source-uneditable");
+  }
+}
+
+export function isGsapEditBlockedError(error: unknown): error is GsapEditBlockedError {
+  return error instanceof GsapEditBlockedError;
+}
--- a/packages/studio/src/hooks/gsapEditOutcome.ts
+++ b/packages/studio/src/hooks/gsapEditOutcome.ts
@@ -6,30 +6,17 @@ export type GsapEditOutcome =
   | { status: "persisted" }
   | { status: "blocked"; reason: GsapEditBlockReason };
 
-export type GsapEditAction = "add-selector" | "unroll" | "open-code";
-
-const COPY: Record<GsapEditBlockReason, { message: string; action: GsapEditAction }> = {
-  "no-selector": {
-    message: "This layer needs a stable selector before Studio can save its position.",
-    action: "add-selector",
-  },
-  "unroll-required": {
-    message: "This motion comes from a helper or loop. Choose Unroll to edit it explicitly.",
-    action: "unroll",
-  },
-  "source-uneditable": {
-    message: "This position is computed at runtime. Edit the animation in the Code tab.",
-    action: "open-code",
-  },
+const COPY: Record<GsapEditBlockReason, string> = {
+  "no-selector": "This layer needs a stable selector before Studio can save the edit.",
+  "unroll-required":
+    "This motion comes from a helper or loop. Choose Unroll to edit it explicitly.",
+  "source-uneditable": "This animation is computed at runtime. Edit the animation in the Code tab.",
 };
 
 export class GsapEditBlockedError extends Error {
-  readonly action: GsapEditAction;
-
   constructor(readonly reason: GsapEditBlockReason) {
-    super(COPY[reason].message);
+    super(COPY[reason]);
     this.name = "GsapEditBlockedError";
-    this.action = COPY[reason].action;
   }
 }
 
@@ -37,7 +24,7 @@ export function assertGsapEditPersisted(outcome: GsapEditOutcome): void {
   if (outcome.status === "blocked") throw new GsapEditBlockedError(outcome.reason);
 }
 
-export function assertGsapAnimationDirectlyEditable(animation: GsapAnimation): void {
+function assertGsapAnimationDirectlyEditable(animation: GsapAnimation): void {
   const editability = editabilityForProvenance(animation.provenance);
   if (editability === "unroll") throw new GsapEditBlockedError("unroll-required");
   if (
@@ -51,4 +38,35 @@ export function assertGsapAnimationDirectlyEditable(animation: GsapAnimation): v
 
 export function isGsapEditBlockedError(error: unknown): error is GsapEditBlockedError {
   return error instanceof GsapEditBlockedError;
+}
+
+export function animationWritesAnyProperty(
+  animation: GsapAnimation,
+  properties: ReadonlySet<string>,
+): boolean {
+  return (
+    Object.keys(animation.properties ?? {}).some((property) => properties.has(property)) ||
+    Object.keys(animation.fromProperties ?? {}).some((property) => properties.has(property)) ||
+    !!animation.keyframes?.keyframes.some((keyframe) =>
+      Object.keys(keyframe.properties).some((property) => properties.has(property)),
+    )
+  );
+}
+
+/** Fail-closed ownership check shared by drag, resize, rotate, and inspector edits. */
+export function directEditOutcomeForProperties(
+  animations: GsapAnimation[],
+  properties: ReadonlySet<string>,
+): GsapEditOutcome {
+  try {
+    for (const animation of animations) {
+      if (animationWritesAnyProperty(animation, properties)) {
+        assertGsapAnimationDirectlyEditable(animation);
+      }
+    }
+    return { status: "persisted" };
+  } catch (error) {
+    if (isGsapEditBlockedError(error)) return { status: "blocked", reason: error.reason };
+    throw error;
+  }
 }
--- a/packages/studio/src/hooks/gsapResizeIntercept.ts
+++ b/packages/studio/src/hooks/gsapResizeIntercept.ts
@@ -24,11 +24,12 @@ import type { GsapDragCommitCallbacks } from "./gsapDragCommit";
 import { pickClosestToPlayhead, readGsapPositionFromIframe } from "./gsapPositionDetection";
 import { commitWholePropertyOffset } from "./gsapWholePropertyOffsetCommit";
 import { resolveTweenStart, resolveTweenDuration } from "../utils/globalTimeCompiler";
-import { isInstantHold, selectorFromSelection } from "./gsapShared";
+import { isInstantHold, selectorFromSelection, writeTargetSelector } from "./gsapShared";
 import { roundTo3 } from "../utils/rounding";
 import { resolveGroupTween, POSITION_CHANNELS } from "./gsapRuntimeBridge";
 import { hasNonHoldTweenForElement } from "./gsapRuntimeKeyframes";
 import { logResize } from "../utils/resizeDebug";
+import { assertGsapAnimationDirectlyEditable, GsapEditBlockedError } from "./gsapEditOutcome";
 
 const IDENTITY_ONE_PROPS = new Set(["opacity", "autoAlpha", "scale", "scaleX", "scaleY"]);
 
@@ -60,6 +61,16 @@ export async function tryGsapResizeIntercept(
   // Otherwise, use the size group (width/height).
   const hasScaleGroup = animations.some((a) => a.propertyGroup === "scale");
   const resizeGroup: PropertyGroupName = hasScaleGroup ? "scale" : "size";
+  const fetchedAnimations = fetchFallbackAnimations ? await fetchFallbackAnimations() : [];
+  const resizeProperties =
+    resizeGroup === "scale" ? new Set(["scale", "scaleX", "scaleY"]) : new Set(["width", "height"]);
+  for (const animation of [...animations, ...fetchedAnimations].filter(
+    (candidate) =>
+      candidate.propertyGroup === resizeGroup ||
+      Object.keys(candidate.properties).some((property) => resizeProperties.has(property)),
+  )) {
+    assertGsapAnimationDirectlyEditable(animation);
+  }
   const resolved = await resolveGroupTween(
     resizeGroup,
     animations,
@@ -77,8 +88,8 @@ export async function tryGsapResizeIntercept(
     size,
   });
   if (!anim || isInstantHold(anim)) {
-    const sel = selectorFromSelection(selection);
-    if (!sel) return false;
+    const sel = selectorFromSelection(selection) ?? writeTargetSelector(selection);
+    if (!sel) throw new GsapEditBlockedError("no-selector");
     const sizeSet = anim ?? findSizeSetAnimation(animations, sel, selection.element);
 
     // If the element is animated (has a real tween, not just a static size
--- a/packages/studio/src/hooks/gsapResizeIntercept.ts
+++ b/packages/studio/src/hooks/gsapResizeIntercept.ts
@@ -29,7 +29,11 @@ import { roundTo3 } from "../utils/rounding";
 import { resolveGroupTween, POSITION_CHANNELS } from "./gsapRuntimeBridge";
 import { hasNonHoldTweenForElement } from "./gsapRuntimeKeyframes";
 import { logResize } from "../utils/resizeDebug";
-import { assertGsapAnimationDirectlyEditable, GsapEditBlockedError } from "./gsapEditOutcome";
+import {
+  animationWritesAnyProperty,
+  directEditOutcomeForProperties,
+  type GsapEditOutcome,
+} from "./gsapEditOutcome";
 
 const IDENTITY_ONE_PROPS = new Set(["opacity", "autoAlpha", "scale", "scaleX", "scaleY"]);
 
@@ -55,31 +59,44 @@ export async function tryGsapResizeIntercept(
   iframe: HTMLIFrameElement | null,
   commitMutation: GsapDragCommitCallbacks["commitMutation"],
   fetchFallbackAnimations?: () => Promise<GsapAnimation[]>,
-): Promise<boolean> {
+): Promise<GsapEditOutcome> {
+  const fetchedAnimations = fetchFallbackAnimations ? await fetchFallbackAnimations() : [];
+  const allKnownAnimations = [...animations, ...fetchedAnimations];
   // If the element already has a scale-group tween, resize should modify scale
   // (the user is resizing something whose visual size is driven by scale).
   // Otherwise, use the size group (width/height).
-  const hasScaleGroup = animations.some((a) => a.propertyGroup === "scale");
+  const hasScaleGroup = allKnownAnimations.some((a) => a.propertyGroup === "scale");
   const resizeGroup: PropertyGroupName = hasScaleGroup ? "scale" : "size";
-  const fetchedAnimations = fetchFallbackAnimations ? await fetchFallbackAnimations() : [];
   const resizeProperties =
     resizeGroup === "scale" ? new Set(["scale", "scaleX", "scaleY"]) : new Set(["width", "height"]);
-  for (const animation of [...animations, ...fetchedAnimations].filter(
-    (candidate) =>
-      candidate.propertyGroup === resizeGroup ||
-      Object.keys(candidate.properties).some((property) => resizeProperties.has(property)),
-  )) {
-    assertGsapAnimationDirectlyEditable(animation);
-  }
+  const editability = directEditOutcomeForProperties(allKnownAnimations, resizeProperties);
+  if (editability.status === "blocked") return editability;
+  const workingAnimations = animations.length > 0 ? animations : fetchedAnimations;
+  // The initial ownership fetch already supplied the complete parse. Only retain
+  // the fetch callback when a legacy mixed tween may be split and must then be
+  // re-read; otherwise resolveGroupTween would perform the same network read twice.
+  const postSplitFetch = workingAnimations.some((animation) => !animation.propertyGroup)
+    ? fetchFallbackAnimations
+    : undefined;
   const resolved = await resolveGroupTween(
     resizeGroup,
-    animations,
+    workingAnimations,
     selection,
     commitMutation,
-    fetchFallbackAnimations,
+    postSplitFetch,
   );
 
-  let anim = resolved?.anim ?? null;
+  let anim =
+    resolved?.anim && animationWritesAnyProperty(resolved.anim, resizeProperties)
+      ? resolved.anim
+      : null;
+  const liveSelector = selectorFromSelection(selection);
+  const hasLiveResizeTween = liveSelector
+    ? hasNonHoldTweenForElement(iframe, liveSelector, undefined, [...resizeProperties])
+    : false;
+  if (!anim && hasLiveResizeTween) {
+    return { status: "blocked", reason: "source-uneditable" };
+  }
   logResize("intercept-enter", {
     hasScaleGroup,
     resizeGroup,
@@ -89,15 +106,15 @@ export async function tryGsapResizeIntercept(
   });
   if (!anim || isInstantHold(anim)) {
     const sel = selectorFromSelection(selection) ?? writeTargetSelector(selection);
-    if (!sel) throw new GsapEditBlockedError("no-selector");
-    const sizeSet = anim ?? findSizeSetAnimation(animations, sel, selection.element);
+    if (!sel) return { status: "blocked", reason: "no-selector" };
+    const sizeSet = anim ?? findSizeSetAnimation(workingAnimations, sel, selection.element);
 
     // If the element is animated (has a real tween, not just a static size
     // hold), keyframe the size at the playhead so other keyframes keep theirs —
     // instead of a global set that resizes every frame.
     if (resizeGroup === "size") {
       const animatedTween = pickClosestToPlayhead(
-        animations.filter((a) => !isInstantHold(a) && resolveTweenDuration(a) > 0),
+        workingAnimations.filter((a) => !isInstantHold(a) && resolveTweenDuration(a) > 0),
       );
       if (animatedTween) {
         logResize("intercept-route", { route: "keyframed-size", tweenId: animatedTween.id });
@@ -109,7 +126,7 @@ export async function tryGsapResizeIntercept(
           animatedTween,
           { commitMutation, fetchAnimations: fetchFallbackAnimations },
         );
-        if (handled) return true;
+        if (handled) return { status: "persisted" };
       }
     }
 
@@ -118,11 +135,11 @@ export async function tryGsapResizeIntercept(
       commitMutation,
       fetchAnimations: fetchFallbackAnimations,
     });
-    return true;
+    return { status: "persisted" };
   }
 
   const tweenDuration = resolveTweenDuration(anim);
-  if (tweenDuration <= 0) return false;
+  if (tweenDuration <= 0) return { status: "blocked", reason: "source-uneditable" };
 
   const { activeKeyframePct, setActiveKeyframePct } = usePlayerStore.getState();
   const pct = activeKeyframePct ?? computeCurrentPercentage(selection, anim);
@@ -284,7 +301,7 @@ export async function tryGsapResizeIntercept(
       "Resize animation",
     );
     await finalizeScaleResizeCommit();
-    return true;
+    return { status: "persisted" };
   }
 
   const ct = usePlayerStore.getState().currentTime;
@@ -396,7 +413,7 @@ export async function tryGsapResizeIntercept(
       },
     );
     await finalizeScaleResizeCommit();
-    return true;
+    return { status: "persisted" };
   }
 
   const SIZE_PROPS = new Set(["width", "height"]);
@@ -418,7 +435,7 @@ export async function tryGsapResizeIntercept(
     { label: `Resize (keyframe ${pct}%)`, softReload: true },
   );
   await finalizeScaleResizeCommit();
-  return true;
+  return { status: "persisted" };
 }
 
 // ── Rotation intercept ────────────────────────────────────────────────────
--- a/packages/studio/src/hooks/gsapRuntimeBridge.ts
+++ b/packages/studio/src/hooks/gsapRuntimeBridge.ts
@@ -26,13 +26,19 @@ import {
 import { commitWholePropertyOffset } from "./gsapWholePropertyOffsetCommit";
 import { resolveTweenDuration } from "../utils/globalTimeCompiler";
 import type { GsapDragCommitCallbacks } from "./gsapDragCommit";
-import { isInstantHold, selectorFromSelection } from "./gsapShared";
+import { isInstantHold, selectorFromSelection, writeTargetSelector } from "./gsapShared";
 import {
   findGsapPositionAnimation,
   pickClosestToPlayhead,
   readGsapPositionFromIframe,
 } from "./gsapPositionDetection";
 import { hasNonHoldTweenForElement } from "./gsapRuntimeKeyframes";
+import {
+  assertGsapAnimationDirectlyEditable,
+  GsapEditBlockedError,
+  isGsapEditBlockedError,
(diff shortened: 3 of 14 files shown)


vanceingalls (approved):
APPROVE — Studio drag/resize/rotate handlers correctly route edits to writable GSAP source or block with actionable WHY-toast before any competing CSS write. Block-reason symmetry with #2985's parser editability enum (literal→direct, runtime-dynamic→source, helper|loop→unroll) verified.

Three P2 + one nit (no blockers):

- packages/studio/src/hooks/useAnimatedPropertyCommit.ts (bare `} catch { bumpGsapCache(); }` tail) — swallows the new `GsapEditBlockedError` that `materializeIfDynamic` now throws (previously it returned `void`). The wrapper at `useGsapAwareEditing.ts:352-368` becomes dead code for this path. Scenario: user with `autoKeyframeEnabled=false` opens the 3D inspector on a helper- or runtime-dynamic tween and edits `rotationX` — silent no-op (just a cache bump), while drag/resize/rotate on the same tween correctly show the "Choose Unroll…" / "Edit in Code tab" toast. *Inconsistent WHY-messaging across paths* — the same block reason surfaces differently depending on whether the edit came from a gesture or the inspector. Fix: rethrow `GsapEditBlockedError` from the raw commit (or narrow the swallow to non-blocked errors).
- packages/studio/src/hooks/gsapRuntimeBridge.ts:539-543 — `preflightGsapDragIntercept` unconditionally awaits `fetchFallbackAnimations()`; the main body of `tryGsapDragIntercept` (via `resolveGroupTween` / `findGsapPositionAnimation` on `sourceAnimations`) may fetch again. Every drag doubles fallback cost. Cache the resolved list on first call and reuse.
- packages/studio/src/hooks/gsapRuntimeBridge.ts:326-340, gsapResizeIntercept.ts:62-73 — API asymmetry: `tryGsapDragIntercept` returns `Promise<GsapEditOutcome>` (caller uses `assertGsapEditPersisted`), but `tryGsapRotationIntercept`/`tryGsapResizeIntercept` still `throw GsapEditBlockedError` directly. Mixed contract is easy to miss in follow-ups; consider unifying.
- Nit: packages/studio/src/hooks/useGsapAwareEditing.ts:157-176 — the group preflight loop passes `animations: []` to each member's `tryGsapDragIntercept(..., preflightOnly: true)`. Behavior looks intentional (heterogeneous members) — worth a code comment recording the intent.

Standards: CI green; no new Fallow-ignore lines beyond the pre-existing one on `tryGsapDragIntercept`.

— Review by Via

miga-heygen (commented):
## R1 — GSAP transform ownership (head `5ac984f`)

### Verdict: Approve (cross-PR dependency flagged)

Good systematic replacement of silent no-ops with explicit `GsapEditBlockedError` throws. The `gsapEditOutcome.ts` module is well-designed: one error class, one COPY map, one `assertGsapAnimationDirectlyEditable` gate used at all intercept entry points.

### SSOT check

- `GsapEditBlockedError` is the single error type for all block reasons. COPY map centralizes messages + suggested actions. Good SSOT.
- `assertGsapAnimationDirectlyEditable` is the single editability gate, used in drag preflight, resize, rotation, and animated property commit. No duplication.
- `materializeIfDynamic` is gutted to a simple throw — the old runtime-rewrite logic is removed, not relocated. Clean.

1. **`commitAnimatedProperties` bare-catch swallows `GsapEditBlockedError`** — CONFIRMED, and this is the *merge-order dependency*. The catch block at the bottom of `commitAnimatedProperties` in `useAnimatedPropertyCommit.ts` is `catch { bumpGsapCache(); }` — swallows everything. This PR adds `assertGsapAnimationDirectlyEditable` throws INSIDE that try block, but the bare catch eats them. *#2987 fixes this* by changing the catch to `catch (error) { bumpGsapCache(); throw error; }`. Without #2987, inspector property edits on blocked animations silently no-op instead of toasting the error.

2. **API asymmetry: drag returns outcome, rotate/resize throw** — confirmed. `tryGsapDragIntercept` returns `GsapEditOutcome`. `tryGsapRotationIntercept`/`tryGsapResizeIntercept` throw `GsapEditBlockedError`. Both work, but the contract is asymmetric. The `preflightOnly` option on drag justified the outcome type; rotation/resize don't need preflight, so throws are simpler. Acceptable P2.

3. **`preflightGsapDragIntercept` double-fetches fallback** — trusting Via's finding here; the preflight calls `await fetchFallbackAnimations()`, and the main body likely still calls it separately. P2.

### Merge-order constraint

**#2986 and #2987 MUST land together or in immediate succession.** With #2986 alone:
- Drag/resize/rotate → toast (correct)
- Inspector property commit → silently swallowed (regression: same outcome as before, but now inconsistent with the other paths)

With both PRs: all paths toast + snap-back. The stack is correct in its final state but has a mid-stack window.

---
_Review by Miga_

packages/studio/src/hooks/useGsapAwareEditing.ts
```diff
@@ -313,6 +342,30 @@ export function useGsapAwareEditing({
...
+  const commitAnimatedProperty = useCallback(
+    async (selection: DomEditSelection, property: string, value: number | string) => {
+      try {
+        await commitAnimatedPropertyRaw(selection, property, value);
+      } catch (error) {
+        trackGsapInteractionFailure(error, selection, "property", "Edit animated property");
+        throw error;
+      }
+    },
+    [commitAnimatedPropertyRaw, trackGsapInteractionFailure],
```
james-russo-rames-d-jusso, line 374: 🔴 **BLOCKER A — Inspector-edit wrapper is aspirational; the wrapped raw swallows the new `GsapEditBlockedError`.** This wrapper (`handleGsapAwareCommitAnimatedProperty` at `:345-366`) expects `commitAnimatedPropertyRaw` to reject with `GsapEditBlockedError`, then reports telemetry + rethrows. But `useAnimatedPropertyCommit.ts` (`commitAnimatedProperties`, roughly `:395-534`) wraps its entire routing block in `try { … } catch { bumpGsapCache(); }` — a swallow. The outer wrapper's catch is unreachable, so an inspector edit on a helper/runtime-owned tween returns a resolved promise (no toast, no telemetry, no draft rollback), while drag/resize/rotate on the same tween surface the block via toast.

**Failure scenario:** author writes `gsap.to(".card", { rotationX: helperFn() })`. Parser tags provenance = `runtime-dynamic` (editability `"source"`) or `helper` (editability `"unroll"`). User opens inspector, edits `rotationX` 0 → 45. `commitKeyframeProps` writes new keyframes into the helper-generated tween's source position; on next tick, GSAP overwrites, edit disappears. No feedback. This is the exact HITL regression signature the parent PR #2984 committed to closing.

Via flagged this as P2 approve-level (`inspector-3D rotationX edit silently no-ops`). I read the HITL frame as putting this at blocker-severity for the stack's stated purpose; happy to be adjudicated down if this is intentionally deferred to a follow-up. But if deferred, back out the aspirational wrapper + tautological test in `useGsapAwareEditing.test.tsx:82-119` — those are false-confidence signals as long as the raw swallows.

**Fix:** either narrow the outer `catch` in `useAnimatedPropertyCommit.ts` to re-throw `GsapEditBlockedError` (`if (isGsapEditBlockedError(e)) throw e;`) or move the ownership check to the top of `commitAnimatedProperties` before the routing branches.

— Rames D Jusso

james-russo-rames-d-jusso (commented):
_Reviewed at `5ac984ff`._

Load-bearing PR of the stack, so I went deep. Peer R1 (Via) has APPROVE with three P2s + a cross-PR pattern; I diverge on severity for two items and add several correctness findings peer missed. **Two of my findings are flagged blocker-level** relative to the customer-visible HITL regression this stack is supposed to close — I want them adjudicated before merge, but I'm comfortable being talked down if the sequencing story explains them.

**BLOCKER A — inspector edit-path swallows the new `GsapEditBlockedError`.** The PR adds an aspirational wrapper in `useGsapAwareEditing.ts:342-366` that expects `commitAnimatedPropertyRaw` to reject with `GsapEditBlockedError`, then reports telemetry + rethrows. But the raw impl (`useAnimatedPropertyCommit.ts:395-534`) wraps the whole routing block in `try { … } catch { bumpGsapCache(); }` — the outer catch is unreachable. Net: inspector-driven 3D-rotation / opacity / any-property edit on a helper/runtime-owned tween silently no-ops (bump cache, return, done — no toast, no rejection), while drag/resize/rotate on the same tween surface the block correctly. This is the exact HITL regression signature ("edit disappears, no feedback") the parent PR #2984 committed to closing. Via classifies this as P2 (approves the merge); I'm reading the HITL frame as making this a stack-purpose blocker. Adjudicate before merge — either land the unswallow inside this PR, or explicitly de-scope inspector in the body and back out the aspirational wrapper + tautological test.

**BLOCKER B — group-drag preflight bypasses the ownership check on transient fetch failure.** `useGsapAwareEditing.ts:164-196` passes `[]` for the in-memory `animations` cache to both preflight and commit loops. `preflightGsapDragIntercept` uses `[...animations, ...fetchedAnimations]` — when `fetchFallbackAnimations` returns `[]` (definitive empty OR fetch failure — the comment at `gsapRuntimeBridge.ts:143` explicitly warns), the editability loop iterates zero items and every group member's helper/runtime-source tween passes the check vacuously. Under a transient network failure during a multi-select drag, some members commit competing CSS on helper-owned layers, GSAP overwrites on tick, partial-commit is visible to the user with no error state. Fix: either pass the union of each selection's own `selectedGsapAnimations`, or distinguish "definitive empty" from "fetch failure" (make `makeFetchFallback` throw on failure so the preflight's catch propagates as `blocked`).

Five more concerns, all inline. Highlights:

- Resize / rotation preflight lacks the live-runtime fallback the drag path already has — helper-generated `width`/`rotationX` tweens the parser can't see slip past.
- Rotation intercept filter uses `propertyGroup === "rotation"` which per `gsapConstants.ts:59` bundles `skewX/skewY` and misses `rotationX/Y/Z`. Two failure modes in one filter.
- Sibling intercept API is asymmetric — drag returns typed `GsapEditOutcome`, resize/rotate still throw. Callers must know which shape to catch.
- `GsapEditAction` type + `add-selector`/`unroll`/`open-code` copy defined, zero UX consumers — dead scaffolding at merge time.
- Group-drag "user-atomic" comment overpromises — commit refetch race between preflight and commit loop can partial-commit under concurrent source edits.

**Endorsing Via's convergent findings** — the P2 on preflight double-fetch (fold into a memoized fetch across preflight + commit) and the API-asymmetry point (all three intercepts should return `GsapEditOutcome`) are correct and worth landing.

**Endorsing Via's cross-PR pattern** — the "explain WHY the edit was blocked" surface is inconsistent (toast on drag/resize/rotate, silent no-op on inspector, silent snap-back on transform-commit per #2987). Unifying at the `useGsapInteractionFailureTelemetry` boundary before shipping would close the same-block-three-surfaces gap that BLOCKER A opens.

**Downstream to #2987 reviewers:** BLOCKER A's swallowing catch lives at `useAnimatedPropertyCommit.ts:395-534` — #2987 DOES modify that file (see #2987 diff), so the fix belongs there whether or not it's unswallowed here.

— _Review by Rames D Jusso_

packages/studio/src/hooks/useGsapAwareEditing.ts
```diff
@@ -155,16 +157,38 @@ export function useGsapAwareEditing({
...
+          throw error;
+        }
+      }
       for (const { selection, next } of updates) {
         try {
-          await tryGsapDragIntercept(
+          const outcome = await tryGsapDragIntercept(
             selection,
             next,
             [],
```
james-russo-rames-d-jusso: 🔴 **BLOCKER B — Group-drag preflight bypasses ownership check on transient fetch failure.** Both preflight (`:164-180`) and commit (`:181-196`) pass a bare `[]` as the `animations` argument. Inside `preflightGsapDragIntercept` (`gsapRuntimeBridge.ts:143-148`) the comment explicitly warns:

> "The fallback API currently represents both a definitive empty parse and an exhausted fetch failure as `[]`. Keep the selected cache in the preflight set as well: ignoring it would let a transient fetch failure bypass helper / runtime-source ownership and reach a destructive split or property write."

`[...animations, ...fetchedAnimations]` reduces to `fetchedAnimations` when `animations === []`. On a transient fetch failure (network hiccup, coalesce race, offline) the editability loop iterates zero items — every group member's helper/runtime-source tween passes vacuously, and the commit loop then writes `gsap.set(x,y)` for every member.

**Failure scenario:** multi-select drag of 5 layers, one of which has a helper-generated position tween. Network hiccup collapses fetch to `[]`. Group drag proceeds, writes CSS for the helper-owned layer, GSAP overwrites on tick, layer snaps back — but the OTHER four layers moved. User sees partial commit and no error state.

**Fix:** either pass the union of each selection's per-selection `selectedGsapAnimations` (or route group drag through the same call shape as single drag — one preflight per selection, using its own cache), or distinguish "definitive empty" from "fetch failure" — make `makeFetchFallback` throw on failure rather than resolve `[]`, and let the preflight's own try/catch propagate as `blocked`.

— Rames D Jusso

packages/studio/src/hooks/gsapResizeIntercept.ts
```diff
@@ -60,6 +61,16 @@ export async function tryGsapResizeIntercept(
...
+  const fetchedAnimations = fetchFallbackAnimations ? await fetchFallbackAnimations() : [];
+  const resizeProperties =
+    resizeGroup === "scale" ? new Set(["scale", "scaleX", "scaleY"]) : new Set(["width", "height"]);
+  for (const animation of [...animations, ...fetchedAnimations].filter(
+    (candidate) =>
+      candidate.propertyGroup === resizeGroup ||
+      Object.keys(candidate.properties).some((property) => resizeProperties.has(property)),
+  )) {
+    assertGsapAnimationDirectlyEditable(animation);
+  }
```
james-russo-rames-d-jusso: 🟠 **Resize / rotation preflight lacks the live-runtime fallback the drag path already has.** `preflightGsapDragIntercept` at `gsapRuntimeBridge.ts:171` guards: `if (hasLivePosition && !posAnim) return { status: "blocked", reason: "source-uneditable" };` — this catches helper-generated tweens the parser doesn't see. `tryGsapResizeIntercept` (this file) and `tryGsapRotationIntercept` (`gsapRuntimeBridge.ts:320-337`) only inspect `[...animations, ...fetchedAnimations]` — no `hasNonHoldTweenForElement` check on scale/size/rotation channels.

**Failure scenario:** author writes `gsap.to(el, { width: computeWidth() })` in a helper the parser can't statically resolve — tween never appears in `animations`. User resizes → preflight iterates zero size-touching animations → `resolveGroupTween` returns null → the "static-size-set" branch writes `gsap.set(width: 300)` → helper's runtime tween wins → resize disappears. Same bug the drag path guards against, unguarded here.

**Fix:** add `hasNonHoldTweenForElement(iframe, selector, undefined, ["width","height","scale","scaleX","scaleY"])` before the static-set branch in `tryGsapResizeIntercept`, and analogous for `rotation`/`skewX`/`skewY` in `tryGsapRotationIntercept`. Block with `source-uneditable` when live runtime motion exists on the target channels but no editable authored tween does.

— Rames D Jusso

packages/studio/src/hooks/gsapRuntimeBridge.ts
```diff
@@ -268,8 +325,15 @@ export async function tryGsapRotationIntercept(
...
-  if (!selector) return false;
+  const selector = selectorFromSelection(selection) ?? writeTargetSelector(selection);
+  if (!selector) throw new GsapEditBlockedError("no-selector");
+
+  const fetchedAnimations = fetchFallbackAnimations ? await fetchFallbackAnimations() : [];
+  for (const animation of [...animations, ...fetchedAnimations].filter(
+    (candidate) => candidate.propertyGroup === "rotation" || "rotation" in candidate.properties,
+  )) {
+    assertGsapAnimationDirectlyEditable(animation);
+  }
```
james-russo-rames-d-jusso: 🟠 **Rotation filter has two overlapping bugs — misses `rotationX/Y/Z` AND overblocks on skew tweens.** Filter is `candidate.propertyGroup === "rotation" || "rotation" in candidate.properties`. Per `gsapConstants.ts:59`, `rotation` group bundles `{ rotation, skewX, skewY }` — and `rotationX/Y/Z` classify as `"other"`.

**Miss #1 (under-block):** helper-generated `rotationX` tween (very common on 3D card flips) escapes the filter → `assertGsapAnimationDirectlyEditable` never runs against it. Combined with BLOCKER A, the 3D-rotation inspector path double-fails silently.

**Miss #2 (over-block):** if the element has an authored `rotation` tween AND a separate helper-driven `skewX` ambient jitter, the rotation gizmo blocks with `unroll-required` because the skew helper fails `assertGsapAnimationDirectlyEditable` — even though the gizmo only touches `rotation`. User is forced to "Unroll" a skew helper they didn't intend to touch.

**Fix:** filter by the property being written, not the group. For rotation intercept: only iterate animations that carry `rotation` (or the specific rotation axis) as a channel — drop `skewX/skewY` from the filter; add `rotationX/Y/Z` explicitly.

— Rames D Jusso

packages/studio/src/hooks/gsapRuntimeBridge.ts
```diff
@@ -121,24 +127,75 @@ export type { GsapDragCommitCallbacks };
...
+  );
+  if (preflight.status === "blocked" || options?.preflightOnly) return preflight;
   const selector = selectorFromSelection(selection);
-  if (!selector) {
-    return false;
-  }
+  // The preflight above proves this; retain a defensive result for DOM churn.
+  if (!selector) return { status: "blocked", reason: "no-selector" };
 
   // Self-heal: enforce a single position write BEFORE committing. A corrupted
```
james-russo-rames-d-jusso, line 189: 🟠 **Sibling intercept API is asymmetric — drag returns typed `GsapEditOutcome`, resize/rotate still throw.** After this PR:
- Drag: `Promise<GsapEditOutcome>` — preflight *catches* the throw and returns `{ status: "blocked", reason }`.
- Resize (`gsapResizeIntercept.ts:51-73`): `Promise<boolean>` — throws `GsapEditBlockedError` unrecovered.
- Rotation (`gsapRuntimeBridge.ts:320-336`): `Promise<boolean>` — same throw path.

Callers must know to handle BOTH signal shapes (`useGsapAwareEditing.ts:114-130, 240-276, 300-320`). The type system enforces `assertGsapEditPersisted` on drag but nothing enforces block-handling on resize/rotation — a caller that forgets the try/catch simply leaks the throw.

Per `[[feedback_sibling_primitive_pattern_divergence_check]]`, this is a first-class concern for the ownership boundary: same block reason, two different signal shapes across three sibling intercepts. Fix: have all three return `GsapEditOutcome`. The preflight-then-catch pattern already exists in drag; mirror it in resize/rotation. (Endorsing Via's convergent P2.)

— Rames D Jusso

packages/studio/src/hooks/gsapEditOutcome.ts
```diff
@@ -0,0 +1,54 @@
...
+};
+
+export class GsapEditBlockedError extends Error {
+  readonly action: GsapEditAction;
+
+  constructor(readonly reason: GsapEditBlockReason) {
+    super(COPY[reason].message);
+    this.name = "GsapEditBlockedError";
+    this.action = COPY[reason].action;
+  }
```
james-russo-rames-d-jusso, line 20: 🟠 **Dead scaffolding — `GsapEditAction` + copy registry defined, zero UX consumers.** Grep across `packages/studio/src` shows nothing reads `.action` on `GsapEditBlockedError` or imports `GsapEditAction`. `useGsapInteractionFailureTelemetry.ts:22-25` surfaces `error.message` only — the action code (`add-selector` / `unroll` / `open-code`) is defined but never rendered as an "Add selector / Unroll / Open Code" button. Confirmed against `/tmp/hf-stack/2987.patch` — also no consumer there. `handleUnroll` (`useGsapAwareEditing.ts:413-418`) is wired but not reason-gated by the action code.

Per `[[feedback_component_defined_but_no_app_consumer]]`, this is dead scaffolding at merge time. Fix: either drop the `action` field + `GsapEditAction` type from #2986 and add it in the same PR that wires the buttons, OR add a docstring naming the consumer file that will land later so the intent is auditable.

— Rames D Jusso

packages/studio/src/hooks/useGsapAwareEditing.test.tsx
```diff
@@ -79,6 +81,46 @@ function mountResizeHandler(animations: GsapAnimation[]) {
...
+      selector: "#clip",
+    } as unknown as DomEditSelection;
+    let commit!: (selection: DomEditSelection, property: string, value: number) => Promise<void>;
+    function Harness() {
+      commit = useGsapAwareEditing({
+        domEditSelection: selection,
+        selectedGsapAnimations: [],
+        gsapCommitMutation: vi.fn(),
+        previewIframeRef: { current: null },
+        showToast: vi.fn(),
```
james-russo-rames-d-jusso: 🟠 **Tautological inspector-rejection test — mocks the function under test.** The test at `:82-119` sets `mocks.commitAnimatedProperty.mockRejectedValue(error)` and then asserts the outer wrapper reports + rethrows. But in production the RAW `commitAnimatedProperty` implementation swallows every error (see BLOCKER A). The test proves the wrapper's telemetry plumbing but NOTHING about actual editability-block behavior at this surface. Cross-links `[[feedback_mock_cascade_fakeout_in_test_names]]` — the test name ("reports … block reasons …") promises verification of a chain the mock replaces.

**Fix:** replace with an integration test that uses the REAL `useAnimatedPropertyCommit` implementation, mounts a selection whose `selectedGsapAnimations` contain an `editabilityForProvenance === "source"` or `"unroll"` tween, and asserts (a) `gsapCommitMutation` is NOT called and (b) the wrapper's telemetry fires with the `GsapEditBlockedError` reason. If BLOCKER A stays deferred, mark this test `.todo` or delete it — as-is it grants false confidence.

— Rames D Jusso

packages/studio/src/hooks/useGsapAwareEditing.ts
miguel-heygen: Fixed in .  now performs the real ownership preflight before routing, rethrows every persistence/ownership failure after refreshing the cache, and the outer wrapper receives the rejection for telemetry/rollback. Real-hook tests assert helper ownership prevents mutation and persistence failures reject. Focused suite: 46/46; Studio typecheck and pre-commit gate pass.

packages/studio/src/hooks/useGsapAwareEditing.ts
miguel-heygen: Fixed in . Group drag now fetches each member with , caches that parse, preflights every member before the first mutation, then reuses the proven ownership result for the write. Fetch failure rejects the whole gesture; no member mutation occurs. Regression tests cover both a blocked second member and ownership-fetch failure.

packages/studio/src/hooks/gsapResizeIntercept.ts
miguel-heygen: Fixed in . Resize and rotation now mirror drag’s live-runtime defense: exact size/scale or rotation-channel runtime motion with no editable authored mapping returns  before any static write. Dedicated resize and rotation regressions assert zero mutations.

packages/studio/src/hooks/gsapEditOutcome.ts
miguel-heygen: Fixed in . Removed the unused  type and  field; the error now carries only the reason/message actually consumed by telemetry and toast UX.

packages/studio/src/hooks/gsapRuntimeBridge.ts
miguel-heygen: Fixed in . Rotation ownership is now property-exact across , , , and ; skew-only helpers do not block a 2D rotation edit. Tests pin both the 3D helper block and skew non-interference.

packages/studio/src/hooks/useGsapAwareEditing.test.tsx
miguel-heygen: Fixed in . Removed the tautological mocked wrapper test. The replacement mounts the real : helper-authored ownership rejects before , and a real persistence rejection is rethrown. The outer routing suite separately verifies blocked resize restoration.

packages/studio/src/hooks/useGsapAwareEditing.ts
miguel-heygen: Correction with exact symbols: fixed in ca5befb5e. Group drag fetches every member with failOnFetchError=true, caches that parse, preflights all members before the first mutation, then reuses the proven ownership result for writes. Fetch failure rejects the entire gesture with zero member mutations.

packages/studio/src/hooks/useGsapAwareEditing.ts
miguel-heygen: Correction with exact symbols (the prior reply lost code spans): fixed in ca5befb5e. useAnimatedPropertyCommit now performs the real ownership preflight before routing and rethrows every ownership/persistence failure after refreshing the cache. Real-hook tests assert helper ownership prevents mutation and persistence failures reject. Focused suite: 46/46; Studio typecheck and pre-commit gate pass.

packages/studio/src/hooks/gsapResizeIntercept.ts
miguel-heygen: Correction with exact symbols: fixed in ca5befb5e. Resize and rotation mirror drag's live-runtime defense: exact size/scale or rotation-channel runtime motion with no editable authored mapping returns source-uneditable before any static write. Dedicated regressions assert zero mutations.

packages/studio/src/hooks/gsapRuntimeBridge.ts
miguel-heygen: Fixed in ca5befb5e. Drag, resize, and rotation now all return Promise<GsapEditOutcome>; every caller passes the result through assertGsapEditPersisted. Blocked paths no longer mix boolean/throw/outcome contracts, and the focused sibling-route tests pass.

packages/studio/src/hooks/gsapRuntimeBridge.ts
miguel-heygen: Correction with exact symbols: fixed in ca5befb5e. Rotation ownership is property-exact across rotation, rotationX, rotationY, and rotationZ; skew-only helpers do not block a 2D rotation edit. Tests pin both the 3D helper block and skew non-interference.

packages/studio/src/hooks/useGsapAwareEditing.test.tsx
miguel-heygen: Correction with exact symbols: fixed in ca5befb5e. Removed the tautological mocked wrapper test. The replacement mounts the real useAnimatedPropertyCommit: helper-authored ownership rejects before gsapCommitMutation, and a real persistence rejection is rethrown.

packages/studio/src/hooks/gsapEditOutcome.ts
miguel-heygen: Correction with exact symbols: fixed in ca5befb5e. Removed the unused GsapEditAction type and action field; the error now carries only the reason/message consumed by telemetry and toast UX.

miguel-heygen:
Fresh stacked browser evidence (production workspace build + real Chromium):

- Direct `rotationX: 10` persisted and reloaded beside a separate 2D `rotation` tween, confirming the exact-axis filter does not overblock unrelated 2D rotation.
- A helper-owned tween declared through `addTween(target, vars = {...})` rejected an Inspector X edit with the Unroll guidance, issued no mutation request, restored the field, and left source unchanged.
- A forced GSAP-mutation 500 restored the Inspector and preview DOM and surfaced the shared error telemetry/toast path.

Full browser transcript is summarized on #2987.

miga-heygen (commented):
## R2 — Consistent GSAP edit ownership (head `ca5befb`)

### Verdict: Approve

All R1 findings resolved plus the cross-PR dependency concern:

1. **API asymmetry fixed** — All three intercepts (drag, resize, rotation) now return `GsapEditOutcome` consistently. Tests use `.resolves.toEqual({status: "blocked"})` instead of `.rejects.toMatchObject()`. The `GsapEditAction` type is removed — simplified COPY map.

2. **Shared ownership check** — `directEditOutcomeForProperties(animations, properties)` is the single gate for ownership verification, used by drag, resize, rotation, and inspector commits. `animationWritesAnyProperty` is the shared property matcher. `assertGsapAnimationDirectlyEditable` is now private (not exported). Clean SSOT.

3. **Double-fetch fixed** — Resize calls `fetchFallbackAnimations` once, result reused in `allKnownAnimations`. `postSplitFetch` only passed to `resolveGroupTween` when a legacy mixed tween might require re-read after split. Test added: "reuses the ownership parse instead of fetching a resolved size group twice" with `expect(fetchAnimations).toHaveBeenCalledTimes(1)`.

4. **Merge-order dependency resolved** — With rotation/resize returning outcome (not throwing), the bare-catch issue is no longer the primary concern. `directEditOutcomeForProperties` catches `GsapEditBlockedError` internally and returns outcome. The stack is safe to merge in any order now.

New test coverage for runtime size motion blocking and rotation 3D property isolation is strong.

No new findings.

---
_Review by Miga_

james-russo-rames-d-jusso (commented):
_Reviewed at `ca5befb5` — delta since `5ac984ff`._

All seven R1 findings — including both BLOCKERs — addressed with real (non-tautological) test coverage.

**BLOCKER A (inspector edit-path swallow)** — resolved with defense-in-depth across the stack. The `commitAnimatedPropertyRaw` layer in #2987 now (a) narrows its outer catch to re-throw and (b) hoists `assertGsapEditPersisted(directEditOutcomeForProperties(...))` to the top of `commitAnimatedProperties` before the try/routing. New test at `useAnimatedPropertyCommit.test.tsx:63-101` uses the real hook with a helper-provenance anim + `rotationX` prop and asserts (a) commit rejects with `GsapEditBlockedError/reason:'unroll-required'`, (b) `mutations.length === 0`. No swallow, no silent no-op.

**BLOCKER B (group-drag preflight vacuous on transient fetch failure)** — two-part fix:

1. New `failOnFetchError` option in `useGsapAnimationFetchFallback.ts:82` throws `"GSAP animation ownership could not be verified"` when the parse endpoint fails — eliminating the vacuous `[]` case.
2. Group-drag now runs an explicit preflight loop proving editability of every member before the first mutation, caching parsed animations in a `preflightAnimations` Map. The write loop reads from that cache instead of passing bare `[]`.

Tests at `useGsapAwareEditing.test.tsx:236-282` ("preflights every group member before the first mutation") and `:284-302` ("fails a group preflight closed when ownership cannot be fetched") prove commit is never called on block / fetch-fail.

Five concerns also landed:

- **Resize / rotation live-runtime fallback** — resize and rotation intercepts now run `directEditOutcomeForProperties` over `allKnownAnimations` for their respective property groups AND fall back to `hasNonHoldTweenForElement` to block when a live runtime tween exists but no editable source. Dedicated tests for each intercept prove both branches.
- **Rotation filter (rotationX/Y/Z + skew leak)** — new `ROTATION_CHANNELS = ['rotation', 'rotationX', 'rotationY', 'rotationZ']` set; filter uses `animationWritesAnyProperty(anim, ROTATION_CHANNEL_SET)`. Both R1 sub-findings covered: 3D-rotation blocks correctly (tests at `gsapRuntimeBridge.test.ts:246-262`), and an unrelated helper-authored skew tween no longer blocks 2D rotation (`:264-284`).
- **API asymmetry** — all three intercepts now return `Promise<GsapEditOutcome>`. Every call site consumes via `assertGsapEditPersisted(outcome)`; ~25 test-site updates from `.toBe(true)` → `.toEqual({ status: 'persisted' })`.
- **Dead scaffolding** — `GsapEditAction` gone. The `COPY` registry has a real UX consumer: `useGsapInteractionFailureTelemetry.ts:22-24` threads `error.message` into `showToast`, so each blocked reason surfaces a reason-specific toast. Test at `useGsapInteractionFailureTelemetry.test.tsx:17-42` proves the `unroll-required` message reaches `showToast`.
- **Tautological inspector-rejection test** — the mocked-hook test is replaced by the real integration-style test at `useAnimatedPropertyCommit.test.tsx:63-101` (see BLOCKER A). `useGsapAwareEditing.test.tsx` still mocks its dependency, but that layer no longer claims to test inspector-rejection semantics.

Additional hardening spotted during the walk-through: `commitStaticGsap{Position,Rotation,Size}` and `materializeIfDynamic` now throw `GsapEditBlockedError` instead of silently returning; every catch in `useGsapAwareEditing.ts` tracks-and-rethrows.

46 focused tests + typecheck/lint green. Load-bearing PR of the stack, delivered clean. Looks good from my side, leaving as COMMENTED.

— _Review by Rames D Jusso_

jrusso1020 (approved):
Approving to satisfy `require_last_push_approval` at `ca5befb5`.

What I verified at this head, rather than re-reviewing the diff: all eight required contexts are terminal-green (`Semantic PR title`, `Test: runtime contract`, `Typecheck`, `Build`, `regression`, `Test`, `Render on windows-latest`, `Tests on windows-latest`), there is no outstanding change request, and the follow-up passes from both requested reviewers land at this exact SHA with both earlier blockers recorded as closed.

Two notes for whoever runs the merge:

1. The earlier approval was never dismissed. `dismiss_stale_reviews_on_push` is `false` here, so it still reads `APPROVED`; it just predates the latest push, which is what `require_last_push_approval` actually gates on.
2. **Expect this approval to disappear when #2985 lands.** This PR is based on `fix/gsap-helper-defaults`, so merging #2985 retargets it onto `main`, and a base-branch change *is* the one thing that dismisses approvals on this branch (`dismiss_stale_reviews_on_push: false`, but base changes still dismiss). That will need a fresh approval at the retargeted head before this can merge. It is expected behavior, not a failed gate.

vanceingalls (approved):
R2 APPROVE at `ca5befb5`. All 4 findings from R1 addressed at the correct boundary.

- **F1 (`commitAnimatedProperties` swallowing `GsapEditBlockedError`)** — FIXED. `useAnimatedPropertyCommit.ts:543-546` now `catch (error) { bumpGsapCache(); throw error; }`; `useGsapAwareEditing.ts:356-360` wraps with `trackGsapInteractionFailure` + rethrow. Upstream fail-fast via `assertGsapEditPersisted(directEditOutcomeForProperties(...))`.
- **F2 (double `fetchFallbackAnimations` per drag)** — FIXED. `preflightPassed` flag lets group-drag caller reuse the preflight fetch. Resize/rotate use `postSplitFetch = workingAnimations.some((a) => !a.propertyGroup) ? fetchFallbackAnimations : undefined` with an explicit "would perform the same network read twice" comment.
- **F3 (API asymmetry: drag returns outcome, rotate/resize throw)** — FIXED. `tryGsapResizeIntercept` and `tryGsapRotationIntercept` now both return `Promise<GsapEditOutcome>`; all three callers converge on `assertGsapEditPersisted(outcome)` in `useGsapAwareEditing.ts`.
- **F4 (group preflight `animations: []` sentinel)** — FIXED. Group preflight loop now fetches real animations via `makeFetchFallback(selection, { failOnFetchError: true })`, caches per-selection, and passes them to both preflight and commit. WHY comment at `useGsapAwareEditing.ts:165-168` documents "Editability is user-atomic."

No new findings. Standards: CI green (Fallow SUCCESS, Typecheck SUCCESS, Lint SUCCESS, Format SUCCESS). No Golden Rule violations touched.

— Review by Via

PR #2987, opened 2026-08-04
fix(studio): make inspector commits transactional

miguel-heygen:
Stack 2/2 after GSAP ownership. Adds atomic preflight, rollback, and stale-generation protection to inspector and grouped transform commits. 321 changed lines. Split from #2984.

### DIFF
--- a/packages/studio/src/components/editor/propertyPanelCommitField.tsx
+++ b/packages/studio/src/components/editor/propertyPanelCommitField.tsx
@@ -21,7 +21,7 @@ export function CommitField({
   liveCommit?: boolean;
   align?: "left" | "right";
   onPreview?: (nextValue: string) => void;
-  onCommit: (nextValue: string) => void;
+  onCommit: (nextValue: string) => void | Promise<void>;
 }) {
   const [draft, setDraft] = useState(value);
   const valueRef = useRef(value);
@@ -29,6 +29,7 @@ export function CommitField({
   const inputRef = useRef<HTMLInputElement>(null);
   const focusedRef = useRef(false);
   const dirtyRef = useRef(false);
+  const commitGenerationRef = useRef(0);
   valueRef.current = value;
   draftRef.current = draft;
 
@@ -67,14 +68,25 @@ export function CommitField({
     }, 250);
   };
   const cancelGesture = () => {
+    commitGenerationRef.current += 1;
     clearGestureSettleTimer();
     gestureActiveRef.current = false;
     gestureTransaction.cancel();
   };
   const commitDraft = (nextValue: string) => {
+    const generation = ++commitGenerationRef.current;
     setDraft(nextValue);
     onPreview?.(nextValue);
-    if (nextValue !== valueRef.current) onCommit(nextValue);
+    if (nextValue !== valueRef.current) {
+      const baseline = valueRef.current;
+      void Promise.resolve(onCommit(nextValue)).catch(() => {
+        if (generation !== commitGenerationRef.current) return;
+        // The source write is authoritative. A rejected mutation must not leave
+        // the field showing an optimistic value that will disappear on seek.
+        setDraft(baseline);
+        onPreview?.(baseline);
+      });
+    }
   };
   const cancelGestureFromKeyEvent = (event: React.KeyboardEvent<HTMLInputElement>) => {
     if (!gestureActiveRef.current) return false;
@@ -89,6 +101,7 @@ export function CommitField({
     const nextDraft = adjustNumericToken(draftRef.current, direction, event);
     if (!nextDraft) return;
     event.preventDefault();
+    commitGenerationRef.current += 1;
     dirtyRef.current = false;
     gestureActiveRef.current = true;
     gestureTransaction.preview(nextDraft);
@@ -111,6 +124,7 @@ export function CommitField({
   };
 
   useEffect(() => {
+    commitGenerationRef.current += 1;
     if (focusedRef.current && dirtyRef.current) return;
     setDraft(value);
   }, [value]);
@@ -148,6 +162,7 @@ export function CommitField({
         focusedRef.current = true;
       }}
       onChange={(event) => {
+        commitGenerationRef.current += 1;
         settleGesture();
         dirtyRef.current = true;
         setDraft(event.target.value);
--- a/packages/studio/src/components/editor/propertyPanelCommitField.tsx
+++ b/packages/studio/src/components/editor/propertyPanelCommitField.tsx
@@ -30,6 +30,18 @@ export function CommitField({
   const focusedRef = useRef(false);
   const dirtyRef = useRef(false);
   const commitGenerationRef = useRef(0);
+  const pendingCommitRef = useRef<{
+    baseline: string;
+    optimistic: string;
+  } | null>(null);
+  const lastValueRef = useRef(value);
+  if (!Object.is(lastValueRef.current, value)) {
+    lastValueRef.current = value;
+    if (!Object.is(pendingCommitRef.current?.optimistic, value)) {
+      commitGenerationRef.current += 1;
+      pendingCommitRef.current = null;
+    }
+  }
   valueRef.current = value;
   draftRef.current = draft;
 
@@ -79,13 +91,22 @@ export function CommitField({
     onPreview?.(nextValue);
     if (nextValue !== valueRef.current) {
       const baseline = valueRef.current;
-      void Promise.resolve(onCommit(nextValue)).catch(() => {
+      pendingCommitRef.current = { baseline, optimistic: nextValue };
+      const rollback = () => {
         if (generation !== commitGenerationRef.current) return;
+        pendingCommitRef.current = null;
         // The source write is authoritative. A rejected mutation must not leave
         // the field showing an optimistic value that will disappear on seek.
         setDraft(baseline);
         onPreview?.(baseline);
-      });
+      };
+      try {
+        void Promise.resolve(onCommit(nextValue)).then(() => {
+          if (generation === commitGenerationRef.current) pendingCommitRef.current = null;
+        }, rollback);
+      } catch {
+        rollback();
+      }
     }
   };
   const cancelGestureFromKeyEvent = (event: React.KeyboardEvent<HTMLInputElement>) => {
@@ -124,7 +145,6 @@ export function CommitField({
   };
 
   useEffect(() => {
-    commitGenerationRef.current += 1;
     if (focusedRef.current && dirtyRef.current) return;
     setDraft(value);
   }, [value]);
--- a/packages/studio/src/components/editor/propertyPanelTransformCommit.ts
+++ b/packages/studio/src/components/editor/propertyPanelTransformCommit.ts
@@ -41,13 +41,13 @@ export function createTransformCommitHandlers({
   // Route a transform value into the GSAP animation (or a new keyframe) when the
   // element is animated. Returns true when handled, so callers fall through to
   // the manual-transform path only for non-animated elements.
-  const commitAnimatedTransformValue = (
+  const commitAnimatedTransformValue = async (
     property: string,
     value: number,
     noCallbacksMessage: string,
-  ): boolean => {
+  ): Promise<boolean> => {
     if (onCommitAnimatedProperty && hasGsapAnimation) {
-      void onCommitAnimatedProperty(element, property, value);
+      await onCommitAnimatedProperty(element, property, value);
       return true;
     }
     if (gsapKeyframes && gsapAnimId && onAddKeyframe) {
@@ -62,11 +62,11 @@ export function createTransformCommitHandlers({
     return false;
   };
 
-  const commitManualOffset = (axis: "x" | "y", nextValue: string) => {
+  const commitManualOffset = async (axis: "x" | "y", nextValue: string) => {
     const parsed = parsePxMetricValue(nextValue);
     if (parsed == null) return;
     if (
-      commitAnimatedTransformValue(
+      await commitAnimatedTransformValue(
         axis,
         parsed,
         "Cannot edit position — animation callbacks not available",
@@ -74,20 +74,20 @@ export function createTransformCommitHandlers({
     )
       return;
     const current = readStudioPathOffset(element.element);
-    void Promise.resolve(
+    await Promise.resolve(
       onSetManualOffset(element, {
         x: axis === "x" ? parsed : current.x,
         y: axis === "y" ? parsed : current.y,
       }),
-    ).catch(() => undefined);
+    );
   };
 
   // fallow-ignore-next-line complexity
-  const commitManualSize = (axis: "width" | "height", nextValue: string) => {
+  const commitManualSize = async (axis: "width" | "height", nextValue: string) => {
     const parsed = parsePxMetricValue(nextValue);
     if (parsed == null || parsed <= 0) return;
     if (onCommitAnimatedProperty && hasGsapAnimation) {
-      void onCommitAnimatedProperty(element, axis, parsed);
+      await onCommitAnimatedProperty(element, axis, parsed);
       return;
     }
     if (hasGsapAnimation) {
@@ -103,26 +103,26 @@ export function createTransformCommitHandlers({
       current.height > 0
         ? current.height
         : (parsePxMetricValue(styles.height ?? "") ?? element.boundingBox.height);
-    void Promise.resolve(
+    await Promise.resolve(
       onSetManualSize(element, {
         width: axis === "width" ? parsed : width,
         height: axis === "height" ? parsed : height,
       }),
-    ).catch(() => undefined);
+    );
   };
 
-  const commitManualRotation = (nextValue: string) => {
+  const commitManualRotation = async (nextValue: string) => {
     const parsed = Number.parseFloat(nextValue);
     if (!Number.isFinite(parsed)) return;
     if (
-      commitAnimatedTransformValue(
+      await commitAnimatedTransformValue(
         "rotation",
         parsed,
         "Cannot edit rotation — animation callbacks not available",
       )
     )
       return;
-    void Promise.resolve(onSetManualRotation(element, { angle: parsed })).catch(() => undefined);
+    await Promise.resolve(onSetManualRotation(element, { angle: parsed }));
   };
 
   return { commitManualOffset, commitManualSize, commitManualRotation };
--- a/packages/studio/src/components/editor/useInspectorGestureTransaction.ts
+++ b/packages/studio/src/components/editor/useInspectorGestureTransaction.ts
@@ -1,5 +1,9 @@
 import { useCallback, useEffect, useRef, useState } from "react";
 
+function isPromiseCommit(result: void | Promise<void>): result is Promise<void> {
+  return Boolean(result && typeof result.then === "function");
+}
+
 /** One owner for continuous inspector edits: preview freely, persist once. */
 export function useInspectorGestureTransaction<T>({
   sourceValue,
@@ -16,16 +20,23 @@ export function useInspectorGestureTransaction<T>({
   const commitRef = useRef(onCommit);
   const generationRef = useRef(0);
   const pendingRef = useRef<{ before: T; latest: T } | null>(null);
+  const awaitingSourceAckRef = useRef<{ generation: number; value: T } | null>(null);
   const lastSourceValueRef = useRef(sourceValue);
   if (!Object.is(lastSourceValueRef.current, sourceValue)) {
     lastSourceValueRef.current = sourceValue;
+    const matchesSourceAck = Object.is(awaitingSourceAckRef.current?.value, sourceValue);
     const matchesOptimisticValue =
       (activeRef.current && Object.is(activeRef.current.latest, sourceValue)) ||
-      (pendingRef.current && Object.is(pendingRef.current.latest, sourceValue));
+      (pendingRef.current && Object.is(pendingRef.current.latest, sourceValue)) ||
+      matchesSourceAck;
+    if (matchesSourceAck) awaitingSourceAckRef.current = null;
     if (!matchesOptimisticValue) {
       generationRef.current += 1;
       activeRef.current = null;
       pendingRef.current = null;
+      awaitingSourceAckRef.current = null;
+      sourceRef.current = sourceValue;
+    } else {
       sourceRef.current = sourceValue;
     }
   }
@@ -48,6 +59,16 @@ export function useInspectorGestureTransaction<T>({
     previewRef.current(value);
   }, []);
 
+  const rollbackCommit = useCallback((active: { before: T; latest: T }, generation: number) => {
+    if (generation !== generationRef.current) return;
+    pendingRef.current = null;
+    if (awaitingSourceAckRef.current?.generation === generation) {
+      awaitingSourceAckRef.current = null;
+    }
+    sourceRef.current = active.before;
+    previewRef.current(active.before);
+  }, []);
+
   const settle = useCallback(() => {
     const active = activeRef.current;
     activeRef.current = null;
@@ -55,27 +76,29 @@ export function useInspectorGestureTransaction<T>({
       const generation = ++generationRef.current;
       sourceRef.current = active.latest;
       pendingRef.current = active;
+      awaitingSourceAckRef.current = { generation, value: active.latest };
       try {
         const result = commitRef.current(active.latest);
-        void Promise.resolve(result).then(
-          () => {
-            if (generation === generationRef.current) pendingRef.current = null;
-          },
-          () => {
-            if (generation !== generationRef.current) return;
-            pendingRef.current = null;
-            sourceRef.current = active.before;
-            previewRef.current(active.before);
-          },
-        );
+        if (isPromiseCommit(result)) {
+          void result.then(
+            () => {
+              if (generation === generationRef.current) pendingRef.current = null;
+            },
+            () => rollbackCommit(active, generation),
+          );
+        } else if (generation === generationRef.current) {
+          pendingRef.current = null;
+          // Synchronous inspector consumers historically restore their preview
+          // after persisting (color pickers close, curves release the pointer).
+          // Async source mutations keep the optimistic preview until the write
+          // resolves so they do not flash back to the baseline while pending.
+          previewRef.current(active.before);
+        }
       } catch {
-        if (generation !== generationRef.current) return;
-        pendingRef.current = null;
-        sourceRef.current = active.before;
-        previewRef.current(active.before);
+        rollbackCommit(active, generation);
       }
     }
-  }, []);
+  }, [rollbackCommit]);
 
   const cancel = useCallback(() => {
     generationRef.current += 1;
--- a/packages/studio/src/components/editor/useInspectorGestureTransaction.ts
+++ b/packages/studio/src/components/editor/useInspectorGestureTransaction.ts
@@ -8,24 +8,27 @@ export function useInspectorGestureTransaction<T>({
 }: {
   sourceValue: T;
   onPreview: (value: T) => void;
-  onCommit: (value: T) => void;
+  onCommit: (value: T) => void | Promise<void>;
 }) {
   const sourceRef = useRef(sourceValue);
   const activeRef = useRef<{ before: T; latest: T } | null>(null);
   const previewRef = useRef(onPreview);
   const commitRef = useRef(onCommit);
+  const generationRef = useRef(0);
   if (!activeRef.current) sourceRef.current = sourceValue;
   previewRef.current = onPreview;
   commitRef.current = onCommit;
 
   const begin = useCallback(() => {
     if (!activeRef.current) {
+      generationRef.current += 1;
       activeRef.current = { before: sourceRef.current, latest: sourceRef.current };
     }
   }, []);
 
   const preview = useCallback((value: T) => {
     if (!activeRef.current) {
+      generationRef.current += 1;
       activeRef.current = { before: sourceRef.current, latest: sourceRef.current };
     }
     activeRef.current.latest = value;
@@ -36,16 +39,30 @@ export function useInspectorGestureTransaction<T>({
     const active = activeRef.current;
     activeRef.current = null;
     if (active && !Object.is(active.before, active.latest)) {
+      const generation = ++generationRef.current;
       sourceRef.current = active.latest;
       // Restore the captured baseline before the persistent commit captures
-      // rollback state. The commit reapplies `latest` synchronously, so this
-      // is not visible but a failed save can now correctly restore `before`.
+      // rollback state. The commit owner remains responsible for applying its
+      // durable/optimistic value; some controls intentionally leave the
+      // baseline preview in place until their source value advances.
       previewRef.current(active.before);
-      commitRef.current(active.latest);
+      try {
+        const result = commitRef.current(active.latest);
+        void Promise.resolve(result).catch(() => {
+          if (generation !== generationRef.current) return;
+          sourceRef.current = active.before;
+          previewRef.current(active.before);
+        });
+      } catch {
+        if (generation !== generationRef.current) return;
+        sourceRef.current = active.before;
+        previewRef.current(active.before);
+      }
     }
   }, []);
 
   const cancel = useCallback(() => {
+    generationRef.current += 1;
     const active = activeRef.current;
     activeRef.current = null;
     if (active && !Object.is(active.before, active.latest)) {
@@ -53,6 +70,10 @@ export function useInspectorGestureTransaction<T>({
       previewRef.current(active.before);
     }
   }, []);
+
+  useEffect(() => {
+    generationRef.current += 1;
+  }, [sourceValue]);
 
   useEffect(() => cancel, [cancel]);
 
@@ -66,7 +87,7 @@ export function useInspectorGestureDraft<T>({
 }: {
   sourceValue: T;
   onPreview: (value: T) => void;
-  onCommit: (value: T) => void;
+  onCommit: (value: T) => void | Promise<void>;
 }) {
   const [draft, setDraft] = useState(sourceValue);
   const transaction = useInspectorGestureTransaction({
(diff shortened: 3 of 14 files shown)


vanceingalls (approved):
APPROVE — Robust generation-counter design correctly handles concurrent-gesture races (traced 4 scenarios including `sourceValue` advancing via a successful prior commit while a later commit is in flight — the `sourceRef.current = sourceValue` resync on every render covers the tight window). Return-value propagation through `propertyPanel3dTransform.tsx` and `propertyPanelPrimitives.tsx` correctly turns fire-and-forget `void` calls into awaitable chains the transaction can catch. Tests exercise the exact races the PR claims to fix.

Three P2 (no blockers):

- packages/studio/src/hooks/useAnimatedPropertyCommit.ts `commitStaticSet` (~L155-215, `commitSetProps` L143-149) — "Transactional" claim is oversold. The loop `for (const [targetWrite, batch] of byTargetWrite) { await commitSetProps(...) }` (then a second loop for `newSetBatches`) is not atomic: a rejection on the second-or-later mutation leaves earlier writes persisted server-side. Field-level rollback + `bumpGsapCache()` restore *eventual* consistency, not atomicity. Scenario: grouped-transform commit spanning two property groups (position + size on element with legacy mixed writes) — server 500 on second write → position persisted, size not, inspector shows baseline while server has mixed state until cache refetch. Consider narrowing the PR body ("adds error detection + field rollback + stale-generation guard") or wrapping the loop in a server-side batch endpoint.
- packages/studio/src/components/editor/useInspectorGestureTransaction.ts:73-75 — `useEffect(() => { generationRef.current += 1; }, [sourceValue])` uses an effect purely to sync a ref against a prop change. AGENTS.md rule "No `useEffect` for state syncing" applies. Rewrite as a render-time detector: `const lastSourceRef = useRef(sourceValue); if (!Object.is(lastSourceRef.current, sourceValue)) { lastSourceRef.current = sourceValue; generationRef.current += 1; }`. Also strictly-mode-safe.
- packages/studio/src/components/editor/propertyPanelTransformCommit.ts:71-98 + gsapEditOutcome.ts — when `onCommitAnimatedProperty` throws `GsapEditBlockedError` (rich `.message` + `.action`), the transform handlers let the promise reject with no `showToast(err.message)` at this layer. The field silently rolls back and the user sees a value snap-back with no explanation of *why*. Since the error carries actionable copy specifically for user display, either catch-and-toast at the handler boundary or verify an upstream boundary already does. Otherwise, silent no-op UX from the X-user's original complaint persists for the "no-selector"/"unroll-required"/"source-uneditable" paths at the transform-commit boundary.

**Cross-PR pattern (#2986 + #2987): the "explain WHY the edit was blocked" UX is inconsistent across paths.** Same block reason surfaces as a toast on drag/resize/rotate, silent no-op on inspector 3D property, and silent snap-back on transform-commit. Worth unifying at a common boundary before shipping the X-user-facing fix.

Standards: Lint/preview-regression/player-perf green; Fallow ignores on `pickBestAnimation` and `commitAnimatedProperties` (complexity) justified.

— Review by Via

miga-heygen (commented):
## R1 — Transactional inspector commits (head `7143d07`)

### Verdict: Approve (confirming Via's P2s)

This PR completes the error-surfacing story started by #2986. The key fix is changing the bare `catch { bumpGsapCache(); }` in `commitAnimatedProperties` to `catch (error) { bumpGsapCache(); throw error; }` — which unblocks the `GsapEditBlockedError` throws that #2986 added.

### What was verified

**Generation counter for stale-rejection detection:** Both `CommitField` and `useInspectorGestureTransaction` use a `generationRef` that bumps on cancel, new input, new gesture, and source-value changes. Rejection handlers check `generation !== generationRef.current` before restoring. This prevents an old rejection from overwriting a newer draft — tested explicitly in both files.

**Async commit propagation chain:**
- `onCommit` signatures widened from `void` to `void | Promise<void>` through `CommitField`, `FlatRow`, `MetricField`, and `useInspectorGestureTransaction`. Return values are propagated with `return onCommit(...)` at each layer.
- `propertyPanelTransformCommit.ts`: `void Promise.resolve(...)` → `await`. `.catch(() => undefined)` removed. Errors now propagate.
- `CommitField.commitDraft`: rejection → `setDraft(baseline)` + `onPreview?.(baseline)` (snap-back). Generation check prevents stale snap-back.

**Full error flow (with both #2986 + #2987):**
1. Inspector 3D edit → `commitAnimatedProperty` wrapper (from #2986: toast + rethrow) → `CommitField` catch → snap-back
2. Drag/resize/rotate → outcome/throw → `useGsapAwareEditing` catch → toast → `gestureTransaction` rollback
3. All paths: toast + visual restoration. Consistent.

1. **"Transactional" oversold** — agreed as a naming concern. `commitStaticSet` loop is `await`-per-group, not atomic. A mid-loop failure leaves partial state. But the PR title refers to the field-level optimistic-update pattern (draft → commit → reject → snap-back), not database ACID. The field-level pattern IS transactional. The multi-property path is sequential-with-best-effort, which is the pre-existing contract.

2. **`useEffect` generation bump** — confirmed. Both `CommitField` and `useInspectorGestureTransaction` bump `generationRef` in effects, not render-time. The timing gap (render → effect) is sub-frame and hasn't caused issues, but a render-time `useMemo` or inline ref-write would be more correct. P2.

3. **`propertyPanelTransformCommit` doesn't catch/toast** — partially disagree. The error flows through `commitAnimatedProperty` (the #2986 wrapper), which DOES toast via `trackGsapInteractionFailure` before rethrowing. The `CommitField` then catches the rejection and snaps back. The user sees toast + snap-back, same as other paths. Via may have been analyzing the paths without the wrapper in scope.

### Tests

Strong coverage: rejected commit → snap-back, stale rejection → no snap-back, `useInspectorGestureTransaction` rollback, `commitAnimatedProperties` rejection propagation, `bumpGsapCache` called on error.

---
_Review by Miga_

packages/studio/src/hooks/useAnimatedPropertyCommit.ts
```diff
@@ -401,12 +406,30 @@ export function useAnimatedPropertyCommit(deps: CommitAnimatedPropertyDeps) {
...
         selectedGsapAnimations,
         selector,
         primaryProp,
       );
+      if (!anim && !writeTargetSelector(selection)) {
+        throw new GsapEditBlockedError("no-selector");
+      }
       // Whether the element is animated at all. A 3D edit only creates/edits
       // keyframes when it IS — a static element (no keyframes on any of its tweens)
       // gets a `tl.set`, never new keyframes (matches manual drag / resize / rotate).
```
james-russo-rames-d-jusso, line 459: 🔴 **BLOCKER — Swallowing catch is the landing site for #2986's HITL regression fix.** This function is wrapped in `try { … } catch { bumpGsapCache(); }` — swallow. #2986 adds an aspirational `handleGsapAwareCommitAnimatedProperty` wrapper (`useGsapAwareEditing.ts:345-366`) that expects rejection with `GsapEditBlockedError`, but this catch makes the outer wrapper unreachable. Inspector-driven edit on a helper/runtime-owned tween silently no-ops (bump cache, return, done — no toast, no rejection, no draft rollback), while drag/resize/rotate on the same tween surface the block via toast.

See #2986 review's BLOCKER A for full scenario. This is the exact HITL regression signature ("edit disappears, no feedback") the parent PR #2984 committed to closing. Peer (Via) classifies as P2; I'm reading the HITL frame as putting this at blocker-severity for the STACK's stated purpose.

**Fix here:** narrow the catch to re-throw block errors — `catch (e) { if (isGsapEditBlockedError(e)) throw e; bumpGsapCache(); }` — AND add preflight to the top of `commitAnimatedProperties` before the routing branches, using `assertGsapAnimationDirectlyEditable` from `gsapEditOutcome.ts`. Also drop the tautological test at `useGsapAwareEditing.test.tsx:82-119` (mocks the function under test) once the raw path actually throws.

— Rames D Jusso

packages/studio/src/hooks/useAnimatedPropertyCommit.ts
```diff
@@ -191,9 +192,12 @@ async function commitStaticSet(
...
-  );
+  const staticWrites = selector
+    ? animations.filter(
+        (a) =>
+          isInstantHold(a) && tweenTargetsElement(a.targetSelector, selector, selection.element),
+      )
+    : [];
   // Resolve every group's target BEFORE committing anything, and coalesce
   // groups that land on the SAME write into one commit: the snapshot is captured
   // once, so if two groups resolved to one legacy mixed write, a first
```
james-russo-rames-d-jusso, line 237: 🟠 **`commitStaticSet` multi-batch loop is NOT atomic — PR body oversells "transactional".** Around `:216-222` the loop `for (const [targetWrite, batch] of byTargetWrite) { await commitSetProps(...); } for (const batch of newSetBatches) { await addGlobalStaticSet(...); }` — sequential awaits, each a distinct server round-trip.

**Failure scenario:** commit `{x: 50, scale: 1.2}` where `x` and `scale` land on distinct existing writes. First `commitSetProps` succeeds (server persists `x=50`). Second `commitSetProps` throws (id 404's — panel cache stale between preflight and mid-loop). Outer try catches, calls `bumpGsapCache()`, and rethrows — but batch 1 is ALREADY persisted server-side. User asked for both atomically, got `x=50` only, plus a UI error. PR body advertises "atomic preflight, rollback, and stale-generation protection to inspector and grouped transform commits" — this violates the grouped-commit invariant.

Additionally, preflight at `:415-423` only checks `targetGroup = classifyPropertyGroup(primaryProp)` — a multi-group commit gets one group's preflight coverage, but the loop happily writes batches from other groups without their per-group preflight. Second-group animation could be helper-authored yet still receive a write.

**Fix:** either narrow the API contract to single-group commits (`assert groups.size === 1`), or preflight EVERY group in `propEntries` before the write loop, and add compensation logic (rollback of persisted batches) if you want true multi-group atomicity. (Converging with Via's P2 on non-atomic behaviour.)

— Rames D Jusso

packages/studio/src/components/editor/propertyPanelCommitField.tsx
```diff
@@ -67,14 +68,25 @@ export function CommitField({
...
-    if (nextValue !== valueRef.current) onCommit(nextValue);
+    if (nextValue !== valueRef.current) {
+      const baseline = valueRef.current;
+      void Promise.resolve(onCommit(nextValue)).catch(() => {
+        if (generation !== commitGenerationRef.current) return;
+        // The source write is authoritative. A rejected mutation must not leave
+        // the field showing an optimistic value that will disappear on seek.
+        setDraft(baseline);
+        onPreview?.(baseline);
+      });
```
james-russo-rames-d-jusso: 🟠 **DOM mutation from `applyStudio*` is NOT rolled back on rejection.** The catch here restores `draft` (input reverts) but the sync DOM mutation upstream isn't undone.

**Failure scenario:** user types `99px` into W, blurs, `commitDraft` fires. Under the manual (non-GSAP) path, `useDomGeometryCommits.ts:54-58` calls `applyStudioBoxSize(selection.element, next)` SYNCHRONOUSLY (preview iframe resizes to 99px visually), then persists via `commitPositionPatchToHtml`. Persist rejects (write-locked / offline / stale coalesce). This catch runs `setDraft(baseline)` — input reverts to `22px`, but the preview iframe still renders the element at 99px because nothing un-applies the DOM mutation. State divergence between inspector and canvas until reload / re-selection.

The transactional invariant in the PR body ("partial failures roll back cleanly") isn't upheld for the DOM-mutation-then-persist path. **Fix:** either have `commitPositionPatchToHtml`'s rejection snapshot-and-restore the DOM (snapshot before `applyStudio*` and re-apply on catch), or have CommitField's rollback call a caller-supplied `onRollback(baseline)` that both the transaction and the geometry handlers can wire to.

— Rames D Jusso

packages/studio/src/components/editor/useInspectorGestureTransaction.ts
```diff
@@ -36,16 +39,30 @@ export function useInspectorGestureTransaction<T>({
...
     if (active && !Object.is(active.before, active.latest)) {
+      const generation = ++generationRef.current;
       sourceRef.current = active.latest;
       // Restore the captured baseline before the persistent commit captures
-      // rollback state. The commit reapplies `latest` synchronously, so this
-      // is not visible but a failed save can now correctly restore `before`.
+      // rollback state. The commit owner remains responsible for applying its
+      // durable/optimistic value; some controls intentionally leave the
+      // baseline preview in place until their source value advances.
       previewRef.current(active.before);
```
james-russo-rames-d-jusso: 🟠 **Settle-time flash to `before` widens under async commits.** At settle, `previewRef.current(active.before)` fires (drives `setDraft(active.before)` inside CommitField), then `commitRef.current(active.latest)` is awaited. Because `onCommit` is now genuinely async (Promise on the network path), the parent's source value can't advance until the persist ACKs. React repaints CommitField with `draft="before"` for the ENTIRE network round-trip; only when the parent's `value` prop finally advances to `latest` does `useEffect(..., [value])` set draft back to `latest`. Every metric-field release now flashes `latest → before → latest`.

Pre-PR this window was ~one microtask (commit was `void`/synchronous). It's now full network duration. `useInspectorGestureDraft` sidesteps this by calling `setDraft(next)` inside its wrapping `onCommit` at `:100-102`, but `CommitField` and `propertyPanelColor` don't — they pass `onCommit` through raw.

Under Miao's asset-flicker debt lens this is user-visible and a regression against a pattern the codebase already knows how to hide (see `useInspectorGestureDraft`). **Fix:** either omit `previewRef.current(active.before)` when `sourceRef.current` already equals `active.latest` and let the caller drive the final paint, or mirror the `useInspectorGestureDraft` wrap in `CommitField`/`propertyPanelColor` so the settle → commit sync block re-asserts `latest` and batching hides the flash.

— Rames D Jusso

packages/studio/src/components/editor/useInspectorGestureTransaction.ts
```diff
@@ -54,6 +71,10 @@ export function useInspectorGestureTransaction<T>({
     }
   }, []);
 
+  useEffect(() => {
+    generationRef.current += 1;
+  }, [sourceValue]);
```
james-russo-rames-d-jusso: 🟠 **`useEffect(() => generationRef.current += 1, [sourceValue])` gates legitimate rollback.** If a caller synchronously advances the parent state feeding `sourceValue` (e.g. a future `onSetManualOffset` that dispatches an optimistic store update alongside its DOM write), this effect fires and bumps `generationRef`. Later, when the async persist rejects, the catch's `generation !== generationRef.current` check now fires — skipping the transaction's rollback. If the caller doesn't own its own rollback, parent state permanently reflects the never-persisted `latest`. Same shape holds for CommitField's `useEffect(..., [value])` at `:126-130`.

There's an implicit contract ("any onCommit that mutates parent state synchronously must own its own rollback") that isn't documented on the hook's props. **Fix:** either document the invariant on the exported type, or distinguish "bump triggered by settle-publish" from "bump triggered by unrelated parent update" — e.g. compare `sourceRef.current === sourceValue` before bumping.

(Converging with Via's P2 flagging this as an AGENTS.md anti-pattern; peer suggests rewriting as a render-time detector, which resolves the semantic issue as a side-effect.)

— Rames D Jusso

packages/studio/src/components/editor/propertyPanelTransformCommit.ts
```diff
@@ -62,32 +62,32 @@ export function createTransformCommitHandlers({
...
-  const commitManualSize = (axis: "width" | "height", nextValue: string) => {
+  const commitManualSize = async (axis: "width" | "height", nextValue: string) => {
     const parsed = parsePxMetricValue(nextValue);
     if (parsed == null || parsed <= 0) return;
     if (onCommitAnimatedProperty && hasGsapAnimation) {
-      void onCommitAnimatedProperty(element, axis, parsed);
+      await onCommitAnimatedProperty(element, axis, parsed);
       return;
     }
     if (hasGsapAnimation) {
```
james-russo-rames-d-jusso, line 93: 🟠 **Transform-commit boundary doesn't catch `GsapEditBlockedError` → silent snap-back.** `propertyPanelTransformCommit.ts` awaits `onSetManualOffset`/`onSetManualSize`/`onSetManualRotation` via `Promise.resolve(...)` but doesn't route rejection to any user surface. When the underlying handler throws `GsapEditBlockedError` (which #2986's ownership checks now do), the promise rejects, the caller sees no handling — the CommitField catch does eventually restore `draft`, but the transform-commit surface (drag handles, resize corners, rotation gizmo when routed through here) shows nothing and the element snaps back with no explanation.

Same UX-inconsistency pattern as Via's cross-PR item: same block reason, three surfaces (drag toast / inspector silent / transform snap-back). Wrap the await in `try { … } catch (e) { if (isGsapEditBlockedError(e)) { trackGsapInteractionFailure(e, selection, "transform", label); return; } throw e; }` — surface at the `useGsapInteractionFailureTelemetry` boundary like drag does.

(Converging with Via's P2 on silent snap-back.)

— Rames D Jusso

james-russo-rames-d-jusso (commented):
_Reviewed at `7143d07d`._

Solid intent — async plumb-through + rejection rollback + generation gating are the right primitives, and the sibling shape for `generationRef` matches the codebase (see `useSdkSession.ts:165, 211, 254, 271`). Peer R1 (Via) has APPROVE with three P2s; converging on two, adding four correctness findings peer missed, and **escalating one item to blocker** because it's the landing site for #2986's HITL regression fix.

**BLOCKER — inspector swallowing catch lives here and this PR doesn't unswallow it.** `useAnimatedPropertyCommit.ts:395-534` wraps the entire property-routing block in `try { … } catch { bumpGsapCache(); }`. #2986 adds an aspirational `handleGsapAwareCommitAnimatedProperty` wrapper (`useGsapAwareEditing.ts:345-366`) that expects the raw impl to reject with `GsapEditBlockedError` — but this catch swallows it. Net: inspector-driven edit on a helper/runtime-owned tween silently no-ops while drag/resize/rotate on the same tween surface the block via toast. That's the exact HITL regression signature the parent #2984 committed to closing. This PR modifies `useAnimatedPropertyCommit.ts` (hunks at `:24-30, 69-75, 179-184, 192-203, 258-264, 397-403, 406-435, 521-527, 551-560`), so the fix belongs here: narrow the catch to `if (isGsapEditBlockedError(e)) throw e;` and add preflight to `commitAnimatedProperties` before the routing branches (using `assertGsapAnimationDirectlyEditable` from #2986). See #2986 BLOCKER A for the full failure scenario.

Four more concerns, all inline:

- **`commitStaticSet` multi-batch loop is not atomic** — `useAnimatedPropertyCommit.ts:216-222` iterates `commitSetProps` per target-write with sequential `await`. Server 500 mid-loop leaves partial state (`x=50` persisted, `scale=1.2` still in-flight). PR body advertises "atomic preflight, rollback, and stale-generation protection to grouped transform commits" — this violates it. (Converging with Via's P2.)
- **DOM mutation from `applyStudioPathOffset` is NOT rolled back on rejection** — `propertyPanelCommitField.tsx:82-88` restores `draft` in the catch, but `applyStudioBoxSize` / `applyStudioPathOffset` in `useDomGeometryCommits.ts:54-58` synchronously mutate the DOM before the async persist. On persist rejection, the input reverts to baseline but the preview iframe still renders at the failed value. Real transaction gap.
- **Settle-time preview flash to `before` widens under async commits** — `useInspectorGestureTransaction.ts:41-48` fires `previewRef.current(active.before)` at settle. Now that `onCommit` is genuinely async (returns a Promise on the network path), the parent value can't advance until the persist ACKs — every metric-field release paints stale-before-final for the full round-trip. `useInspectorGestureDraft` sidesteps this at `:100-102`; `CommitField` / `propertyPanelColor` do not. Miao's asset-flicker debt lens: user-visible.
- **`useEffect(() => generationRef.current += 1, [sourceValue])` gates legitimate rollback if a caller synchronously advances parent state.** (Converging with Via's P2 flagging this as an AGENTS.md anti-pattern.) A caller doing optimistic-parent-update + async-persist would see the effect bump generation, then the persist rejection's catch would see `generation !== generationRef.current` and skip the transaction rollback — parent state permanently reflects the never-persisted `latest`.
- **`propertyPanelTransformCommit` doesn't catch `GsapEditBlockedError` → silent snap-back** — the transform-commit boundary drops the block on the floor. Same UX-inconsistency pattern Via's cross-PR item calls out. Converging.

**Cross-PR UX pattern (endorsing Via):** the "explain WHY the edit was blocked" surface is inconsistent — toast on drag/resize/rotate, silent no-op on inspector, silent snap-back on transform-commit. Same block reason, three different surfaces. Unifying at `useGsapInteractionFailureTelemetry` before shipping would close the gap the BLOCKER opens.

Small nit worth surfacing (body only, no inline): `propertyPanelTypes.ts:74-76` declares `onSetManualOffset` / `onSetManualSize` / `onSetManualRotation` as `=> void` while `propertyPanelTransformCommit.ts` does `await Promise.resolve(...)`. Sibling handlers in the same file already use `void | Promise<void>`. Align the three to match runtime.

**Test-coverage gaps** — no test covers `commitStaticSet` partial rejection mid-loop, no test covers multi-group preflight where two animations share `targetGroup` but only one is `unroll-required`, no test covers `commitAnimatedProperties({x, opacity})` where x's group passes preflight but opacity's group has a helper-authored tween. All three would have caught the concerns above.

— _Review by Rames D Jusso_

packages/studio/src/hooks/useAnimatedPropertyCommit.ts
miguel-heygen: Fixed across the stacked head. The raw commit path now rethrows all persistence and ownership failures (covered by useAnimatedPropertyCommit rejection tests), so the telemetry wrapper receives the error and the field transaction can roll back. Evidence: ca5befb5e + 66132fafb.

packages/studio/src/hooks/useAnimatedPropertyCommit.ts
miguel-heygen: Fixed in 66132fafb. commitStaticSet now plans every property-group mutation before dispatch and uses the existing server batch capability for one atomic source write. If batch capability is unavailable, it rejects before sending any mutation. Regression tests pin both the atomic x+width path and the fail-before-mutate path.

packages/studio/src/components/editor/propertyPanelCommitField.tsx
miguel-heygen: Fixed in 66132fafb. Each manual geometry persistence owner now captures the exact DOM snapshot before applyStudioPathOffset/applyStudioBoxSize/applyStudioRotation and restores it on rejection, including anchored resize offset. A dedicated hook test rejects all three saves and verifies exact offset, size, and rotation restoration.

packages/studio/src/components/editor/useInspectorGestureTransaction.ts
miguel-heygen: Fixed in 66132fafb. Settle keeps the latest preview visible while persistence is pending; the baseline is applied only on actual rejection. The regression test asserts the preview call sequence is latest while pending, then baseline only after failure.

packages/studio/src/components/editor/useInspectorGestureTransaction.ts
miguel-heygen: Fixed in 66132fafb. The source-value generation effect was removed. A render-time source detector now distinguishes optimistic echoes from unrelated authoritative updates: matching optimistic values preserve rollback eligibility, while a different external value invalidates the stale transaction. Tests cover optimistic echo plus rejection and older rejection versus a newer gesture.

packages/studio/src/components/editor/propertyPanelTransformCommit.ts
miguel-heygen: Fixed across ca5befb5e + 66132fafb. Transform handlers await and rethrow GsapEditBlockedError, the common useGsapAwareEditing wrapper records telemetry and shows the block reason, and CommitField restores the durable baseline. Parametrized tests pin position, size, and rotation propagation with no manual fallback.

miguel-heygen:
Browser verification on stacked head `66132faf` (local CLI production build + real Chromium, `design-panel-qa` fixture):

- Direct GSAP X edit emitted `gsap-mutations`, persisted `gsap.set(... { x: 20 })`, and showed `20px` after a full reload.
- Injected one-shot HTTP 500 for the next GSAP mutation: Inspector rolled back to `0px`, preview DOM rolled back to `translate(0px, 0px)`, disk retained the prior `x: 20`, and Studio showed both `Couldn't save animation: e2e injected fault` and `Failed to save animated edit.`
- `rotationX: 10` persisted beside the fixture's separate 2D `rotation` tween and reloaded as `10°`, exercising the exact-axis ownership filter.
- Replaced the direct tween in the scratch fixture with `addTween(target, vars = {...}); addTween("#qa-tween-box")`: attempting X=25 showed `This motion comes from a helper or loop. Choose Unroll to edit it explicitly.`, sent no `gsap-mutations` request, rolled the field back to `0px`, and left source unchanged.

The repository's older generic `design-panel.mjs` fixture can still select real elements, but its legacy `data-panel-section` field locators no longer match the current flat Inspector DOM; the assertions above were therefore driven directly through the current accessible UI and CDP mouse/keyboard events.

miga-heygen (commented):
## R2 — Atomic inspector persistence (head `66132fa`)

### Verdict: Approve

R1 findings resolved:

1. **Error propagation fixed** — The bare catch pattern is now `.catch((error) => { bumpGsapCache(); throw error; })` at all three mutation sites in `commitAnimatedProperties`. Errors propagate through the wrapper in `useGsapAwareEditing` → toast → rethrow → `CommitField` catch → snap-back.

2. **Transform-commit test coverage** — New `propertyPanelTransformCommit.test.ts` verifies the async error propagation through the transform commit chain.

3. **`useDomGeometryCommits`** — New file cleanly separates geometry commit wiring.

The merge-order concern from R1 is no longer critical since #2986's v2 moves intercepts to outcome-based returns (not throwing). Both PRs are safe to merge independently now, though landing together still gives the cleanest deployment.

No new findings.

---
_Review by Miga_

miguel-heygen:
CI follow-up at `68cb244d`: the full-suite `Test` job exposed two existing synchronous color-editor expectations after the async transaction rewrite (`propertyPanelColor` incomplete-hex settle and `propertyPanelColorCurves` release preview).

Fixed by separating transaction completion semantics:
- Promise-returning persistence keeps its optimistic preview while pending and rolls back only on rejection.
- Synchronous inspector consumers restore their preview immediately after persistence, preserving the established color-picker/curve behavior.
- A source acknowledgement for an earlier commit is tracked separately so it cannot cancel a newer active gesture.

Verification:
- 78/78 focused tests across transaction, CommitField/flat primitives, transform commit, animated-property commit, DOM geometry, color field, and color curves.
- Studio typecheck, format, lint, tracked-artifact gate, file-size gate, and Fallow all pass (inherited duplication remains warning-only).

miga-heygen (commented):
## R3 — Sync/async commit split (head `68cb244`)

### Verdict: Approve

Targeted fix for CI-caught color editor compatibility failures. The R2 `settle()` wrapped all commit results in `Promise.resolve(result).then(...)`, which deferred cleanup to the next microtask — breaking synchronous consumers (color pickers, curve editors) that expect the baseline preview to restore immediately after persist.

### What was verified

**Sync/async split:** `isPromiseCommit` (duck-type thenable check) routes the two paths:
- *Sync commits* (void return): immediately clear `pendingRef` and restore baseline preview — matching historical behavior. Color picker closes, source value arrives in the next render.
- *Async commits* (Promise return): keep optimistic preview visible until resolution. No flash-back to baseline while the write is in flight.

**`awaitingSourceAckRef`:** Tracks the generation + value of a pending async commit. When the incoming `sourceValue` matches the awaited ack value, the render-time detector recognizes it as the commit's own propagation (not an external override) and doesn't bump the generation counter. Clean: cleared on ack, rollback, and cancel.

**`rollbackCommit` extraction:** Shared callback deduplicates the two error-handling blocks from R2 (catch block + promise rejection). Correctly clears `pendingRef`, `awaitingSourceAckRef`, restores source + preview.

**Edge cases:**
- Sync throw → `catch` → `rollbackCommit` → restore. Correct.
- Non-thenable truthy return → `isPromiseCommit` returns false → sync path. Correct.
- Stale rollback (generation mismatch) → no-op. Correct.

No new findings.

---
_Review by Miga_

james-russo-rames-d-jusso (commented):
_Reviewed at `68cb244d` — delta since `7143d07d`, with a targeted sync-compat follow-up from `66132faf` → `68cb244d`._

All six R1 findings — the BLOCKER and all five concerns — addressed.

**BLOCKER (`useAnimatedPropertyCommit.ts` swallowing catch)** — narrowed to re-throw at `:583-585`, AND preflight hoisted to the top of `commitAnimatedProperties` at `:439-444`: `assertGsapEditPersisted(directEditOutcomeForProperties(selectedGsapAnimations, new Set(propEntries.map(([property]) => property))))` runs before any try / routing. Test at `useAnimatedPropertyCommit.test.tsx` (`rethrows a persistence failure to the telemetry wrapper`) asserts `rejects.toBe(failure)` AND `bumpGsapCache` called exactly once. HITL regression path closed at the layer where the swallow lived.

Five concerns landed:

- **`commitStaticSet` multi-batch atomicity** — plans all calls first via `planStaticSetCalls`, then routes single-group calls through `commit(...)` OR multi-group through atomic `commit.batch(calls, ...)`. Throws pre-flight if `commit.batch` is unavailable — zero partial-write. Tests cover single-call, atomic-batch, and missing-batch-preflight-throw paths.
- **DOM mutation rollback on rejection** — all three geometry commits (path offset, box size, rotation) now capture before-state via `captureStudio*`, and on rejection restore via `restoreStudio*` and rethrow. Test at `useDomGeometryCommits.test.tsx` (`restores every optimistic geometry mutation when persistence rejects`) verifies all three via real DOM read-back on `readStudioPathOffset` / `readStudioBoxSize` / `readStudioRotation` after rejection.
- **Settle-time flash to `before`** — at `66132faf` the pre-commit `previewRef.current(active.before)` was removed and only the async rejection callback restores the baseline. `68cb244d` refines further: an `isPromiseCommit` type-guard splits sync-vs-async paths so legacy sync inspector consumers (color pickers, curve editors) preserve their historical after-persist preview-restore, while async commits keep the optimistic preview through the network round-trip. My R1 concern was specifically about the flash *widening under async*; the async path is clean.
- **`generationRef` unconditional bump on `[sourceValue]`** — the `useEffect(() => { generationRef.current += 1 }, [sourceValue])` is gone, replaced with a render-time guard: generation only bumps when the source-value change doesn't match `activeRef.current.latest`, `pendingRef.current.latest`, OR the new `awaitingSourceAckRef.current.value`. Parent echoing the committed value no longer aborts an in-flight rollback path. Test at `propertyPanelFlatPrimitives.test.tsx` (`restores its durable value when an async commit rejects`) re-renders parent with the optimistic value before the rejection fires and still asserts rollback.
- **Transform-commit boundary swallow** — solved via a cleaner mechanism than R1 proposed: instead of catching `GsapEditBlockedError` locally, all three transform handlers `await` and let rejection propagate up to CommitField's uniform rollback. Telemetry fires once at the `useGsapAwareEditing` seam (`useGsapAwareEditing.ts:357` — `trackGsapInteractionFailure(error, selection, "property", "Edit animated property"); throw error;`). Test (`propagates blocked ${axis} edits so the field can roll back`) verifies all three handlers propagate `GsapEditBlockedError` AND the manual setters are not called during a blocked flow.

**One small note on `68cb244d`** — the new sync-branch in `settle()` (`else if (generation === generationRef.current) { pendingRef.current = null; previewRef.current(active.before); }`) is not exercised by a dedicated test in `useInspectorGestureTransaction.test.tsx` (all three tests there are async). The color-editor suites (`propertyPanelColor*.test.tsx`) do cover it indirectly — which is what caught the earlier failures the compat fix addresses. Not blocking; a targeted sync-path unit test on the hook itself would guard against future regression of the split.

82 focused tests + full build/typecheck/lint/precommit + real-browser success/reload/fault paths green. Ready from my side, leaving as COMMENTED.

— _Review by Rames D Jusso_

vanceingalls (approved):
R2 APPROVE at `68cb244d`. All 3 P2 findings + the cross-PR UX consistency callout are FIXED.

- **F1 (atomicity oversold)** — FIXED. A real batch primitive was added: `commit.batch(calls)` posts a single `POST /projects/:id/gsap-mutations-batch/:file`; server-side `applyGsapMutations` applies all mutations in memory then does ONE `writeFileSync` — errors return early with no bytes written. `commitStaticSet` now throws when `commit.batch` is unavailable rather than looping per-target-write. `gsap-mutation-rollback` endpoint added for higher-level orchestration. "Transactional" is now accurate.
- **F2 (`useEffect` for state syncing anti-pattern)** — FIXED. `useInspectorGestureTransaction.ts:24-42` replaces the effect with a render-time detector (`lastSourceValueRef` + `Object.is` guard at the top of the hook body). Preserves optimistic-value ack semantics via `awaitingSourceAckRef`. Matches AGENTS.md.
- **F3 (no toast on `GsapEditBlockedError` at transform-commit)** — FIXED. Toast wiring landed at the correct layer: `useGsapInteractionFailureTelemetry` catches at the `useGsapAwareEditing.commitAnimatedProperty(ies)` wrapper and calls `showToast(error.message, "error")` when `isGsapEditBlockedError(error)`, else "Failed to save animated edit." Then rethrows so `useInspectorGestureTransaction.settle`'s rollback still restores the field. `propertyPanelTransformCommit.test.ts` asserts the error propagates to enable rollback.

**Cross-PR UX consistency** — FIXED at the right boundary. All five gesture surfaces (drag, group-drag, resize, rotate, inspector property/transform commit) now funnel through `trackGsapInteractionFailure` at the same code point in `useGsapAwareEditing.ts` — same error-copy source (`GsapEditBlockedError.message` from the shared COPY table in `gsapEditOutcome.ts`) and same UX (toast + rollback). Silent snap-back is gone across the stack.

No new findings. Standards: CI green modulo in-progress Windows tests, no blockers. Fallow audit SUCCESS. New tests added for batch atomicity + telemetry + geometry paths.

— Review by Via

PR #2988, opened 2026-08-04
test(studio): pin text-field Backspace routing

miguel-heygen:
Independent regression for #2981. Current main does not reproduce the reported deletion, so this pins repeated Backspace to the text field across save and selection refresh while preserving intentional canvas deletion. 215 changed lines. Refs #2981. Split from #2984.

### DIFF
--- a/packages/studio/src/hooks/useAppHotkeys.textEditing.test.tsx
+++ b/packages/studio/src/hooks/useAppHotkeys.textEditing.test.tsx
@@ -1,0 +1,215 @@
+// @vitest-environment happy-dom
+
+import React, { act, useRef, useState } from "react";
+import { createRoot, type Root } from "react-dom/client";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { TextAreaField } from "../components/editor/propertyPanelSections";
+import type { DomEditSelection } from "../components/editor/domEditing";
+import type { LeftSidebarHandle } from "../components/sidebar/LeftSidebar";
+import { usePlayerStore } from "../player/store/playerStore";
+import { useAppHotkeys } from "./useAppHotkeys";
+
+(globalThis as unknown as { IS_REACT_ACT_ENVIRONMENT: boolean }).IS_REACT_ACT_ENVIRONMENT = true;
+
+const timelineDelete = vi.fn(async () => undefined);
+const domDelete = vi.fn(async () => undefined);
+const keyframeDelete = vi.fn();
+const textCommits = vi.fn();
+let root: Root | null = null;
+
+function parentWithTextChild(): DomEditSelection {
+  const element = document.createElement("section");
+  element.id = "selected-card";
+  const child = document.createElement("span");
+  child.textContent = "Kicker";
+  element.append(child);
+  return {
+    element,
+    id: "selected-card",
+    selector: "#selected-card",
+    selectorIndex: 0,
+    label: "Selected card",
+    tagName: "section",
+    sourceFile: "index.html",
+    compositionPath: "index.html",
+    isCompositionHost: false,
+    isInsideLockedComposition: false,
+    boundingBox: { x: 0, y: 0, width: 320, height: 180 },
+    textContent: "Kicker",
+    dataAttributes: {},
+    inlineStyles: {},
+    computedStyles: {},
+    textFields: [
+      {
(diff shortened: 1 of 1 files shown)


miguel-heygen:
Real-browser verification on current head `a1096e2f` (production Studio + CLI build, real Chromium, `design-panel-qa` fixture):

- Selected the real `Qa Zone Headline` element and focused its Inspector text field (`Static Headline`).
- Pressed Backspace, waited through the scheduled save/selection refresh, then pressed Backspace again. The field remained focused with caret ownership, became `Static Headli`, the preview updated, and `index.html` persisted `<h1 ...>Static Headli</h1>` without deleting the selected element.
- Then focused the timeline element itself and pressed Backspace once; the timeline clip and source element were removed, confirming intentional canvas/timeline deletion is still routed.

Focused Vitest: 1/1 passed. Studio typecheck and production Studio/CLI builds pass. The prior timeline viewport check was an unrelated performance-gate flake (3/5 samples passed); its failed job has been rerun.

james-russo-rames-d-jusso (commented):
_Reviewed at `a1096e2f`._

Test-only regression pin for #2981. Uses the real `useAppHotkeys` hook + real `KeyboardEvent` dispatch through happy-dom; mocks only the effect handlers (`handleTimelineElementDelete` / `handleDomEditElementDelete` / `onDeleteSelectedKeyframes` / `textCommits`) so they act as witnesses on which routing decision the hook made — not a mock-cascade fakeout. The three invariants under test map cleanly to the reported bug:

- Backspace on the focused `TextAreaField` before the debounced save (120ms window) — must NOT be preventDefault'd, text commit fires, activeElement + selection remain on the textarea, no delete handler runs.
- Backspace on the same field after the commit's selection-refresh cycle re-renders parent state — same invariants, proves ownership doesn't transfer on the refresh.
- Backspace on the canvas button with a selected timeline element — MUST be preventDefault'd and route to `timelineDelete` exactly once.

The harness comment at `:117-119` explicitly names the failure mode it's guarding ("Text persistence rebuilds the selected parent's text-field model from the preview. Model that refresh: it must not transfer keyboard ownership away from the still-mounted Content editor.") — good future-reader signal.

One small nit worth mentioning (not blocking): the three invariants are packed into a single `it("keeps two Backspaces … across its scheduled save")` block. Splitting into three `it()`s (before-save, after-refresh, canvas-fallback) would give sharper failure attribution if someone regresses just one — but the composed form has value in proving the sequence, so I'd leave it if you prefer. `handleTimelineElementSplit` / `handleCopy` / `handlePaste` / `handleCut` mocks are wired for shape compatibility but never verified — expected in a Backspace-focused test.

All checks green. Clean pin — looks good from my side, leaving as COMMENTED.

— _Review by Rames D Jusso_

vanceingalls (approved):
APPROVE at `a1096e2f`. Test covers both invariants (Backspace stays in Inspector field across the two consecutive presses; timeline-focused Backspace still deletes) with semantics-asserting assertions — `defaultPrevented`, `document.activeElement` identity, exact `selectionStart/End` walk 5→4, exact `textCommits` argument sequence, exact mock-call counts on `timelineDelete`/`domDelete`/`keyframeDelete`. Not mutation-escapable via a no-op handler.

Two nits (non-blocking):

- **Selection-refresh harness thin** — `setSelectionRefreshed(true)` flips a `data-*` attribute on the canvas button and re-renders the parent, but `selectionRef.current` (the `DomEditSelection` object) is initialized once and never rewritten. The real regression path likely rebuilds the parent's text-field model — i.e., replaces `selectionRef.current` with a new object. If the guarded regression involved ref-object replacement, add a second `act()` that assigns a new `parentWithTextChild()` to `selectionRef.current` between the two Backspaces. If the regression was strictly a parent re-render (no ref replacement), then the current harness is fine.
- **Two invariants share one `it` block** — if the inspector-focus-stability assertion fails, the timeline-focused branch never runs. Splitting into two `it` blocks would improve failure diagnostics but is non-blocking.

Standards: CI green (46/46 SUCCESS or SKIPPED, 0 failures). Lint/Format/Typecheck/Test/CodeQL/Fallow-audit/preview-regression all SUCCESS. PR body honestly notes "current main does not reproduce" — this is a forward-looking pin, not a repro, which is the right framing.

— Review by Via

PR #2989, opened 2026-08-04
fix(studio): drain pending edits before reload

miguel-heygen:
External-change stack 1/5. Adds typed draining and recovery for Code, DOM, and queued Studio saves before reload. 422 changed lines. Split from #2984.

### DIFF
--- a/packages/studio/src/utils/studioFileVersion.ts
+++ b/packages/studio/src/utils/studioFileVersion.ts
@@ -1,3 +1,28 @@
+const WRITE_TOKEN_TTL_MS = 10_000;
+const studioWriteTokens = new Map<string, number>();
+
+function pruneStudioWriteTokens(now: number): void {
+  for (const [token, createdAt] of studioWriteTokens) {
+    if (now - createdAt >= WRITE_TOKEN_TTL_MS) studioWriteTokens.delete(token);
+  }
+}
+
+export function markStudioWriteToken(token: string, now: number = Date.now()): void {
+  pruneStudioWriteTokens(now);
+  studioWriteTokens.set(token, now);
+}
+
+export function consumeStudioWriteToken(token: string | null, now: number = Date.now()): boolean {
+  pruneStudioWriteTokens(now);
+  if (!token || !studioWriteTokens.has(token)) return false;
+  studioWriteTokens.delete(token);
+  return true;
+}
+
+export function resetStudioWriteTokens(): void {
+  studioWriteTokens.clear();
+}
+
 /** Browser-safe SHA-256 version matching studio-server's strong ETag format. */
 export async function studioFileContentVersion(content: string): Promise<string> {
   const bytes = new TextEncoder().encode(content);
--- a/packages/studio/src/utils/studioFileVersion.ts
+++ b/packages/studio/src/utils/studioFileVersion.ts
@@ -1,4 +1,6 @@
-const WRITE_TOKEN_TTL_MS = 10_000;
+// A token is marked before the request starts, so its lifetime must cover slow writes,
+// retries, and the subsequent file-change echo without retaining abandoned tokens forever.
+const WRITE_TOKEN_TTL_MS = 5 * 60_000;
 const studioWriteTokens = new Map<string, number>();
 
 function pruneStudioWriteTokens(now: number): void {
@@ -7,11 +9,13 @@ function pruneStudioWriteTokens(now: number): void {
   }
 }
 
+/** Marked by the Studio file writer introduced in external-change stack PR #2990. */
 export function markStudioWriteToken(token: string, now: number = Date.now()): void {
   pruneStudioWriteTokens(now);
   studioWriteTokens.set(token, now);
 }
 
+/** Consumed by the external-change coordinator introduced in stack PR #2991. */
 export function consumeStudioWriteToken(token: string | null, now: number = Date.now()): boolean {
   pruneStudioWriteTokens(now);
   if (!token || !studioWriteTokens.has(token)) return false;
@@ -19,6 +23,7 @@ export function consumeStudioWriteToken(token: string | null, now: number = Date
   return true;
 }
 
+/** Resets module-level echo identity between PR #2991 coordinator tests. */
 export function resetStudioWriteTokens(): void {
   studioWriteTokens.clear();
 }
--- a/packages/studio/src/utils/studioPendingEdits.ts
+++ b/packages/studio/src/utils/studioPendingEdits.ts
@@ -14,6 +14,19 @@ function waitForPostBlurEffects(): Promise<void> {
   return new Promise((resolve) => setTimeout(resolve, 0));
 }
 
+function inspectDrainFailures(results: PromiseSettledResult<unknown>[]): {
+  conflict?: StudioFileConflictError;
+  firstFailure?: PromiseRejectedResult;
+} {
+  let firstFailure: PromiseRejectedResult | undefined;
+  for (const result of results) {
+    if (result.status !== "rejected") continue;
+    if (result.reason instanceof StudioFileConflictError) return { conflict: result.reason };
+    firstFailure ??= result;
+  }
+  return { firstFailure };
+}
+
 export function trackStudioPendingEdit(
   result: Promise<unknown> | unknown,
 ): Promise<unknown> | undefined {
@@ -43,24 +56,16 @@ export async function flushStudioPendingEdits(): Promise<StudioPendingEditsDrain
   window.dispatchEvent(
     new CustomEvent<StudioFlushPendingEditsDetail>(STUDIO_FLUSH_PENDING_EDITS_EVENT, { detail }),
   );
-  let firstFailure: unknown;
-  let hasFailure = false;
+  let firstFailure: PromiseRejectedResult | undefined;
   while (detail.promises.length > 0 || pendingEditPromises.size > 0) {
     const promises = [...detail.promises, ...pendingEditPromises];
     detail.promises = [];
     const results = await Promise.allSettled(promises);
-    for (const result of results) {
-      if (result.status !== "rejected") continue;
-      if (result.reason instanceof StudioFileConflictError) {
-        return { status: "conflict", error: result.reason };
-      }
-      if (!hasFailure) {
-        firstFailure = result.reason;
-        hasFailure = true;
-      }
-    }
+    const batchFailures = inspectDrainFailures(results);
+    if (batchFailures.conflict) return { status: "conflict", error: batchFailures.conflict };
+    firstFailure ??= batchFailures.firstFailure;
   }
-  return hasFailure ? { status: "failed", error: firstFailure } : { status: "clean" };
+  return firstFailure ? { status: "failed", error: firstFailure.reason } : { status: "clean" };
 }
 
 export function addStudioPendingEditFlushListener(
--- a/packages/studio/src/utils/studioPendingEdits.ts
+++ b/packages/studio/src/utils/studioPendingEdits.ts
@@ -1,4 +1,4 @@
-import { StudioFileConflictError } from "./studioSaveDiagnostics";
+import { StudioFileConflictError, type StudioSaveDrainResult } from "./studioSaveDiagnostics";
 
 const STUDIO_FLUSH_PENDING_EDITS_EVENT = "hf-studio-flush-pending-edits";
 
@@ -6,12 +6,13 @@ interface StudioFlushPendingEditsDetail {
   promises: Array<Promise<unknown>>;
 }
 
-export type StudioPendingEditsDrainResult =
-  | { status: "clean" }
-  | { status: "conflict"; error: StudioFileConflictError }
-  | { status: "failed"; error: unknown };
+export type StudioPendingEditsDrainResult = StudioSaveDrainResult;
 
 const pendingEditPromises = new Set<Promise<unknown>>();
+
+function waitForPostBlurEffects(): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, 0));
+}
 
 export function trackStudioPendingEdit(
   result: Promise<unknown> | unknown,
@@ -33,28 +34,33 @@ export async function flushStudioPendingEdits(): Promise<StudioPendingEditsDrain
     active.matches('input, textarea, select, [contenteditable="true"], [role="textbox"]')
   ) {
     active.blur();
-    // React blur handlers enqueue their save synchronously, while state-driven
-    // registrations can land in the next microtask.
+    // ponytail: Preserve synchronous/microtask blur commits, then cross one task boundary
+    // so React effects triggered by the blur can register their flush listener.
     await Promise.resolve();
+    await waitForPostBlurEffects();
   }
   const detail: StudioFlushPendingEditsDetail = { promises: [] };
   window.dispatchEvent(
     new CustomEvent<StudioFlushPendingEditsDetail>(STUDIO_FLUSH_PENDING_EDITS_EVENT, { detail }),
   );
+  let firstFailure: unknown;
+  let hasFailure = false;
   while (detail.promises.length > 0 || pendingEditPromises.size > 0) {
     const promises = [...detail.promises, ...pendingEditPromises];
     detail.promises = [];
     const results = await Promise.allSettled(promises);
-    const rejected = results.find(
-      (result): result is PromiseRejectedResult => result.status === "rejected",
-    );
-    if (rejected) {
-      return rejected.reason instanceof StudioFileConflictError
-        ? { status: "conflict", error: rejected.reason }
-        : { status: "failed", error: rejected.reason };
+    for (const result of results) {
+      if (result.status !== "rejected") continue;
+      if (result.reason instanceof StudioFileConflictError) {
+        return { status: "conflict", error: result.reason };
+      }
+      if (!hasFailure) {
+        firstFailure = result.reason;
+        hasFailure = true;
+      }
     }
   }
-  return { status: "clean" };
+  return hasFailure ? { status: "failed", error: firstFailure } : { status: "clean" };
 }
 
 export function addStudioPendingEditFlushListener(
--- a/packages/studio/src/utils/studioPendingEdits.ts
+++ b/packages/studio/src/utils/studioPendingEdits.ts
@@ -1,8 +1,15 @@
+import { StudioFileConflictError } from "./studioSaveDiagnostics";
+
 const STUDIO_FLUSH_PENDING_EDITS_EVENT = "hf-studio-flush-pending-edits";
 
 interface StudioFlushPendingEditsDetail {
   promises: Array<Promise<unknown>>;
 }
+
+export type StudioPendingEditsDrainResult =
+  | { status: "clean" }
+  | { status: "conflict"; error: StudioFileConflictError }
+  | { status: "failed"; error: unknown };
 
 const pendingEditPromises = new Set<Promise<unknown>>();
 
@@ -19,7 +26,17 @@ export function trackStudioPendingEdit(
   return promise;
 }
 
-export async function flushStudioPendingEdits(): Promise<void> {
+export async function flushStudioPendingEdits(): Promise<StudioPendingEditsDrainResult> {
+  const active = document.activeElement;
+  if (
+    active instanceof HTMLElement &&
+    active.matches('input, textarea, select, [contenteditable="true"], [role="textbox"]')
+  ) {
+    active.blur();
+    // React blur handlers enqueue their save synchronously, while state-driven
+    // registrations can land in the next microtask.
+    await Promise.resolve();
+  }
   const detail: StudioFlushPendingEditsDetail = { promises: [] };
   window.dispatchEvent(
     new CustomEvent<StudioFlushPendingEditsDetail>(STUDIO_FLUSH_PENDING_EDITS_EVENT, { detail }),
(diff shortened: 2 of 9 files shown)


miguel-heygen:
* **#2989** <a href="https://app.graphite.com/github/pr/heygen-com/hyperframes/2989?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/heygen-com/hyperframes/2989?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/studio/src/utils/studioFileVersion.ts
```diff
@@ -1,3 +1,28 @@
+const WRITE_TOKEN_TTL_MS = 10_000;
+const studioWriteTokens = new Map<string, number>();
+
+function pruneStudioWriteTokens(now: number): void {
+  for (const [token, createdAt] of studioWriteTokens) {
+    if (now - createdAt >= WRITE_TOKEN_TTL_MS) studioWriteTokens.delete(token);
+  }
+}
+
+export function markStudioWriteToken(token: string, now: number = Date.now()): void {
```
james-russo-rames-d-jusso, line 13: 🟠 **Write-token API has zero app consumers at merge time.** `markStudioWriteToken` / `consumeStudioWriteToken` / `resetStudioWriteTokens` (this file lines 10-24) are only referenced by `studioFileVersion.test.ts` — grep across `packages/` confirms no runtime callsite marks or consumes tokens. The PR body's framing ("External-change stack 1/5") explains this: the wiring lands in PRs 2-5. Per `[[feedback_component_defined_but_no_app_consumer]]`, the ask is either land alongside the first consumer OR add a docstring naming the downstream PR that will wire it.

Same concern applies to `useEditorSave.ts:138-165` (`flushPendingSave` / `discardPendingSave` / `getPendingCandidate`) — the `EditorSaveHandle` surface is fully exposed but no caller invokes the drain APIs in this PR. If 1/5 merges alone before 2-5 land, we ship orphaned scaffolding.

**Fix (light):** doc-comment on each of the three write-token exports naming which stack PR wires them (e.g. `// Consumed by PR#XXXX — external-change detector, stack 3/5`). That converts orphan-at-merge from a real risk into an auditable trail. Same treatment for `EditorSaveHandle`'s new methods.

— Rames D Jusso

packages/studio/src/utils/studioFileVersion.ts
```diff
@@ -1,3 +1,28 @@
+const WRITE_TOKEN_TTL_MS = 10_000;
```
james-russo-rames-d-jusso: 🟡 **10-second write-token TTL is tight for slow-network Studio writes.** `saveProjectFilesWithHistory` awaits `writeFile` (a POST to the studio server, per the coordinator's contract) — under lossy connectivity or large HTML payloads, that's easily >10s. When the token expires mid-flight, the eventual echo-notification arriving from the server would fail `consumeStudioWriteToken` and get classified as an external change, triggering the very reconcile flow this scaffolding is designed to suppress.

Suggest either (a) tie the TTL to the actual save-timeout window (`serializeStudioFileMutations` presumably has one), or (b) refresh the token's `createdAt` on the request completing so it survives round-trip time and only ages during the echo-wait window.

Not blocking — 10s is fine for the normal case — but wanted to name it before it gets wired downstream.

— Rames D Jusso

packages/studio/src/utils/studioPendingEdits.ts
```diff
@@ -19,16 +26,35 @@ export function trackStudioPendingEdit(
...
 }
 
-export async function flushStudioPendingEdits(): Promise<void> {
+export async function flushStudioPendingEdits(): Promise<StudioPendingEditsDrainResult> {
+  const active = document.activeElement;
+  if (
+    active instanceof HTMLElement &&
+    active.matches('input, textarea, select, [contenteditable="true"], [role="textbox"]')
+  ) {
+    active.blur();
```
james-russo-rames-d-jusso, line 49: 🟡 **Blur → single-microtask wait may miss state-driven pending-edit registrations.** The comment at `:36-37` acknowledges the split — synchronous blur handlers register within the microtask, state-driven handlers may not. But `await Promise.resolve()` is only one microtask tick. A blur handler that does `setState(...)` triggering a `useEffect` that calls `trackStudioPendingEdit` runs the effect AFTER React's commit phase — which, under React 18's automatic batching, can be more than one microtask after the blur event.

Concrete failure: a `<TextField>` whose blur handler calls `setEditing(false)` and the pending-edit registration happens in a `useEffect(() => { if (!editing) trackStudioPendingEdit(persist()) }, [editing])`. Blur fires → setEditing scheduled → microtask ticks → flush proceeds without registration → the effect runs → persist promise is scheduled but not awaited by the drain.

Probably fine for the current call sites (Miguel presumably tested against real callers, and the test at `studioPendingEdits.test.ts` uses a synchronous blur handler). But worth confirming there's no `useEffect`-registered pending edit path in production, OR loop the flush a second time to catch late arrivals (`while (detail.promises.length > 0 || pendingEditPromises.size > 0)` re-runs the drain, but only after the first tick has already registered promises — a state-driven registration two-microtasks-out would still be missed).

— Rames D Jusso

james-russo-rames-d-jusso (commented):
_Reviewed at `1d66603e`._

Foundation PR (1/5 per the body) for typed drain + recovery across Code, DOM queue, focused-field blur, and write-token identity. Design shape is right: discriminated `{status:"clean"|"conflict"|"failed"}` return across three drain APIs, focused-field blur added to `flushStudioPendingEdits` to commit pending debounced work before the drain, and a module-level write-token map so downstream can distinguish own writes from external ones.

Three concerns inline (all 🟠/🟡, no blocker):

- **Zero app consumers at merge time** — write-token API + `EditorSaveHandle`'s new drain methods are exposed but no runtime callsite invokes them. Expected for a 1/5 stack, but recommend doc-comments naming the downstream PR that will wire each surface so 1/5-alone-merged doesn't leave orphaned scaffolding.
- **10s write-token TTL** may be tight for slow-network / large-payload Studio writes — a 12s POST would let its own echo classify as external. Consider tying TTL to the save-timeout window or refreshing on request completion.
- **Blur → single-microtask** wait may miss `useEffect`-driven pending-edit registrations under React 18 batching. Probably fine for current call sites but worth confirming.

**Verified — no fix needed:** the concurrent-Code-save race path Miguel flagged in the delegation message. Re-typing during an in-flight persist cancels the rAF, launches a fresh `persistCandidate` on the new content, and `saveProjectFilesWithHistory`'s shared coordinator serializes the writes. The `inFlightRef === task` guard in `persistCandidate`'s `.finally` correctly protects against overwriting a newer in-flight task's ref cleanup by a stale older task.

**Nits (body-only):**

- The `React blur handlers enqueue their save synchronously, while state-driven registrations can land in the next microtask.` comment at `studioPendingEdits.ts:36-37` is textbook `ponytail:` territory — non-obvious timing invariant that future contributors will otherwise remove as "unnecessary microtask await." Consider prefixing with `// ponytail:` per the HF idiom.
- Three structurally identical drain-result types (`EditorSaveDrainResult` / `DomEditSaveDrainResult` / `StudioPendingEditsDrainResult`) could consolidate to `type StudioSaveDrainResult<E = unknown> = ...` for uniform downstream handling. Not required — foundation-shape is fine — but a compose-site alias would help stack PRs 2-5 avoid triple-narrowing.
- Test coverage adds happy + failure paths for each new drain, plus the join-in-flight scenario for `useEditorSave`. Missing: `flushPendingSave` conflict branch (asserts `{status:"conflict"}`), write-token TTL expiry, and `discardPendingSave` (unused in this PR — expected to be exercised in stack 2/5).

**CI note:** the `Tests on windows-latest > useStudioTestHooks.test.tsx > timeline performance fixture` failure is a 5s timeout on the 50k keyframe fixture generation, with a preceding happy-dom network abort trying to fetch Google Fonts CSS. Classic flake pattern, unrelated to `useEditorSave` / drain logic. Magi noted the shard is rerunning.

Ready from my side, leaving as COMMENTED.

— _Review by Rames D Jusso_

vanceingalls (approved):
Read the diff at `1d66603e` and cross-referenced the drain machinery against the existing `serializeStudioFileMutations` coordinator. Foundation PR (1/5 in the external-change stack) reads clean.

**Concurrent Code-save race — VERIFIED**
`packages/studio/src/utils/studioFileMutationCoordinator.ts` maintains a `WeakMap<writer, Map<path, Promise>>`. `saveProjectFilesWithHistory` (packages/studio/src/utils/studioFileHistory.ts:61) wraps the entire `readFile → writeFile → recordEdit` sequence inside `serializeStudioFileMutations`, so concurrent Code saves to the same file serialize FIFO with the read-modify-write inside the critical section. No stale-clone problem — Magi's claim holds.

**Drain internals I traced through:**
- `useEditorSave.flushPendingSave` — cancels the rAF, joins the in-flight task via identity (`candidate === inFlightCandidateRef.current`), or fires a fresh `persistCandidate`. `.finally` uses `inFlightRef.current === task` closure-capture to avoid stomping a newer in-flight save. Correct.
- `pendingCandidateRef` clear-on-settle guards against clearing a newer candidate (`if (pendingCandidateRef.current === candidate)`). Correct.
- Write-token echo (`markStudioWriteToken` / `consumeStudioWriteToken`) is string-keyed, TTL-pruned on every mark/consume — bounded memory.

**Observations for the downstream stacks (non-blocking):**

- P2 — `flushStudioPendingEdits` uses `results.find(status === "rejected")`, so first-encountered rejection wins. If a batch contains both a `StudioFileConflictError` and a generic failure, the reported severity depends on Set/promise ordering, not severity. Consider two passes (`find(reason instanceof StudioFileConflictError) ?? find(...)`) so conflict beats failed for the reload-decision path.
- P2 — `DomEditSaveQueue.drainError` is set on every rejection but only cleared on `reset()`. A rejected save followed by successful saves leaves `waitForIdle()` reporting `{status: "failed", error: staleError}`. Defensible ("since last reset, at least one save failed") but worth documenting or tying to breaker state.
- P2 — `active.blur() + await Promise.resolve()` catches synchronous blur handlers (as tested). React state-driven blur→setState→useEffect→`trackStudioPendingEdit` may not fit in a single microtask tick. The while-loop mitigates because iterations continue draining, but only for promises that made it into `pendingEditPromises` before dispatch. Follow-up: consider `await flushSync` or a second tick.
- nit — `useEditorSave.flushPendingSave` / `getPendingCandidate` / `discardPendingSave` and `markStudioWriteToken` / `consumeStudioWriteToken` have no production caller yet (grep across `packages/`). Expected for foundation 1/5 — wiring lives in later stacks. Just flagging so reviewers of stacks 2-5 know these entry points arrived here.
- nit — 10s write-token TTL: if a real API write takes >10s (rare, but slow-server case), the token is evicted and the receipt gets treated as external. Would tighten if the generator is deterministic; if UUID-v4, low collision risk.

**Tests (16 focused):**
- Cover individual queues + typed drain shapes (clean/conflict/failed).
- Do NOT cover mixed conflict+failed in one drain.
- Do NOT cover the race between Code editor rAF-buffered save and DOM edit queue draining concurrently — but `serializeStudioFileMutations` has its own coverage.

**CI (`1d66603e`, run 30880080197):** Typecheck / Lint / Format / Fallow audit / Producer unit + integration / Studio load smoke / Studio timeline viewport gate / Windows Tests / 8 of 9 regression shards — all green. Shard-4 (`style-16-prod style-4-prod plan-v2-partial-color audio-mux-parity subtitles ...`) still `in_progress` ~2h in — visual-regression shard, orthogonal to the Studio drain code paths.

Approving on the drain-correctness axis. Follow-ups above can land in stacks 2-5 or as separate polish PRs.

— Review by Via

packages/studio/src/utils/studioFileVersion.ts
miguel-heygen: > Write-token API has zero app consumers at merge time.

Addressed in `ac89e2d1a`: each write-token export now names its exact downstream owner (#2990 mark, #2991 consume/test reset), and the pending Editor save methods name their #2993 reconciliation consumer.

packages/studio/src/utils/studioFileVersion.ts
miguel-heygen: > 10-second write-token TTL is tight for slow-network Studio writes.

Addressed in `ac89e2d1a`: retention is now five minutes so it covers slow writes, retries, and the echo window while still pruning abandoned tokens. Added boundary coverage proving a 60s write remains consumable and a five-minute-old token expires.

packages/studio/src/utils/studioPendingEdits.ts
miguel-heygen: > Blur → single-microtask wait may miss state-driven pending-edit registrations.

Addressed in `ac89e2d1a`: the drain now preserves the microtask checkpoint and crosses one bounded zero-delay task boundary before dispatch, allowing post-commit effects to register. The new regression registers after blur on the next task and proves the save is included.

miguel-heygen:
> "Zero app consumers" / "10s write-token TTL" / "Blur → single-microtask" / "conflict beats failed" / "drainError is stale" / shared drain types and missing conflict, expiry, and discard coverage

Addressed in `ac89e2d1a`: the stacked APIs now name their exact #2990/#2991/#2993 consumers; token retention covers slow writes with expiry coverage; blur draining waits one bounded post-commit task and carries the `ponytail:` invariant; conflict wins independent of ordering; a recovered DOM queue clears stale errors without clearing an open breaker; all three drain shapes share `StudioSaveDrainResult`; and regressions now cover mixed failures, delayed registration, conflict flush, expiry, discard, and queue recovery.

Validation: oxfmt/oxlint and `git diff --check` passed; focused Studio drain tests passed 22/22. The broader direct Studio run reached 3,139 passes, with only pre-existing environment/harness failures in unrelated timeline CSS globals and missing linked dependencies.

vanceingalls (approved):
### R2 delta re-review at `ac89e2d1` (vs R1 head `1d66603e`)

All three P2s and both nits from R1 land cleanly, with driving tests for each branch.

**F1 — flush conflict-priority — FIXED.** `flushStudioPendingEdits` now iterates all rejected results per batch and returns conflict before falling back to `firstFailure` (`studioPendingEdits.ts:36-58`). Test `prioritizes a conflict when pending edits fail with mixed errors` drives the mixed-error path against two listeners — one failing, one conflicting — and asserts conflict wins.

**F2 — DomEditSaveQueue drainError recovery — FIXED.** `run()` now clears `drainError = null` alongside `consecutiveFailures = 0` when a save succeeds and the breaker is closed (`domEditSaveQueue.ts:66-71`). Test `clears a stale drain failure after a successful save` exercises the failure→success sequence and asserts `waitForIdle()` returns `{ status: "clean" }`.

**F3 — post-blur React-effect capture — FIXED.** `flushStudioPendingEdits` now awaits both a microtask *and* one macrotask boundary (`waitForPostBlurEffects` = `setTimeout(0)`) after `active.blur()`, and its comment names the two commit paths (`studioPendingEdits.ts:34-39`). Test `waits for a post-blur effect to register its pending edit listener` uses a blur handler whose `setTimeout(0)` registers `addStudioPendingEditFlushListener` — the flush must observe that listener, and it does.

**Nit-1 — no-caller foundation piece — SKIPPED (documented).** JSDoc on `EditorSaveHandle.{getPendingCandidate,flushPendingSave,discardPendingSave}` and on the token helpers now cites the downstream consumers (`#2990`, `#2991`, `#2993`). Fine as a stack-context anchor.

**Nit-2 — write-token TTL — FIXED.** `WRITE_TOKEN_TTL_MS` is now `5 * 60_000` with a header comment on why (`studioFileVersion.ts:1-3`). Test `keeps a token through a slow write and expires abandoned identity state` proves consume-at-61s succeeds and consume-at-301s prunes.

**Shared type — DONE.** New `StudioSaveDrainResult<Failure = unknown>` in `studioSaveDiagnostics.ts:59-62`. `EditorSaveDrainResult` / `DomEditSaveDrainResult` / `StudioPendingEditsDrainResult` collapse to aliases of it. No ad-hoc union left.

**New (non-blocking) — Fallow audit is red at ac89e2d1.** Eight *minor* findings: seven `fallow/code-duplication` in `useEditorSave.test.tsx` (the new `Probe`/`captured.handle`/`root` scaffolding is copy-pasted across the two new tests + the pre-existing one — a small test helper would eliminate it), and one `fallow/high-cognitive-complexity` at `flushStudioPendingEdits:30` (16 vs threshold 15) — a direct consequence of the F1 conflict-priority loop. Both are cosmetic against the correctness win. Flagging for follow-up, not blocking merge.

Signed-off by Via.

— Review by Via

miguel-heygen:
Follow-up at d066cbe7: the fresh Fallow audit exposed duplicated hook mounts and a one-point cognitive-complexity overage. I extracted the shared test harness and settled-failure inspection without changing drain behavior or assertions.

Validation:
- focused Studio tests: 11/11
- bunx fallow audit --base origin/main --fail-on-issues: pass (duplication 0, complexity 0)
- oxfmt, oxlint, and git diff --check: pass

Hosted CI is running on this exact head.

vanceingalls (approved):
R3 exact-head verify at `d066cbe7` (delta from R2 `ac89e2d1`).

**Delta scope (2 files)**
- `packages/studio/src/hooks/useEditorSave.test.tsx` — purely extractive: `mountEditorSave(writeProjectFile)` helper folds 4 duplicated `Probe`/`createRoot`/`render`/`unmount` blocks into one. Every test's hook args, ordering, and assertions are byte-preserved.
- `packages/studio/src/utils/studioPendingEdits.ts` — `inspectDrainFailures(results)` extracted from the drain loop. Complexity delta only.

**Drain-semantics equivalence (F1 / F2 / F3)**
- *F1 conflict-first classification*: preserved. Helper linearly scans each batch's settled results — returns `{conflict}` on the first `StudioFileConflictError`, else `{firstFailure}` at first non-conflict rejection. Same "conflict wins in batch" order as before. Cross-batch first-failure preservation retained via `firstFailure ??= batchFailures.firstFailure` at the loop level. Final result payload is identical (`firstFailure.reason` after extract == `firstFailure` before extract — the wrapper is `PromiseRejectedResult` whose `.reason` is what the pre-refactor `hasFailure` branch already returned).
- *F2 `DomEditSaveQueue.run` drainError clearing*: `packages/studio/src/utils/domEditSaveQueue.ts` is NOT in the R2→R3 delta. Byte-identical at both HEADs.
- *F3 blur post-effect capture*: `waitForPostBlurEffects` + `await Promise.resolve(); await waitForPostBlurEffects();` unchanged at HEAD; verified via file read at `ref=d066cbe7`.

**CI**
All completed checks SUCCESS at `d066cbe7`. Analyze (js/ts), Preflight (lint+format), Preview parity still IN_PROGRESS — approval predicated on those finishing green (they exercise the refactored paths).

Approving.

— Review by Via

james-russo-rames-d-jusso (commented):
_Delta-reviewed `1d66603e..d066cbe7`._

Verified each of my R1 concerns against the current head. All resolved. Three orthogonal improvements went in beyond what I flagged.

| # | R1 concern | R2 verdict | Anchor |
|---|---|---|---|
| I1 | Write-token API has zero app consumers — add docstrings naming the downstream PR | 🟢 RESOLVED | `studioFileVersion.ts:12,18,26` name PR #2990/#2991; `useEditorSave.ts:38,40,42` name PR #2993 |
| I2 | 10-second write-token TTL is tight for slow-network Studio writes | 🟢 RESOLVED | `studioFileVersion.ts:3` — `WRITE_TOKEN_TTL_MS = 5 * 60_000`, comment names slow-writes + retries + subsequent echo. Test at `studioFileVersion.test.ts:53-59` asserts 61s → still valid, 301s → expired |
| I3 | Blur → single-microtask may miss useEffect-driven pending-edit registrations under React 18 | 🟢 RESOLVED | `studioPendingEdits.ts:52-53` — `Promise.resolve()` + `waitForPostBlurEffects()` (setTimeout-0 task boundary). Test at `studioPendingEdits.test.ts:39-63` explicitly registers listener in `setTimeout(0)` from a blur handler and asserts flush waits for it |
| B1 | `ponytail:` marker on the timing comment | 🟢 RESOLVED | `studioPendingEdits.ts:50` — `// ponytail:` prefix landed, comment now names the react-effects invariant |
| B2 | Three structurally identical drain-result types — consolidate | 🟢 RESOLVED | `studioSaveDiagnostics.ts:59-62` — shared `StudioSaveDrainResult<Failure = unknown>`; all three sites (`studioPendingEdits.ts:9`, `useEditorSave.ts:34`, `domEditSaveQueue.ts:27`) alias to it |
| B3 | `flushPendingSave` conflict-branch test | 🟢 RESOLVED | `useEditorSave.test.tsx:105-121` preserves conflict details; plus `studioPendingEdits.test.ts:103-127` prioritizes conflict over mixed errors |
| B4 | Write-token TTL expiry test | 🟢 RESOLVED | `studioFileVersion.test.ts:53-59` covers both hold-through-slow-write and abandoned-token-expiry |
| B5 | `discardPendingSave` test | 🟢 RESOLVED | `useEditorSave.test.tsx:126-140` asserts null candidate, no writeProjectFile, `cancelAnimationFrame` called |

**Orthogonal improvements beyond my findings (all correct):**

- **Conflict-priority drain semantics** (`studioPendingEdits.ts:17-28,64-68`) — `inspectDrainFailures` scans all rejections; any `StudioFileConflictError` short-circuits to `{status:"conflict"}`. Previously "first rejection wins" would return `{status:"failed"}` even when a subsequent listener rejected with a conflict. User impact of a conflict is higher (needs external-content reconciliation before commit), so it dominates. The `firstFailure ??= …` accumulator across loop iterations correctly preserves the earliest non-conflict failure across drain rounds. Good.
- **Recovered-queue error clearing** (`domEditSaveQueue.ts:66-70`) — `drainError = null` on successful save when the breaker is closed. Previously one transient failure would leave `waitForIdle()` reporting `{status:"failed", error:<stale>}` indefinitely. Test at `domEditSaveQueue.test.ts:118-131` proves recovery.
- **Extracted helpers** — `waitForPostBlurEffects()` and `inspectDrainFailures()` make the drain sequence testable and self-documenting.

**Nothing else surfaced in the delta pass.** Foundation ships clean. Ready from my side, leaving as COMMENTED.

PR #2990, opened 2026-08-04
feat(studio): preserve external file conflicts

miguel-heygen:
External-change stack 2/5. Adds exact Studio write identity, overwrite support, and durable IndexedDB conflict snapshots. 247 changed lines. Split from #2984.


miguel-heygen:
* **#2990** <a href="https://app.graphite.com/github/pr/heygen-com/hyperframes/2990?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/heygen-com/hyperframes/2990?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

miguel-heygen:
Fixed a lost-response retry identity race at `d40296721`.

Each PUT attempt now creates and marks its own `X-Hyperframes-Write-Token`. If an earlier request committed but its response was lost, both filesystem receipts have independent tokens and both self-write echoes can be suppressed. The regression forces the first PUT response to disappear, advances the retry timer, and proves the two request tokens differ; it failed on the prior head and passes now.

Also removed the inherited duplicate hook harness surfaced by the commit gate and documented the four #2990 stack-boundary exports in Fallow config (first consumed by #2991).

Verification:
- 10/10 focused tests across file-manager retry ownership, editor-save drain, file versions, and IndexedDB conflict storage
- Studio typecheck
- targeted oxlint + oxfmt
- `git diff --check`
- Fallow audit vs `origin/main`: zero issues
- full pre-commit and commit-message hooks green

packages/studio/src/hooks/useFileManager.ts
```diff
@@ -201,6 +208,24 @@ export function useFileManager({
...
+  const overwriteExternalConflict = useCallback(
+    async (conflict: StudioFileConflictError) => {
+      if (conflict.currentContent != null) {
+        await writeProjectFile(
+          conflict.filePath,
+          conflict.attemptedContent,
+          conflict.currentContent,
+        );
+      } else {
+        fileVersions.set(conflict.filePath, conflict.currentVersion);
```
james-russo-rames-d-jusso, line 222: 🟠 **`overwriteExternalConflict`'s null-content fallback (`:220-224`) can silently overwrite a third-party edit.** The `else` branch runs when the conflict payload didn't carry `currentContent` — it sets `fileVersions.set(conflict.filePath, conflict.currentVersion)` (the version we ORIGINALLY conflicted against) and writes `attemptedContent` without an expected-content hint.

**Failure scenario:** A edits `index.html` → save → 409 with `currentVersion: v2`. Between receiving that 409 and A clicking "Overwrite," collaborator B has already saved v3. A clicks Overwrite → this fallback sets our expected version to v2 (stale) and PUTs `attemptedContent`. Depending on server-side check behavior, either (a) the write succeeds silently, destroying B's v3 work with no user awareness, or (b) it 409s again on v2-vs-v3 mismatch (better outcome) — but the code path doesn't distinguish.

The primary path (`currentContent != null`) is safe because it threads `currentContent` as the `expectedContent` hint, so the server can enforce the intended-overwrite invariant. The null-content fallback bypasses that.

**Fix (light):** either (a) always re-fetch the current file version + content before overwrite when `currentContent == null` (single extra GET), OR (b) mark the fallback as an intentional force-overwrite in the request (server flag), and add a ponytail comment explaining why the version-set-then-blind-write is safe for the specific case the fallback is designed for. As-is, the two branches look symmetric but have divergent safety envelopes.

— Rames D Jusso

packages/studio/src/utils/externalConflictStorage.ts
```diff
@@ -0,0 +1,162 @@
...
+        store.put(snapshot, key(snapshot.projectId, snapshot.filePath)),
+      );
+    },
+    async delete(projectId, filePath) {
+      await withStore<undefined>("readwrite", (store) => store.delete(key(projectId, filePath)));
+    },
+  };
+}
+
+const indexedDbStorage = createIndexedDbExternalConflictStorage();
```
james-russo-rames-d-jusso, line 113: 🟡 **IndexedDB persist is best-effort; snapshot loss is silent unless the caller handles rejection.** `persistExternalConflictSnapshot` / `persistExternalFailureSnapshot` (`:115-140`) are `async` with no internal try/catch — an IndexedDB failure (Safari private mode, storage quota exceeded, transaction aborted by browser) rejects the promise. If the downstream #2991 caller does `void persistExternalConflictSnapshot(...)` (fire-and-forget), the user gets a conflict UI but the recovery snapshot never lands. On reload, the recovery flow finds nothing and the user's `attemptedContent` is silently lost.

Not a bug in this PR — the primitive correctly propagates errors. But: (a) the downstream wiring PR needs to await + fallback (e.g. keep the conflict in memory + surface a warning); (b) `openDb` rejects with `"IndexedDB is unavailable; conflict recovery snapshot was not saved"` (`:41-43`) which reads as an internal note — it needs to reach a user-visible surface via the downstream wiring; (c) worth documenting on each exported `persist*` function that the promise SHOULD be awaited by the caller, not fire-and-forgot, and that rejections indicate loss of durable recovery.

— Rames D Jusso

james-russo-rames-d-jusso (commented):
_Reviewed at `d4029672`._

Stack 2/5. Retry-race fix + external-conflict storage primitive + `overwriteExternalConflict` action wired into `FileManagerContext`. Two concerns inline (both non-blocker), plus explicit endorsement of the pattern shift.

**Endorsing the `.fallowrc.jsonc` proactive-consumer-naming pattern.** The new allowlist entry at `:172-181` names #2991 as the downstream consumer for the four `externalConflictStorage` exports. This is exactly the pattern I asked for on #2989 R1 (name downstream wiring PR in the scaffolding surface). Please carry this forward for `EditorSaveHandle`'s new drain methods and the `studioFileVersion` write-token API too — either land them in the same file's Fallow entry or a doc-comment on each export. That converts orphan-at-merge from a real risk to an auditable trail across the whole stack.

**Retry-race fix — verified.** The write-token generation moved from outside `retryStudioSave` (one token per save call) to inside (`useFileManager.ts:122-127` — one token per attempt). The comment names the failure mode precisely ("a committed request loses its response, the retry can produce a second filesystem receipt that must be suppressed independently"). Regression test at `useFileManager.projectOwnership.test.tsx:207-247` asserts distinct `X-Hyperframes-Write-Token` headers across the two PUT calls after a `TypeError("response lost")` retry — this is the right shape.

Independent of my #2989 concern about the 10s TTL: that concern was about clock-based expiry during a single slow write; this fix is about identity uniqueness across retries. They're complementary, both worth landing.

Two inline concerns — see comments — mostly about the `overwriteExternalConflict` null-content fallback (stale-version race window) and IndexedDB persist being best-effort at merge time.

**Nits (body-only):**

- `externalConflictStorage.test.ts` only exercises the memory-backed variant via `createMemoryExternalConflictStorage`. The IndexedDB path (`createIndexedDbExternalConflictStorage`, `openDb`, `withStore`) is untested at unit level. happy-dom / fake-indexeddb would make a smoke test cheap — asserts open-store-put-get-delete lifecycle for the real path.
- `persistExternalConflictSnapshot` etc. are hardcoded to a module-level `indexedDbStorage` singleton (`:113`). Downstream integration tests would benefit from being able to inject a memory storage (via factory or context param). Not blocking — foundation-shape is fine — but flag if the wiring PR wants to unit-test the interaction without touching real IndexedDB.
- `overwriteExternalConflict` has no test in this PR. Both branches (currentContent present vs null) deserve coverage before #2991 wires the UI.

**Endorsing the tests:** the `useFileManager.projectOwnership.test.tsx` retry-race test is clean — mocks the transport layer, drives real fetch calls, inspects header identity directly. No mock-cascade fakeout (per `[[feedback_mock_cascade_fakeout_in_test_names]]`).

Ready from my side, leaving as COMMENTED.

— _Review by Rames D Jusso_

vanceingalls (approved):
**R1 verdict: APPROVE.** Fix mechanism verified against \`d40296721\`.

**Race + fix trace (verified):**
- Before: \`createStudioWriteToken()\` fired ONCE outside \`retryStudioSave\` in \`useFileManager.ts\`; the same UUID rode both the lost-response PUT and its retry. Server appended two \`FileWriteReceipt\` entries sharing that token; client \`studioWriteTokens\` Map was marked once → \`consumeStudioWriteToken\` popped true on the first watcher echo, false on the second → second self-write reclassified as external.
- After: token generation + \`markStudioWriteToken\` moved INSIDE the retry callback. Each attempt mints its own \`crypto.randomUUID()\`, marks it in the client Map, and sends it as \`X-Hyperframes-Write-Token\`. Server \`createWriteToken(header)\` respects the client-supplied UUID, so both server receipts and both client marks are distinct.
- Token semantics: UUID string keyed in a Map — value-equality on a primitive is effectively strict identity; collision negligible. Symmetric 10s TTL on both sides (\`WRITE_TOKEN_TTL_MS\` client / \`RECEIPT_TTL_MS\` server).

**Regression fidelity:** \`useFileManager.projectOwnership.test.tsx\` forces PUT #1 to reject with \`TypeError("response lost")\`, drains \`retryStudioSave\` timers, then asserts the two PUTs' \`X-Hyperframes-Write-Token\` headers differ. On the prior head both would be the pre-loop constant → assertion fails. Test drives the exact failure mode.

**Failure-mode coverage:**
- Parallel writes to different files: each \`writeProjectFile\` has its own retry loop → distinct tokens. Server FIFO is per-absPath.
- Compound retry (client retry after server retry): each new client attempt mints fresh; server \`createWriteToken\` accepts each. Clean.
- Classifier consumer (\`consumeStudioWriteToken\`) still has value-equality on string tokens — the fresh-token-per-retry regime feeds it correctly. No update needed here; wiring lands in #2991 per PR body.

**Nits (non-blocking):**
- \`externalConflictStorage.ts\` IndexedDB path has no direct test — only \`createMemoryExternalConflictStorage\` is exercised. Coverage gap on the production persistence layer.
- \`openDb()\` opens+closes a DB connection per get/set/delete; wasteful but safe.
- \`overwriteExternalConflict\` else-branch when \`currentContent == null\` writes with no \`expectedContent\` and \`If-None-Match: *\` — inherent race if an external re-create lands in the gap, but caller will just receive a fresh conflict. Acceptable.

**Standing #2989 findings:** not touched by this PR (\`flushStudioPendingEdits\` first-rejection-wins, \`DomEditSaveQueue.drainError\` non-reset, \`active.blur() + microtask\` skip, 10s TTL eviction). All still open for their own PR — none regressed here.

**CI:** SUCCESS on player-perf + change-detects; \`Preview parity\` + Graphite mergeability still in progress; UNSTABLE from downstack #2989. No red required checks.

— Review by Via

packages/studio/src/hooks/useFileManager.ts
miguel-heygen: > `overwriteExternalConflict`'s null-content fallback can silently overwrite a third-party edit.

Addressed with regression evidence: this path is already conditional, not blind. It stores the observed version, then `writeProjectFile` emits `If-Match`; a deleted-file conflict emits `If-None-Match: *`. New client tests prove that a stale v2 overwrite receives and propagates a fresh v3 `StudioFileConflictError`, and that the deleted-file path is create-only. The existing server route tests also verify stale `If-Match` and racing create-only requests preserve the newer bytes. No force flag or extra GET was added because either would weaken or duplicate the existing CAS contract.

packages/studio/src/utils/externalConflictStorage.ts
miguel-heygen: > The downstream wiring PR needs to await + fallback, and each exported `persist*` function should document that rejection means loss of durable recovery.

Addressed: both persist exports now state that callers must await them, retain the Studio draft in memory, and surface a rejection because durable recovery was not written. The tests now cover both the real IndexedDB put/get/delete lifecycle and the explicit unavailable-IndexedDB rejection. Downstream #2991 already injects and awaits these callbacks and converts rejection into retained blocked state rather than fire-and-forgetting it.

miguel-heygen:
Addressed the body-level follow-ups at `9969887d7`:

> `externalConflictStorage.test.ts` only exercises the memory-backed variant.

Added `fake-indexeddb` coverage through the production helper path: database/store creation, put, get, delete, and the explicit unavailable-IndexedDB rejection all run directly.

> Downstream integration tests would benefit from being able to inject a memory storage.

Kept the storage module API small because the injection seam already exists at the correct boundary in #2991: the coordinator accepts persist/load/delete callbacks and its tests inject fakes; #2993 supplies the real IndexedDB helpers.

> Both `overwriteExternalConflict` branches deserve coverage.

Added coverage for the content-derived `If-Match` branch, the content-less stale-version branch propagating a fresh v3 409, and the deleted-file `If-None-Match: *` branch.

> Carry proactive consumer naming forward for `EditorSaveHandle` and the write-token API.

Now inherited from the rebased #2989 parent: each drain method names #2993, `markStudioWriteToken` names #2990, `consumeStudioWriteToken` names #2991, and the reset helper names its #2991 test consumer.

Validation: focused Studio/storage/server tests pass (76/76), Studio typecheck passes, focused oxlint/oxfmt pass, and Fallow reports no issues across the stacked diff.

james-russo-rames-d-jusso (commented):
_Delta-reviewed `d40296721..9969887d`._

Verified both R1 concerns against the current head. Both resolved. Also carrying the whole #2989 R2 delta (docstrings, TTL-5min, post-blur task boundary, shared drain type, conflict-priority, recovered-queue clearing) via the restack onto `d066cbe7`.

| # | R1 concern | R2 verdict | Anchor |
|---|---|---|---|
| P1 | `overwriteExternalConflict`'s null-content fallback (`useFileManager.ts:221-224`) can silently overwrite a third-party edit | 🟢 RESOLVED WITH REGRESSION EVIDENCE — CAS invariant catches this | `useFileManager.projectOwnership.test.tsx:204-224` "preserves a newer third-party edit when a content-less conflict version is stale". Stale v2 conflict → server 409 with v3+external content → `overwriteExternalConflict` rejects with a fresh `StudioFileConflictError`. My R1 missed the CAS wire — `writeProjectFile` still emits `If-Match: {conflict.currentVersion}` from `fileVersions` in the null-content branch, so server-side CAS catches the third-party edit. Additional coverage: `:187-202` content-carrying overwrite emits `If-Match` with hash of external content; `:226-240` deleted-file path emits `If-None-Match: *` |
| P2 | IndexedDB persist is best-effort; snapshot loss is silent unless caller handles rejection — wiring PR needs to await + surface | 🟢 RESOLVED | `externalConflictStorage.ts:115-121,133-139` — both `persistExternalConflictSnapshot` and `persistExternalFailureSnapshot` carry docstrings stating "Callers must await this promise: rejection means the Studio draft was not saved durably and must remain available in memory while the storage failure is surfaced to the user." Test at `externalConflictStorage.test.ts:57-82` verifies persist/load/delete through real IndexedDB (via `fake-indexeddb`); `:84-95` verifies "rejects persistence when IndexedDB is unavailable" — proves the rejection contract with a concrete error message |

**Correction on my P1 R1:** I flagged the null-content fallback as a silent-overwrite risk without tracing through `writeProjectFile`'s wire. The CAS contract (If-Match / If-None-Match on the PUT) is the invariant that catches the third-party edit, and Miguel's `useFileManager.projectOwnership.test.tsx:204-224` regression proves it end-to-end. Fair rebuttal; retracting the concern.

**Nothing else surfaced in the delta pass.** Foundation ships clean, and #2990 correctly re-emits a fresh conflict rather than clobbering. Ready from my side, leaving as COMMENTED.

vanceingalls (dismissed):
**R3 verdict: APPROVE.** Fresh-head verify at `9969887d` after restack onto merged #2989 `d066cbe7`.

**Thread 1 — overwrite atomicity via `If-Match` / `If-None-Match` (FIXED).**

Traced `overwriteExternalConflict` (`useFileManager.ts:213-228`) → `writeProjectFile` (`:101-177`). Three atomic branches, each covered:

1. `currentContent != null` → `writeProjectFile(path, attempted, currentContent)` → `studioExpectedFileVersion` returns `studioFileContentVersion(currentContent)` (SHA-256 of the exact reviewed bytes) → `If-Match: <sha256>`. Server 409s if content on disk isn't byte-for-byte the reviewed version.
2. `currentContent == null` (stale-version-only) → `fileVersions.set(path, currentVersion)`; PUT rides with `If-Match: <currentVersion>`. Fresh third-party edit produces 409 with fresh `currentVersion`/`currentContent`, re-thrown as `StudioFileConflictError` back into reconciliation — no silent overwrite.
3. `currentVersion == null` (deleted-file) → `If-None-Match: *` (create-only).

Tests at `useFileManager.projectOwnership.test.tsx:187-240` inspect the actual `Headers` object of the outgoing PUT for all three branches (`.get("If-Match")` / `.get("If-None-Match")`). Not classList-vs-CSS; the real emitted header is asserted. Path 2 asserts the 409 propagates the fresh `v3`/`THIRD PARTY` back to caller — third-party edit preservation verified.

**Thread 2 — IndexedDB persistence contract (FIXED).**

`externalConflictStorage.test.ts` now runs `persistExternalConflictSnapshot` / `loadExternalConflictSnapshot` / `deleteExternalConflictSnapshot` through the production `indexedDbStorage` singleton against `fake-indexeddb`'s `IDBFactory` (standard, not hand-roll — per memory reference). Exercises `openDb` upgrade path, `withStore` readonly/readwrite transactions, and the `db.close()` on `oncomplete`. Unavailable-storage branch is tested by stubbing `globalThis.indexedDB = undefined` → `persistExternalConflictSnapshot` rejects with the documented message. Doc comments on both persist helpers commit to the awaited-rejection contract ("draft was not saved durably and must remain available in memory"). Callers must not silently drop.

**Prior R2 nits:**
- IndexedDB path untested at unit level [FIXED] — fake-indexeddb suite covers open/put/get/delete + unavailable-storage.
- `openDb()` per-operation open+close [SKIPPED] — non-blocker; per-op lifecycle now proven by the fake-indexeddb test, so the pattern is at least verifiably correct.
- `overwriteExternalConflict` else-branch race [FIXED] — see Thread 1 path 2 above; fresh 409 propagates newer state.

**No regression of #2989 F1** (conflict-first drain). This PR does not touch `flushStudioPendingEdits`, `DomEditSaveQueue.drainError`, or the 10s TTL from that stack layer.

**CI:** `UNSTABLE` due to in-progress `Graphite mergeability_check` (downstack #2991-#2993 open) — expected for a mid-stack Graphite PR. Zero failed required checks; player-perf + preview-regression jobs still IN_PROGRESS at time of review.

— Review by Via

vanceingalls (approved):
Byte-clean re-stamp at unchanged head `9969887d` — GitHub dismissed the R3 approval when the base retargeted to main after #2989 merged, but the tree hash is identical to what I reviewed. All prior findings remain resolved (both inline threads FIXED, prior R2 nits addressed, #2989 F1 drain semantics untouched, `fake-indexeddb`-driven persistence contract, `If-Match`/`If-None-Match` overwrite atomicity). No delta since R3 — mechanical stamp only.

— Review by Via

vanceingalls (approved):
Byte-clean re-stamp at unchanged head `9969887d` — GitHub dismissed the R3 approval when the base retargeted to main after #2989 merged, but the tree hash is identical to what I reviewed. All prior findings remain resolved (both inline threads FIXED, prior R2 nits addressed, #2989 F1 drain semantics untouched, `fake-indexeddb`-driven persistence contract, `If-Match`/`If-None-Match` overwrite atomicity). No delta since R3 — mechanical stamp only.

— Review by Via

PR #2992, opened 2026-08-04
feat(studio): add external conflict recovery UI

miguel-heygen:
External-change stack 4/5. Adds review, copy, download, take external, overwrite, and retry recovery actions. 371 changed lines. Split from #2984.

### DIFF
--- a/packages/studio/src/components/ExternalFileConflictBanner.tsx
+++ b/packages/studio/src/components/ExternalFileConflictBanner.tsx
@@ -1,0 +1,255 @@
+import { useRef, useState } from "react";
+import type {
+  ExternalFileChangeBlockedState,
+  ExternalFileChangeCoordinatorHandle,
+} from "../hooks/useExternalFileChangeCoordinator";
+import { useDialogBehavior } from "./ui/useDialogBehavior";
+
+function errorMessage(error: unknown): string {
+  return error instanceof Error ? error.message : String(error);
+}
+
+function downloadText(filename: string, content: string): void {
+  const url = URL.createObjectURL(new Blob([content], { type: "text/html;charset=utf-8" }));
+  const anchor = document.createElement("a");
+  anchor.href = url;
+  anchor.download = filename;
+  anchor.click();
+  URL.revokeObjectURL(url);
+}
+
+function ConflictReview({
+  conflict,
+  onClose,
+}: {
+  conflict: Extract<ExternalFileChangeBlockedState, { status: "conflict" }>;
+  onClose: () => void;
+}) {
+  const containerRef = useRef<HTMLDivElement>(null);
+  useDialogBehavior({ open: true, onClose, containerRef });
+  const external = conflict.error.currentContent ?? "(The server did not return file contents.)";
+  const studio = conflict.error.attemptedContent;
+
+  return (
+    <div
+      className="fixed inset-0 z-[110] flex items-center justify-center bg-black/70 px-5 backdrop-blur-sm"
+      onClick={onClose}
+    >
+      <div
+        ref={containerRef}
+        role="dialog"
+        aria-modal="true"
+        aria-labelledby="external-conflict-title"
+        tabIndex={-1}
+        className="max-h-[88vh] w-full max-w-5xl overflow-auto rounded-xl border border-amber-400/30 bg-neutral-950 p-5 text-neutral-100 shadow-2xl outline-none"
+        onClick={(event) => event.stopPropagation()}
+      >
+        <div className="flex items-start justify-between gap-4">
+          <div>
+            <h2 id="external-conflict-title" className="text-base font-semibold">
+              Review both versions of {conflict.error.filePath}
+            </h2>
+            <p className="mt-1 text-xs text-neutral-400">
+              Reviewing or exporting does not change either version.
+            </p>
+          </div>
+          <button type="button" onClick={onClose} className="rounded px-2 py-1 text-neutral-400">
+            Close
+          </button>
+        </div>
+        <div className="mt-4 grid gap-4 md:grid-cols-2">
+          {[
+            { title: "File on disk", content: external, suffix: "external" },
+            { title: "Unsaved Studio version", content: studio, suffix: "studio" },
+          ].map((side) => (
+            <section key={side.suffix} aria-label={side.title}>
+              <div className="mb-2 flex items-center justify-between gap-2">
+                <h3 className="text-sm font-medium">{side.title}</h3>
+                <div className="flex gap-2">
+                  <button
+                    type="button"
+                    className="rounded border border-neutral-700 px-2 py-1 text-[11px]"
+                    onClick={() => void navigator.clipboard.writeText(side.content)}
+                  >
+                    Copy
+                  </button>
+                  <button
+                    type="button"
+                    className="rounded border border-neutral-700 px-2 py-1 text-[11px]"
+                    onClick={() =>
+                      downloadText(`${conflict.error.filePath}.${side.suffix}.html`, side.content)
+                    }
+                  >
+                    Download
+                  </button>
+                </div>
+              </div>
+              <textarea
+                readOnly
+                value={side.content}
+                className="h-80 w-full resize-y rounded border border-neutral-800 bg-neutral-900 p-3 font-mono text-[11px] leading-relaxed text-neutral-300"
+              />
+            </section>
+          ))}
+        </div>
+      </div>
+    </div>
+  );
+}
+
+function FailedDraftReview({
+  path,
+  content,
+  onClose,
+}: {
+  path: string;
+  content: string;
+  onClose: () => void;
+}) {
+  const containerRef = useRef<HTMLDivElement>(null);
+  useDialogBehavior({ open: true, onClose, containerRef });
+  return (
+    <div
+      className="fixed inset-0 z-[110] flex items-center justify-center bg-black/70 px-5 backdrop-blur-sm"
+      onClick={onClose}
+    >
+      <div
+        ref={containerRef}
+        role="dialog"
+        aria-modal="true"
+        aria-labelledby="failed-draft-title"
+        tabIndex={-1}
+        className="max-h-[88vh] w-full max-w-3xl overflow-auto rounded-xl border border-amber-400/30 bg-neutral-950 p-5 text-neutral-100 shadow-2xl outline-none"
+        onClick={(event) => event.stopPropagation()}
+      >
+        <div className="flex items-start justify-between gap-4">
+          <div>
+            <h2 id="failed-draft-title" className="text-base font-semibold">
+              Recover unsaved Studio draft for {path}
+            </h2>
+            <p className="mt-1 text-xs text-neutral-400">
+              Copy or download this draft before choosing to discard it.
+            </p>
+          </div>
+          <button type="button" onClick={onClose} className="rounded px-2 py-1 text-neutral-400">
+            Close
+          </button>
+        </div>
+        <div className="mt-4 flex justify-end gap-2">
+          <button
+            type="button"
+            className="rounded border border-neutral-700 px-2 py-1 text-[11px]"
+            onClick={() => void navigator.clipboard.writeText(content)}
+          >
+            Copy
+          </button>
+          <button
+            type="button"
+            className="rounded border border-neutral-700 px-2 py-1 text-[11px]"
+            onClick={() => downloadText(`${path}.studio.html`, content)}
+          >
+            Download
+          </button>
+        </div>
+        <textarea
+          readOnly
+          value={content}
+          className="mt-2 h-80 w-full resize-y rounded border border-neutral-800 bg-neutral-900 p-3 font-mono text-[11px] leading-relaxed text-neutral-300"
+        />
+      </div>
+    </div>
+  );
+}
+
+export function ExternalFileConflictBanner({
+  coordinator,
+}: {
+  coordinator: ExternalFileChangeCoordinatorHandle;
+}) {
+  const [reviewing, setReviewing] = useState(false);
+  const blocked = coordinator.blocked;
+  if (!blocked) return null;
+
+  const conflict = blocked.status === "conflict" ? blocked : null;
+  const failure = blocked.status === "failed" ? blocked : null;
+  return (
+    <>
+      <div
+        role="alert"
+        className="absolute left-1/2 top-14 z-[94] flex max-w-[calc(100vw-32px)] -translate-x-1/2 flex-wrap items-center gap-3 rounded-md border border-amber-400/30 bg-amber-950/95 px-4 py-2 text-[12px] font-medium text-amber-50 shadow-lg"
+      >
+        <span>
+          {conflict
+            ? `${conflict.error.filePath} changed outside Studio. Preview is paused so neither version is lost.`
+            : `Studio could not safely finish local saves: ${errorMessage(blocked.error)}. Preview is paused.`}
+        </span>
+        {conflict && (
+          <button type="button" onClick={() => setReviewing(true)} className="underline">
+            Review or export both
+          </button>
+        )}
+        {failure?.studioContent != null && (
+          <button type="button" onClick={() => setReviewing(true)} className="underline">
+            Review or export Studio draft
+          </button>
+        )}
+        {failure && !failure.recovered && (
+          <button type="button" onClick={() => void coordinator.retry()} className="underline">
+            Retry save
+          </button>
+        )}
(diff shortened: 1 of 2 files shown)


miguel-heygen:
* **#2992** <a href="https://app.graphite.com/github/pr/heygen-com/hyperframes/2992?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/heygen-com/hyperframes/2992?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

vanceingalls (approved):
### R1 review at `d56b21974` — APPROVE

Fresh R1 (no prior reviews on this PR). Reviewed as stack 4/5: recovery UI banner + review/copy/download/overwrite/retry surface for the `useExternalFileChangeCoordinator` handle introduced in #2991.

**Correctness / shape**
- `ExternalFileConflictBanner` consumes `ExternalFileChangeCoordinatorHandle` correctly. All four handle fields (`blocked`, `retry`, `useExternalFile`, `keepStudioFile`) exist on the type shipped in `useExternalFileChangeCoordinator.ts`. Nullability handled: `conflict.error.currentContent` has the `?? "(The server did not return file contents.)"` fallback; `failure.studioContent` is null-guarded before rendering the review button and dialog.
- Single owner semantics from #2991 preserved — this PR only reads from the handle and calls its methods; no independent state.
- Drain-echo-suppress-conflict-reload semantics untouched (banner is a leaf UI over the state machine).

**Tests are semantic, not presence-only**
- Test 3 explicitly asserts the anti-fake-retry invariant: `not.toContain("Retry save")` when `recovered=true`, and confirms the alternative overwrite button calls `keepStudioFile` after `window.confirm` returns true. Good coverage of the "post-remount, don't offer a stale retry" contract.
- Test 1 asserts both textareas render both full versions when reviewing conflict — matches the "no destructive default" copy invariant.
- Test 2 asserts studio-draft-only review works after a drain failure with `studioContent` set.

**Lifecycle / a11y**
- Uses the shared `useDialogBehavior` for both modals (Escape, focus trap, restore on close). Only one modal open at a time (via `reviewing` state), so no duplicate document-level keydown listeners.
- Element-identity keying in `ConflictReview`'s `.map()` uses `key={side.suffix}` — string-unique, not array index.
- No `useEffect` state-syncing; the component is purely derived from `coordinator.blocked`.
- No non-component exports from the `.tsx` file — `ConflictReview` / `FailedDraftReview` / `errorMessage` / `downloadText` are all module-local. Fast Refresh clean.

**Fallow**
- Banner has no `fallow-ignore-file unused-file` marker, but Fallow audit passes because the sibling `.test.tsx` imports it as a live consumer. Same shape as #2991's coordinator.

**Deferred parity**
- Grep at HEAD: no other `hf:file-change` listeners or conflict-banner surfaces outside this file + `useExternalFileChangeCoordinator`. Single-owner invariant preserved.
- Banner is not yet mounted anywhere (expected — mount lands in the 5/5 wire-up PR). `useDialogBehavior` still carries its `fallow-ignore-file unused-file` marker with the "consumers land in the shell/sidebar PRs later in this stack" comment — same drain still applies.

**Non-blocking nits (follow-up, do not block this merge)**
1. `downloadText` hardcodes `.html` extension and `text/html` MIME (`${filePath}.${suffix}.html`). If Studio ever edits non-HTML compositions, a user downloading `styles.css`'s versions gets `styles.css.external.html`. Consider deriving from the file extension. In practice Studio is HTML-first so this is cosmetic.
2. `void navigator.clipboard.writeText(...)` swallows rejections. Clipboard blocked in insecure contexts (or by browser policy) silently fails with no user feedback. A small toast/`Message.error` on `.catch` would surface it.
3. `document.createElement("a"); anchor.click()` without appending to DOM works in modern Chromium/WebKit but not Firefox in all cases. Appending + removing after click is the safer idiom.

CI green through Fallow / Player perf / regression / preview-regression at time of review; Typecheck / Lint / Build / integration still in progress but the surface (2 new files, +371/-0, all self-contained) is low-risk.

— Review by Via

packages/studio/src/components/ExternalFileConflictBanner.tsx
```diff
@@ -0,0 +1,255 @@
...
+            { title: "Unsaved Studio version", content: studio, suffix: "studio" },
+          ].map((side) => (
+            <section key={side.suffix} aria-label={side.title}>
+              <div className="mb-2 flex items-center justify-between gap-2">
+                <h3 className="text-sm font-medium">{side.title}</h3>
+                <div className="flex gap-2">
+                  <button
+                    type="button"
+                    className="rounded border border-neutral-700 px-2 py-1 text-[11px]"
+                    onClick={() => void navigator.clipboard.writeText(side.content)}
```
james-russo-rames-d-jusso, line 72: 🟡 **`void navigator.clipboard.writeText(side.content)` swallows rejections and provides no success feedback — this is the *primary* export path a user takes right before deciding to discard local edits.**

Same pattern at `:142` (failed-draft dialog) and both are user-critical moments. Two problems in one:

1. **Silent rejection.** `navigator.clipboard.writeText` rejects when:
   - Document is not focused (rare but happens if the copy is triggered from a modal after a focus-related bug)
   - Insecure context (non-HTTPS, non-localhost)
   - Permission denied (rare on `writeText`, but the user's browser policy or extension can deny)
   - The tab has lost user activation (some browsers require a recent gesture)
   
   `void` throws away the rejection. User clicks Copy, thinks it worked, pastes into an editor, gets... whatever they had before. Then discards Studio changes. Lost work.

2. **No success feedback.** Even when it works, the button shows no visual state change. Common pattern: swap the button label to "Copied!" for 2s, or add a checkmark icon. Without feedback the user often clicks twice, worried the first click missed.

**Recommend:**
```ts
const [copied, setCopied] = useState(false);
// ...
onClick={async () => {
  try {
    await navigator.clipboard.writeText(side.content);
    setCopied(true);
    setTimeout(() => setCopied(false), 2000);
  } catch (err) {
    // fallback: create a temporary textarea, select, execCommand('copy'),
    // or surface via showToast if that's available
    console.error("Copy failed:", err);
  }
}}
```

At minimum: catch the rejection and log/toast. Ideally: pair with success feedback.

— Rames D Jusso

packages/studio/src/components/ExternalFileConflictBanner.tsx
```diff
@@ -0,0 +1,255 @@
...
+          className="rounded border border-amber-200/30 px-2 py-1"
+        >
+          Discard Studio edits and reload file
+        </button>
+        {conflict && (
+          <button
+            type="button"
+            onClick={() => {
+              if (
+                window.confirm(
```
james-russo-rames-d-jusso, line 213: 🟡 **`window.confirm(...)` for the two overwrite paths (`:213-215` and `:230-232`) is a blocking synchronous browser dialog — the file already has a well-designed ARIA `role="dialog"` + `useDialogBehavior` pattern in *this same file* (used for `ConflictReview` and `FailedDraftReview`).**

Problems with `window.confirm`:
- Blocking synchronous — freezes the event loop, main thread, React batching. Bad UX in a modern React app.
- Not styleable — breaks the app's amber/red visual language.
- Not focus-managed the same way — `useDialogBehavior` traps focus, restores it, and handles Escape; `window.confirm` bypasses that whole contract.
- Screen readers announce it differently than a role=dialog modal.

**Recommend:** refactor to a `ConfirmDialog` component using the same `useDialogBehavior({ open, onClose, containerRef })` pattern already in this file. Two dialogs already use it (`ConflictReview` at `:29`, `FailedDraftReview` at `:110`); adding a third for confirms is trivial and cleaner than the browser alert.

The two current messages are:
- `"Overwrite the externally changed file with the Studio version? The server will preserve its normal backup before writing."` (verified — `packages/studio-server/src/helpers/backupJournal.ts` + `packages/studio-server/src/routes/files.ts:414` do call `snapshotBeforeWrite` on the write path, so this messaging is factually correct ✓)
- `"Overwrite the file with the recovered Studio draft? The current file will be preserved by the server's normal backup before writing."`

Both are red-flag destructive; both would benefit from a proper confirmation modal that matches the rest of the recovery UI.

— Rames D Jusso

packages/studio/src/components/ExternalFileConflictBanner.tsx
```diff
@@ -0,0 +1,255 @@
...
+function errorMessage(error: unknown): string {
+  return error instanceof Error ? error.message : String(error);
+}
+
+function downloadText(filename: string, content: string): void {
+  const url = URL.createObjectURL(new Blob([content], { type: "text/html;charset=utf-8" }));
+  const anchor = document.createElement("a");
+  anchor.href = url;
+  anchor.download = filename;
+  anchor.click();
```
james-russo-rames-d-jusso, line 17: 🟡 **`downloadText` uses a detached-anchor click + synchronous `URL.revokeObjectURL`. Works in modern Chrome/Firefox but has cross-browser + timing fragility worth naming.**

```ts
const url = URL.createObjectURL(new Blob([content], { type: "text/html;charset=utf-8" }));
const anchor = document.createElement("a");
anchor.href = url;
anchor.download = filename;
anchor.click();
URL.revokeObjectURL(url);
```

Two subtle points:

1. **Anchor is not attached to `document.body` before `click()`.** Historically some browsers (older Firefox, some Safari versions) required the anchor to be in the DOM for `.click()` to trigger a download. Current Chrome/Firefox handle detached anchors fine, but the safer pattern still `appendChild` → `click()` → `remove()`.

2. **`URL.revokeObjectURL(url)` fires synchronously *immediately after* `.click()`.** The browser initiates the download during the click event, so this generally works — but the download-initiation-vs-revoke ordering isn't spec-guaranteed. In slow-io scenarios (large blobs, low-memory devices), synchronously revoking can cancel the download mid-request. Content here is HTML files that can be 100KB+ from a real Studio project, which is exactly the kind of size where this matters.

**Safer pattern:**
```ts
function downloadText(filename: string, content: string): void {
  const url = URL.createObjectURL(new Blob([content], { type: "text/html;charset=utf-8" }));
  const anchor = document.createElement("a");
  anchor.href = url;
  anchor.download = filename;
  document.body.append(anchor);
  anchor.click();
  anchor.remove();
  // Give the download a full task to start before revoking
  setTimeout(() => URL.revokeObjectURL(url), 0);
}
```

Non-blocking — modern Chromium likely handles the current code fine, and Studio's user base is Chromium-heavy — but the download IS a user-critical path (they clicked Download right before discarding local edits). Worth hardening.

— Rames D Jusso

james-russo-rames-d-jusso (commented):
_Reviewed at `d56b2197`. External-change stack 4/5 — the UI for the `#2991` coordinator. Two new files, 371 additions: `ExternalFileConflictBanner.tsx` (256 lines: banner + `ConflictReview` dialog + `FailedDraftReview` dialog) and `ExternalFileConflictBanner.test.tsx` (117 lines, 3 cases)._

Overall shape is good — clean separation between banner (persistent alert with action buttons) and modal review dialogs (side-by-side diff for conflict, single-panel review for failed drafts). Correctly gates destructive actions to the right states:
- `Retry save` shown only for `failed && !recovered` — matches `#2991`'s `retry()` guard.
- `Overwrite file with recovered Studio draft` shown only for `failed?.recovered && studioContent != null` — matches the `keepStudioFile` gate I flagged in `#2991` R1. **UI-mitigates the \"silent no-op on live-failed\" concern.**
- `ConflictReview` shows both external and Studio content side-by-side; `FailedDraftReview` shows only the recovered draft (external not surfaced for the recovered-failed case — simplification of the stale-content concern from `#2991` R1).

Three inline concerns, all 🟡 non-blocking. UX correctness — no security or state-machine gaps.

**Verified end-to-end:**

- ✅ **State gating**: `conflict && …` at `:186`, `:208`, `:243` correctly bounds conflict-only UI; `failure?.recovered && …` at `:225`, `:246` correctly bounds recovered-failed-only UI. `Retry save` uses `!failure.recovered`. No mis-labeled buttons in the wrong states.
- ✅ **Server backup claim in the confirm text is TRUE** — `packages/studio-server/src/helpers/backupJournal.ts` + `packages/studio-server/src/routes/files.ts:367, :414, :1241` call `snapshotBeforeWrite` on write paths, snapshots to `.hyperframes/backup/{timestamp}-{key}`. \"The server will preserve its normal backup\" is factually correct.\n- ✅ **ARIA/dialog behavior**: `role=\"dialog\"` + `aria-modal=\"true\"` + `aria-labelledby` on both review dialogs; `useDialogBehavior` handles focus-trap + Escape + restore-focus. Banner itself uses `role=\"alert\"`. All correct.\n- ✅ **`errorMessage` unwrap** at `:8-10` handles the `blocked.error: unknown` type cleanly (Error instance → `.message`, else String coercion). No unsafe cast.\n- ✅ **Test coverage of gate logic**: 3 tests cover the three main state shapes (live conflict / live failed / recovered failed). Regression at `:105` proves \"no Retry button on recovered\" — the load-bearing gate that mirrors `#2991`'s hook contract.\n- ✅ CI: 30+ green checks, focused suite Magi cited (\"UI tests + Studio typecheck pass\").\n\n**Body-only nits:**\n\n- **Modal backdrop click can close on drag-select-release-outside.** `ConflictReview:35-36` and `FailedDraftReview:112-114` both use `<div onClick={onClose}>` for the backdrop. A user drag-selecting text inside the read-only textarea and releasing the mouse outside the modal's inner container triggers backdrop `onClick` — modal closes, text selection lost. Standard fix: switch backdrop to `onMouseDown={onClose}` (or track mousedown target and only close if mousedown+mouseup were both on backdrop). Load-bearing for the review dialog specifically because copying via drag-select is a common flow.\n\n- **Copy button has no success/error state.** Same locations as inline finding #1 (`:72`, `:142`): after clicking Copy the button visually doesn't change. Users often click multiple times, unsure whether it worked. Recommend a 2s \"Copied!\" swap on success, plus toast-or-console on failure (see inline).\n\n- **`<textarea readOnly value={content}>` with large HTML content.** Studio project HTML files can be 100KB+. `<textarea>` rendering of that much text in a `h-80` (320px) box is slow on first render, and the browser layouts every line. Non-blocking for typical files; consider a virtualized viewer OR a `if (content.length > 100_000) truncate` warning for the rare large-project case. `textarea` also disables syntax highlighting, which for HTML review is a real loss. A `<pre><code>` block with monospace styling + optional syntax highlighting (later PR) would be a better long-term shape.\n\n- **Test coverage gaps.**\n  - `retry()` click for `failed && !recovered` not exercised — only asserted button *exists* by-negation in test 3.\n  - `useExternalFile()` click (`Discard Studio edits and reload file`) not tested.\n  - `keepStudioFile()` on LIVE conflict not tested (only on recovered failed via test 3).\n  - `window.confirm` returning FALSE (user cancels overwrite) not tested — the current test always mocks `mockReturnValue(true)`.\n  - `downloadText` mechanic not directly asserted — no test spies on `URL.createObjectURL` or asserts the anchor is created.\n  - `navigator.clipboard.writeText` not asserted called (or that rejection is handled).\n  - Focus trap + ESC key + return-focus from `useDialogBehavior` untested here (assumed tested at the hook level).\n  - `errorMessage` with a non-Error thrown value not exercised.\n\n- **`Retry save` label matches `#2991`'s hook semantics only if `resetSaveQueues` preserves the failed candidate.** Wire-up dependency: in `#2991`, `retry()` calls `resetSaveQueues?.() → processChange(payload, true) → drainPendingChanges()`. If `resetSaveQueues` clears the queue *including* the failed candidate, `drainPendingChanges` returns clean, blocked clears, and the user's local draft is silently orphaned in the editor (never re-persisted). If `resetSaveQueues` clears only the *error state* and preserves the pending write, retry works as labeled. **Verify in `#2993` wire-up** — the button label \"Retry save\" is a load-bearing promise and depends entirely on how the wire-up defines `resetSaveQueues` semantics. Recommend `#2993` review pays attention to this specifically.\n\n- **`z-[110]` (dialogs) vs `z-[94]` (banner) magic numbers.** Studio likely has a z-index scale somewhere; if so, use those tokens. If not, this PR is a good moment to name them (`Z_MODAL = 110`, `Z_TOAST = 94`).\n\nReady from where I sit once the three inline concerns are considered. The clipboard-no-feedback (`:72`) is the load-bearing UX one for a critical export moment; the `window.confirm` (`:213`) is the pattern-inconsistency one; the `downloadText` (`:17`) is the cross-browser hardening one. Leaving as COMMENTED.\n\n— _[Review by Rames D Jusso](https://github.com/james-russo-rames-d-jusso)_\n

PR #2983, opened 2026-08-04
fix(studio): resolve sibling asset paths in sub-composition previews

miguel-heygen:
Fixes #2956

## What

A composition in a subdirectory that references a **sibling** file (`<link rel="stylesheet" href="_shared.css">`, not `../_shared.css`) is now resolved against the composition's own directory when building the standalone sub-composition preview page.

## Why

The preview page borrows the project-root `<base href="/api/projects/:id/preview/">`, but the path rewriter only rewrote `../`-prefixed paths. So `design/styleframes/frame-01.html` referencing `_shared.css` was served unrewritten and the browser requested `/preview/_shared.css` → **404**.

With its stylesheet missing, the frame renders unstyled: `body` has no background, and the thumbnail generator's transparent-body fallback paints it `#1c2028`. Result: dark navy thumbnail with unreadable dark text on every styleframe in the Board view.

The report attributed this to project scale (~21 sibling files). It is not scale-related: a 2-file project reproduces identically, and the same file moved to the project root renders correctly. The trigger is **composition-in-a-subdirectory + relative sibling asset ref**.

Reproduced before the fix (single `curl` against the thumbnail endpoint, plus a direct headless capture of the preview URL):

```
HTTP 404 http://localhost:5190/api/projects/big/preview/_shared.css
body bg: rgb(28, 32, 40)
```

## How

`resolvePreviewAssetPath` in `packages/studio-server/src/helpers/subComposition.ts`, applied through the single rewrite pass all three dispatch branches (template / full-doc / fragment) already share, so `src`, `href`, inline `style` urls, and `<style>` blocks all get the same rule:

1. `../` paths keep resolving against the composition dir (unchanged, shared with the producer's inliner so preview and render agree).
2. Any other relative path is re-pointed at the composition's directory **only when that sibling file exists on disk**.

The disk check is what keeps the two conventions apart: registry blocks are installed into a subdirectory but reference project-root assets (`assets/logo.png`), which are already correct under the root base and have no sibling on disk, so they are left untouched.

Not changed: the `#1c2028` transparent-body fallback in the thumbnail generator. It is correct for genuinely transparent compositions; the illegibility was a downstream symptom of the 404.

## Test plan

- [x] Unit tests added/updated — two tests in `subComposition.test.ts`: a red-first regression guard for the sibling `<link>` / `` / `url()` case, and a guard that project-root-relative refs with no sibling on disk stay untouched.
[...]

Issue #2956, opened 2026-08-02
Studio thumbnail endpoint renders dark/illegible background on projects with many sibling styleframe files


reachadam:
## Summary

Studio's live thumbnail-generation endpoint (`GET /api/projects/:project/thumbnail/:path?t=...`) renders styleframe/composition HTML with a broken dark background and illegible (near-invisible) text, even though the same HTML renders correctly when captured directly with the same headless Chrome binary the CLI vendors (bypassing this specific endpoint). This makes the Storyboard/Board view unusable for reviewing designs on any project once it reaches a certain scale.

## What's ruled out

- Not a hyperframes version issue — reproduces identically on 0.7.76, 0.7.87, and the latest 0.7.89.
- Not GPU/rendering-related — reproduces with both default (auto) GPU and `--no-browser-gpu` (SwiftShader).
- Not caching — reproduces on a freshly-started server, on a brand-new port, with `.thumbnails` cache cleared.
- Not CSS-content-specific — a page using only inline `<style>`, or an external stylesheet with CSS custom properties and `color-scheme: light only`, renders correctly through this same endpoint **in a small project**.
- **It is scale-triggered, not content-triggered.** The exact same trivial HTML (plain white/lavender gradient background, one text element) renders correctly via this endpoint in a project with 1-2 sibling files, and renders with the broken dark background in a project with ~21+ sibling files in the same directory — with zero difference in the file's own content or CSS.

## Reproduction (fully generic, no project-specific content)

1. Create a project directory:
   ```
   hyperframes.json:
   {"$schema": "https://hyperframes.heygen.com/schema/hyperframes.json"}

   index.html: (minimal valid composition scaffold, 1920x1080, empty #root)

   design/styleframes/_shared.css:
   :root { --ink:#18181b; --lav2:#dad0ff; }
   * { margin:0; padding:0; box-sizing:border-box; }
   html, body { width:1920px; height:1080px; overflow:hidden; }
   body { font-family: system-ui, sans-serif; color: var(--ink); }
   .stage { position:absolute; inset:0; background: linear-gradient(#ffffff 0%, #ffffff 52%, var(--lav2) 100%); }
   .txt { position:absolute; left:96px; top:140px; font-size:56px; font-weight:700; }
   ```
2. Create 21 near-identical sibling files `design/styleframes/frame-01.html` … `frame-21.html`, each:
   ```html
   <!doctype html>
   <html lang="en">
[...]

PR #2994, opened 2026-08-04
fix(core): resolve sub-composition sibling asset paths in render and asset discovery

miguel-heygen:
Follow-up to #2983, which fixed the Studio preview half of #2956. Same root cause, the rest of the surface.

## What

`rewriteAssetPath` takes an optional `assetExists` probe. A plain relative ref authored inside a sub-composition (`_shared.css`, `clip.mp4`) resolves against the composition's own directory when that sibling exists on disk; a project-root ref with no sibling (the registry's `assets/logo.png` convention) stays as authored. Callers that can see the filesystem supply the probe, so the module keeps no `node:fs` import.

It also fixes a second, separate defect in the inliner: `<head>` `<link>` hrefs and external `<script src>` are hoisted into the root document but never went through the rewrite **at all** — so even the documented `../` form escaped the project and 404'd at render time.

## Why

#2983 fixed `buildSubCompositionHtml` only. The producer/bundler inliner shares the same resolver and had the same defect, so a project whose styleframes lived in a subdirectory rendered unstyled in the actual MP4, not just in thumbnails.

Reproduced on a generated project (`design/styleframes/frame-01.html` referencing sibling `_shared.css` + `frame-01.png`, and `../shared/theme.css`). Before, the inlined root document emitted:

```
<link href="_shared.css" rel="stylesheet">
<link href="../shared/theme.css" rel="stylesheet">
url("frame-01.png")
src="frame-01.png"
```

All four resolve against the project root: three 404, and the `../` one escapes the project entirely.

The same misresolution silently degraded the asset-discovery utilities, which answer "where is this file on disk" for a ref in a sub-composition:

- **HEVC preview lint** skips any video it cannot resolve, so a real HEVC sibling was never reported — and when a same-named file existed at the project root, it probed the **wrong file**.
- **Project lint** asset scans, **publish proxy baking**, and **media-treatment** source resolution have the same resolution step.

## How

- `packages/parsers/src/rewriteSubCompPaths.ts` — the `AssetExists` probe, threaded through `rewriteAssetPaths` / `rewriteInlineStyleAssetUrls` / `rewriteCssAssetUrls`. Query strings and hashes survive the rewrite (`frame.png?v=2`).
- `packages/core/src/compiler/inlineSubCompositions.ts` — accepts the probe; hoisted `<link href>` and external script `src` now go through the same resolver as the body pass.
- Probe supplied by: the preview bundler, the producer compiler, the studio preview builder, the HEVC preview lint, the project lint's asset scans, publish proxy baking, media-treatment.

Behavior is unchanged for every caller that does not pass a probe.

## Test plan

- [x] Unit tests added/updated, red first: 5 in `rewriteSubCompPaths.test.ts` (sibling, query/hash, root-relative left alone, `../` unaffected, no-probe default), 3 in `inlineSubCompositions.test.ts` (hoisted links, hoisted script srcs, markup + hoisted CSS + inline styles), 3 new in `hevcPreviewLint.test.ts` (sibling resolved, sibling preferred over a same-named root file, root-relative still resolved).
- [x] Manual testing performed — same generated project: the inlined root document now emits `design/styleframes/_shared.css`, `design/shared/theme.css`, and `design/styleframes/frame-01.png`; the bundler preview renders styled, and a real `hyperframes render` produces a correctly styled MP4 frame (was unstyled).
- [x] Suites green: parsers 908, core 1434, studio-server 402, lint 514, producer unit lane 556. The 2 pre-existing `transcribe` failures in the CLI package reproduce on main and are untouched here.
- [ ] Documentation updated — n/a

## Not covered

`lintProject`'s missing-asset scan never reaches sub-composition markup in the shape I tested (a genuinely missing ``/`<audio>` in a sub-composition is not reported either). That is a separate gap, not this bug; the probe is wired into its resolution step regardless, so it is correct if and when that scan reaches those sources.

### DIFF
--- a/packages/cli/src/commands/media-treatment.ts
+++ b/packages/cli/src/commands/media-treatment.ts
@@ -1,5 +1,5 @@
 import { existsSync, readFileSync, writeFileSync } from "node:fs";
-import { relative, resolve } from "node:path";
+import { join, relative, resolve } from "node:path";
 import {
   HF_COLOR_GRADING_ATTR,
   getHfColorGradingCapabilities,
@@ -502,7 +502,7 @@ export function resolveMediaTreatmentSource(
   if (!cleanSource) throw new Error("Selected media has no analyzable local src");
   const projectRelative = cleanSource.startsWith("/")
     ? cleanSource
-    : rewriteAssetPath(compositionFile, cleanSource);
+    : rewriteAssetPath(compositionFile, cleanSource, (path) => existsSync(join(projectDir, path)));
   const asset = resolveExistingLocalAsset(projectDir, projectRelative);
   if (!asset) throw new Error(`Media file not found: ${source}`);
   return asset.resolved;
--- a/packages/cli/src/utils/publishProxyBake.ts
+++ b/packages/cli/src/utils/publishProxyBake.ts
@@ -25,6 +25,7 @@
  * structured manifest rather than silently shipping an unplayable asset.
  */
 
+import { existsSync } from "node:fs";
 import { readFile } from "node:fs/promises";
 import { basename, dirname, resolve } from "node:path";
 import { parseHTML } from "linkedom";
@@ -147,7 +148,9 @@ export async function bakeMediaProxies(
         // (rewriteAssetPath to root-relative, then decodeUrlPathVariants via
         // resolveLocalAssetCandidates) so percent-encoded and root-absolute
         // srcs match the map keys the scan produced.
-        const rootRelativeSrc = rewriteAssetPath(entryPath, cleaned);
+        const rootRelativeSrc = rewriteAssetPath(entryPath, cleaned, (path) =>
+          existsSync(resolve(absProjectDir, path)),
+        );
         for (const candidate of resolveLocalAssetCandidates(absProjectDir, rootRelativeSrc)) {
           const archivePath = proxyByAbsolutePath.get(candidate);
           if (archivePath) return archivePath;
--- a/packages/core/src/compiler/htmlBundler.ts
+++ b/packages/core/src/compiler/htmlBundler.ts
@@ -898,6 +898,13 @@ export async function bundleToSingleHtml(
     parseHtml: parseHTMLContent,
     hostIdentityMap: hostIdentityByElement,
     rewriteInlineStyles: true,
+    // A sub-composition's SIBLING assets (`_shared.css` next to it) must be
+    // re-pointed at its own directory when its content moves to the root
+    // document; project-root refs with no such sibling stay as authored.
+    assetExists: (path: string) => {
+      const resolved = resolveEntryPath(path);
+      return resolved !== null && existsSync(resolved);
+    },
     flattenInnerRoot: prepareFlattenedInnerRoot,
     readVariableDefaults: readDeclaredDefaults,
     parseHostVariables: parseHostVariableValues,
--- a/packages/core/src/compiler/inlineSubCompositions.test.ts
+++ b/packages/core/src/compiler/inlineSubCompositions.test.ts
@@ -553,3 +553,71 @@ describe("inlineSubCompositions – recursive host discovery", () => {
     ]);
   });
 });
+
+describe("inlineSubCompositions – sub-composition asset paths", () => {
+  // Every asset ref a sub-composition in a subdirectory carries has to be
+  // re-pointed when its content moves into the project-root document. Hoisted
+  // <head> <link>/<script src> used to bypass the rewrite entirely (so even the
+  // documented `../` form escaped the project), and sibling refs (`_shared.css`)
+  // silently 404'd. Both render the frame unstyled.
+  const SUB_COMP = `<!doctype html>
+<html><head>
+  <link rel="stylesheet" href="_shared.css">
+  <link rel="stylesheet" href="../shared/theme.css">
+  <script src="helper.js"></script>
+  <style>.badge { background-image: url("frame.png"); }</style>
+</head><body>
+  <div data-composition-id="frame" data-width="1920" data-height="1080">
+    <img src="frame.png" alt="">
+    <div style="background-image: url('frame.png')"></div>
+  </div>
+</body></html>`;
+
+  const PROJECT_FILES = [
+    "design/styleframes/_shared.css",
+    "design/styleframes/frame.png",
+    "design/styleframes/helper.js",
+    "design/shared/theme.css",
+  ];
+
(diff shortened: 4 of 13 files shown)


vanceingalls (approved):
**Verdict: APPROVE** at `48a04f1b87`.

Two-defect fix, consolidation win, thorough tests. Clean follow-up to #2983 — same discriminator (`assetExists` probe) now shared by every asset-resolution site (preview + render + inliner + bundler + HEVC lint + project lint + publish-proxy-bake + media-treatment). No blockers.

## Defect 1 — Sibling resolution now covers render + discovery (extended from #2983's thumbnail path)

`rewriteAssetPath(compSrcPath, relativePath, assetExists?)` in `parsers/rewriteSubCompPaths.ts:79-98` adds the probe. Semantics:

1. Non-relative URL (absolute / data: / protocol) → unchanged.
2. `compDir === "." || !compDir` (composition at project root) → unchanged.
3. `../` or `..` → route through existing `join + resolve` (parent-relative, unchanged).
4. Plain relative WITHOUT probe → return unchanged (**backward-compat: no caller breakage**).
5. Plain relative WITH probe: split path/suffix via `splitPathSuffix`, compute `sibling = resolve("/", join(compDir, filePart)).slice(1)`, check `assetExists(sibling)`. If true → return `sibling + suffix` (query/hash preserved). Else → return unchanged.

The disk-check-as-discriminator moves from studio-server-local (#2983) into shared parsers. Every caller now uses the SAME resolver, so preview + render + discovery agree on which convention (sibling vs project-root) a given ref means — closing the studio-server-vs-producer divergence risk I raised in the #2983 review.

**Test coverage** at `rewriteSubCompPaths.test.ts:57-101`:
- Sibling resolves against comp dir ✓
- Query/hash preserved (`frame.png?v=2` → `design/styleframes/frame.png?v=2`) ✓
- Project-root-relative left alone when no sibling (`assets/logo.png` from `blocks/hero.html`) ✓
- `../` still routes through the existing `needsRewrite` path without consulting the probe ✓
- No-op without probe (backward compat) ✓

## Defect 2 — Hoisted `<link href>` / external `<script src>` were bypassing the rewriter ENTIRELY

`inlineSubCompositions.ts:302-320` — head-hoisted `<script src>` and `<link href>` extraction previously used `(s.getAttribute("src") || "").trim()` with NO rewrite. Even the documented `../` form was being pushed as-authored → escaped the project → 404 at render time. Fix: `resolveSubAssetPath = raw => rewriteAssetPath(src, raw, assetExists)` wraps the extraction so every hoisted attribute goes through the same resolver as the body-attribute pass.

The content-region `<script>` extraction at `:335` got the same fix.

Test coverage at `inlineSubCompositions.test.ts:576-618` pins all three cases (hoisted links, hoisted scripts, markup + hoisted CSS + inline styles) — and specifically asserts `../shared/theme.css` becomes `design/shared/theme.css` (the previously-escaping case), not just the sibling case.

## Consolidation

`studio-server/subComposition.ts` loses 60 lines: removes the local `resolvePreviewAssetPath` and `rewriteCssUrls` helpers, delegates to shared `rewriteAssetPaths` / `rewriteInlineStyleAssetUrls` / `rewriteCssAssetUrls` from `@hyperframes/core`. One resolver, uniform behavior across preview / render / discovery — this is the "same root cause, the rest of the surface" outcome from the PR body.

## HEVC lint sibling-vs-same-named-root — the strongest test

`hevcPreviewLint.test.ts:33-45`:
```
const dir = makeProject(["clip.mp4", "design/styleframes/clip.mp4"]);
collectLocalVideoCandidates(dir, [{ html: `<video src="clip.mp4"></video>`, compSrcPath: "design/styleframes/frame-01.html" }]);
expect([...candidates.keys()]).toEqual([join(dir, "design/styleframes/clip.mp4")]);
```

Both a sibling AND a same-named project-root file exist. Pre-fix behavior would probe the wrong file (project root). Post-fix correctly prefers the sibling. This is the "wrong-file" case the PR body calls out, and the test pins the fix.

## Every caller consistently wires the probe

Verified `existsSync(join(projectDir, path))` (or equivalent `resolve`) at each site:
- `producer/htmlCompiler.ts:971` — `existsSync(resolve(projectDir, path))`
- `studio-server/subComposition.ts:38` — `existsSync(join(projectDir, path))`
- `core/htmlBundler.ts:902-905` — `resolveEntryPath(path) !== null && existsSync(resolved)` (bundler-side path resolution)
- `cli/publishProxyBake.ts:149-150` — `existsSync(resolve(absProjectDir, path))`
- `cli/media-treatment.ts:502` — `existsSync(join(projectDir, path))`
- `lint/hevcPreviewLint.ts:95` — `existsSync(join(projectDir, path))`
- `lint/project.ts:133/287/335` — three sites, all consistent

Callers that CANNOT see the filesystem (browser-facing) simply omit the probe → backward-compat behavior.

## Non-blockers

1. *`existsSync` per attribute per composition.* For N sub-comps × M asset refs, this is N×M sync stat calls. For preview/thumbnail path this is fine (already tolerated in #2983). For render + discovery paths that touch large projects, marginal per-render cost — probably in the low-ms range and not a hot loop. Same non-blocker I raised on #2983.

2. *Windows path separator handling.* The shared implementation uses `node:path`'s native `join` / `dirname` / `resolve` (not `path/posix`). On Windows, `join` produces `\\`-separated strings. `existsSync` accepts either separator on Windows so the probe still returns true; the concern is only if the rewritten path (`sibling + suffix`) with backslashes gets written into HTML — some browsers/renderers might not handle backslashes cleanly. Pacific's dev/prod matrix (Mac dev + Linux prod) avoids this in practice. Non-blocker; noting for anyone touching this on Windows.

3. *`splitPathSuffix` matches the first `?` OR `#`.* If `foo.png#frag?v=2` (fragment before query — non-standard but possible), suffix would be `#frag?v=2`. Correctly preserved as one blob. Not a concern; noting for completeness.

4. *"Not covered" section acknowledges `lintProject`'s missing-asset scan doesn't reach sub-composition markup.* Separate pre-existing gap; probe is wired regardless so it's correct if/when that scan reaches those sources. Author's own callout — not this PR's regression.

## Test-plan checkboxes match my read

Author's test plan: 5 parser tests, 3 inliner tests, 3 HEVC lint tests. Suites green (parsers 908, core 1434, studio-server 402, lint 514, producer unit 556). The 2 pre-existing transcribe failures in CLI are unrelated. CI: 41 passing, 11 running, 0 failing at time of review.

— Review by Via

PR #2995, opened 2026-08-04
test(studio): add a dense project fixture backed by real reused media

miguel-heygen:
First of a sequence splitting a large performance branch into reviewable pieces. This one is pure test infrastructure — **0 logic lines, 618 test lines** — and nothing else in the sequence depends on it except the runtime-cost gate.

## What it adds

A deterministic on-disk project generator that writes a dense project into `data/projects/`, and — the point of the second commit — backs it with real media: a tiny PNG, WAV, and MP4 emitted to disk and referenced by many clips.

## Why real media matters

A fixture of 3,000 empty divs measures manifest derivation and nothing else. The behaviour that actually costs users time on real projects is **media request amplification** — the same source fetched once per referencing clip.

This fixture reproduces that: **4.31× requests per distinct media source, against 4.27× measured on a real staged project.** A `MIN_REFERENCES_PER_SOURCE` guard is what produces it; without that guard the generator emits a 1:1 reference pattern and the amplification disappears.

## Reviewing this

Run the generator and look at the output tree — the assertion that matters is that each media source is referenced by at least the enforced minimum number of clips, and that two runs with the same parameters produce byte-identical output.

## Verification

`packages/studio` suite green (3,285 tests). No product code changed.


miguel-heygen:
Closing this unused stand-alone load-fixture branch. Its only dependency is the old runtime-cost gate, and current timeline CI already has an exercised fixture. Any fresh media optimization can bring a small real-source reproduction tailored to the behavior being changed.

PR #2996, opened 2026-08-04
test(studio): gate idle CPU, scrub cost, request volume and switch retention

miguel-heygen:
Gates what a project costs while it is **merely open** — a thing open-time measurement cannot see, because open time is one number at one instant.

**238 logic lines, 1,120 test lines.** The harness is the deliverable here, which is why the raw diff is large and the decision surface is small.

> Stacked on the fixture PR (hard dependency — the gate drives the fixture it writes). Please do not delete that branch while this is open.

## The four windows

| Window | What it asserts |
|---|---|
| idle | CPU as a fraction of a core, sampled past quiescence |
| scrub | per-frame work during a scrub |
| open | media requests per distinct source |
| switch | heap retained across a project switch |

## Two things that decide whether this gate means anything

**Idle CPU is sampled past a quiescence condition, not a fixed sleep.** The first 10s window reads ~60% of a core while the thumbnail-extraction backlog drains, then settles flat at 16-17% with exactly 1,200 style recalculations per 10s. A fixed sleep measures the transient, and the gate asserts nothing useful. There is a test proving the detector does **not** settle during the backlog window — that is the guard against the whole gate being decorative.

**One commit here fixes a defect in the gate itself.** The request-volume window originally counted ranged reads as duplication, and 39 of 40 media requests carry `Range` headers. It reported a 4.63× regression that did not exist. `count logical media fetches, not ranged reads` is the fix: N ranged reads of one URL count as one logical fetch. It is kept as its own commit rather than squashed, because a gate that once lied about a regression should show that in its history.

## Notes for review

- Budgets live in TypeScript and are imported by the gate, so the app and the gate read one owner rather than two drifting copies.
- A window that cannot run **fails**; it never falls back to a partial or synthesized number.
- Adds one CI job; does not modify existing ones.

## Verification

`packages/studio` suite green (3,299 tests).


miguel-heygen:
Closing the old standalone runtime-cost gate. Current performance/lifecycle code and existing timeline CI have moved on, and this harness needs corrected project-readiness and ranged-fetch attribution before its numbers are trustworthy. Future focused performance fixes should carry measurements for their actual current workload.

PR #2998, opened 2026-08-04
perf(studio,core): stop the overlay loops and the paused transport from running when nothing moves

miguel-heygen:
Three render loops that ran continuously regardless of whether anything was moving.

**Measured: idle animation frames 303/s → 1/s.**

## The heaviest review in this sequence — 713 logic lines

I tried to split it and could not do so honestly. The overlay-loop change and the paused-transport change both modify `SnapGuideOverlay` and `useDomEditCompositionRect`, and unlike the other overlaps in this sequence their hunks genuinely **interleave** rather than sitting in separate regions. The natural cut line falls between "the loop" and "what the loop reads", which splits an implementation from its own consumers — a worse review than one large PR.

**The three loops are independent. Read them in any order:**

1. **Canvas overlay loops** — park after a run of idle frames, resume on the next mutation.
2. **Motion-path loop** — stops re-running when no tracked node has moved.
3. **Paused transport** — the subtle one. See below.

## The one behavioural change

The transport previously ticked on `requestAnimationFrame` unconditionally. It now ticks on rAF **while playing** and on a 500ms timer **while paused**. Playback cadence is untouched; only the paused state changes.

This is the only change of the three that alters runtime behaviour rather than suppressing repaints, so it carries the most test weight: a paused runtime that receives a seek must still update within one timer period, and the rebind policy must not attempt a periodic bind while playing with a captured timeline.

## What this does not claim

The idle-frame number is measured with the runtime-cost gate, which is a separate PR in this sequence. Without it, you can still verify the behaviour from the tests; you cannot reproduce the 303 → 1 figure from this PR alone.

## Verification

`packages/core` green (1,442 tests), `packages/studio` green (3,297 tests).

### DIFF
--- a/packages/core/src/runtime/init.pausedTransport.test.ts
+++ b/packages/core/src/runtime/init.pausedTransport.test.ts
@@ -1,0 +1,316 @@
+// fallow-ignore-file code-duplication
+// (the timeline/rAF harness mirrors init.test.ts on purpose — these tests pin
+// the paused-state transport contract and must not move when that file does.)
+/**
+ * What the transport loop owes a PAUSED runtime.
+ *
+ * Idle Studio burned a flat ~120 style recalcs/second because `transportTick`
+ * rescheduled itself every frame regardless of play state and re-seeked the
+ * timeline on every one of those frames. The frame loop can stop while paused,
+ * but three of its passengers cannot: they are the only reason a composition
+ * that registers its timeline late, or is edited while parked, ever converges.
+ *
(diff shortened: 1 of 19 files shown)


miguel-heygen:
Closing the stale broad idle-loop rewrite. Current main has evolved paused-edit and render-seek behavior; the old frame-count result is insufficient to justify replacing those schedules and overlay invalidation paths. Any future optimization should start from a fresh current-main profile and a smaller scope.

PR #3000, opened 2026-08-04
perf(runtime): fetch and sync the playhead neighbourhood, not the whole composition

miguel-heygen:
The runtime set `preload="auto"` and called `load()` on **every** media element in a composition, and re-synced all of them on every tick. Both now scale with a window around the playhead rather than with the size of the document.

**Measured: scrub DOM scans 156 → 9.**

## Commit order is the point

The two characterization commits come first and change no behaviour — they pin what media buffering and active-window transitions guarantee **today**. They are included rather than squashed because they are the evidence that the behaviour change preserved what mattered. Read them first; the two perf commits are much easier to judge once you know what was nailed down.

## The window

Preload covers roughly +10s ahead and −2s behind the playhead. Elements leaving the window are evicted and paused.

The one non-obvious rule: the transition set spans the **union of the last tick and this tick**, not just the current one. An element that crosses the window boundary within a single tick would otherwise be dropped without being paused — it would keep decoding, unreferenced, which is exactly the cost this change exists to remove.

## Note on a shared file

Touches `packages/core/src/runtime/init.ts`, which another PR in this sequence also touches. The hunks are in disjoint regions of the file (this one around lines 1842-2119, the other from 2353 on), so they merge cleanly in either order. Worth a glance after whichever lands second.

## Verification

`packages/core` green (1,459 tests), `packages/studio` green (3,273 tests).

### DIFF
--- a/packages/core/src/runtime/entry.ts
+++ b/packages/core/src/runtime/entry.ts
@@ -1,5 +1,6 @@
 import { initSandboxRuntimeModular } from "./init";
 import { installAuthoredOpacityCapture } from "./colorGrading";
+import { installRuntimeMediaObserver } from "./media";
 import { fitTextFontSize } from "../text/fitTextFontSize";
 import { getVariables } from "./getVariables";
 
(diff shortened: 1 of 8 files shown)


miguel-heygen:
Closing the stale performance branch. Main still uses eager media preparation, but the old measurements do not establish a current bottleneck, and the proposed preload/cache policy changes playback behavior broadly. No replacement PR is planned without a fresh profile of current main.

PR #3001, opened 2026-08-04
refactor(studio): one media resolution layer for the Studio document

miguel-heygen:
Thumbnails, waveforms, and media probes each resolved media independently, with three near-duplicate code paths for the same job. They now share one resolver.

**This is a replacement, not an addition.** The diff reads as 900 added / 256 removed, but the added lines subsume the removed ones — read it as one layer replacing three, not as new surface area.

**681 logic lines**, the second-heaviest review in this sequence.

## What changes for callers

Each of the three consumers previously constructed its own shape from raw probe output. They now receive that shape from the resolver. The consumer-side diffs are mostly deletions.

## What to check

The failure path is the part worth reading closely. A failed resolution surfaces to the caller rather than being swallowed into a placeholder — the old paths differed on this, and the consolidated one had to pick a behaviour. It picks surfacing.

The other is cancellation: resolution is cancelled cleanly when its consumer unmounts mid-flight, which matters because thumbnails mount and unmount constantly under timeline virtualization.

## Related

A follow-up PR stacked on this one adds request de-duplication on top of this layer (each media source fetched once rather than once per referencing clip). That is where the request-volume numbers come from; this PR is the structural precondition and claims no measurement of its own.

## Verification

`packages/studio` suite green (3,288 tests). Thumbnails and waveforms render in a real project.

### DIFF
--- a/packages/studio/src/components/ui/VideoFrameThumbnail.tsx
+++ b/packages/studio/src/components/ui/VideoFrameThumbnail.tsx
@@ -1,8 +1,16 @@
-import { useState, useEffect } from "react";
+import { useState, useEffect, useMemo } from "react";
+import {
+  extractVideoFrames,
+  mediaCacheKey,
+  thumbnailResolver,
+} from "../../player/lib/mediaResolver";
+
+/** Seek to ~10% of the *source's* duration to avoid black opening frames. */
+const posterTimestamps = (sourceDuration: number) => [Math.min(2, sourceDuration * 0.1 || 2)];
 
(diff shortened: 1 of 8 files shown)


miguel-heygen:
Closing as superseded by the thumbnail scheduler and bounded metadata registry now on main. Timeline thumbnails and waveforms already use shared leases, cancellation, byte/concurrency limits and source-window-aware keys. Restoring this older resolver would lose current lifecycle and failure behavior.

PR #3002, opened 2026-08-04
perf(studio): fetch each media source once, and enforce the timeline budgets

miguel-heygen:
Two related changes to how the Studio treats media at open.

**Measured: media requests per distinct source 8.5× → 1.13×.**

> Stacked on the media resolution PR (hard dependency — same module). Please do not delete that branch while this is open.

## 1. Fetch each source once

A media source referenced by N clips was fetched N times while opening a project. It is now fetched once and shared.

The de-duplication is deliberately **not** a permanent cache: a source that is fetched, evicted, and re-requested fetches again. Making it permanent would trade a request problem for a memory problem, which is the wrong trade for an editor that keeps projects open for hours.

## 2. Make the declared timeline budgets real — or delete them

This is the half worth reading carefully.

`timelineViewportBudgets` declared `concurrentVideoDecodes: 2`, `concurrentMetadataJobs: 4`, `concurrentCompositionFetches: 2`, and `concurrentServerPages: 1`. Not one of them was enforced anywhere in the application. They were read only by diagnostics assertions — that is, by code that checked the numbers against themselves.

A declared-but-unenforced budget is worse than no budget, because every reader assumes it is a guarantee. Each one is now enforced at its stated limit (exceeding it queues rather than drops), explicitly annotated as advisory, or deleted.

## Note on a shared file

Touches `timelineElementHelpers.ts` around line 189; another PR in this sequence touches it around line 412. Disjoint, merges cleanly either way.

## Verification

`packages/player` green (334 tests), `packages/studio` green (3,333 tests).

### DIFF
--- a/packages/player/src/parent-media.test.ts
+++ b/packages/player/src/parent-media.test.ts
@@ -162,3 +162,56 @@ describe("ParentMediaManager audio-src proxy lifecycle", () => {
     expect(mgr.entries[0]).toBe(adopted);
   });
 });
+
+describe("ParentMediaManager adopted-proxy preloading", () => {
+  /** A document holding one timed `<audio>` the runtime has promoted to load. */
+  function iframeDocWithPromotedAudio(src: string): Document {
+    const iframe = document.createElement("iframe");
+    document.body.appendChild(iframe);
+    const doc = iframe.contentDocument!;
+    const audio = doc.createElement("audio");
+    audio.setAttribute("data-start", "0");
+    audio.setAttribute("data-duration", "5");
+    audio.preload = "auto";
+    audio.src = src;
+    doc.body.appendChild(audio);
+    return doc;
+  }
+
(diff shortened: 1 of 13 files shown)


miguel-heygen:
Closing this stale stacked branch. Main now has a real bounded thumbnail scheduler and metadata queue, so the generic resolver/budget rewrite is superseded. The narrow remaining candidate is avoiding eager loads of runtime-owned fallback audio proxies; that can be verified and shipped independently from current main.

miguel-heygen:
Fresh audit decision: Current thumbnail scheduling already replaces the main cache/budget proposal. The remaining eager parent-media proxy loading is a separate measurement candidate; delaying it can affect fallback playback and needs a current network/playback comparison. No verified net benefit has been established for that residual, so no immediate successor is planned.

PR #3003, opened 2026-08-04
perf(studio): take the Node-only AST and DOM stack off the first-paint path

miguel-heygen:
The eager browser bundle carried a Node-only AST and DOM stack — `@babel/parser`, `esprima`, `acorn`, `recast`, `ast-types`, `source-map`, `linkedom` — reachable from the SDK entry point on the first-paint path.

**Measured: eager entry chunk 1,116 → 604 KB gzip.**

## Two changes

1. **Defer the source editor and markdown preview** off first paint. Neither is visible on load.
2. **Move the AST stack behind a lazy boundary.** It is only needed when a composition is opened for editing, not when the shell renders.

## A correction worth recording

An early investigation concluded `linkedom` was already tree-shaken out. That was a **grep false negative on minified output** — the identifier does not survive minification in a form a text search finds. Sourcemap analysis showed 123 of its modules still present in the eager chunk.

The bundle-budget harness in this PR encodes the lesson: it asserts against the **module graph and gzip size**, never against a text search of the built bundle. If you take one thing from this PR, take that — a grep of minified output cannot prove absence.

## What to verify

- The eager entry chunk stays under the ratchet, asserted on gzip size.
- The AST stack is reachable only through the lazy path.
- The source editor, storyboard source editor, and render queue all still resolve their lazy imports and function after deferral.

## Note on a shared file

Adds a script entry to `packages/studio/package.json` adjacent to one added by the runtime-gate PR in this sequence. Whichever lands second may need a one-line conflict resolution.

## Verification

`packages/studio` suite green (3,279 tests); bundle budget harness passes.

### DIFF
--- a/packages/parsers/package-subpaths.json
+++ b/packages/parsers/package-subpaths.json
@@ -14,6 +14,12 @@
       "types": null,
       "environments": ["browser", "bun", "node"]
     },
+    "./types": {
+      "source": "./src/types.ts",
+      "runtime": "./dist/types.js",
+      "types": "./dist/types.d.ts",
+      "environments": ["browser", "bun", "node"]
+    },
     "./gsap-parser": {
       "source": "./src/gsapParserExports.ts",
       "runtime": "./dist/gsapParserExports.js",
--- a/packages/parsers/package.json
+++ b/packages/parsers/package.json
@@ -22,6 +22,12 @@
       "types": "./src/index.ts"
     },
     "./package.json": "./package.json",
+    "./types": {
+      "bun": "./src/types.ts",
+      "node": "./dist/types.js",
+      "import": "./src/types.ts",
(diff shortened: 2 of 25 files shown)


miguel-heygen:
Closing this stale bundle branch. Several unnecessary eager import edges still exist, so the useful part will be rebuilt as a small fresh import-boundary change and measured from the current production module graph. The historical bundle figures and broad chunking recipe will not be carried over without revalidation.

miguel-heygen:
Fresh audit decision: The current import-boundary audit identifies potential first-load reductions, but the old bundle measurements are stale and no current user-visible bottleneck or fresh size/latency result has been established. I am retaining the analysis as a benchmark candidate rather than recreating the broad bundle refactor without current evidence.

PR #3004, opened 2026-08-04
perf(studio-server): memoize the built preview document on the project signature

miguel-heygen:
The preview document was rebuilt from scratch on every request, including requests that changed nothing.

**Measured: preview build 441ms → 12ms. Repeat requests now return 304 instead of 200.**

## Two commits, and the second is the interesting one

The first memoizes the built document on a project signature.

The second fixes a **real bug in the first**: the signature was computed before hf-ids were minted, so minting them left the memo stale — the server would keep serving a document that no longer matched the project. It is kept as a separate commit rather than squashed because the sequence is the useful record. A memo whose invalidation is subtly wrong is worse than no memo, and this is exactly how that happens.

## What to check

- Two requests against an unchanged project build the document once.
- Editing the project invalidates and rebuilds.
- **Minting hf-ids invalidates the signature** — the regression the second commit fixes.
- A repeat request with a matching ETag returns 304 with no body.

## Verification

`packages/studio-server` suite green (410 tests).

### DIFF
--- a/packages/cli/src/server/studioServer.ts
+++ b/packages/cli/src/server/studioServer.ts
@@ -414,6 +414,10 @@ export function createStudioServer(options: StudioServerOptions): StudioServer {
       return cachedProjectSignature;
     },
 
+    invalidateProjectSignature(dir: string): void {
+      if (resolve(dir) === resolve(projectDir)) cachedProjectSignature = null;
+    },
+
     async lint(html: string, opts?: { filePath?: string }) {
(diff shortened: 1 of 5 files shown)


miguel-heygen:
Closing this old preview memo proposal. Current preview routes already support ETag/304 revalidation, and a new server document cache needs current workload measurements plus validation against today’s runtime, codec and subcomposition paths. The old benchmark alone does not justify carrying this cache forward.

PR #3005, opened 2026-08-04
fix(shader-transitions): survive WebGL context loss, and create the context only when a shader draws

miguel-heygen:
Two changes to the same question: when does a WebGL context exist, and what happens when the browser takes it away.

## 1. Survive context loss

Previously there was no path back from a lost context — the renderer relied on garbage collection to reclaim contexts and had no handling for the browser reclaiming one itself. A lost context left the preview permanently broken until reload.

Context loss is now detected and the renderer recovers on the next draw. A draw issued while the context is lost does not throw.

Browsers drop WebGL contexts under memory pressure and when too many exist on one page. In an editor that keeps a preview open for hours, this is not a hypothetical.

## 2. Create the context only when a shader draws

A composition with no shader transition previously still created a context. It no longer does — creation is deferred until the first shader actually draws, and releasing the last shader releases the context rather than waiting for GC.

This directly reduces how often the browser hits its context limit, which makes the loss handling above matter less often. The two changes are worth reading together.

## Verification

`packages/shader-transitions` suite green (21 tests), covering: recovery after loss, no context for shader-free compositions, exactly-once creation on first draw, release without GC, and no throw on draw-while-lost.

### DIFF
--- a/packages/shader-transitions/src/hyper-shader.test.ts
+++ b/packages/shader-transitions/src/hyper-shader.test.ts
@@ -20,6 +20,7 @@ import { init, type TransitionConfig } from "./hyper-shader.js";
 let drawArraysCalls = 0;
 let createProgramCalls = 0;
 let loseContextCalls = 0;
+let webglContextCount = 0;
 
 function createMockWebGl(): WebGLRenderingContext {
   return {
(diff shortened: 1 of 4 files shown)


miguel-heygen:
Closing the stale branch. A small independently reproduced defect remains: the module-global vertex shader is reused across WebGL contexts, so the second context fails to link. Carry only that context-ownership fix into a fresh PR; the broader lazy-allocation/recovery rewrite is not being recreated.

miguel-heygen:
Verified narrow successor on current main: https://github.com/heygen-com/hyperframes/pull/3672

PR #3006, opened 2026-08-04
fix(studio): release the previous project's render scopes on switch

miguel-heygen:
Switching projects retained the previous project's render scopes. In an editor where people switch between projects repeatedly in one session, that accumulates.

**Measured: retained heap 0.237 → 0.061 MB per switch cycle.**

## The cause

A `useCallback` closed over a scope that outlived the project it belonged to, chaining the previous project's state into the new one's callback identity. Plus one substring reference that was never explicitly detached.

## A caveat for anyone reproducing this

An earlier measurement of this same leak reported **+554 retained DOM nodes per switch**. That number was wrong, and wrong in an instructive way: it was the *measurement harness's own retained element handle* pinning an ancestor chain, not the application.

If you reproduce this, dispose every element handle before sampling the heap. A harness that holds a handle to a node in the tree it is measuring will report a leak that exists only in the harness.

## What to check

- Switching projects releases the previous project's selection scope.
- The lint modal's listeners detach on unmount.
- A switch-and-return cycle does not accumulate retained scopes across N iterations.
- Selection still works normally after a switch — the release is not over-eager, which is the obvious way to break this while making the number look good.

## Verification

`packages/studio` suite green (3,276 tests). The heap figure comes from the switch window of the runtime-cost gate, a separate PR in this sequence; the behaviour is verifiable from the tests here alone.

### DIFF
--- a/packages/studio/src/components/nle/NLEContext.tsx
+++ b/packages/studio/src/components/nle/NLEContext.tsx
@@ -25,6 +25,19 @@ export function shouldDisableTimelineWhileCompositionLoading(compositionLoading:
   return compositionLoading;
 }
 
+/**
+ * Copy a substring off its parent.
+ *
+ * V8 represents a substring of 13+ characters as a SlicedString that keeps the
+ * WHOLE parent alive. A handful of `compositions/*.html` paths cut out of a
+ * 36 KB `index.html` therefore pin that entire document for as long as the map
+ * holding them lives — one document per project opened. Concatenating and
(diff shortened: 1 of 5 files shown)


miguel-heygen:
Closing this old retention proposal. Project/source-map lifecycle and selection synchronization have changed on main, and the old callback-identity tests do not establish a current memory leak. Any remaining growth needs a fresh switch/GC retaining-path reproduction before a narrowly targeted fix.

PR #3007, opened 2026-08-04
perf(studio): scan the preview document once per timeline derivation

miguel-heygen:
Deriving timeline elements from a clip manifest called a resolver once per clip, and **every call rebuilt the entire candidate set** — a `[data-start]` query plus a four-selector `closest()` walk on every node it returned. Quadratic in clip count.

On a 273-clip project this was the single largest task in the application.

**Measured: idle main-thread script time 2,047ms → 882ms per 15s window (−57%),** three runs per arm at 4× CPU throttle.

## Why this shows up at idle

The derivation runs on a **timer at roughly 1.3/s whether or not anyone is interacting**. This is not a cost paid on click; it is a cost paid continuously, which is why the number above comes from a window with no input at all.

## What this does NOT buy

**Click latency is unchanged** — 7.4ms median before, 8.3ms after. Inside noise.

That deserves stating plainly because the obvious framing of this change is wrong. An earlier investigation reported selection as janky at "p95 41.8ms, 53% of frames dropped". Both halves were artifacts: the frame metric counted 16.8ms frames as dropped against a 16.7ms threshold (i.e. it counted vsync jitter), and the p95 came from a sample short enough that a single first-click property-panel mount dominated it. Steady-state selection was always ~8ms.

The win here is real and it is the idle path. It is not a responsiveness win.

## The scan is deliberately pass-scoped

Built once per derivation and dropped with it. It must **never** be cached across derivations: the preview document mutates between them, and a cached scan that outlived one would resolve clips to detached nodes — silently binding to the wrong element, which is far worse than a slow scan. There is a test asserting exactly this: a pass built before a mutation still resolves to the detached node; a fresh pass resolves to the live one.

Node-claiming moved into the resolver so a shared pass cannot hand two clips the same element.

## One commit here is a guard, not a change

`pin clip identity across the manifest and clip-tree paths` adds tests and no behaviour. It pins that the two paths compute clip identity **differently and cannot be unified**: the clip-tree path has an ordinal fallback the manifest path does not.

Unifying them looks like an obvious cleanup and is a data-loss bug. It would change `clip.id` for clips that currently resolve to null, which downstream diffing reads as element replacement — silently dropping selection, expanded state, the keyframe cache, and lint findings. The test exists so the next person to notice the asymmetry finds out why before changing it.

## Note on a shared file

Touches `timelineElementHelpers.ts` around line 412; another PR in this sequence touches it around line 189. Disjoint, merges cleanly either way.

## Verification

`packages/core` green (1,437 tests), `packages/studio` green (3,277 tests).

### DIFF
--- a/packages/core/src/runtime/clipIdentity.test.ts
+++ b/packages/core/src/runtime/clipIdentity.test.ts
@@ -1,0 +1,105 @@
+import { describe, it, expect, afterEach } from "vitest";
+import { createClipTree } from "./clipTree";
+import { collectRuntimeTimelinePayload } from "./timeline";
+
+/**
+ * Characterization of the ONE contract two runtime paths publish under the same
+ * name: `clip.id` in `__clipManifest` (timeline.ts) and `node.id` in
+ * `__clipTree` (clipTree.ts). Studio joins the two id spaces — the tree builds
+ * `clipParentMap`, the manifest supplies the clip that looks a parent up in it
+ * (see useTimelineSyncCallbacks.ts) — so any change to either side is a change
+ * to element identity, and studio diffs identity to decide "same element" vs
+ * "replaced element". A replacement drops selection, expanded rows, the
+ * keyframe cache and lint findings.
+ *
+ * These tests pin what each path emits TODAY. They are a guard for anyone
+ * unifying the two fallbacks, not an endorsement of the current split.
+ */
+
+const treeParams = {
+  startResolver: { resolveStartForElement: () => 0 },
+  timelineRegistry: {},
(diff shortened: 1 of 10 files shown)


miguel-heygen:
Closing this stale performance branch. The repeated fallback candidate scan still exists, but the live caller has moved to timelineSyncHydration. The useful fix will be rebuilt narrowly there, alongside correct DOM identity binding, with current query-count and browser evidence. The old idle percentages will not be reused.

miguel-heygen:
Verified narrow successor: https://github.com/heygen-com/hyperframes/pull/3681

PR #2977, opened 2026-08-03
docs: add the shared page components and the Reference Project

ukimsanov:
First of three commits that replace #2973. Everything here is additive — nothing imports these files yet, navigation is untouched, and no published page changes.

**The page components** (`docs/snippets/`, `docs/custom.css`)

Six React snippets the rebuilt pages compose against: `DocsVideo` / `ShowcaseWall`, `LiveReferenceProject` (which embeds a composition through `<hyperframes-player>`), `WorkflowChooser`, `AgentAction`, and two grid snippets.

The scrub indicator is a timecode bubble, not a thumbnail. Driving a preview frame meant mounting a second `<video>` with the same `src`, so every page carrying a film downloaded the whole file twice.

**The Reference Project** (`examples/docs-reference-project/`)

One real 10-second project the docs can point at instead of describing a hypothetical one — a live capture of example.com, synthesised narration, caption timings measured from that narration.

Verified independently, not quoted from a report:

```
hyperframes lint    0 errors, 0 warnings
hyperframes check   passed — 0 errors, 0 warnings, 2 info
contrast            28/28 text checks pass WCAG AA
```

Only the two WAV masters exceed the 500 KB non-LFS limit, so only those go through LFS. The MP3 stings (11 KB, 18 KB) and the capture PNG (21 KB) stay plain, so the example still works after a clone without `git lfs pull`.

`bun run docs:bundle-reference` regenerates the single-file embed the Introduction page loads from the CDN.

## Review notes

- 0 dangling navigation entries, 0 broken internal links.
- The one broken link the checker reports at this commit — `/concepts` from the Lambda migration guide — already exists on `main`; the next PR fixes it.

PR #2710, opened 2026-07-21
fix(studio): scope timeline ease focus lifecycle

miguel-heygen:
## Summary

Scopes edit-ease focus to the exact project, inspector session, selected element, animation, and request. A delayed or remounted inspector cannot consume a stale request, focus another animation, or clear a newer request.

This final Family E head also carries the valid post-merge hardening found while reviewing the preceding context-menu and beat-drag changes.

## Changes

- Stamp ease-focus requests with project, session, element, animation, and a monotonic nonce.
- Consume a request only when it matches the active inspector and clear only the consumed nonce.
- Share one animation-list implementation between the classic and flat property panels.
- Keep the one-shot focus consumer stable across inspector renders and consolidate lifecycle store reads to one shallow subscription per panel layer.
- Clear selector-only keyframe caches at the committed-result boundary so deleted diamonds disappear immediately without lying on failed saves.
- Exercise context-menu auto-dismiss and live-model dispatch through the rendered overlay.
- Keep beat dragging allocation-free across seek updates, scope viewport observation, key the active beat by time, reject a second touch, and let Escape bubble.
- Document why beat drag owns immediate source-removal interruption, why session and nonce identities survive reset, and why the development state handle exists.
- Inject the keyframe slice through its exact session-identity dependency and use typed commit-test tuples.
- Remove an ineffective context-menu memo boundary and an unimplemented callback declaration.

## Stack

Family E, 7 of 7 and final PR. Base: main.

## Validation

- Full Studio suite: 301 test files passed; 3,285 tests passed; 18 todos.
- Studio typecheck passed.
- Oxlint and oxfmt passed across all 28 changed files.
- File-size, diff, and pre-commit Fallow gates passed with no gated findings.

PR #3009, opened 2026-08-04
chore: release v0.7.92

vanceingalls:
## What

Stable release **v0.7.92**, cut through the release-PR flow so `publish.yml` actually fires.

Covers everything since v0.7.90 — headline is percentage-based canary rollouts, plus 29 new registry theme items, virtualized Studio timeline performance work, and sub-composition sibling asset path resolution.

## Why this is 0.7.92 and not 0.7.91

v0.7.91's version bump and changelog reached `main` directly, without a release PR. Stable releases publish **only** from a merged `release/v*` PR — `publish.yml`'s push trigger is `v*-*`, prerelease-only — so publish never ran and 0.7.91 was never on npm.

Rather than revert a commit that by then sat under two unrelated merges (#2710, #2977), the notes fold forward: this release covers the full v0.7.90..HEAD range, the stray `v0.7.91` tag has been deleted from the remote, and `releases/v0.7.91.md` plus its changelog entry are removed. No published version is skipped — 0.7.91 simply never existed as a release.

## Also fixes the cause

`set-version.ts` printed the prerelease flow (`git push origin main` + `git push origin v<version>`) as the way to publish a **stable** release. Following it is what produced the unpublishable tag, and a stray tag then fails the next release's `verify_remote_tag` check. It now prints the branch + PR commands, says not to push the local tag, and points at `docs/contributing/release-channels.mdx`.

## Merging this

The workflow checks out the exact merge SHA, creates the `v0.7.92` tag there, publishes to npm under `latest`, and cuts the GitHub release. No tag was pushed from local.


miguel-heygen (approved):
Validated the stable release artifacts: all 13 publishable package manifests are at 0.7.92, v0.7.91 artifacts are folded forward cleanly, no stale v0.7.91 references remain outside changelog history, the release diff passes git diff --check, and no v0.7.92 remote tag exists. The original Build failure was an aws-cdk-lib tarball extraction/download failure; rerun is in progress.

PR #3011, opened 2026-08-04
fix(engine): keep BeginFrame time monotonic across frame rates

### LINKED ISSUE #3012: BeginFrame can hang at 60fps after warmup moves Chrome time backwards
## Describe the bug

A BeginFrame render can hang during initialization at 60fps and other high frame rates.

Warmup advances Chrome in fixed 33ms steps, but the timestamps used after warmup are calculated from the output fps. At 60fps this makes Chrome time jump backwards:

- last warmup tick: 1947ms
- init commit tick: about 1067ms
- liveness probe tick: about 1083ms
- frame zero: about 1167ms

`HeadlessExperimental.beginFrame` can stall after receiving that sequence.

## Reproduction

Run a 60fps render on the BeginFrame path with locked warmup enabled. The timestamp calculation on current `main` produces the sequence above.

A regression test and the fix are included in #3011.

## Expected behavior

The init commit, liveness probe, and first captured frame should all use timestamps later than the final warmup tick.

## Actual behavior

The first post-warmup timestamp jumps back by roughly 880ms at 60fps, and the BeginFrame call may never complete.

## Environment

Current `main` at `8e41fa17b985b24cb3b4719e553481a583124b68`. The affected path is Linux/headless-shell BeginFrame capture.

zihanyang-dev:
Fixes #3012

## What

Fixes `HeadlessExperimental.beginFrame` hanging at 60fps and other high frame rates.

## Why

Warmup always advances Chrome in 33ms steps, but the first capture timestamp was calculated from the output fps.

At 60fps, the last warmup tick is 1947ms, while the next commit tick used to jump back to about 1067ms. Chrome sees time moving backwards and `beginFrame` can stall.

## How

Keep the old timestamps when they are already safe. If the commit tick would go backwards, move the capture baseline past warmup first.

The init commit and both producer probes now use the same timestamp helpers, so they cannot calculate different values.

## Test plan

- [x] Added tests for 24, 30, 31, 32, 33, 60, 120, 240, and 59.94fps
- [x] Added coverage for the actual session baseline, commit parameters, and both producer probe paths
- [x] Engine tests: 1383 passed, 3 skipped
- [x] Producer unit tests passed
- [x] Engine and producer typechecks and builds passed
- [ ] Manual render test

### DIFF
--- a/packages/engine/src/index.ts
+++ b/packages/engine/src/index.ts
@@ -120,6 +120,7 @@ export {
   getCapturePerfSummary,
   percentileOf,
   prepareCaptureSessionForReuse,
+  deriveBeginFrameProbeTimeTicks,
   type CaptureSession,
   isTransientBrowserError,
   isMemoryExhaustionError,
--- a/packages/engine/src/services/frameCapture-warmupTicks.test.ts
+++ b/packages/engine/src/services/frameCapture-warmupTicks.test.ts
@@ -12,7 +12,10 @@
 import { describe, expect, it } from "vitest";
 import {
   LOCKED_WARMUP_TICKS,
+  deriveBeginFrameTimelineTicks,
+  deriveBeginFrameTimeTicks,
   driveWarmupTicks,
+  prepareBeginFrameTimeline,
   warmupFrameTimeTicks,
   type WarmupTickState,
 } from "./frameCapture.js";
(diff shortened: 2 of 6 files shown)


miga-heygen (commented):
## R1 — Monotonic BeginFrame timing (head `22bc4d6`)

### Verdict: Approve

Clean fix for a real hang at high frame rates. The root cause is clear: warmup runs at 33ms intervals, but the capture baseline was calculated from the output fps. At 60fps, the commit tick falls BEFORE the last warmup tick (1067ms vs 1947ms), Chrome sees time go backwards, and `beginFrame` stalls.

### What was verified

**`deriveBeginFrameTimeTicks` logic — correct at all tested rates:**

| fps | captureInterval | legacyCommit | lastWarmup | legacy safe? | result |
|-----|----------------|-------------|------------|-------------|--------|
| 30 | 33.33ms | 2133ms | 1947ms | yes | legacy (2333ms) |
| 60 | 16.67ms | 1067ms | 1947ms | *no* | monotonic (2147ms) |
| 120 | 8.33ms | 533ms | 1947ms | *no* | monotonic (2063ms) |
| 240 | 4.17ms | 267ms | 1947ms | *no* | monotonic (2022ms) |

The crossover is at ~33fps (where captureInterval matches warmupInterval). Everything at or below ~33fps preserves the legacy baseline (backward-compatible). Everything above gets rebased past warmup. The invariant `warmup < commit < probe < capture` holds in all cases.

**SSOT improvements:**
- Probe tick calculation was inline in TWO places (`renderChunk.ts` and `probeStage.ts`) — both now use `deriveBeginFrameProbeTimeTicks`. Good dedup.
- Commit tick calculation was inline in `initializeSession` — now through `preparedBeginFrameTimeline.commitParams`.
- Three magic constants (10, 6, 5) are now named: `BEGIN_FRAME_CAPTURE_HEADROOM_INTERVALS`, `BEGIN_FRAME_COMMIT_LEAD_INTERVALS`, `BEGIN_FRAME_PROBE_LEAD_INTERVALS`.

**Tests:** Strong coverage — parametric tests across 9 frame rates (24, 30, 31, 32, 33, 60, 120, 240, 59.94), monotonicity assertions, legacy-preservation assertions, unlocked warmup, and session-init wiring. Producer probe test asserts the exact `frameTimeTicks` passed to `probeBeginFrameLiveness`.

**Edge cases:**
- `LOCKED_WARMUP_TICKS = 0`: legacy commit = `4 * captureIntervalMs > 0` → always safe. Correct.
- Unlocked warmup with fewer ticks: explicitly tested.
- `Math.max(0, ...)` in probe derivation guards against negative ticks at very low capture rates. Correct.

No findings. Ship it.

---
_Review by Miga_

miguel-heygen (approved):

zihanyang-dev:
@miguel-heygen This is ready to merge.

PR #3013, opened 2026-08-04
feat(audio): audio FX registry, Web Audio graphs, and offline render

vanceingalls:
First of six. Adds the audio FX layer: a registry of effects, the Web Audio
graphs that realise them, the offline render path, and the panel that drives
them.

- **Registry** (`audioFx.ts`) — 14 effects across filter / dynamics / non-linear
  / time, each declaring its full parameter surface: range, step, unit, scale
  and a one-line hint. Chains serialise onto the element as `data-fx-chain`,
  the way colour grading carries its config, so there is no side-car file to
  find, resolve or lose.
- **Graphs** (`audio/audioFxGraph.ts`) — one builder per registry entry, plus
  four AudioWorklet processors for the dynamics. Biquad coefficients follow the
  RBJ cookbook; a one-pole filter is built from raw IIR coefficients because
  `BiquadFilterNode` is always two-pole.
- **Render** (`engine/audioFxRender.ts`) — runs the *same* builders in an
  `OfflineAudioContext` inside the headless browser the engine already drives.
  Preview and render agree by construction rather than by measured tolerance,
  which is why the graph code is bundled once and shared rather than
  reimplemented per side.
- **Carve** (`audioCarve.ts`) — analyses a voice track and dips the bed where it
  sits, via a dependency-free radix-2 FFT. `intelligibilityBias` exists because
  ranking bands by raw voice power lands on the fundamental, where a bed does
  not actually mask a voice; weighting toward 1–3 kHz is what makes the cut
  audible.
- **Panel** — generated entirely from the registry, so adding an effect or a
  knob upstream needs no change in the UI, and the panel cannot offer a value
  the renderer would reject.

Failure posture throughout: an unreadable chain **plays dry** in preview but
**fails the render**, because a render that silently drops effects ships
something plausible and wrong.


vanceingalls:
Superseded by stack #3027. This PR carried the whole audio FX foundation in one diff (+3539); it is now split into four PRs, each under 1000 LOC: #3019 registry, #3020 Web Audio graphs, #3021 offline render + carve, #3022 panel. Same commits, rebased onto current main.

PR #3014, opened 2026-08-04
feat(audio): hear the FX chain while previewing

vanceingalls:
Stacked on #3013.

Splices the chain into the preview transport, so a chain is heard while
scrubbing rather than only after a render.

The splice point matters: the transport plays each track from a decoded
`AudioBuffer` and mutes the `<audio>` element to avoid doubling, so
`createMediaElementSource` would have processed a stream nothing is listening
to. Effects go between the decoded source and its gain — the same order the
offline render uses.

Two fixes found by using it:

- **Parameter drags restarted playback.** The panel wrote through the persisting
  attribute path on every `input` event, so each pixel of a knob drag triggered
  a preview refresh, a composition reload, and an audio restart. Dragging now
  writes live and commits once on release.
- **Worklet nodes were constructed before their processor was registered.**
  `AudioWorkletNode` throws in that case and the whole chain is lost. Worse, a
  module-level promise cache made every context after the first believe it was
  ready when it was not — registration is now tracked per context, and a chain
  that needs an unregistered worklet plays dry until the module lands, then
  swaps itself in.

Also splits the chain-node row out of `FxSection`: the render callback had grown
into a 70-line closure over half the component's state, and the two reorder
arrows were the same button written twice.


terencecho (commented):
**Verdict: BLOCK** — required "File size check" is red and PR-caused. Otherwise the splice and worklet-late-swap architecture verify cleanly; four foundation-cohort seeds should be flagged for pre-emption.

**PR-body claim verification (head `15dfbda5`):**
- Transport builds the graph on schedule: `packages/core/src/runtime/webAudioTransport.ts:195` calls `attachElementFxChain(this._ctx, el, sourceNode, gainNode)` before `gainNode.connect(this._masterGain)`.
- Splice sits between the decoded source and the volume gain (matches offline order): `packages/core/src/runtime/audioFx.ts:81-84` wires `source.connect(handle.input)` and `handle.output.connect(destination)`.
- Worklet-not-ready plays dry then hot-swaps: `runtime/audioFx.ts:66-96` (verified via `WeakMap`+`WeakSet` per-context registration in `audio/audioFxWorklets.ts:229-249`).
- Live re-parameterisation via `MutationObserver` on `HF_AUDIO_FX_ATTR`: `runtime/audioFx.ts:113-121`.
- Dispose on `stopAll()`: `webAudioTransport.ts:290` — `source.fx?.dispose()`.

**Foundation-cohort seed audit:**
1. **LFO dispose leak (#3183 later fixes with `retireLfo`) — PRESENT.** `chorusLfo` and `allpassPhaser` in `audio/audioFxGraph.ts` call `lfo.stop()` but the LFO is NOT in the `[input, out, dl, depth, wet, dry, …].forEach(disconnect)` list. This PR activates the leak by wiring `buildFxChain` into preview. Seed the retire helper here.
2. **Worklet lifecycle (#3175) — PRESENT.** `HfCompressor` / `HfLimiter` / `HfGate` / `HfBitcrush` in `audio/audioFxWorklets.ts` all `return true` unconditionally in `process()`. `workletBuilder`'s dispose only calls `node.disconnect()` — no `port.postMessage` stop signal, no `process(): false` self-retire. Preview accumulates processors across clip lifetimes.
3. **Chunked PCM (#3174) — N/A.** Preview decodes via `AudioContext.decodeAudioData(arrayBuffer)`; the single `data:base64` URL (`audioFxWorklets.ts:243`) carries the worklet JS source, not PCM.
4. **Duck-before-quantise (#3174) — N/A.** Real-time Web Audio path; no offline sample clamp.

**Extra seed I found (worth mentioning here so it's on the wa-18 sweep):** natural track-end doesn't dispose fx. `schedulePlayback`'s `sourceNode.addEventListener("ended", …)` at `webAudioTransport.ts:238-244` splices `_activeSources` but never calls `source.fx?.dispose()`. Only `stopAll()` disposes. A clip that plays to its natural boundary leaves the MutationObserver + LFOs + worklet nodes attached until the next stop — compounds with (1) and (2).

**`HfAudioFxNode` field / allow-list check:** New optional `fromCarve?: boolean` at `audioFx.ts:738`; covered in both `parseAudioFxChain` (`:784, :790`) and `serializeAudioFxChain` (`:812`). Correctly gated. The hand-constructed allow-list pattern (#3019's root concern — not typed against `keyof HfAudioFxNode`) is unchanged, but this field IS covered.

**Test-shape audit:**
- `runtime/audioFx.test.ts` uniquely pins the splice (dry path must no longer exist), pins graceful degradation (empty / unreadable / unknown effect / no-`getAttribute`), pins dispose. Uses a fake `Ctx`/`Node` — cannot exercise worklet late-swap or the MutationObserver path.
- `propertyPanelFxSection.test.tsx` pins the drag-preview-vs-commit regression the PR body calls out (`onChainPreview` ≥3, `onChainChange` 0 during drag; 1 on pointerup). Good.
- Missing pins: the worklet-not-ready → dry → late-swap branch, and the observer callback ignoring `handle.update()`'s `false` return on shape change (documented as "wait for next schedule" — but no test guards it).

**Nits:**
- `runtime/audioFx.ts:120` — `if (next.chain.nodes.length > 0) handle.update(next.chain);` drops `handle.update()`'s return value; a shape change silently falls through with the old graph running. Intended per PR body, but worth a code comment referencing the "waits for next schedule" contract.
- `audio/audioFxGraph.ts` `onePoleBuilder`'s `simple(node, () => {})` returns `dispose: () => node.disconnect()` — fine, but ties into the shape-signature carrying `frequency` for one-pole rebuilds.

**CI:**
- FAILURE — File size check: `PropertyPanelFlat.tsx has 672 lines (max 600)`. This PR added +60 net lines there. Real, blocking, PR-caused. Split before merge.
- FAILURE — CI Test (`@hyperframes/engine` `audioFxRender.test.ts:95`, `spawnSync ffmpeg ENOENT`): missing `ffmpeg` on the runner — infra drift, not PR signal. The affected test is offline-render territory this PR doesn't touch.
- All Preflight, regression shards 1-9, Perf (load/fps/scrub/drift/parity), Windows render+tests, Preview parity, CodeQL, Typecheck, Studio smoke, Producer/SDK/CLI: green.

— Review by tai (pr-review)

PR #3015, opened 2026-08-04
feat(audio): automation envelope model

vanceingalls:
Stacked on #3014.

The data model behind Ableton-style automation: breakpoint envelopes over track
volume or one knob of one effect, stored on the element as `data-automation`.

```
data-automation='{"version":1,"lanes":[
  {"target":"volume",          "points":[{"t":0,"v":0.55},{"t":1.5,"v":0.15}]},
  {"target":"fx.n2.frequency", "points":[{"t":0,"v":400},{"t":11.5,"v":18000}]}]}'
```

Decisions worth reviewing:

- **Times are clip-local**, so an envelope travels with the clip when it moves —
  the clip-envelope model, not arrangement automation. Stated rather than
  implied, since the alternative changes where the data lives.
- **`sampleAutomationLane` is the single interpolator.** The lane drawing, the
  preview scheduler and the render bake all call it, so the picture and the
  sound cannot disagree about a curve.
- **Log-scaled parameters interpolate in log space**, so a 200 Hz → 8 kHz sweep
  is an even sweep — what the knob's own scale already promises.
- **Chain nodes gain a stable `id`**, minted by count rather than randomly so the
  document is identical on every machine. Lanes address nodes by id, so
  reordering a chain never re-points a lane, and a lane whose effect was deleted
  is dropped rather than left to reattach if an unrelated effect later takes the
  same id.
- **`automatable` flags on the registry**, with each graph builder handing back
  the AudioParam behind them. A knob is not always one param — a wet/dry mix is
  two gains moving in opposition, a millisecond knob drives a delay time in
  seconds — so each target carries its own mapping. A test builds every effect
  and checks the exposure both ways, so the flag cannot lie.

`lint` warns when a track carries both a volume lane and a GSAP volume tween,
since only the lane is heard.


terencecho (commented):
**Verdict: LGTM_PENDING_CI** (both red checks look pre-existing / base-inherited — confirm before merge)

**Parse/serialize allow-list (Via's flag) — CLEAN.** One new `HfAudioFxNode` field: `id?: string` (`audioFx.ts:770`). Both allow-lists updated symmetrically:
- Parse: cast at `audioFx.ts:820-826`, emitted at `audioFx.ts:831` via `...(typeof node.id === "string" && node.id ? { id: node.id } : {})`.
- Serialize: emitted at `audioFx.ts:852` via `...(node.id ? { id: node.id } : {})`.
- Test pins the parse-side: `audioAutomation.test.ts:281-287` (`parseAudioFxChain(...).nodes[0]!.id === "n7"`).

No new field on `HfAudioFxChain` itself. Automation is stored as a separate `data-automation` element attribute with its own `parseAutomation` / `serializeAutomation` (`audioAutomation.ts:214,246`) plus a full round-trip test at `audioAutomation.test.ts:101-120`.

**PR-body claims — verified.**
- Single interpolator: `sampleAutomationLane` at `audioAutomation.ts:319` is the only public sampler; `sampleAutomationCurve` (`:344`) delegates to it.
- Log-space interpolation: `lerpValue` at `audioAutomation.ts:299-306`, guarded by `a > 0 && b > 0`.
- `automatable` flags: `audioFx.ts:37-45` (interface field), test at `audio/audioFxGraph.test.ts:344-407` builds every effect and checks both ways (exposed ⇔ flagged).
- Counted ids: `mintAudioFxNodeId` at `audioFx.ts:864-871`, walks `n1, n2, …` filling the first gap. Test pins the gap-fill (`audioAutomation.test.ts:275-278`).
- Lint rule: `findVolumeDoubleAutomationFindings` at `packages/lint/src/rules/media.ts:604-641`, warning severity.

**Envelope semantics — sound.**
- NaN guard: `numberOrNull` (`audioAutomation.ts:139-146`) refuses non-finite and empty strings before they reach `AudioParam`; `cleanPoint` drops the point rather than coercing (`:167-176`).
- Curve exponent: `Math.pow(x, Math.pow(2, 2 * curve))` at `:298` — for curve ∈ [-1,1], exponent ∈ [0.25, 4] always positive; pinned at both endpoints (test `:206-216`).
- Hold-at-edges outside points (`:322-325`); binary-search interior (`:326-334`); pathological-lane cap at 512 (`:33`, test `:243-253`).

**Foundation-cohort seed audit.**
- viaConic NaN — no conic / bezier interpolation exists at this SHA. `audioAutomation.ts` is 368 lines. Not seeded here.
- Rate-change reschedule — times are clip-local (`audioAutomation.ts:63`, `t: seconds from the start of the clip`). This is the correct base for rate-change fixup at the scheduler layer (later PR).

**Test-shape audit — strong.** Each new test pins one behavior: target parsing (`:31-40`), registry-driven range (`:42-56`), normalize sort/dedup (`:60-79`), NaN drop (`:81-100`), volume clamp at parse (`:102-118`), malformed rejection (`:120-129`), round-trip (`:132-149`), resolve drop + range clamp (`:153-170`), no-chain drop (`:172-178`), hold-outside (`:184-190`), lerp (`:192-195`), log-space sweep (`:197-207`), curve bend endpoints (`:209-223`), bisection under density (`:225-231`), pathological cap (`:233-254`), curve endpoints (`:256-262`).

**Nits (non-blocking):**
1. No `parse(serialize(chain))` round-trip test with `id` (only parse-side pinned at `:281-287`). Given the symmetric emit/read, risk is low, but a two-line round-trip test would fully close Via's pattern for future maintainers.
2. `sampleAutomationCurve` clamps `count < 2` to 2 (`:339`) with no test pin.
3. `lerpValue` silently falls to linear when a log-scale segment touches a non-positive value (`:302`). Volume can't hit this (linear scale); fx params are clamped to `p.min > 0` post-resolve, so the fallback is defensive-only. Consider a code comment.
4. Lint regex `#${id}(?![\w-])[^;]{0,200}?\bvolume\s*:` (`media.ts:635`) is best-effort — compound selectors like `"#a, #bgm"` may miss. Author explicitly labels this a "warning" where "a miss costs nothing"; fine as-is.

**CI status.** 2 red:
- `File size check` — fails on `packages/studio/src/components/editor/PropertyPanelFlat.tsx` (672 lines > 600). Not modified by this PR. Base-inherited from `wa-2-fx-preview` / earlier.
- `Test` (`bun run --filter '!@hyperframes/producer' test`) — exit-1 from CI job aggregator; visible per-package summaries (studio 329/329, player individual suites all pass, core/parsers show only stderr noise). `@hyperframes/player` shows `ECONNREFUSED ::1:3000` stderr, but its test suites still print `✓` — looks like a flake or pre-existing baseline failure. Recommend confirming both against `wa-2-fx-preview` head before merging; if both reproduce on the parent, this PR is clean.

— Review by tai (pr-review)

PR #3016, opened 2026-08-04
feat(audio): play automation envelopes in preview and bake them at render

vanceingalls:
Stacked on #3015.

Makes the envelopes actually run, on both sides.

**Preview** — each lane is scheduled onto the AudioParams behind its knob using
native ramps and value curves. Nothing evaluates the envelope per frame: it is
handed to the audio thread once, so it stays sample-accurate however busy the
main thread is. Timing comes from the transport, so an envelope survives seeking
into the middle of a clip, a clip that has not started yet, and a playback rate
that compresses clip seconds into context seconds.

A segment is only scheduled as a straight ramp when nothing bends it — no
curvature, a linear parameter scale, and no unit mapping. Log-scaled and mapped
segments are sampled instead, because a delay knob in milliseconds and a wet/dry
pair moving in opposition are not linear in the parameter they drive.

**Render** — the offline path schedules with the same scheduler, in the
`OfflineAudioContext` that already runs the same graph builders. The input WAV is
the clip's own audio from its first sample, so clip-local time is offline time.
Volume lanes take the existing PCM bake rather than a second mechanism: the lane
becomes keyframes, so a straight fade stays two of them and only a bent segment
is sampled — the baker interpolates linearly and would otherwise quietly
straighten the curve.

Lanes with nowhere to write are skipped rather than reported: a one-pole filter
exposes no frequency param, and the worklet effects expose none at all.

A browser test sweeps a lowpass from below a 2 kHz tone to well above it and
measures both ends. Parsing an envelope is not the same as scheduling it, and
only running the real thing tells the two apart — this test is what caught that
the injected runtime bundle needed rebuilding.


terencecho (commented):
**Verdict: LGTM_WITH_NITS pending CI clarification** — foundation-cohort seeds are all present as expected for this Batch stage; scheduling shape is correct and physically verified.

**Verified PR-body claims (file:line at head `c651302`):**
- (a) Preview via Web Audio scheduling — `packages/core/src/audio/audioFxAutomation.ts` `emit()` emits `setValueAtTime` / `linearRampToValueAtTime` / `setValueCurveAtTime`; wired into `packages/core/src/runtime/webAudioTransport.ts` `scheduleVolumeLane` + `attachElementFxChain(..., timing)`.
- (b) Offline bake — `packages/core/stubs/audio-fx-runtime-entry.ts` `render()` invokes `scheduleChainAutomation(..., { scheduledAt: 0, elapsed: 0, rate: 1 })` inside `OfflineAudioContext`. Volume lane bakes to PCM via `packages/engine/src/services/audioMixer.ts` `volumeLaneKeyframes` + `applyVolumeEnvelopeToWav`.

**AudioParam scheduling correctness (per invariant):**
- Cancellation on lane update: PASS — `scheduleParamLane` calls `target.param.cancelScheduledValues(timing.scheduledAt)` before every write; `attachElementFxChain.rescheduleAutomation` fires `cancelParamLane` + re-schedule on `data-automation`/`data-fx-chain` MutationObserver.
- Time base: PASS — `scheduledAt = this._ctx.currentTime` at play time; `toCtx(t) = scheduledAt + (t - elapsed) / rate` is absolute AudioContext-relative. Offline uses `scheduledAt: 0` which is the correct `startRendering()` origin.
- `exponentialRampToValueAtTime` zero-value: N/A — code never calls it. Log-scaled ranges (e.g. filter frequency) go through `setValueCurveAtTime` sampled by `sampleAutomationCurve`, sidestepping the spec's throw-on-zero.
- First-point seed: PASS — `planOps` unshifts `set(timing.scheduledAt, sampleAutomationLane(...))`; suppressed only when a curve begins exactly at `scheduledAt` (Web Audio forbids event overlap with a value curve). Test "does not seed on top of a curve that starts at the same instant" pins this.
- Retryable/swallowed scheduling: no try/catch wraps the AudioParam calls. `readAutomation` (`packages/core/src/runtime/audioFx.ts`) catches parse errors and silently returns `NO_AUTOMATION` — see nit.

**Foundation-cohort seeds (expected to be fixed by #3174):**
- Reschedule on transport rate change: SEED — `rate` is captured once at `WebAudioTransport.play()`. Rate/seek changes after schedule don't retrigger `rescheduleAutomation` (only `data-*` attribute mutations do). Envelope will drift on a 2× rate flip mid-play.
- Duck-before-quantise: PASS — envelope multiplies into samples in `applyVolumeEnvelopeToWav` BEFORE `writeWav`'s ±1 clamp.
- Chunked PCM: SEED — `packages/engine/src/services/audioFxRender.ts` transfers each plane as a single base64 blob through `page.evaluate`. Long clips will punish the serializer.
- Silent scheduling failure: SEED — `readAutomation` catch, `scheduleChainAutomation`'s `byId.get(...)?.automation?.[param]` miss, and `resolveAutomationRange` returning null all drop lanes with no dev-log.

**Bake correctness:** `scheduledAt: 0, elapsed: 0, rate: 1` yields `toCtx(t) = t` on the offline timeline; `source.start()` implicitly begins at 0. Clip-local time == offline time as claimed. Correct.

**Test-shape audit:** Both shapes present. `audioFxAutomation.test.ts` uses `FakeParam` for scheduling-shape assertions (correct for the pure planner). `audioFxRender.test.ts` "sweeps a filter across the clip" **physically** measures `db(sliceRms(after, 0.05, 0.25))` vs `1.2..1.45` and asserts `tail > head + 15` dB — this is the load-bearing test the PR body brags caught the injected-runtime rebuild issue. Good.

**Nits (non-blocking):**
- Consider a dev-only `swallow`/diagnostics log on `readAutomation` catch and on lane-target miss — helps authors debug "why isn't my envelope playing".
- `readElementAutomation` re-parses the chain attribute to resolve; the resolved chain from `readChain` above it in `attachElementFxChain` isn't threaded down. Minor.

**CI:**
- Passing: Lint, Typecheck, Build, Producer unit/integration, Windows render, all regression shards.
- FAILURE `CI / Test`: all 3513 studio tests exit 0, but overall job exits 1 with `ECONNREFUSED 127.0.0.1:3000` noise — looks like an engine-suite localhost stub, not this PR's code. Worth confirming.
- FAILURE `File size check`: expected bundle growth from the new automation planner + runtime entry. Confirm gate threshold.
- CANCELLED `Preview parity`: needs re-run.

— Review by tai (pr-review)

PR #3017, opened 2026-08-04
fix(audio): two things the envelopes needed to be heard at all

vanceingalls:
Stacked on #3016. Both found by listening rather than by testing.

**1. The volume lane was inaudible in preview.** It was scheduled onto the
transport's gain AudioParam — but the runtime rewrites that gain *every tick*
from `data-volume` and the GSAP-seeked value, so the envelope was erased within
a frame. Volume automation was correct in the render and silent in preview.

The lane now feeds the per-tick path where the probed volume keyframes already
sit, checked ahead of them so the two cannot fight, and the transport no longer
schedules volume at all: one mechanism instead of two racing.

The cost is stated plainly: in preview the level now steps per tick rather than
per sample, exactly as the existing keyframe path does. The render still bakes it
sample-accurately, and FX parameters are still scheduled on their own
AudioParams, since nothing rewrites those. Measured on a real duck —
`0.550` open, `0.150` ducked, clean ramps at both ends.

**2. Structural chain edits never applied to the running graph.**
`buildFxChain`'s `update` reports `false` when a change is structural rather than
new values, and the attribute observer ignored that. So adding, removing or
bypassing an effect did nothing until the persisting write reloaded the
composition — and that reload restarted every playing track, which is what was
audible as the chop.

The graph is now swapped in place: old effects detached, new ones built and
connected between the same source and gain, lanes re-scheduled onto the new
nodes. The source node is never touched, so playback does not restart. A track
with no chain is watched too, so adding its first effect is heard the same way —
which means the attach function now always returns a disposer instead of `null`
for the empty case.


terencecho (commented):
**Verdict: LGTM_WITH_NITS.** Both fixes cleanly wire what compiled but wasn't reaching the graph. Falsification tests uniquely pin each mode. CI failures are stack-inherited or infrastructure.

**Two fixes (head SHA `68ce910`):**

1. **Volume lane feeds the per-tick fader instead of the doomed gain-AudioParam.** New `audioAutomationVolume.ts:38-57` samples the lane per tick; wired into `media.ts:274-279` ahead of `volumeKeyframes`, and `scheduleVolumeLane` deleted from `webAudioTransport.ts` (previously ~line 200). Runtime's per-tick rewrite of `gainNode.gain` from `data-volume`/GSAP was erasing the scheduled envelope within a frame — now the lane owns the fader on the same path.
2. **Structural FX-chain edits swap the graph in place while the source keeps playing.** `audioFx.ts:141-181`: `rebuild()` detaches old handle, attaches new one between the same source and gain; `attachElementFxChain` no longer returns `null` for empty chain (`ln 124`) so the first-effect add is watched. Observer at `ln 224-228` now branches on `handle.update`'s `false` return to invoke `rebuild`.

**Falsification tests (all fail pre-fix):**
- Fix 1: `media.test.ts:328-336` "drives the transport gain from the lane, not from data-volume" (`0.8`/`0.1` vs `data-volume=0.55`); `media.test.ts:365-386` "supersedes keyframes probed from the timeline" (`0.1` vs keyframe-forced `1`). Both would return the pre-fix value.
- Fix 2: `audioFx.test.ts:135-146` first-effect add (empty chain used to return `null` w/ no observer); `ln 148-162` remove-last (early-return on empty chain in old observer); `ln 164-186` structural swap (old code called `handle.update` and ignored the `false`, leaving connections unchanged).

**Foundation cohort:** No AudioParam ordering changes (#3174/#3175/#3178/#3213 scope untouched). No LFO/worklet dispose invariant touched — worklet-not-ready path at `ln 96-119` retains prior behaviour (still no observer install; existing narrow gap, not a regression). No parse allow-list changes.

**Cross-fix interaction:** Clean. `scheduleChainAutomation` (`audioFxAutomation.ts:189-208`) filters `parsed.kind === "fx"`, so it never touches volume — Fix 2's rebuild scheduler cannot resurrect the deleted AudioParam-schedule footgun.

**Test-shape:** Each new test pins exactly one fix mode; no overlap; no hidden coverage from higher-level helpers.

**Nits:**
- `webAudioTransport.ts:195-198` comment "volume automation rides on their output — the same order the offline render uses" is now slightly stale (preview goes per-tick; render still bakes). Consider aligning with the PR-body framing.
- Worklet-not-ready branch at `audioFx.ts:96-119` still skips the MutationObserver, so a chain edited before worklet registration lands stays static until re-attach. Pre-existing; call-out only.

**CI:** `File size check` fails on `PropertyPanelFlat.tsx` (672 lines) — untouched here, stack-inherited from base. `Fallow audit` re-flags the `audioFx.ts` refactor; may need an extra `.fallowrc.jsonc` entry. `Test` fails on `packages/engine/…/audioFxRender.test.ts` with `spawnSync ffmpeg ENOENT` — runner-infrastructure gap, unrelated. All load-bearing lanes (Producer unit/integration, regression shards 1-9, preview-regression, player-perf, typecheck, lint) green.

— Review by tai (pr-review)

PR #3018, opened 2026-08-04
feat(studio): automation lanes in the timeline, and automate/un-automate from the panel

vanceingalls:
Stacked on #3017. The UI.

**Lanes.** Each automated parameter draws its own lane under the audio clip, on
the same disclosure caret the keyframe lanes use — that caret is the DAW
automation triangle. One lane per parameter rather than a selector to swap
between them, so two envelopes can be read and edited without hiding either.
Double-click the line to add a point, drag to shape it, right-click a point to
remove it.

Everything the lane knows comes from the FX registry — range, unit, log scale —
so a log knob is drawn *and dragged* on a log axis, and the curve on screen is
the curve the audio thread plays.

**Panel.** An automated parameter's control is disabled, since a value set there
would be overwritten by the envelope on the next tick. Each automatable
parameter gets a toggle: adding a lane seeds it with a single point at the value
the control already holds, so switching to an envelope never changes the sound,
only where the value comes from. Parameters no envelope can drive have no toggle
at all.

### Things that took more than one attempt

- **Dragging a point did not move it.** The live write deliberately skips the
  refresh, so the stored value does not move under the pointer; the lane needed a
  local draft. Releasing it then snapped back until the draft was made to survive
  until the *prop* changed rather than until the drag ended — which in turn
  needed the parsed attribute cached, or a fresh object each playhead tick threw
  the drag away.
- **Deleting one effect made every later delete a no-op.** The live path persists
  but does not resync the selection, so the panel kept computing from the
  snapshot it was built with. Added a commit that skips the reload *and* resyncs;
  both halves are needed, and they were fighting each other.
- **Automating a volume would have deleted every FX lane.** The volume section
  does not parse the chain, and resolving against an absent chain drops every
  lane it cannot resolve. Now "no chain" means "do not resolve" rather than "drop".
- **Carve glitched the track** by opening a second `AudioContext` just to decode;
  it decodes in an `OfflineAudioContext`, which touches no device. Turning carve
  off now also drops the filters it generated.

`AudioFxGroup` moves into its own module and `AUTOMATION_LANE_H` into its own —
`PropertyPanelFlat` and `timelineLayout` were both at the 600-line studio limit.


vanceingalls:
Superseded by stack #3027. This PR carried the studio UI in one diff (+2397); it is now #3023 lane geometry and plumbing, #3024 the lane component, #3025 the non-reloading write path + volume, #3026 the per-parameter surface — each under 1000 LOC, with the pure lane maths extracted into its own module along the way.

PR #3019, opened 2026-08-04
feat(core): the audio FX registry

vanceingalls:
The registry every other piece of the audio FX work reads from: what effects exist, what parameters each one has, their ranges, units and defaults.

One list rather than knowledge spread across the panel, the graph builder and the renderer — those three have to agree about what a `peaking` node is, and the only way they agree is by asking the same source.

It is also what makes the panel possible without hand-writing a control per effect: the panel is generated from this, so adding an effect here is the whole job of shipping it.

Includes a fix worth calling out — several effects declared parameters that moved nothing. A knob that does not change the sound is worse than a missing one, because the author spends time on it before concluding they cannot hear the difference.

### DIFF
--- a/packages/core/package-subpaths.json
+++ b/packages/core/package-subpaths.json
@@ -86,6 +86,12 @@
       "types": "./dist/compiler/index.d.ts",
       "environments": ["bun", "node"]
     },
+    "./audio-fx": {
+      "source": "./src/audioFx.ts",
+      "runtime": "./dist/audioFx.js",
+      "types": "./dist/audioFx.d.ts",
+      "environments": ["browser", "bun", "node"]
+    },
     "./color-grading": {
       "source": "./src/colorGrading.ts",
       "runtime": "./dist/colorGrading.js",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -100,6 +100,12 @@
       "import": "./src/compiler/index.ts",
       "types": "./src/compiler/index.ts"
     },
+    "./audio-fx": {
+      "bun": "./src/audioFx.ts",
+      "node": "./dist/audioFx.js",
+      "import": "./src/audioFx.ts",
+      "types": "./src/audioFx.ts"
+    },
     "./color-grading": {
       "bun": "./src/colorGrading.ts",
       "node": "./dist/colorGrading.js",
@@ -357,6 +363,10 @@
       "./compiler": {
         "import": "./dist/compiler/index.js",
         "types": "./dist/compiler/index.d.ts"
+      },
+      "./audio-fx": {
+        "import": "./dist/audioFx.js",
+        "types": "./dist/audioFx.d.ts"
       },
       "./color-grading": {
         "import": "./dist/colorGrading.js",
--- a/packages/core/src/audioFx.test.ts
+++ b/packages/core/src/audioFx.test.ts
@@ -3,6 +3,7 @@ import {
   AudioFxChainError,
   defaultAudioFxParams,
   enabledAudioFxNodes,
+  getAudioFxDef,
   HF_AUDIO_FX,
   HF_AUDIO_FX_CHAIN_VERSION,
   HF_AUDIO_FX_IDS,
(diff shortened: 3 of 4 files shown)


terencecho (commented):
**Verdict: LGTM_PENDING_CI** — code is a clean, well-designed root, but Lint / Test / regression shard-4 are all currently FAILURE and a load-bearing future-hazard is present at parse/serialize (see below).

**Public API surface** (`packages/core/src/audioFx.ts`). Downstream stack (#3020+) will bind to all of these:
- Constants: `HF_AUDIO_FX_ATTR` (`:214`), `HF_AUDIO_FX_CHAIN_VERSION` (`:217`), `HF_AUDIO_FX` (`:323`), `HF_AUDIO_FX_IDS` (`:894`).
- Types (public contract — expensive to evolve): `HfAudioFxGroup` (`:219`), `HfAudioFxNumberParam` (`:221`), `HfAudioFxEnumParam` (`:237`), `HfAudioFxParam` (`:246`), `HfAudioFxParamValues` (`:248`), `HfAudioFxDef` (`:250`), `HfAudioFxNode` (`:932`), `HfAudioFxChain` (`:940`).
- Functions: `getAudioFxDef` (`:890`), `defaultAudioFxParams` (`:897`), `normalizeAudioFxParams` (`:912`), `parseAudioFxChain` (`:957`), `enabledAudioFxNodes` (`:995`), `serializeAudioFxChain` (`:1000`). Error class: `AudioFxChainError` (`:945`).
- `mintAudioFxNodeId` / per-node `id` field NOT here — will need to be added when downstream reorder/panel PRs (#3022+) land.

**PR-body claim verification.**
- "14 effects across filters/dynamics/nonlinear/time": counted at `:323-886` — 5+3+2+4 = 14. ✓
- "Parameters carry range, step, unit, default, scale": `HfAudioFxNumberParam` at `:221-235`. ✓
- "Parsing rejects unknown ids rather than skipping": `:979-981` throws, tested `:142-144`. ✓
- "Data only — no browser globals": `web` is a string id (`:262`). ✓
- "Panel is generated from this": labels/units/hints/scale all present on defs (e.g. `freq()` `:265-281` sets `scale: "log"`). ✓
- Fix: chorus `decay` and bitcrush `aa` removed, shelf `q` removed — regression-pinned at `:172-190`. ✓

**Parse/serialize allow-list assessment (LOAD-BEARING).** This is the EXACT miss shape that later caused #3177/#3178 and #3212/#3213 to be load-bearing pairs — and it is built in HERE at the root:
- `parseAudioFxChain` constructs the node literal by hand at `:982-989` (`{ type, enabled, params }`). Any field later added to `HfAudioFxNode` (e.g. `fromPreset`, `fromCarve`, `label`, `presetAmount`) is silently dropped on load unless BOTH allow-lists are edited.
- `serializeAudioFxChain` mirrors this at `:1003-1007`. Silent field-drop on save.
- **Neither list is typed against `keyof HfAudioFxNode`.** TS excess-property checks don't fire on the return object literal, so extending `HfAudioFxNode` with an optional field compiles clean while parse/serialize discard it — the reviewer sees green tests, ships, and the field silently disappears on save→reload.
- Cheap defense at the ROOT (much cheaper than a per-field pair PR every time): type both allow-lists via a `satisfies` guard or a `Required<Pick<HfAudioFxNode, ...>>` helper, and add a structural round-trip pin (see §Test-shape). Without either, every field addition in downstream PRs is a landmine.

**Test-shape audit** (audioFx.test.ts). All tests exercise real mechanism, not "constant is defined" tautology — but:
- "round-trips a chain" at `:135` only asserts `enabled` and `params.gain`. It does NOT do `expect(parse(serialize(chain))).toEqual(chain)`. A stub `parseAudioFxChain` returning `{ nodes: [{ type: "peaking", enabled: true, params: {gain:-6, frequency:1000, q:1} }] }` would pass. Add a structural full-object round-trip.
- `serializeAudioFxChain` has no direct test — the disabled-flag write path at `:1005` is untested.
- Reject-unknown/bad-JSON/bad-version tests (`:142-155`) are strong.

**Type safety.**
- `HfAudioFxNode.type: string` is NOT a discriminated union — `HF_AUDIO_FX_IDS` exists but no `type HfAudioFxId = typeof HF_AUDIO_FX[number]["id"]` alias. Callers constructing chains get zero static help; only runtime `BY_ID.has` catches typos. Cheap to add, huge downstream ergonomics win.
- `normalizeAudioFxParams` on unknown id returns `{}` silently (`:917`) — asymmetric with `parseAudioFxChain` throwing. Deliberate for defensive callers, but worth a JSDoc line so downstream PRs don't add duplicate guards.
- `enabled: node.enabled !== false` (`:984`) treats every non-`false` value (incl. `"false"` string) as enabled. Comment at `:935` only covers absent; document that only literal `false` disables.

**Nits.** (1) `HfAudioFxParamValues = Record<string, number | string>` accepts any key at compile time — combined with untyped `type`, panel/graph builders will lose autocomplete. (2) Consider `HF_AUDIO_FX satisfies readonly HfAudioFxDef[]` to preserve literal id types.

**CI.** `Lint`, `Test`, `regression-shards (shard-4, ... audio-mux-parity)`, and `Semantic PR title` are FAILURE; several other shards are CANCELLED. Since this PR only adds greenfield files, most of that reads as stack-wide drift, but Lint/Test could plausibly be `audioFx.ts` itself — please confirm before merge that `oxfmt --check`/vitest are green on `packages/core/src/audioFx.{ts,test.ts}`. Semantic PR title is title-format-only ("feat(core): the …" starts with lowercase article).

**Note on retrospective vs actionable.** If this PR is being merged now (fresh review post-Vance's authorization), the parse/serialize typing hardening is *actionable and would eliminate the recurring load-bearing-pair pattern for the whole downstream stack*. If #3019 is already merged, the finding is retrospective — flagged for future greenfield roots. Assuming actionable given Via's "vance authorized 17 more slices" framing.

— Review by tai (pr-review)

terencecho (approved):
**Converting my prior `--comment` at `8e0f63d` to `--approve` at the same head.**

The prior review's architectural observation on hand-typed `parseAudioFxChain`/`serializeAudioFxChain` allow-lists (`audioFx.ts:982-989/1003-1007`) was seed-level, not blocker-level — it named a class of downstream fixes needed at consumer boundaries. Those load-bearing follow-ups (#3178, #3213) already landed, so the root-PR's code-review substance is clean at `8e0f63d`.

Head SHA unchanged since my original review. Required checks at `8e0f63d` are currently green (`Lint`, `Test`, `regression`, `regression-shards`, `Preflight (lint + format)`, `File size check`, `Typecheck`, `preview-regression`, `player-perf` all report SUCCESS in the latest workflow run per PR's `statusCheckRollup`). Earlier-run failures shown in the raw check-runs list are stale and superseded.

No new substantive verifications warranted — the delta since my prior sign-off is zero. Approve stands on the same evidence I documented at `8e0f63d` plus the confirmed landing of #3178 / #3213.

— Review by tai (pr-review)

PR #3020, opened 2026-08-04
feat(core): Web Audio graphs for the FX registry

vanceingalls:
The Web Audio graph for each registry entry — turning a `data-fx-chain` node into real audio nodes, wired in signal order.

This is the piece both runtimes share. The studio builds it in a live `AudioContext` and the engine builds the same graph in an `OfflineAudioContext`, so what an author hears while scrubbing is what gets rendered. One implementation, two hosts; a second one would drift and the drift would only show up in a finished video.

Four effects needed worklets rather than native nodes (compressor, limiter, gate, bitcrush) because Web Audio has no primitive for them.

Fixes folded in: reverb level, phaser wiring, per-channel dynamics, and a one-pole filter case — each found by listening to the graph rather than reading it.

### DIFF
--- a/.fallowrc.jsonc
+++ b/.fallowrc.jsonc
@@ -146,6 +146,17 @@
       "file": "packages/studio/src/utils/studioHelpers.ts",
       "exports": ["resolveDroppedAssetDimensions"],
     },
+    // Audio FX worklets sit near the bottom of the audio stack: the worklet
+    // source and its test reset are consumed by the runtime and engine PRs
+    // upstack, so a per-PR audit against the merge base sees them as unused.
+    {
+      "file": "packages/core/src/audio/audioFxWorklets.ts",
+      "exports": ["AUDIO_FX_WORKLET_SOURCE", "__resetAudioFxWorkletsForTests"],
+    },
+    {
+      "file": "packages/core/src/audio/audioFxGraph.ts",
+      "exports": ["ensureAudioFxWorklets"],
+    },
     // drawElementService is the bottom of the fast-capture Graphite stack
     // (#1917): its consumers (frameCapture in #1919) land two PRs upstack, so
     // a per-PR audit diffing against the merge base sees these exports as
--- a/packages/core/package-subpaths.json
+++ b/packages/core/package-subpaths.json
@@ -86,6 +86,12 @@
       "types": "./dist/compiler/index.d.ts",
       "environments": ["bun", "node"]
     },
+    "./audio-fx": {
+      "source": "./src/audioFx.ts",
+      "runtime": "./dist/audioFx.js",
+      "types": "./dist/audioFx.d.ts",
+      "environments": ["browser", "bun", "node"]
+    },
     "./color-grading": {
       "source": "./src/colorGrading.ts",
       "runtime": "./dist/colorGrading.js",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -100,6 +100,12 @@
       "import": "./src/compiler/index.ts",
       "types": "./src/compiler/index.ts"
     },
+    "./audio-fx": {
+      "bun": "./src/audioFx.ts",
+      "node": "./dist/audioFx.js",
+      "import": "./src/audioFx.ts",
+      "types": "./src/audioFx.ts"
+    },
     "./color-grading": {
       "bun": "./src/colorGrading.ts",
       "node": "./dist/colorGrading.js",
@@ -357,6 +363,10 @@
       "./compiler": {
         "import": "./dist/compiler/index.js",
         "types": "./dist/compiler/index.d.ts"
+      },
+      "./audio-fx": {
+        "import": "./dist/audioFx.js",
+        "types": "./dist/audioFx.d.ts"
       },
       "./color-grading": {
         "import": "./dist/colorGrading.js",
--- a/packages/core/src/audio/audioFxGraph.test.ts
+++ b/packages/core/src/audio/audioFxGraph.test.ts
@@ -1,0 +1,276 @@
+// @vitest-environment happy-dom
+import { describe, expect, it } from "vitest";
+import { defaultAudioFxParams, HF_AUDIO_FX, type HfAudioFxChain } from "../audioFx.js";
+import {
+  buildFxChain,
+  buildFxNode,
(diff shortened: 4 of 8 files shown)


terencecho (commented):
**Verdict: LGTM_WITH_NITS**

Clean, deliberate work. One graph builder per registry `web` id, `update()` on every handle so param changes reparameterise in place, and a shape signature that forces a rebuild when node type changes. Ships with the four workflow fixes the PR body advertises, each anchored by a test.

**Verified PR-body claims (at `c3291f2`)**

- One builder per `web` id, updater on each handle: `packages/core/src/audio/audioFxGraph.ts:61-66` (`FxNodeHandle`) and BUILDERS map at `:334-349` — 14 `web` ids, 14 entries; every registry `web` id in `packages/core/src/audioFx.ts` (grep hit lines 133/144/152/160/168/253/303/375/433/473/516/565/636/687) is present.
- Update in place vs rebuild: `buildFxChain` returns `update(chain): boolean`; false when `shapeOf(next) !== shape` (`audioFxGraph.ts:406-419`). Tests at `audioFxGraph.test.ts:194-215` exercise value-only, node-add, node-remove, bypass, and pole-count-change cases.
- One-pole = IIRFilterNode, two-pole = BiquadFilterNode: `audioFxGraph.ts:91-107`, dispatched at `:376-378`. Test `:129-136`.
- Four worklets (compressor/limiter/gate/bitcrush): `audioFxWorklets.ts` registerProcessor calls at hf-compressor / hf-limiter / hf-gate / hf-bitcrush; wired at `audioFxGraph.ts:334-346`.
- data: URL not blob: URL for worklet module: `audioFxWorklets.ts:198-204`, with reason in the docstring.
- Fixes: reverb level normalisation (`audioFxGraph.ts:47-56`, test `:254-263`), phaser in_gain/out_gain as trims not wet/dry (`:174-231`, test `:265-274`), per-channel envelope followers (`audioFxWorklets.ts` `EnvBank` + comment), one-pole cutoff forces rebuild (`audioFxGraph.ts:355-362`, test `:288-303`).

**#3019 consumer verification**

Imports from `../audioFx.js`: `getAudioFxDef`, `normalizeAudioFxParams`, `HfAudioFxChain`, `HfAudioFxParamValues` (`audioFxGraph.ts:11-16`). Tests additionally consume `HF_AUDIO_FX`, `defaultAudioFxParams`. No duplication — parameter clamp/default logic stays in #3019; this PR only reads normalised values via `.value =`.

**Params flow**

`.value =` throughout — no `setValueAtTime`. Every non-worklet builder pulls its registry params directly: biquad reads `frequency/q/gain`, waveshaper reads `type/threshold/oversample/output` (with dB→linear at `:141` and `:151`), delayFeedback reads `time/feedback/mix` (ms→s at `:166`), chorusLfo reads `delay/depth/speed/mix` (ms→s at `:196-197`), allpassPhaser reads `delay/decay/speed/type/in_gain/out_gain`, convolver reads `size/damping/wet/dry`. Worklet builders forward the whole param bag through `processorOptions`/`postMessage`. One-pole drops `q` (fixed 6 dB/oct slope; the drop is silent — worth a code comment). No stale param that the registry declares but the builder ignores.

**Dispose/lifecycle audit**

`FxNodeHandle.dispose` contract: disconnect every internally-created node. Delivered by biquad (via `simple`), onePole, worklet, waveshaper (3 nodes), delayFeedback (6 nodes), convolver (5 nodes).

**Nit (dispose leak):** `chorusLfo` dispose (`:212-217`) and `allpassPhaser` dispose (`:236-241`) call `lfo.stop()` inside try/catch but do *not* call `lfo.disconnect()`. The oscillator still holds an outgoing edge to `depth`. Low impact (GC once the depth node drops), but strictly violates the "disconnect everything you made" invariant. **This is EXACTLY the leak that later got explicitly fixed in wa-20b2 #3183 via `retireLfo` — landing the disconnect at root would preempt that whole downstream fix.**

**Worklet retry**

`ensureAudioFxWorklets` caches `modulePromise` at module scope (`audioFxWorklets.ts:191-207`). A rejected promise sticks for the process lifetime — no retry path. Given the source is a data: URL (no network), realistic failure modes reduce to insecure-context or parse error, both deterministic; fine but worth a comment. **This sticky-rejection concern is what later drove #3175's registration-retry fix — again, cheap to land at root.**

**Test-shape verdict**

`FakeNode.disconnect` sets a single boolean rather than popping the specific edge (`audioFxGraph.test.ts:41-43`). The dispose invariant test at `:218-222` uses `c.created.every((n) => n.disconnected)` — it catches "did every created node get `.disconnect()` called at least once", but does not catch partial disconnects. The dispose test also runs only against `chain("delay")`, so it would NOT catch the chorus/phaser LFO omission above — extending the test to iterate `HF_AUDIO_FX` (like the build test does) would surface these.

**Nits**

1. Add `lfo.disconnect()` inside chorus/phaser dispose lambdas — preempts #3183's fix.
2. Extend the "disconnects everything" test to loop over every `HF_AUDIO_FX` entry.
3. Comment near `onePoleBuilder` noting that `q` is intentionally ignored.
4. Comment near `modulePromise ??=` noting rejection is sticky by design.

**CI status**

All required checks green at `c3291f2` — Preflight (lint+format), Build, Lint, Typecheck, Fallow audit, Producer unit + integration tests, Test, regression-shards 1-9, Perf lanes (load/fps/scrub/drift/parity), Preview parity, Windows tests, Studio smoke + timeline viewport gate, CLI smoke, SDK unit+contract+smoke, CodeQL. No stack-wide drift.

— Review by tai (pr-review)

terencecho (approved):

terencecho (approved):
Refreshing my prior `--approve` at the SAME head SHA `c3291f29` — a graphite force-push touched the ref (SHA unchanged) and `require_last_push_approval: true` dismissed my earlier stamp from 2026-08-12T01:55:03Z. Code content is byte-identical to what I already signed off. CI at `c3291f29` per `statusCheckRollup` (deduped by name-latest): 46 SUCCESS, 8 SKIPPED (advisory), 0 in-progress, 0 FAILED. Converting BLOCKED → APPROVED with a fresh signature.

— Review by tai (pr-review)

PR #3021, opened 2026-08-04
feat(engine): render the FX chain offline, and the carve analysis behind it

vanceingalls:
Rendering the FX chain offline, so a mix survives the trip from preview to file.

The engine builds the same graph as the studio in an `OfflineAudioContext` and renders it ahead of encode. Without this the effects are a preview-only illusion — audible while scrubbing, absent from the MP4.

Also lands the voiceover carve analysis (`audioCarve.ts`): the measurement that finds which bands a voice occupies so a music bed can be dipped in exactly those and nowhere else.

Three fixes, all about the render being honest rather than merely producing a file:
- keep the FX render 16-bit, stereo and correctly sized
- report an FX render failure as an audio error rather than letting it pass silently
- a chain that cannot be realised fails the render instead of shipping a dry track that sounds plausible and is wrong


terencecho (commented):
**Verdict: LGTM pending CI — code is sound; the new tests fail in the Test lane and must be green before merge.**

**PR-body claims (verified at head `17c654d0`):**
- "Same graph as the studio in an OfflineAudioContext" — verified: `packages/core/stubs/audio-fx-runtime-entry.ts:24-27` imports `buildFxChain`, `chainNeedsWorklets`, `ensureAudioFxWorklets` from #3020's `audio/audioFxGraph.ts`. Real reuse, no duplicate builders.
- "Keep the FX render 16-bit, stereo and correctly sized" — verified: `packages/engine/src/services/audioFxRender.ts:1031-1060` writes `WAVE_FORMAT_PCM` bits=16, `channels` param passed through; `deinterleave`/`interleave` preserve plane count. Header math (`sampleRate*channels*2` byte rate, `channels*2` block align) is correct.
- "Report FX render failure as an audio error" — verified: `audioMixer.ts` (line ~1311) rethrows `AudioFxRenderError` out of the per-track catch; `audioStage.ts:1408-1430` catches `processCompositionAudio` rejections into `audioError` field and preserves abort shape via `assertNotAborted()`. Positive + negative pin tests in `audioStage.test.ts`.
- "Chain that cannot be realised fails the render" — verified: `applyAudioFxChain` throws `AudioFxRenderError` on missing input, empty output, or wrapped browser errors.

**OfflineAudioContext correctness:** `audio-fx-runtime-entry.ts:679-700` constructs `new OfflineAudioContext(channels, frames, sampleRate)` with `channels = max(1, planes.length)`, `frames = planes[0]?.length ?? 0`. Awaits `startRendering()`, copies each `getChannelData(c)` into a fresh `Float32Array`. Explicit doc note that reverb/delay tails are cut at input length — acknowledged product decision, matches ConvolverNode+unit-energy normalisation in #3020. `chainNeedsWorklets` → `ensureAudioFxWorklets(ctx)` before source connect. File://-URL host page for secure-context AudioWorklet is a good touch.

**Carve mechanism (`audioCarve.ts:337-604`):** FFT=4096, HOP=2048 (75% overlap), Hann window, Welch-style spectrum average, radix-2 iterative FFT. Third-octave binning across nine `CANDIDATE_CENTERS_HZ` from 160-6000 Hz. Scoring in dB minus `intelligibilityPenaltyDb` (Gaussian around 2 kHz, authority 30 dB * bias). Dead bands score `-Infinity` so a well-favoured empty band can never outrank a real one. Depth-per-band from `pow(10, (scoreDb - topDb)/10)`, floored at `maxCutDb/2`. Fresh `Float64Array(FRAME)` per hop; ~10 hops for a 500 ms fixture, no allocation hotspot at analysis-time scale.

**Test-shape (render):** Real deterministic inputs. `readWav/writeWav` roundtrip with 16-bit quantisation tolerance, stereo interleave preservation, integration with `applyVolumeEnvelopeToWav` (the load-bearing reason for the format flip), and full-scale clamp. Browser tests use physical assertions: -30 dB peaking at 440 Hz drops rms > 15 dB on a 440 Hz tone; compressor threshold=-40/ratio=20 drops rms > 3 dB; multi-effect + reverb produces non-empty output.

**Test-shape (carve):** Known-input assertions — `bands: 1` on tilted voice must reach ≥1000 Hz at default bias (calibrates the "knob that only works at extreme is broken" invariant); 250 Hz vs. 1600 Hz two-tone with 9 dB tilt pins that bias-0 picks 250 Hz and default-bias picks 1600 Hz. Not shape-checks — genuine numerical pins with commentary that names the physics.

**Consumer relationship with #3019/#3020:** Runtime bundle IIFE (`build-audio-fx-runtime.ts`) mirrors the position-edits-render pattern already present. `audioCarve` uses `defaultAudioFxParams`, `HF_AUDIO_FX_CHAIN_VERSION`, and produces `HfAudioFxChain` from #3019. Both engine and studio consume the same `buildFxChain`. Zero duplication.

**Chunked-PCM:** Not chunked in this PR — samples cross the CDP boundary as base64 in a single `page.evaluate`. Fine for typical clip lengths; consistent with #3174 arriving as a later, targeted fix.

**Scope-containment:** +1236 LOC is contained: core FX runtime bundle + carve module + engine render service + mixer wiring + producer stage wiring + `fallowrc` allowlist + package-subpaths. No unrelated changes.

**Nits:**
- `audioFxRender.ts:1015` divisor `32768` on read vs. `32767` scale on write — asymmetric by ~0.00003 dB, standard PCM16, ignore.
- Test at `audioFxRender.test.ts:801-823` calls literal `execFileSync("ffmpeg", ...)` — passing `process.env.FFMPEG_BIN ?? "ffmpeg"` would let the same `prepare-ffmpeg-bin` action wiring work without a system install.

**CI status:** Required `Test` lane is FAILING. Four `audioFxRender.test.ts` failures, all environment:
1. `readWav / writeWav > reads 16-bit PCM` — `spawnSync ffmpeg ENOENT`. The `Test` job (`.github/workflows/ci.yml:263-291`) uses `prepare-ffmpeg-bin` which only sets `FFMPEG_BIN` (and writes a self-referential stub when ffmpeg is absent) — it does NOT `apt-get install ffmpeg`. Only `producer-source-tests` does. This is the first engine test to `execFileSync("ffmpeg")` literally.
2-4. Three `browser render` tests — `EACCES` on `~/.cache/hyperframes/chrome/.../chrome-headless-shell`. This is the first engine test to actually launch a browser via `acquireBrowser`; the Chromium in Test-lane cache is not `+x`.

CodeQL FAILURE is stale (2026-08-07, before last push, not re-run). Semantic PR title fails on the "feat(engine):" prefix — likely fine. Stack-wide CANCELLED regressions are supersede-drift, not real.

**Ask before merge:** either (a) add `sudo apt-get install -y --no-install-recommends ffmpeg` + a `chmod +x` on the Puppeteer cache to the `Test` job, or (b) gate these four tests on availability (`it.skipIf(!hasFfmpeg)` and `it.skipIf(!hasChromium)`), or (c) move the browser tests to `producer-source-tests`-style integration lane. Code review approves; block is CI only.

— Review by tai (pr-review)

PR #3022, opened 2026-08-04
feat(studio): the FX panel, generated from the registry

vanceingalls:
The FX panel in the studio, generated from the registry rather than hand-written.

Every control — its label, range, step and unit — comes from the registry entry, so an effect added there arrives in the panel with working controls and nothing to wire up. The alternative is a bespoke component per effect and a slow divergence between what the panel offers and what the graph can do.

Shown only on audio tracks, since the chain lives on the audio element.


terencecho (commented):
**Verdict: BLOCK** — the required `File size check` fails on this PR's changes (`packages/studio/src/components/editor/PropertyPanelFlat.tsx` has 612 lines, max 600). Mechanical fix (~5 min), not a design problem. Code itself is LGTM.

**PR-body claims — all verified at head `10311902`:**

- Registry-driven controls: `propertyPanelFxSection.tsx:18` imports `HF_AUDIO_FX` from `@hyperframes/core/audio-fx`; add menu iterates it (`HF_AUDIO_FX.filter((d) => d.group === g)`, L60-63); nodes render via `def.params.map((p) => <FxParamRow …/>)` (`propertyPanelFxControls.tsx:139`). Nothing per-effect.
- Log-scale mapping: `toSlider`/`fromSlider` gated on `p.scale === "log"` (`propertyPanelFxControls.tsx:22, 30`).
- Range clamp: `FxParamRow.handleNumber` — `Math.min(p.max, Math.max(p.min, raw))` (L52). Tested (typing 999999 into peaking frequency clamps to 20000, test L145).
- Reorder = first-class: `.hf-fx-move` up/down buttons wired to `moveNode` (`Section.tsx:100-104`).
- Carve is not a chain entry: separate `.hf-fx-carve` block with source picker (L215-283). Test asserts `items).not.toContain("Voiceover carve")` (test L166).
- Audio-only gate: `affordances.ts:213` — `audioFx: facts.tag === "audio"`. Tested for audio/video/div (`affordances.test.ts:200-217`).
- Attribute persistence: `AudioFxGroup` in `PropertyPanelFlat.tsx:571-609` writes `HF_AUDIO_FX_ATTR` / `HF_AUDIO_CARVE_ATTR` via `onSetAttribute` — no new server route. Parse-failure falls back to empty chain (`.tsx:588-591`), matching the claim.
- Collapsed summary: `audioFxSummary` returns e.g. `"2 effects + carve"` (L548-563).

**Data-driven verdict:** clean. Grep of the diff for `=== "compressor"|case "highpass":`-style branches returns zero hits. Every effect flows through `getAudioFxDef(node.type)` + `def.params.map(...)`.

**#3019 relationship:** consumer-only. Imports registry types (`HfAudioFxDef`, `HfAudioFxNumberParam`, `HfAudioFxParam`, `HfAudioFxParamValues`, `HfAudioFxChain`, `HfAudioFxGroup`, `HfAudioFxNode`) and factories (`HF_AUDIO_FX`, `getAudioFxDef`, `defaultAudioFxParams`, `parseAudioFxChain`, `serializeAudioFxChain`) from `@hyperframes/core/audio-fx`. Carve types from `@hyperframes/core/audio-carve`. No local re-definition.

**Test-shape audit:** parameterized — `expect(items).toHaveLength(HF_AUDIO_FX.length)` and `for (const def of HF_AUDIO_FX) expect(items).toContain(def.label)` (test L96-100), so a new effect added upstream is automatically covered. Bypass/reorder/remove/clamp all tested. 16 tests, all pass in CI.

**Accessibility:** slider gets `aria-label={param.label}` and native `<label>` wrapping. Bypass has `aria-pressed`, expand has `aria-expanded`. Solid baseline. **Nit:** log-scale sliders set `value={toSlider(param, current)}` — a screen reader announces the linear-position number, not the actual parameter value. Consider `aria-valuenow={current}` + `aria-valuetext={display(param, current) + (param.unit || "")}` on log-scale params (frequency, time). No Enter/Escape commit/cancel on the number input — it commits on every keystroke.

**State:** parent-owns via `onChainChange`/`onCarveChange` → `onSetAttribute` writes attribute; panel re-parses on next render. Rapid slider drag: each event mutates a single `params[key]` off the current parsed chain, so no dropped-update race in practice.

**Nits (non-blocking):**
- Add-menu item buttons don't propagate `disabled` — a read-only panel with the menu open can still add effects (`Section.tsx:190-201`).
- Move-button `disabled` behavior not covered by the read-only test (test L204 queries only `.hf-fx-bypass, .hf-fx-remove`).
- `chain` / `carve` / `sourceOptions` are IIFE'd fresh per render in `AudioFxGroup`; `sourceOptions` also re-queries the DOM — cheap but memoizable.

**CI status:** 3 failures on head SHA:
- `File size check` — **PR-specific, blocking**: `packages/studio/src/components/editor/PropertyPanelFlat.tsx has 612 lines (max 600)`. This PR adds 96 lines (74 in the file body via `audioFxSummary` + `AudioFxGroup`) and pushes it over. Fix: move `audioFxSummary` and `AudioFxGroup` into `propertyPanelFxSection.tsx` (or a small `propertyPanelFxSectionAdapter.tsx`), leaving `PropertyPanelFlat.tsx` with just the import and the `groups.push({ id: "audio-fx", … })` block.
- `Test` — **stack-wide, not this PR**: fails in `packages/engine/src/services/audioFxRender.test.ts:95` with `spawnSync ffmpeg ENOENT`. That file was introduced by #3021 (`wa-3-fx-render`), the base branch; it's a runner/infrastructure gap, unrelated to the FX panel.
- `Fallow audit` — needs a look at the PR-comment output the workflow references; can't determine scope from log alone.

Once the file-size split lands, this reads as a clean LGTM.

— Review by tai (pr-review)

terencecho (approved):
Refreshing to `--approve` at `dad13c8f`. Prior verdict was BLOCK-on-`File size check` (`PropertyPanelFlat.tsx` 612 > 600 cap) with the code itself LGTM; every one of those conditions is now satisfied.

**BLOCK resolved.** `33100d2e refactor(studio): lift audioFxSummary out of PropertyPanelFlat` — 18-line net drop on `PropertyPanelFlat.tsx` via `audioFxSummary` moved to its own module. Verified line-by-line the function body is identical to the pre-move version (same guts, same fallback-to-`"unreadable"` on parse failure, same `n.enabled !== false` filter, same summary shape) — the import replaces the definition, no drift. New file `audioFxSummary.ts` carries a module docstring explaining why it's split out. **Deliberately narrower than the full `AudioFxGroup` extraction** because "every later commit in the stack edits AudioFxGroup, so moving it here would collide with each of them" — pragmatic call, exactly the minimum cut that clears the cap. New line count 595. `File size check` at head reads SUCCESS.

**Canary gate wired.** `956ae2fc feat(core): register the audio-fx-rack canary at 0%` + `550812a0 feat(studio): put the audio FX rack behind its canary`:
- Registry entry: `name: "audio-fx-rack"`, `percentage: 0`, `owner: "vance"`, `sunsetAfter: "2026-11-15"`, with a description that explicitly names the deliberate scope-limit: *authoring surface only — a composition that already carries `data-fx-chain` still plays and renders it*. Correct-shaped choice: a canary should stage who can reach the feature, not silently null a written attribute out from under an off-cohort user.
- Gate site is a single-line predicate: `if (sections.audioFx && isCanaryEnabled("audio-fx-rack"))`.
- Pinned by three tests in `canary.test.ts`: (a) `audio-fx-rack` is registered; (b) ships at 0%; (c) `sunsetAfter` is a valid ISO date in the future. A ramp has to break the second one out loud — correct level of forcing function.
- Candid gap acknowledged in the commit message rather than papered over: no unit test for the gate itself, because `PropertyPanel.test.tsx`'s harness never renders the Audio FX group for its audio fixture (with or without the gate). Wiring the harness to mount the section is a separate change. Reasonable to defer.

**Fallow / audit findings cleared.**
- `d9df8a15 fix(studio): drop dead __testables export` — `propertyPanelFxControls.tsx` 2-line removal, Fallow-flagged (no test imports it).
- `7c410fd0 fix(core,studio): clear remaining Fallow audit findings`:
   - FxSection's per-node row split into `FxNodeRow` + `FxNodeControls` — CRAP score (31.6, threshold 30) split across two smaller units, not just moved wholesale to a sibling. Net `propertyPanelFxSection.tsx` delta +139/-69 = +70 lines but two smaller units.
   - Test-file dedupe: `openAddMenuItems(host)` extracted from two duplicate blocks in `propertyPanelFxSection.test.tsx`. Clean.
   - Build-script consolidation: `build-audio-fx-runtime.ts` (50 lines) + `build-position-edits-render.ts` (52 lines) → `build-inline-artifact.ts` (97 lines) with a config `TARGETS` map keyed by CLI arg (`audio-fx` / `position-edits`). The prior two scripts differed only in 5 names — Fallow's duplication check was re-flagging on every unrelated line shift. Legitimate consolidation.
   - `.fallowrc.jsonc` exemptions: (a) `canary.test.ts` for its deliberate independent `rawFnv` re-implementation used to cross-check `canaryBucket` (its own docstring explains why importing the function-under-test would defeat the assertion); (b) 7 property-panel test files sharing the `renderInto` React-mount helper — pre-existing across 9 files, 2 outside this stack, not introduced here. Both exemptions carry explanatory comments referring to the same class of exemption that already exists for `slideshowPanelHelpers.ts`.
- `dad13c8f fix(ci): allowlist the build-script consolidation in the no-main-deletions guard` — 2 narrow entries in `ALLOWED_DELETIONS`, each with an explicit reason ("merged into build-inline-artifact.ts"). Not a broad allowlist expansion.

**Test-lane isolation.** `384fdc44` + `7d79370f` fix the `Test` lane flake I called out in my prior review: instead of assuming ffmpeg / Chrome are present, the browser-render suite now probes the binary (runs `--version` and requires exit 0, matching the ffmpeg-suite pattern) and skips exactly 3 cases when the browser is absent while running the other 8. Commit messages explicitly check both directions ("with a browser all 11 cases run and pass; with `HYPERFRAMES_BROWSER_PATH` pointed at a broken binary exactly 3 skip and the other 8 still run") — the anti-pattern that a "guard that silently skipped everything would have looked identical in CI" is called out and defended against. `7fdf4dbf` suppresses one `js/insecure-temporary-file` CodeQL alert inline with rationale ("dataflow reach `tmpdir()` and not the mkdtemp in between") that matches the repo's existing convention (`planV2.ts:222` for a different reason on the same rule). Honest self-correction in the commit body about first mis-reporting the alert as not-real because it was queried against the default-branch alert list, not `refs/pull/N/merge` — that's a good calibration note.

**CI at `dad13c8f`** per `statusCheckRollup` (deduped by name-latest): 42 SUCCESS / 8 SKIPPED / 0 IN_PROGRESS / 0 FAILURE — including `File size check`, `Fallow audit`, `Lint`, `Format`, `CodeQL`, `Producer: unit tests`, `Producer: integration tests`, `Preflight`, `Preview parity`, `Perf: {drift, fps, load, parity, scrub}`, `Render on windows-latest`, `SDK: unit + contract + smoke`, `Build`, `CLI smoke (required)`.

**Prior non-blocking nits (recap for the record).** None of these blocked then and they don't block now, but capturing here since I didn't see them addressed in the delta:
- Add-menu item buttons don't propagate the panel's `disabled` state — read-only panel with the add-menu open still lets you add.
- Move-button `disabled` behavior isn't covered by the read-only test (test queries only `.hf-fx-bypass`, `.hf-fx-remove`).
- `chain` / `carve` / `sourceOptions` are IIFE'd fresh per render in `AudioFxGroup`; `sourceOptions` also re-queries the DOM. Cheap but memoizable.
- Log-scale sliders set `value={toSlider(param, current)}` — a screen reader announces the linear-position number, not the parameter value. `aria-valuenow={current}` + `aria-valuetext={display(param, current) + (param.unit || "")}` on log-scale params would tighten it.
- Number input commits on every keystroke; Enter/Escape commit/cancel would be nicer UX.

Fine to keep as separate follow-ups.

— Review by tai (pr-review)

PR #3023, opened 2026-08-04
feat(studio): the geometry and plumbing behind automation lanes

vanceingalls:
The geometry and plumbing under automation lanes: mapping a lane's points to pixels, and carrying the audio FX attributes onto every timeline row so a lane knows which parameter it is drawing.

Split from the lane component itself (upstack) because the maths is worth testing without a DOM — the value/time to x/y mapping is where off-by-one errors hide, and they are invisible in a screenshot but obvious in a unit test.

Includes an eviction fix for the data cache: it grew without bound across a long editing session.


terencecho (dismissed):
**LGTM_WITH_NITS** — pending shard-3 completion; one test-shape nit that does not gate.

Geometry pure and testable; the cache extraction is clean; the DOM read is symmetric on read/write. The one nit is that the eviction test does not exercise the eviction boundary it names.

**Seed audit (downstream cohort):**

- *`HfAudioFxNode` allow-list (root #3019)* — this PR imports `parseAudioFxChain`, `getAudioFxDef`, `HfAudioFxChain` from `@hyperframes/core/audio-fx` but does **not** add fields to `HfAudioFxNode` or touch `parseAudioFxChain` / `serializeAudioFxChain`. Not aggravated.
- *LFO dispose (later fixed by #3183)* — `audioFxGraph.ts` untouched (only referenced in `.fallowrc.jsonc` for `ensureAudioFxWorklets`). Neutral.
- *Rate-change reschedule (#3016 seed, later fixed by #3174)* — no touch to `WebAudioTransport` / scheduling. Lane geometry is pure and upstream of transport. Neutral.
- *`PropertyPanelFlat.tsx` 600-line cap (Batch 9 #3017)* — not touched. Bonus: this PR extracts `TimelineElement` **out** of `playerStore.ts` into `store/timelineElement.ts` explicitly citing the 600-line ceiling (`timelineElement.ts:5-6`) — same hygiene applied to a different at-cap file.

**Correctness spot-checks:**

- LRU eviction (`automationLaneData.ts:36-49`): `Map` insertion-order + delete-and-reinsert on hit + `keys().next()` on evict is a correct LRU. Cache key `${raw}${fxChain ?? ""}` (`:75`) correctly discriminates on chain, so a chain edit re-parses.
- `withLane` (`automationLaneGeometry.ts:97-107`): all four cases (exists×empty, exists×non-empty, !exists×empty, !exists×non-empty) resolve correctly; order preservation rationale on the comment is sound — moving edited lane to end would switch out from under a "first lane" fallback.
- `toUnit` / `fromUnit` (`automationLaneGeometry.ts:70-88`): log-branch guarded on `min > 0 && value > 0`; zero-width guard `max <= min` returns 0; clamp on `fromUnit` prevents overshoot. Round-trip covered by `automationLaneGeometry.test.ts:34`.
- Expanded-child projection (`useExpandedTimelineElements.ts:154-156`): `fxChain` / `automation` join `hidden` / `timelineLocked` / `timelineRole` as host-derived state — the panel-vs-timeline divergence the comment describes is the real bug this closes.
- DOM parse symmetry (`timelineDOM.ts:141-144`, `:341-347`): both the manifest-translation path and the raw-DOM path now read `data-fx-chain` and `data-automation`, matching the write side elsewhere in the stack.

**Test-shape audit:**

Falsifiability check on the new tests — every assertion I looked at flips under the old code, with one exception:

- `automationLaneData.test.ts:24 "keeps a hot entry alive when other elements push the cache past its limit"` — **the test does not actually push past `CACHE_LIMIT`.** Loop inserts 40 distinct entries; combined with `hot`, cache tops out at 41. `CACHE_LIMIT` is 64 (`automationLaneData.ts:28`), so eviction never fires under either strategy — the pre-fix "clear whole map at LIMIT" branch would ALSO pass this test. Bump the loop to ≥`CACHE_LIMIT` (e.g. 80) if you want the test to guard the regression its name promises. **Nit; behavior is correct, the test just under-verifies.**
- All other tests (`automationLaneGeometry.test.ts`, `useAutomationLanes.test.tsx`, `timelineDOM.test.ts:299-329`) pin behavior with falsifying assertions — the "n1 lane dropped when effect deleted" case and the "labels an fx target with its effect and knob" case in particular would break under the old prop-panel-only path.

**CI:** 29+ required checks green (Typecheck, Lint, Tests on windows, Producer unit/integration, Studio smoke/viewport, 9 regression-shards). Only `regression-shards (shard-3, overlay-montage-prod style-3-prod vignelli-stacking, in-process)` is still `pending` — not a failure, and the other 9 shards passed. If shard-3 turns red, re-check against the shard flake pattern before treating as PR-caused.

— Review by tai (pr-review)

PR #3024, opened 2026-08-04
feat(studio): the automation lane itself

vanceingalls:
The automation lane component — the row under a track that draws an envelope and lets you edit it.

Builds on the geometry from the previous PR, which is why this one is mostly rendering and interaction rather than maths.

Includes a fix for disclosure: a track could not open its automation without a tween running, so lanes on a static clip were unreachable.


terencecho (dismissed):
**LGTM.** Empirically verified the whole diff and cross-referenced upstream helpers at head `5328c31`.

**PR-body claim check:**
- "renders and interaction, not maths" — confirmed. Geometry (`PAD_X`, `GRAB_PX`, `POINT_MERGE_SEC`, `DRAW_SAMPLES`, `laneFor`, `withLane`, `toUnit`/`fromUnit`, `formatValue`) all imported from `automationLaneGeometry.ts` (Batch 9 land). No local re-definitions of the automation shape — `HfAutomation`/`HfAutomationLane`/`HfAutomationPoint`/`AutomationRange` come from `@hyperframes/core/audio-automation`.
- "disclosure fix" — `useTimelineTrackLayout.ts:20-22` + `:43-52` + `TimelineTrackHeader.tsx:314-317/:335` widen the "has something to disclose" predicate to `laneCounts + automationLaneCount`, unblocking a static-clip audio track with only envelopes. `automationLaneCountOf` is exported and reused by `useTimelineRowHeights` (replaces the old inline `elementAutomationLanes(active).length`) so both callers share the cached parse — good "no drift" comment at `useTimelineTrackLayout.ts:19`.

**Downstream cohort seeds:**
- Root #3019 `HfAudioFxNode` allow-list — N/A: this PR reads a chain, doesn't add a node field.
- `PropertyPanelFlat.tsx` 600-line cap — untouched.
- #3016 rate-capture-once — N/A: playhead comes in as a prop (`playheadSec` at `TimelineAutomationLane.tsx:114`), no local GSAP/rate subscription.
- Batch 9 #3015 model reuse — clean; imports the shared model.

**Component-lane specific:**
- Effect cleanup: two `useEffect`s (draft-clear + target-reset). Neither subscribes to anything — both are pure state-derived, no cleanup needed. No infinite loop: draft-clear only fires when `basedOn !== automation`, and next render short-circuits.
- Pointer capture: `setPointerCapture` on `e.target` at `:670` is safe — browsers auto-release on `pointerup` for that `pointerId`; capture retargets while still bubbling to the SVG's React handlers. `onPointerCancel` mirrors `endDrag` so a cancelled gesture still clears `dragIndex`/`hint`.
- Coordinate conversion: `xOf`/`yOf` and inverse `pointAt` both consistently subtract `PAD_X`/`pad` — SVG width is `widthPx + PAD_X*2` and `leftPx - PAD_X`, so screen-position ↔ clip-time is invariant across the inset (test `keeps an end point clear of the lane's edges` pins it).
- rAF batching: `onPointerMove` is single-`setDraft` + one `onPreview`; React 18 auto-batches. The live-write path deliberately skips composition refresh (per `useAutomationLanes` `handleDomAttributeLiveCommit` — no reload on drag). Not a per-frame DOM churn concern.
- a11y: `role="group"` + `aria-label` + inline `<title>`; no keyboard nav for adding/dragging points. Consistent with the rest of the timeline — noting rather than blocking.

**Test shape** — falsifying, not tautological: draft state pinned via "moves the dragged point on screen without waiting for the prop" + "follows the prop again once the store catches up"; propagation ownership via real ancestor wrapper (`renderNested`) rather than same-element listener; log-scale via extremal Hz sweep asserting >30px separation. `resolveTrackKeyframeClip` tests cover "audio with automation only", tie-breaking, and selection preference.

**Minor nit** (non-blocking): `endDrag` runs `commitPoints(lane.points, true)` unconditionally when `dragIndex !== null`. A pointerdown-on-point → immediate pointerup with no move commits identical points — a no-op undo entry. Store-level dedup likely masks it; flagging in case undo history shows spurious frames.

**CI:** `mergeable=CONFLICTING` is expected mid-stack (Batch 10/47); only `WIP` + `Mintlify` reporting — the reduced graphite-child matrix. Full lane tests will run when the stack lands; recommend re-verifying against the load-bearing lanes post-restack.

— Review by tai (pr-review)

PR #3025, opened 2026-08-04
feat(studio): automate a parameter without reloading the preview

vanceingalls:
Automating a parameter without reloading the preview.

The reload is the point. Every automation write previously round-tripped the whole composition, which restarts every playing track — heard as the audio chopping mid-edit. The runtime follows the attribute and swaps the envelope in place instead, so drawing a lane while the transport runs is continuous.

The second commit catches the last two write paths that were still reloading.


terencecho (dismissed):
**LGTM.** Downstream Batch-10 of the automation stack: introduces `handleDomAttributeQuietCommit` (persist without preview reload but resync selection) and threads it through the panel so drawing volume / FX-param automation lanes no longer chops playing tracks. Every downstream-cohort seed I audited is clean at this SHA.

**Verified against `28485fe`:**

- *MutationObserver contract preserved.* `packages/core/src/runtime/audioFx.ts:224-234` already filters on `[HF_AUDIO_FX_ATTR, HF_AUDIO_AUTOMATION_ATTR]`. Writing `data-automation` alone triggers `readChain(el)` (which only reads `HF_AUDIO_FX_ATTR` — `audioFx.ts:54-70`), returns an unchanged chain, `handle.update(next.chain)` short-circuits `true` (shape identical in `audioFxGraph.ts:488-495`), and control lands in `rescheduleAutomation`, not `rebuild`. No spurious structural teardown from the new quiet write path.
- *#3014 preview/commit split preserved.* `useAutomationLanes.ts:59-64` — during drag still `handleDomAttributeLiveCommit` (many, no persist), on pointerup was `handleDomAttributeCommit` and is now `handleDomAttributeQuietCommit`. Semantic count (N live during drag / 1 commit on release) unchanged; only the commit's disruption is reduced.
- *Root #3019 (`HfAudioFxNode` allow-list) — N/A.* No schema fields added to `HfAudioFxNode` / `HfAudioFxChain` in this PR; the diff is panel + commit-plumbing only.
- *Batch-9 #3017 `PropertyPanelFlat.tsx` cap — N/A.* File not touched.
- *Quiet resync loop.* `useDomEditAttributeCommits.ts:665-675` uses `skipRefresh: true, refreshAfter: true` — the docstring's rationale (panel would keep computing from pre-edit snapshot without resync) matches `commitDataAttribute`'s `shouldResync` gate at `:150`. Coalesce key `attr-quiet:` cannot collide with `attr:` / `attr-live:`.
- *Constants line up.* `HF_AUDIO_AUTOMATION_ATTR = "data-automation"` (`audioAutomation.ts:16`) matches both the observer filter and the string the tests assert. `VOLUME_TARGET = "volume"` (`audioAutomation.ts:58`) matches `useVolumeAutomation.ts:604`.
- *Test shape — falsifiable, not tautological.* `useVolumeAutomation.test.tsx:493-497` and `:541-547` explicitly guard the `Number("") === 0` seed-at-silence bug the code comment calls out; `:475-481` proves an FX-only lane doesn't false-positive `volumeAutomated`; `:513-528` proves only the volume lane is dropped when others exist. Good coverage of the "seed at unity when raw is empty/non-numeric" edge that the naive `?? "1"` would miss.
- *`.fallowrc.jsonc` allowance.* `propertyPanelAutomation.ts` exports `automatedTargetsOf` / `resolveAutomationRange` — consumers land upstack per the comment; matches the stated Graphite-per-PR pattern.

**Nit (non-blocking):** `useVolumeAutomation.ts:600-608` — `parsed` is already narrowed to `Number.isFinite(parsed) ? parsed : 1`, then `withSeededLane(automation, VOLUME_TARGET, Number.isFinite(current) ? current : 1)` re-checks finiteness. The second guard is dead: `current` is finite by construction. Tightening it to just `write(withSeededLane(automation, VOLUME_TARGET, current))` matches intent and drops the misleading second branch.

**CI:** Only `Mintlify` (skipped) and `WIP` (success) reported — no required test/lint lane firing on this branch yet. Not blocking; flag if Batch-11 lights up a required lane that this SHA hasn't run.

— Review by tai (pr-review)

PR #3026, opened 2026-08-04
feat(studio): automate and un-automate each effect parameter

vanceingalls:
Automating and un-automating each effect parameter from the FX panel, and the carve's own visibility rule.

The carve is only offered when there is a voice to carve against — a bed with nothing over it has no relationship to express, and offering the control there invites an author to configure something that cannot do anything.

Includes a real bug worth reading the commit for: three writes in the panel each read the source file, mutate one attribute and write it back. Fired without ordering they read the same content and only the last one lands.

One consequence: deleting an effect left its automation lanes behind. Ids are minted lowest-free, so the next effect added took the same id and inherited the dead envelope — arriving disabled and "Automated" without the author ever automating it, and baked into the render.


terencecho (dismissed):
**LGTM_WITH_NITS.** Verified the FX-panel refactor + automate/un-automate flip against the diff at `de15099`. Body claims all pin; the two "real bug" fixes (ordered carve-off writes; effect-delete taking its lanes) are load-bearing and directly tested.

**Verified:**

- *`HfAudioFxNode` shape symmetry* (root #3019 concern): no new field lands at the node level. `audioFx.ts:817-823` (parse) and `:850-855` (serialize) both allow-list `{type, id, fromCarve, enabled, params}` symmetrically. Not an asymmetry vector here.
- *Line-cap concern* (#3017 seed): `PropertyPanelFlat.tsx` sheds the inline `AudioFxGroup` (~112 lines), moving *away* from the 600 cap. Net across the PR is −220 with the whole panel extracted to `propertyPanelAudioFxGroup.tsx`.
- *Ordered read-modify-write* (`propertyPanelAudioFxGroup.tsx:675-686`): `setCarve` `await`s the chain write before the carve write, so the second r-m-w sees the file the first landed. Order asserted at `propertyPanelAudioFxGroup.test.tsx:388-394` as `[data-fx-chain, data-fx-carve]`.
- *Delete-node → drop-its-lanes* (`propertyPanelFxSection.tsx:1168-1174` + `propertyPanelAudioFxGroup.tsx:689-695`): `removeNode` calls `onRemoveNodeAutomation?.(removedId)` before chain mutation. Tested end-to-end at `propertyPanelAudioFxGroup.test.tsx:505-527`.
- *Un-id'd nodes can't be automated* (`propertyPanelFxSection.tsx:1013`, `automatedKeysOf` short-circuits on `!node.id`) — tested at `propertyPanelFxSection.test.tsx:919`.
- *Seed at current value* (`propertyPanelAudioFxGroup.tsx:645-654`): `node.params[paramKey]` when numeric, else `range.default` — tested at `propertyPanelAudioFxGroup.test.tsx:258-270`.
- *`automationAttrValue(next) || null`* clears the attribute when lanes go to 0: `propertyPanelAutomation.ts` returns `""` for empty; write maps `""` → `null`. Tested at `propertyPanelAudioFxGroup.test.tsx:322-334`.
- *Carve visibility* (`propertyPanelFxSection.tsx:1142`): `showCarve = sourceOptions.length > 0 || carve !== null` — an existing carve whose voice track was removed stays visible instead of being stranded. Tested at `:958`.
- *`useVolumeAutomation` intentionally passes `undefined` chain* to `readPanelAutomation` — otherwise a volume automate would delete every FX lane. Called out clearly in the hook's docstring; matches `readPanelAutomation`'s contract in `propertyPanelAutomation.ts:38-46`.

**Nits (non-blocking):**

- *Un-automate transient click*: `removeParamAutomation` drops the lane; the runtime then falls back to `node.params[paramKey]`, frozen at automate-on. If the envelope is pushing a value that diverges from that frozen value at un-automate moment, mid-playback un-automate will snap. Panel does its part; `cancelScheduledValues` + `setValueAtTime(currentEnvelopeValue)` belongs in the runtime observer, not here. Worth naming so the follow-up doesn't get lost.
- *Roundtrip pair-test missing*: no `automate → un-automate → automate` test. Coverage is by-component and the ops are pure over the lane set, so risk is low — but it's the direct falsification test for the whole flip.
- *`removeNodeAutomation` prefix brittleness* (`propertyPanelAudioFxGroup.tsx:690`): `startsWith("fx.${nodeId}.")` is safe today because `mintAudioFxNodeId` yields `n<i>`, but a future id format containing `.` would false-positive-match. Structured lane addressing or a doc comment pinning the id shape would future-proof.
- *`Boolean(disabled) || bypassed`* (`propertyPanelFxSection.tsx:1085`): cast is redundant if `disabled: boolean | undefined`. Purely stylistic.

CI on the child is graphite-typical (WIP pass, Mintlify skipped) — trunk gating expected.

— Review by tai (pr-review)

PR #3029, opened 2026-08-04
fix(engine): verify explicit browserGpuMode=hardware instead of trusting it

### LINKED ISSUE #2967: how to enable gpu render on linux
### Describe the bug

I tried to speed up  the processing of hyperframes  render . 
I have a nvida gpu card （rtx3070） on linux，I use command below

PRODUCER_FORCE_SCREENSHOT=true  hyperframes render --gpu --browser-gpu --workers 4 --video-bitrate 5M  --protocol-timeout 3600000  --browser-timeout 86330 --fps 24  -o html_video_1.mp4 

console logs：

....
[Browser:WARN] [GroupMarkerNotSet(crbug.com/242999)!:A0707C01D4370000]Automatic fallback to software WebGL has been deprecated. Please use the --enable-unsafe-swiftshader flag to opt in to lower security guarantees for trusted content.
  ██████░░░░░░░░░░░░░░░░░░░  25%  Capturing frame 90/19186 (4 workers)[Browser:WARN] [.WebGL-0x281c08c99c00]GL Driver Message (OpenGL, Performance, GL_CLOSE_PATH_NV, High): GPU stall due to ReadPixels
  ██████░░░░░░░░░░░░░░░░░░░  25%  Capturing frame 120/19186 (4 workers)[Browser:WARN] [.WebGL-0x281c08c99c00]GL Driver Message (OpenGL, Performance, GL_CLOSE_PATH_NV, High): GPU stall due to ReadPixels
[Browser:WARN] [.WebGL-0x281c08c99c00]GL Driver Message (OpenGL, Performance, GL_CLOSE_PATH_NV, High): GPU stall due to ReadPixels
[Browser:WARN] [.WebGL-0x281c08c99c00]GL Driver Message (OpenGL, Performance, GL_CLOSE_PATH_NV, High): GPU stall due to ReadPixels (this message will no longer repeat)
  ██████░░░░░░░░░░░░░░░░░░░  25%  Capturing frame 210/19186 (4 workers)[Browser:WARN] [GroupMarkerNotSet(crbug.com/242999)!:A0D07C01D4370000]Automatic fallback to software WebGL has been deprecated. Please use the --enable-unsafe-swiftshader flag to opt in to lower security guarantees for trusted content.

the gpu not used, still use cpu  to render,  what should i do to enable gpu render on linux

### Link to reproduction

https://github.com/heygen-com/hyperframes

### Steps to reproduce

PRODUCER_FORCE_SCREENSHOT=true  hyperframes render --gpu --browser-gpu --workers 4 --video-bitrate 5M  --protocol-timeout 3600000  --browser-timeout 86330 --fps 24  -o html_video_1.mp4 

### Expected behavior

1

### Actual behavior

1

### Environment

```shell
hyperframes doctor

  ✗ Version          0.6.102 → 0.7.90 available
                     Run: hyperframes upgrade
  ✓ Node.js          v24.15.0 (linux x64)
  ✓ CPU              20 cores · Intel(R) Core(TM) i9-10850K CPU @ 3.60GHz @ 4799MHz
  ✓ Memory           31.3 GB total · 17.9 GB available
  ✓ Disk             15.7 GB free
  ✓ /dev/shm         8192 MB
  ✓ Environment      Docker
  ✓ FFmpeg           ffmpeg 6.1 at /usr/local/bin/ffmpeg
  ✓ FFprobe          ffprobe 6.1 at /usr/local/bin/ffprobe
  ✓ Chrome           cache: /root/.cache/hyperframes/chrome/chrome-headless-shell/linux-131.0.6778.85/chrome-headless-shell-linux64/chrome-headless-shell
```

### Additional context

_No response_

miguel-heygen:
## What

`resolveBrowserGpuMode("hardware")` now runs the same WebGL probe that `"auto"` already uses, and warns loudly (once per process) when the probe finds no hardware GPU. The requested mode is still honoured verbatim — the probe result never changes what is returned.

## Why

Fixes #2967. A user with an RTX 3070 ran `hyperframes render --gpu --browser-gpu` on Linux and captured 19186 frames on CPU. Nothing in the output said the GPU was not being used.

Chrome's hardware GL args are **advisory**. With no usable GPU visible to the process (no `/dev/dri`, no NVIDIA Container Toolkit, missing EGL/driver libraries) Chrome silently falls back to software WebGL. Before this change `"hardware"` was a pure pass-through with no probe at all, so the only trace was a buried `Automatic fallback to software WebGL has been deprecated` browser warning among thousands of progress lines.

`"auto"` mode already probes and logs. Explicit `"hardware"` — the mode an operator picks precisely *because* they care about the GPU — was the one path that skipped verification.

## How

- `"software"` still short-circuits with no probe (the distributed `renderChunk` path is unaffected).
- `"auto"` and `"hardware"` share the existing cached probe Promise, so this adds no extra Chrome launch when both are reachable, and concurrent workers still probe exactly once.
- On `"hardware"` + probe-says-software, warn with the platform's actual remediation (GPU passthrough flags and driver requirements on Linux) and return `"hardware"` unchanged.
- One-shot latch on the warning: the render path resolves the mode for the probe browser *and* every parallel worker, so an un-deduplicated warning printed 8 times and buried itself. Caught during Linux testing.
- The probe's own log line now reads `browserGpuMode probe →` rather than `browserGpuMode auto →`, since both modes reach it.

Deliberately **not** doing: downgrading the mode to software, or failing the render. The operator asked for hardware; the fix is to stop the fallback being silent, not to override the request.

## Test plan

Unit tests in `packages/engine/src/services/browserManager.test.ts`: hardware passthrough, the warning firing with platform-correct remediation on a SwiftShader probe result, silence on a real-GPU probe result, and once-per-process deduplication.

Verified end-to-end on a Linux host with **no GPU** (`nvidia-smi` absent, no `/dev/dri`) — the same shape as the report:

| | warnings | output |
|---|---|---|
| `main` (control) | **0** — probe never ran, render silently used CPU | `d7bca178…` |
| this branch | **1** — `browserGpuMode probe → software (… llvmpipe …)` plus the remediation block | `d7bca178…` |

Both renders completed and the MP4s are **byte-identical** (same md5), confirming the warning is diagnostic only.

- [x] Unit tests added
- [x] Manual testing performed (Linux, control vs. patched)
- [ ] Documentation updated (n/a — no flag or default changed)

## Not covered

- The positive case (a host with a real GPU where the probe returns hardware and stays quiet) is covered by unit test only; the Linux box available for testing has no GPU.
- This makes the fallback visible. It does not make GPU capture work in a container that lacks passthrough — that is host/image configuration, which the warning now names.

### DIFF
--- a/packages/engine/src/services/browserManager.ts
+++ b/packages/engine/src/services/browserManager.ts
@@ -438,6 +438,7 @@ let _autoBrowserGpuModeCache: Promise<"software" | "hardware"> | undefined;
 /** Test-only: reset the cached probe result. */
 export function _resetAutoBrowserGpuModeCacheForTests(): void {
   _autoBrowserGpuModeCache = undefined;
+  _unverifiedHardwareGpuWarned = false;
 }
 
 async function getPuppeteerOrNull(): Promise<PuppeteerNode | null> {
@@ -544,12 +545,19 @@ export function resolveBrowserGpuMode(
   if (mode === "auto") return _autoBrowserGpuModeCache;
 
   return _autoBrowserGpuModeCache.then((probed) => {
-    if (probed === "software") {
+    // Warn once per process, not once per caller: `createCaptureSession`
+    // resolves the mode for the probe browser AND every parallel worker, so
+    // an un-deduplicated warning prints N+1 times and buries itself.
+    if (probed === "software" && !_unverifiedHardwareGpuWarned) {
+      _unverifiedHardwareGpuWarned = true;
       console.warn(buildUnverifiedHardwareGpuWarning(options.platform ?? process.platform));
     }
     return "hardware";
   });
 }
+
+/** One-shot latch for the explicit-hardware-probed-to-software warning. */
+let _unverifiedHardwareGpuWarned = false;
 
 /**
  * Warning text for "you asked for hardware GPU, the probe found none".
--- a/packages/engine/src/services/browserManager.ts
+++ b/packages/engine/src/services/browserManager.ts
@@ -422,7 +422,25 @@ export const _probeBeginFrameSupportForTests = probeBeginFrameSupport;
 export const _closeBrowserAfterFailedProbeForTests = closeBrowserAfterFailedProbe;
 
 /**
- * Cached *in-flight or resolved* probe Promise for `resolveBrowserGpuMode("auto", ...)`.
+ * Outcome of the one-shot WebGL probe.
+ *
+ * `cause` distinguishes the two ways a probe lands on `"software"`, because
+ * they need OPPOSITE remediation:
+ *   - `"no-gpu"`  — the probe ran and Chrome reported a software renderer
+ *                   (SwiftShader / llvmpipe). Remediation: GPU passthrough.
+ *   - `"probe-error"` — the probe itself failed (Chrome couldn't launch, bad
+ *                   executable path, sandbox denied). We have NO evidence
+ *                   about the GPU either way; telling the operator to fix GPU
+ *                   passthrough would send them chasing the wrong problem.
+ */
+interface GpuProbeOutcome {
+  mode: "software" | "hardware";
+  cause?: "no-gpu" | "probe-error";
+}
+
+/**
+ * Cached *in-flight or resolved* probe Promise, shared by BOTH the `"auto"`
+ * and explicit `"hardware"` entry points of `resolveBrowserGpuMode`.
  *
  * Caching the Promise (rather than the resolved value) deduplicates concurrent
  * callers — the parallel coordinator runs N workers via `Promise.all`, so a
@@ -430,10 +448,8 @@ export const _closeBrowserAfterFailedProbeForTests = closeBrowserAfterFailedProb
  * simultaneous probe Chromes. The first call assigns the Promise and every
  * other concurrent caller awaits the same one, paying the ~240 ms probe cost
  * exactly once per process lifetime.
- *
- * Exported for tests; production callers go through `resolveBrowserGpuMode`.
  */
-let _autoBrowserGpuModeCache: Promise<"software" | "hardware"> | undefined;
+let _autoBrowserGpuModeCache: Promise<GpuProbeOutcome> | undefined;
 
 /** Test-only: reset the cached probe result. */
 export function _resetAutoBrowserGpuModeCacheForTests(): void {
@@ -479,7 +495,7 @@ async function probeAutoBrowserGpuMode(options: {
   chromePath?: string;
   browserTimeout?: number;
   platform?: NodeJS.Platform;
-}): Promise<"software" | "hardware"> {
+}): Promise<GpuProbeOutcome> {
   const platform = options.platform ?? process.platform;
   const browserTimeout = options.browserTimeout ?? DEFAULT_CONFIG.browserTimeout;
   const executablePath = options.chromePath ?? resolveHeadlessShellPath({});
@@ -487,7 +503,7 @@ async function probeAutoBrowserGpuMode(options: {
 
   if (ppt === null) {
     logResolvedBrowserGpuMode("software", "puppeteer unavailable");
-    return "software";
+    return { mode: "software", cause: "probe-error" };
   }
 
   try {
@@ -498,10 +514,10 @@ async function probeAutoBrowserGpuMode(options: {
     });
     const resolved = resolveWebGlProbeMode(info);
     logResolvedBrowserGpuMode(resolved, describeWebGlProbe(info));
-    return resolved;
+    return resolved === "hardware" ? { mode: "hardware" } : { mode: "software", cause: "no-gpu" };
   } catch (err) {
     logResolvedBrowserGpuMode("software", formatProbeFailure(err));
-    return "software";
+    return { mode: "software", cause: "probe-error" };
   }
 }
 
@@ -542,32 +558,53 @@ export function resolveBrowserGpuMode(
   if (mode === "software") return Promise.resolve(mode);
 
   _autoBrowserGpuModeCache ??= probeAutoBrowserGpuMode(options);
-  if (mode === "auto") return _autoBrowserGpuModeCache;
+  if (mode === "auto") return _autoBrowserGpuModeCache.then((probed) => probed.mode);
 
   return _autoBrowserGpuModeCache.then((probed) => {
-    // Warn once per process, not once per caller: `createCaptureSession`
+    // Warn once per cache lifetime, not once per caller: `createCaptureSession`
     // resolves the mode for the probe browser AND every parallel worker, so
     // an un-deduplicated warning prints N+1 times and buries itself.
-    if (probed === "software" && !_unverifiedHardwareGpuWarned) {
+    if (probed.mode === "software" && !_unverifiedHardwareGpuWarned) {
       _unverifiedHardwareGpuWarned = true;
-      console.warn(buildUnverifiedHardwareGpuWarning(options.platform ?? process.platform));
+      console.warn(
+        buildUnverifiedHardwareGpuWarning(options.platform ?? process.platform, probed.cause),
+      );
     }
     return "hardware";
   });
 }
 
-/** One-shot latch for the explicit-hardware-probed-to-software warning. */
+/**
+ * Latch for the explicit-hardware-probed-to-software warning: fires once per
+ * cache lifetime (re-armed by `_resetAutoBrowserGpuModeCacheForTests`).
+ */
 let _unverifiedHardwareGpuWarned = false;
 
 /**
- * Warning text for "you asked for hardware GPU, the probe found none".
+ * Warning text for "you asked for hardware GPU and we could not confirm it".
  *
- * Names the observable symptom (the render still completes, just on CPU) and
- * the platform's actual remediation, so the operator doesn't have to infer it
- * from Chrome's `Automatic fallback to software WebGL` warning. Exported for
- * tests.
+ * Splits on `cause` because the two failure shapes need opposite remediation.
+ * A probe that RAN and saw SwiftShader is a GPU-passthrough problem. A probe
+ * that could not run tells us nothing about the GPU — pointing that operator
+ * at `--gpus all` would send them chasing a phantom while their Chrome
+ * install is the actual fault.
  */
-export function buildUnverifiedHardwareGpuWarning(platform: NodeJS.Platform | string): string {
+function buildUnverifiedHardwareGpuWarning(
+  platform: NodeJS.Platform | string,
+  cause: GpuProbeOutcome["cause"],
+): string {
+  if (cause === "probe-error") {
+    return (
+      "[hyperframes] browserGpuMode=hardware was requested, but the GPU probe could not run, " +
+      "so hardware acceleration is UNVERIFIED — if Chrome falls back to software WebGL the " +
+      "capture will run at CPU speed. Honouring the explicit request anyway.\n" +
+      "  This is a probe failure, not evidence of a missing GPU: see the " +
+      "`browserGpuMode probe → software (probe failed ...)` line above for the underlying " +
+      "error, which usually means Chrome could not launch (bad HYPERFRAMES_BROWSER_PATH, " +
+      "missing shared libraries, or a denied sandbox) rather than a GPU problem.\n" +
+      "  Run `hyperframes doctor` to check the Chrome install."
+    );
+  }
   const remediation =
     platform === "linux"
       ? "Inside Docker, the container needs GPU passthrough: `--gpus all` with the NVIDIA " +
--- a/packages/engine/src/services/browserManager.ts
+++ b/packages/engine/src/services/browserManager.ts
@@ -507,14 +507,25 @@ async function probeAutoBrowserGpuMode(options: {
 /**
  * Resolve `browserGpuMode` to a concrete `"software" | "hardware"` answer.
  *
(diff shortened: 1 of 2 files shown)


vanceingalls (approved):
## R1 adversarial review

**Verdict**: APPROVE
**Verify semantics**: The same WebGL probe used by `"auto"` is now run for explicit `"hardware"`, but the returned mode stays `"hardware"` — the operator's request is honoured verbatim. The probe result only drives a one-shot stderr warning naming the platform remediation. That's the right trade-off for issue #2967 (silent CPU-only render).

**Trust-site audit (grep-verified)**: The only production consumers of `resolveBrowserGpuMode` are `packages/engine/src/services/frameCapture.ts:1062` and `packages/cli/src/browser/gpuPolicy.ts:19`. Both route every request through the new verify path — no site still trusts `"hardware"` bypass. `_autoBrowserGpuModeCache` short-circuits on `"software"` before probing (line 541), so software callers correctly don't populate the cache and can't spuriously trigger the warning.

**Concurrency / dedup**: One-shot latch `_unverifiedHardwareGpuWarned` is set inside a Promise `.then` — sequential microtask ordering in Node's event loop guarantees the second-through-Nth workers see `true`. Test at `browserManager.test.ts:250-269` asserts 3 concurrent hardware calls warn once. Reset helper wired at line 441.

**OS matrix**: `getBrowserGpuArgs()` covers darwin/win32/linux; the Linux-specific remediation text is tested; non-Linux falls to a generic message (acceptable — the bug repro is Linux).

**Failure-mode**: `probeAutoBrowserGpuMode` catches all errors → returns `"software"`, which then triggers the warning. Slight semantic slippage on genuine Chrome-launch failures (the warning says "found no hardware GPU" when technically the probe itself failed), but the remediation advice remains applicable — minor.

**FINDINGS**:
- [nit] `packages/engine/src/services/browserManager.ts:570` — `buildUnverifiedHardwareGpuWarning` is exported but tests use substring matching on `console.warn.mock.calls`, not the exported helper. Fallow major-severity failure (only failing check) blocks merge. Fix: either un-export the helper (private function) or import it in the test file for exact-equality assertions. Either is trivial.
- [nit] `browserManager.ts:576` non-Linux remediation is generic. Windows/macOS operators get less actionable text than Linux. Fine for this PR scope; consider a follow-up if reports arrive from other OSes.

**Positives worth naming**:
- The "warn once, not once per worker" note + test is exactly the class of parallel-branch instrumentation gap that usually ships broken.
- The `logResolvedBrowserGpuMode` label rename from `"auto →"` to `"probe →"` correctly updates the observability surface for the widened trigger scope.
- MP4 byte-equality verification on control-vs-patched in the PR description proves diagnostic-only intent.

**Standards**: Lint/Typecheck/Tests/Build/Windows render/CLI smoke — all green. Only Fallow is red on the unused export (fixable in one line).

_Review by Via_

packages/engine/src/services/browserManager.ts
```diff
@@ -527,21 +539,63 @@ export function resolveBrowserGpuMode(
...
+
+/**
+ * Warning text for "you asked for hardware GPU, the probe found none".
+ *
+ * Names the observable symptom (the render still completes, just on CPU) and
+ * the platform's actual remediation, so the operator doesn't have to infer it
+ * from Chrome's `Automatic fallback to software WebGL` warning. Exported for
+ * tests.
+ */
+export function buildUnverifiedHardwareGpuWarning(platform: NodeJS.Platform | string): string {
```
james-russo-rames-d-jusso: 🔴 **Fallow audit is failing on this export — required CI check red.** The audit report on this PR flags: `Export 'buildUnverifiedHardwareGpuWarning' is never imported by other modules` (major severity). The docstring says "Exported for tests" but `browserManager.test.ts` asserts the warning content via a `console.warn` spy (`expect(warning).toContain("--gpus all")` at :262-263) rather than importing the builder directly, so the export has no consumer at merge.

**Cleanest fix:** drop the export (make it a local function) — the `console.warn` spy path already verifies the observable behavior. **Alternative:** actually import + test the builder directly. That would also let you cover the darwin/win32 branch of the platform-conditional remediation string, which no test currently asserts (`platform: "linux"` is the only case the new tests exercise).

Either way, must land before merge — Fallow audit fail blocks the required-check gate.

— Rames D Jusso

packages/engine/src/services/browserManager.ts
```diff
@@ -527,21 +539,63 @@ export function resolveBrowserGpuMode(
...
+  _autoBrowserGpuModeCache ??= probeAutoBrowserGpuMode(options);
+  if (mode === "auto") return _autoBrowserGpuModeCache;
+
+  return _autoBrowserGpuModeCache.then((probed) => {
+    // Warn once per process, not once per caller: `createCaptureSession`
+    // resolves the mode for the probe browser AND every parallel worker, so
+    // an un-deduplicated warning prints N+1 times and buries itself.
+    if (probed === "software" && !_unverifiedHardwareGpuWarned) {
+      _unverifiedHardwareGpuWarned = true;
+      console.warn(buildUnverifiedHardwareGpuWarning(options.platform ?? process.platform));
```
james-russo-rames-d-jusso: 🟠 **Probe-launch failure gets diagnosed as "no hardware GPU" in the user-facing warning.** `probeAutoBrowserGpuMode` maps any failure — Chrome launch error, navigation timeout, missing canvas API, unreadable executable, all catch-all-ed into the `catch` at `:502-505` — into a `"software"` return. Then in `resolveBrowserGpuMode`'s continuation (`:551-554`), that `"software"` triggers `buildUnverifiedHardwareGpuWarning(...)` which tells the operator to fix GPU passthrough (`--gpus all`, NVIDIA Container Toolkit, EGL libs), etc.

But if the probe failed because Chrome couldn't launch at all (bad `chromePath`, corrupt binary, sandbox denied), the operator's actual remediation is completely different — they'd chase a nonexistent GPU passthrough problem while their Chrome install is broken.

The observability log at `logResolvedBrowserGpuMode` DOES distinguish the two cases (`probe failed (...)` vs. `WebGL renderer vendor=...`), so the info is available; it just doesn't reach the user-facing warning. Cheapest fix: return a discriminated result from `probeAutoBrowserGpuMode` (`{result: "hardware"} | {result: "software", cause: "no-gpu" | "probe-error"}`) and only fire this specific warning on `no-gpu`, with a different (or suppressed) warning on `probe-error` since a broken probe means we have no evidence either way.

Realistically low-probability — most probe failures ARE genuine no-GPU cases. But the whole point of this PR is "stop the silent fallback," and misdirected remediation on a Chrome-install issue would put us in a symmetric-but-different silence.

— Rames D Jusso

james-russo-rames-d-jusso (commented):
_Reviewed at `6703ea7e`._

Small, targeted fix for #2967. Design is right: share the cached probe Promise between `"auto"` and explicit `"hardware"`, warn once per process when hardware was requested but the probe found only SwiftShader, honour the operator's request verbatim (byte-identical MP4 output confirmed in the PR body's test plan). The invariant chain — one probe per process → one warning per process → one platform-appropriate remediation string → mode never downgraded — is intact.

**One CI-blocker:** Fallow audit is red — `buildUnverifiedHardwareGpuWarning` is exported without a consumer. Details inline at `browserManager.ts:570`.

**Also inline:** `browserManager.ts:553` — the "no hardware GPU" warning fires on any probe failure, including Chrome-launch errors that need a different remediation.

**PR-body promise audit — checks out end-to-end:**

- ✅ `"software"` still short-circuits with no probe (`:542`).
- ✅ `"auto"` and `"hardware"` share the cached probe Promise (`:544` — `??=` ensures single probe per process across both entry points).
- ✅ Concurrent workers still probe exactly once (Promise-level cache, unchanged from prior behavior).
- ✅ `"hardware"` + probe-says-software warns with linux-specific remediation and returns `"hardware"` unchanged (`:547-556`).
- ✅ One-shot latch: `_unverifiedHardwareGpuWarned` gate at `:551`, reset in `_resetAutoBrowserGpuModeCacheForTests` at `:441` (verified — the reset covers both cache and latch).
- ✅ Log line rename `browserGpuMode auto → …` → `browserGpuMode probe → …` at `:598` — semantically correct since explicit hardware now hits the same probe.
- ✅ `beforeEach`/`afterEach` in the `resolveBrowserGpuMode` describe block both call the reset, so no cross-test cache pollution.

**Nits (body-only, not worth an inline thread each):**

- **Stale docstring** at `_autoBrowserGpuModeCache` (`:425-435`) still reads "Cached ... probe Promise for `resolveBrowserGpuMode(\"auto\", ...)`" — the cache is now used by both `"auto"` and `"hardware"` entry points, so consider "for `resolveBrowserGpuMode(...)`" or a brief note about the dual consumer.
- **Log-line rename is observable** — `[hyperframes] browserGpuMode auto →` → `[hyperframes] browserGpuMode probe →` at `:598`. If anyone has a grep / dashboard / alert scraping the old string, it breaks silently. Probably fine — it's `console.error` diagnostic output — but naming it in case Miguel's observability owner has a rule to update.
- **Test coverage gap on the non-linux branch of `buildUnverifiedHardwareGpuWarning`:** the "warns when explicit 'hardware' probes to software" test only exercises `platform: "linux"` and asserts `--gpus all`. The `darwin` / `win32` fallback branch ("Check that the host exposes a GPU...") has no assertion. If you go with the "import + test the builder directly" fix for the Fallow finding, this coverage gap closes for free.
- **Docstring** at the latch declaration (`:559`) says "One-shot latch" but the reset function does re-arm it — technically the latch is "one-shot per cache-lifetime." Trivial wording.

Ready from my side once Fallow is green, leaving as COMMENTED.

packages/engine/src/services/browserManager.ts
miguel-heygen: Fixed in f69c4a0e3 — un-exported, it's a local function now.

Worth recording why I went with un-export over the import-and-test-directly alternative: it would not have worked. Fallow's `entry` list in `.fallowrc.jsonc` covers `packages/producer/src/**/*.test.ts`, `packages/aws-lambda/...`, and `packages/gcp-cloud-run/...`, but **not** `packages/engine/src/**/*.test.ts`. So an engine test file is not an audit entry point, and importing the helper there would have left the export with no reachable consumer and the check still red.

Your related point about the darwin/win32 branch having no coverage was correct and I closed it anyway, just through the spy rather than the builder — new test `gives non-linux hosts the generic remediation, not the Docker one` asserts the generic string is present and `--gpus all` is absent.

Verified locally: `bunx fallow audit --base origin/main` → `✓ No issues in 18 changed files`.

packages/engine/src/services/browserManager.ts
miguel-heygen: Good catch, and you were right that it mattered more than the probability suggests. Fixed in f69c4a0e3, essentially the discriminated-result shape you proposed.

`probeAutoBrowserGpuMode` now returns `{ mode, cause }` where `cause` is `"no-gpu"` (probe ran, Chrome reported SwiftShader/llvmpipe) or `"probe-error"` (probe could not run at all — launch throw, or `getPuppeteerOrNull()` returning null, which was the same silent-software path). `"auto"` maps through `.then(o => o.mode)` so its contract is unchanged.

On `"probe-error"` the warning no longer mentions GPU passthrough. It states hardware acceleration is UNVERIFIED, says explicitly that this is a probe failure and not evidence of a missing GPU, points at the `browserGpuMode probe → software (probe failed ...)` line for the underlying error, names the usual causes (bad `HYPERFRAMES_BROWSER_PATH`, missing shared libraries, denied sandbox), and sends them to `hyperframes doctor`.

New test `does not blame the GPU when the probe itself failed to launch` asserts the probe-error text appears and `--gpus all` does not.

One thing your comment made me find: I also had to change the concurrency test. It asserted Promise identity (`p1 === p2`) as a proxy for "the probe runs once", and both entry points now adapt the shared Promise through `.then`, so identity no longer holds even though the invariant does. Rather than delete the check I made it assert the real thing — `launch` called exactly once across concurrent callers — and added a case covering mixed `auto`/`hardware` callers sharing the single probe. That is a stronger assertion than what was there before.

Engine suite: 1375 passed, 3 skipped. Fallow: no issues.

miguel-heygen:
Both inline threads addressed in f69c4a0e3 and resolved. Also picked up the body-only nits from @james-russo-rames-d-jusso:

- **Stale `_autoBrowserGpuModeCache` docstring** — rewritten; it now says the cache is shared by both the `"auto"` and explicit `"hardware"` entry points, instead of claiming it's auto-only.
- **"One-shot latch" wording** — you were right that the reset re-arms it. Now reads "fires once per cache lifetime (re-armed by `_resetAutoBrowserGpuModeCacheForTests`)".
- **Non-linux coverage gap** — closed via the spy path rather than the exported builder, since un-exporting was the Fallow fix (reasoning in the resolved thread).
- **Log-line rename is observable** — flagging this one rather than silently keeping it. `[hyperframes] browserGpuMode auto →` became `… probe →`. It's `console.error` diagnostic output with no dashboard I can find consuming it, and leaving it saying "auto" would be actively wrong now that explicit hardware hits the same probe. Happy to revert to the old string if anyone has a scraper on it.

CI was red on Fallow before; verified locally now with `bunx fallow audit --base origin/main` → `✓ No issues in 18 changed files`, plus the full engine suite at 1375 passed / 3 skipped.

One correction to my own PR body while I'm here: it claims the probe adds "no extra Chrome launch". That's true for `"auto"` (unchanged) and for `"software"` (still short-circuits), but explicit `--browser-gpu` previously did **zero** probing and now pays one ~240 ms probe launch, once per process. Cheap and the point of the fix, but the body undersold it and I'd rather that be on the record than discovered later.

PR #2840, opened 2026-07-28
feat(cli,core,producer): ramp the parallel-DE router through the canary at 5%

vanceingalls:
## What

The parallel drawElement router becomes the default for eligible renders — **ramped through the canary at 5%**, not flipped on for everyone. Three things ship together:

1. **The kill switch actually works.** `isDeParallelRouterEnabled()` parses `false`/`0`/`off`/`no` (case- and space-insensitive); unset or empty means default. The old `!== "false"` failed open on every spelling but the literal one.
2. **The circuit breaker survives a default-ON flag.** It now writes an explicit `"false"` instead of `delete`-ing the var — under default-ON polarity, deleting means *ON*, so the breaker was a no-op on exactly the hosts it exists to protect.
3. **The rollout is a number someone chose.** Enrolment gates the new behaviour; `percentage: 0` in the registry is a full fleet-wide revert with no release.

Also removes the 25-render cap and the telemetry preconditions (see below).

## Why the ramp

Default-ON without one is a **~17× exposure jump** — from ~6% of eligible renders to all of them — landing on profiles today's trial population never covered: **≤4 CPUs (8.18% of eligible renders)** and **Docker (3.99%)**. ("Trial" is not a user opt-in — it arms automatically on the CLI render path. The `enableDeParallelRouterTrial` flag is *call-site* opt-in, excluding programmatic `renderLocal` consumers because it mutates `process.env`. So ~11% of installs route today without anyone choosing it.) 0.7.60–0.7.64 is the precedent: every unclamped render reverted for five consecutive releases and nobody noticed.

Worth being precise about what the canary buys, because it is easy to state wrongly: **~10.85% of rendering installs already route today**, automatically. That number is *emergent* — the product of eligibility rules, the render cap and the breaker latch — so it drifts with fleet composition and cannot be switched off without shipping. The canary does not make exposure smaller. It makes it **chosen and revertible**.

## Current soak (14d, ≥0.7.66)

| | |
|---|---|
| installs engaged | 1,466 |
| never reverted | 1,311 |
| reverted exactly once | 154 |
| reverted **more than once** | **1** |
| per-render revert rate | 2.79% |
| installs hitting ≥1 revert | 10.57% |

154 of 155 reverting installs stop at exactly one — the breaker working. Held steady as the population tripled from 434 installs, so that is a property of the mechanism, not a sample artifact.

**What reverts actually are** (routed renders, same window):

| outcome | renders | installs |
|---|---|---|
[...]

PR #3030, opened 2026-08-04
docs(canary): record the first fleet read — check 3 is dirty

vanceingalls:


miga-heygen (commented):
## Review: canary calibration first-read record

Docs-only, single file. This is well-structured scientific documentation — checks defined before data, results recorded against those definitions, defect acknowledged where the data doesn't match the accepted exceptions. The editorial revision (from "all flips are first-mint shaped" to "only ~23% are") is the important part of this PR.

### What I verified

**Internal consistency of the data.** The numbers hold up across sections:
- 9,702 total installs, 9,455 non-CI — used consistently, stated per-row where the population differs
- Accuracy: 9.22%/48.64% naive → 9.49%/49.76% CI-excluded — the CI bias is real and the denominator rule is the correct fix
- Stability: 25 + 20 + 45 + 17 = 107 flips on calibration-50, matches the stated total
- Independence: 5.18% vs 4.71% expected (z = 2.16) — marginal significance, the "re-check rather than act" call is reasonable

**Logical structure of the argument.** The reasoning chain is tight:
1. Check 3 came back dirty (107 flips)
2. Only ~23% match the first-mint race signature (≤5s minority persistence, ~3 events)
3. 17 installs sustained the minority for 7.7h / ~125 events — this is a real defect by the check's own definition
4. Registry edit is ruled out (enrollment rate consistent across versions)
5. Therefore: #2840 stays held, next step is emitting decision reason

The escape hatch the doc built in ("add reason property only if this check comes back dirty") fires correctly. Good discipline.

**SSOT on the blocking decision.** The consequence is stated once, clearly: "Nothing should ramp on a bucketer whose cohorts are not provably stable." The next action is concrete: emit `forced_on / forced_off / in_cohort / out_of_cohort`, then re-read.

### Observation (non-blocking)

The leading hypothesis for the >1h flips — shared `distinct_id` across containers with separate seeds — connects directly to PR #3065's identity-persistence telemetry. Once that ships, `identity_persistence = process_only | unknown` plus distinct `invocation_id` values under one `distinct_id` would confirm the container hypothesis. Might be worth a forward-reference here so the next reader knows where to look, but not blocking.

CI green. Clean documentation that's honest about what it found. No concerns.

— Miga

docs/contributing/canary-rollouts.mdx
```diff
@@ -178,8 +248,17 @@ correlated and every canary hits the same unlucky cohort).
...
+Checks 1 and 4 should pass outright — they are properties of the design, and
+failing either is a bug to fix before shipping a real rollout. Check 1 passed
+on the first read once CI was excluded from the denominator.
+
+Check 3 should pass apart from the first-mint race. **On the first read it did
+not**, and the gap is not explained by the accepted exceptions — see "First
+read" above. That is an open defect, not a tolerated limit.
+
+The independence bonus came back at 5.18% overlap against 4.71% expected
+(n = 9,523 non-CI, z = 2.16) — a slight positive correlation, small in absolute terms, worth
```
james-russo-rames-d-jusso: 🔴 Same read, two independence numbers. The primary "Accuracy and independence pass" table above (line 211) gives `overlap 5.03%, n = 27,226 non-CI, z = 1.06` — a fresh, within-2σ result on the 28,100-install / 27,226-non-CI base this doc now anchors on. This closing paragraph still reads `5.18% overlap against 4.71% expected (n = 9,523 non-CI, z = 2.16)`, which is from the earlier iteration of this same PR (the one whose flip count was 107 across 9,702 installs, replaced above with 304 across 28,100). A doc that will be quoted at every re-read cannot quote itself with two different independence results for the same window. Two fixes: (a) delete this paragraph entirely — the accuracy table already carries the independence row in the same shape as the other primary results, and the qualitative gloss ("worth re-checking on the next read rather than acting on now") no longer fits at z = 1.06; or (b) update this paragraph's numbers to match line 211. Please fix before merge. — _[Rames D Jusso](https://github.com/james-russo-rames-d-jusso)_

docs/contributing/canary-rollouts.mdx
```diff
@@ -152,18 +152,88 @@ event-weighted separately: render volume is heavily skewed toward a few heavy
...
+for a mean 20 hours across ~131 events** — that is not a first-run race by any
+reading, and by this check's own definition it is a real defect.
+
+The leading unproven hypothesis is a shared `distinct_id` across containers
+that each hold their own seed.
+
+**What this blocks.** Nothing should ramp on a bucketer whose cohorts are not
+provably stable. The next step is emitting the decision reason
+(`forced_on` / `forced_off` / `in_cohort` / `out_of_cohort`) so a deliberate
+override can be told apart from a genuine flip, then re-reading this table.
```
james-russo-rames-d-jusso: 🟠 The `resolveCanary` reason enum has seven values, not four. `packages/core/src/canary.ts:38-48` defines `forced_on | forced_off | in_cohort | out_of_cohort | no_unit_id | excluded | telemetry_opt_out`, and `evaluateCanary` returns each of the first six (`telemetry_opt_out` never emits by construction — its own comment says so). Once wire-emission ships, two of the omitted values will land on real events:
- **`excluded`** — CI installs (`getSystemMeta().is_ci === true` at `packages/cli/src/telemetry/canary.ts:105`) hit `evaluateCanary`'s `if (input.exclude)` branch at `core/canary.ts:129` and emit `reason: "excluded"`, distinct from a real out-of-cohort. Good for filtering — the analytical query today keys on the `is_ci` field to drop them; the new field means one fewer join — but the reader has to know to allowlist it.
- **`no_unit_id`** — the fails-closed corner when `bucketSeed` is missing. Rare, but appears as `enabled: false` with its own reason.

The four listed cover the flip-vs-override attribution question exactly, so this isn't a plan defect — but a doc-of-record that will drive the next re-read query is cheaper to write once with the full enum than to re-derive later. Either list all six wire values, or state explicitly "naming only the values needed for attribution; see `canary.ts:38-48` for the full enum". Non-blocking. — _[Rames D Jusso](https://github.com/james-russo-rames-d-jusso)_

docs/contributing/canary-rollouts.mdx
```diff
@@ -152,18 +152,88 @@ event-weighted separately: render volume is heavily skewed toward a few heavy
...
+| 5–60s | 48 | 7.9 | 7.8h |
+| 1–60min | 126 | 7.8 | 3.9h |
+| **&gt; 1h — post-settled** | **70** | **131.2** | **20.0h** |
+
+Only ~20% match the first-mint signature. **70 installs sustained the minority
+for a mean 20 hours across ~131 events** — that is not a first-run race by any
+reading, and by this check's own definition it is a real defect.
+
+The leading unproven hypothesis is a shared `distinct_id` across containers
+that each hold their own seed.
```
james-russo-rames-d-jusso: 🟡 The hypothesis is viable only under a specific container topology: `distinct_id` (`anonymousId` in `~/.hyperframes/config.json`) shared across containers while `bucketSeed` (in `install-state.json`) mints separately per container — the opposite of what a naive shared-volume bind-mount gives you. `packages/cli/src/telemetry/config.ts` and the bucketing test at `packages/cli/src/telemetry/canary.test.ts:150-158` (`buckets on the bucketSeed when present — the unit that survives config wipes`) show the two live in separate files with separate lifetimes, so divergence *is* mechanically possible: it needs an image-baked `config.json` (fixed `anonymousId`) paired with a runtime-generated `install-state.json` (fresh seed per container run). Reasonable topology for containerized CI runners, less so for arbitrary Docker consumers.

The existing "next step" plan of emitting `reason` will not confirm or falsify this — `reason` says `in_cohort`/`out_of_cohort` on both sides of the flip, not why the seeds diverged. Two properties that would make it testable in the same emission pass: (a) `bucket_seed_origin` (`persisted` vs `fresh_mint` — do we know if the seed came off disk or was minted this run?), and (b) a low-entropy machine fingerprint (`/etc/machine-id` when readable, or a hash of `os.hostname()`). Together, a shared-`anonymousId`-with-diverging-seed pattern would be immediately visible. Cheap to include alongside `reason` if that emission is landing anyway. Non-blocking — otherwise the re-read will most likely land with "still dirty, still can't attribute the Docker cluster" and need a follow-up emission. — _[Rames D Jusso](https://github.com/james-russo-rames-d-jusso)_

james-russo-rames-d-jusso (commented):
_Reviewed at `c0cbebd`._

Docs-only patch to `docs/contributing/canary-rollouts.mdx` recording the first fleet read of the `calibration-10` / `calibration-50` bucketer. Numbers check out where I sampled: table row sums (60+48+126+70 = 304 ✓), first-mint fraction (60/304 ≈ 20% ✓), primary z-scores against the 27,226-non-CI base (calibration-10: SE = √(0.1×0.9/27226) = 0.00182 → z = (0.0979 − 0.10)/0.00182 = −1.15 ≈ doc's −1.16 ✓, calibration-50: SE = 0.00303 → z = −0.10 ✓, overlap: expected `p₁ × p₂ = 0.0979 × 0.4997 ≈ 0.0489` → SE = √(0.0489 × 0.9511/27226) = 0.00131 → z = (0.0503 − 0.0489)/0.00131 ≈ 1.07 ≈ doc's 1.06 ✓). The flip-rate stability across fleet growth (107/9702 = 1.10% → 304/28100 = 1.08%) is the strongest single point in the doc — it rules out the "we caught a burst" reading before the "post-settled" bucket even needs to be presented.

The correctness gap that matters is on the independence result appearing twice in the same doc with two different values — flagged inline as the blocker. The other two are calibration for the re-read query.

**One thing outside the diff:** the PR body still cites the earlier iteration's numbers (9,702 installs, 9.22% / 48.64%, 107 flips, 17 >1h persisters). Body drift after a re-run is common but material here because the doc is quoting itself throughout, and any Slack/PR summary linking to the body will misreport the fleet read. Consider syncing the PR body to the current diff before merge. Also — the PR is `mergeStateStatus: DIRTY` against `main`; rebase needed.

**Downstream reader check:** none in-tree. Verified via a repo-wide sweep for the calibration percentages and canary names: no code path consumes the percentages, no dashboard/CI job reads the doc, and `.github/workflows/canary-sunset.yml` is a weekly sunset-date checker that doesn't ingest thresholds. `de-parallel-router` is registered at `packages/core/src/canaryRegistry.ts:85` (`percentage: 0`) but not called by production code — today's router gate is `HF_DE_PARALLEL_ROUTER` env plus the per-install circuit breaker at `packages/cli/src/commands/render.ts:1079` (`DE_PARALLEL_ROUTER_TRIAL_MAX_RENDERS`). So the doc's "#2840 stays held" language is a sequencing/policy statement authored by the same person, not a code-enforced gate — which is fine, but worth naming so a future reader doesn't chase the wrong dependency graph.

**Data source provenance** (nudge, not blocking): the doc quotes very specific numbers for a re-readable table but doesn't cite which query / warehouse / window definition generated them. A one-line footnote pointing at the source (Snowflake table + HogQL / view) would let the next reader reproduce or spot-check without asking.

PR #3032, opened 2026-08-04
fix: preserve explicit zero caption grain

miguel-heygen:
fixes reported:1785872503.756319; no related claim remains outstanding

Source claim: https://slack.com/archives/C0BGC335AQY/p1785872503756319

Summary:
- preserve an explicit plate grain value of zero instead of replacing it with the default
- omit the FFmpeg noise filter when resolved grain is zero
- retain noise=alls=5 when the grain setting is absent

Regression coverage:
- explicit grain zero compiles without a noise filter
- missing grain continues to compile with the default level

Verification:
- node --check skills/embedded-captions/scripts/make-theme.cjs
- node --test skills/embedded-captions/scripts/make-theme.test.mjs (2 passed)
- bun run test:skills (463 passed)
- bunx oxlint changed files
- bunx oxfmt --check changed files

### DIFF
--- a/skills/embedded-captions/scripts/make-theme.cjs
+++ b/skills/embedded-captions/scripts/make-theme.cjs
@@ -8537,7 +8537,7 @@ if (!bodyInBg || fx.html || fx.js || setp.fgHtml) {
 }
 
 // _postfx.sh: plate reaction after the matte composite (subject+text move as one)
-const P = HEROLESS ? { grain: (dna.plate || {}).grain || 5 } : dna.plate || {};
+const P = HEROLESS ? { grain: (dna.plate || {}).grain ?? 5 } : dna.plate || {};
 // punchOffset: themes whose impact is NOT the hero onset (e.g. flapboard's
 // lock-complete clack) shift the plate punch anchor; default keeps onset+2f
 const anchorT = (heroIn + (P.punchOffset ?? 0.045)).toFixed(3);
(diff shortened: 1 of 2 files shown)


miga-heygen (commented):
Clean, correct fix.

**The bug:** `||` coerces `0` to falsy, so `(dna.plate || {}).grain || 5` replaced an explicit `grain: 0` with `5` — the noise filter always ran even when the theme author explicitly disabled it.

**The fix:** `??` preserves `0` (only null/undefined trigger the default). The noise filter string is now built conditionally — empty when `grain === 0`, which cleanly removes the `,noise=alls=...` segment from the ffmpeg filter graph rather than passing `noise=alls=0` (which would still allocate the filter for no visual effect).

Two things I verified:
- The `filter` + `rgba` + `noise` concatenation still produces a valid filter graph in both the zero and non-zero paths. The `null` filter name in the test output (`[0:v]null[v]`) is expected — it's the upstream `filter` variable's value when no spatial filter is active.
- The `P.grain || 5` on line 8540 (heroless branch) is also replaced with `??`, so the grain value is consistent between the plate object construction and its later use.

Test coverage is proportionate: explicit-zero and missing-grain are the two boundary cases that matter. No notes.

— Miga

miga-heygen (approved):
No blocking issues found in my review. Stamping. — Miga

jrusso1020:
Continuing this contribution in #3656, rebuilt on current main with regression coverage and attribution to this PR. Keeping this PR open until the successor lands. Thanks for the original fix.

jrusso1020:
Closing as implemented in #3656, now merged. The successor preserves this contribution while updating it for current main and adding regression coverage. Thanks for the original fix.

PR #3034, opened 2026-08-04
fix(cli): ignore ancestor motion in pivot drift audit

miguel-heygen:
fixes reported:1785878590.139659; no related claim remains outstanding

Source claim: https://slack.com/archives/C0BGC335AQY/p1785878590139659

The rotation sampler now records the computed ancestor-transform chain. The pivot-drift detector suppresses viewport-center drift while that chain changes, preventing a translating wrapper from being blamed on a correctly centered rotating child.

Verification:
- 116 focused CLI tests passed
- CLI typecheck passed
- oxlint and oxfmt checks passed
- Exact 2-second fixture: zero layout findings (previously rotation_pivot_drift at 601px)

### DIFF
--- a/packages/cli/src/commands/layout-audit.browser.js
+++ b/packages/cli/src/commands/layout-audit.browser.js
@@ -1546,6 +1546,15 @@
     return (Math.atan2(b, a) * 180) / Math.PI;
   }
 
+  function ancestorTransformSignature(element, root) {
+    const transforms = [];
+    for (let ancestor = element.parentElement; ancestor; ancestor = ancestor.parentElement) {
+      transforms.push(getComputedStyle(ancestor).transform || "none");
+      if (ancestor === root) break;
+    }
+    return transforms.join("|");
+  }
+
   window.__hyperframesRotationSample = function collectRotationSample() {
     const root =
       document.querySelector("[data-composition-id][data-width][data-height]") ||
@@ -1572,6 +1581,7 @@
         w: round(box.width),
         h: round(box.height),
         angle: round(angle),
+        ancestorTransformSignature: ancestorTransformSignature(element, root),
       });
     }
     return samples;
--- a/packages/cli/src/utils/checkBrowser.ts
+++ b/packages/cli/src/utils/checkBrowser.ts
@@ -559,10 +559,19 @@ function parseRotationSample(value: unknown, time: number): RotationSample[] {
   const w = numberValue(value, "w");
   const h = numberValue(value, "h");
   const angle = numberValue(value, "angle");
-  if (!selector || cx === null || cy === null || w === null || h === null || angle === null) {
+  const ancestorTransformSignature = stringValue(value, "ancestorTransformSignature");
+  if (
+    !selector ||
+    cx === null ||
+    cy === null ||
+    w === null ||
+    h === null ||
+    angle === null ||
+    ancestorTransformSignature === null
+  ) {
     return [];
   }
-  return [{ time, selector, cx, cy, w, h, angle }];
+  return [{ time, selector, cx, cy, w, h, angle, ancestorTransformSignature }];
 }
 
 async function collectOffPivotRotationSample(page: Page, time: number): Promise<OffPivotFrame> {
--- a/packages/cli/src/utils/checkPipeline.rotationPivotDrift.test.ts
+++ b/packages/cli/src/utils/checkPipeline.rotationPivotDrift.test.ts
@@ -15,7 +15,17 @@ function rotatedAabb(elemW: number, elemH: number, angleDeg: number): { w: numbe
 
 /** One rotation sample; defaults describe a large square. */
 function sample(overrides: Partial<RotationSample> = {}): RotationSample {
-  return { time: 0, selector: "#spokes", cx: 250, cy: 250, w: 200, h: 200, angle: 0, ...overrides };
+  return {
+    time: 0,
+    selector: "#spokes",
+    cx: 250,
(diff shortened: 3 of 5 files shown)


jrusso1020 (approved):
Reviewed at head `cd1581b2cf519f8c2998a8bb55527b8954689cae`.

**Strengths**

- The new gate sits with the other false-positive gates (`checkPipeline.ts:727-735`) and its comment at `:720-721` states the actual ambiguity - viewport motion may be an ancestor carrying a correctly pivoted child - rather than restating the symptom. That is the right framing for the bug.
- The regression test at `checkPipeline.rotationPivotDrift.test.ts:189-206` is a real behavioural test, not a shape assertion: three samples whose `cx` walks 200 -> 450 -> 700 in lockstep with the ancestor's `tx` of 0 -> 250 -> 500, asserting zero findings. Remove the gate and it fails.
- The field is threaded through all three layers rather than inferred downstream, and `parseRotationSample` (`checkBrowser.ts:562-574`) rejects a sample that lacks it instead of defaulting it to `"none"`. Defaulting would have made a missing signature look like a stable chain, which is the one wrong answer here.

**important - the suppression is total rather than residual, so a real defect inside any animated ancestor is now unreportable**

`hasStableAncestorTransforms` (`checkPipeline.ts:722-724`) requires the joined signature to be identical across every sample, and `detectRotationPivotDrift:745` `continue`s the whole group when it is not. So a genuinely wrong `transformOrigin` on a child that lives inside *any* moving ancestor stops being detectable - a spinner inside a sliding card, a dial inside a panning group. That is a common composition shape, and the false negative is silent in the same way the false positive was.

The detector already has what it needs to be narrower. `maxCenterDrift(group)` at `:748` measures the child's center displacement, and the ancestor's own translation is recoverable from the very string being compared: it is a `matrix(a, b, c, d, tx, ty)`, and this PR's own fixture encodes the lockstep case exactly (`tx` 0/250/500 against `cx` 200/450/700). Subtracting the ancestor `tx`/`ty` delta from the center delta and testing the residual against the existing threshold would suppress this PR's false positive while keeping the compound case alive.

A rotating or scaling ancestor is not a simple subtraction, so a blanket skip is defensible *there*. The narrow rule is: skip when the chain's non-translation components change, subtract when only `tx`/`ty` move.

**important - the signature is an unrounded string join, so the gate can fire on jitter rather than on motion**

`ancestorTransformSignature` (`layout-audit.browser.js:1549-1557`) joins raw `getComputedStyle(ancestor).transform` values. Those resolve to `matrix()` with float components, so a sub-pixel difference anywhere in the chain - parallax, an easing curve still settling, a compositor promotion - flips the string and disables the audit for that element.

Every other field in the same sample object is rounded before it leaves the browser: `cx`, `cy`, `w`, `h` and `angle` all go through `round()` at `layout-audit.browser.js:1579-1583`, and this new one at `:1584` does not. Rounding the matrix components (or parsing them and comparing with a tolerance) makes the gate mean "the ancestors actually moved" instead of "the strings differ", which is what the comment claims it means.

**nit - the walk is bounded only by `root`**

`ancestorTransformSignature` breaks when it reaches `root`, so if `element` is not a descendant of `root` - or neither selector in `collectRotationSample` matched and `root` is not what you expect - it walks to `<html>` and collects the document's transforms. Those are stable in practice, so this is legibility rather than behaviour: an explicit `root.contains(element)` precondition, or a depth cap, would state the intent.

**Checked, not a finding**

- Adding a *required* field to `RotationSample` breaks no construction site outside this diff. The only other files naming it are `check.test.ts`, which mocks `collectRotationSample` to return `[]`, and `checkPipeline.offPivotRotation.test.ts`, which uses `OffPivotRotationSample` - a different type that merely contains the substring.
- `isRotationDriftCandidate` has exactly one caller (`checkPipeline.ts:745`), so the new gate is not inherited by any other detector.
- No sampler/parser version skew is possible. The browser script is injected from the package at run time (`checkBrowser.ts:405`, `page.addScriptTag({ content: loadBrowserScript(...) })`) and copied into the build by `packages/cli/scripts/build-copy.mjs`, so the sampler that emits the field and the parser that requires it always ship together. There is no on-disk user cache of this script.
- Base drift: this branch is 490 commits behind `main`, and `main` has since touched all four production files (2, 38, 35 and 5 lines respectively). It still composes - the three regions this PR edits are byte-identical between its merge base and `main`: the `RotationSample` interface, `isRotationDriftCandidate`, and `parseRotationSample`. The stale part is the verification in the body, not the change.

**Verdict:** APPROVE
**Reasoning:** The diagnosis is right and the plumbing is done properly at all three layers, with a test that actually exercises the gate. Both important findings are about the gate being blunter than its own comment claims - a blanket skip where a residual would do, on a signature sensitive enough to trip on rounding - and neither breaks anything that works today.

— Rames Jusso

miga-heygen (approved):
Reviewed at exact head. The rotation pivot drift audit was attributing ancestor transform motion (e.g. a parent translating across the canvas) to the child element's rotation pivot, producing false positives. Fix captures the computed-transform chain up to the composition root as a signature string, and the drift candidate gate now requires that signature to be stable across all samples. Test covers the exact case: a child rotating while its ancestor translates, correctly producing zero findings.

— Miga

miguel-heygen:
Closing this stale branch. Current main still measures local spin against viewport center displacement, so a translating parent can cause a false pivot-drift finding. A fresh focused fix remains useful, but should avoid blanket string-based suppression that also hides real pivot defects under animated ancestors.

PR #3036, opened 2026-08-04
feat(engine): let an FX tail decay instead of cutting it at the clip

vanceingalls:
Letting an effect tail decay past the end of its clip instead of being cut at the boundary.

A reverb or delay outlives the sound that caused it, but the offline render ended at the last input sample. Measured on a 1.5 s tone through a default reverb: the render cut at 1.524 s with the tail still at −29.7 dB. That is an audible chop, and it was the one place the render did not match preview.

The engine now renders far enough past the clip for the tail to fall away on its own.


miga-heygen (commented):
### Review: feat(engine): let an FX tail decay instead of cutting it at the clip — #3036

**Verdict: LGTM**

**Tail math is provably correct against the actual generators.** Reverb: `0.6 + clamp(size, 0, 1) * 2.6` exactly mirrors `synthesizeReverbImpulse` in `audioFxGraph.ts`. Delay: -60dB floor (`TAIL_FLOOR = 0.001`) uses `ceil(log(0.001) / log(fb)) * gap` — verified: 250ms at 0.35 feedback = 7 repeats = 1.75s. Tests verify the identity against the actual impulse generator at four sizes.

**Render pipeline extension is clean.** `OfflineAudioContext` created with `frames + tail` samples. Automation parsed once (earlier) and shared between tail calculation and scheduling — eliminates redundant parse.

**No silence trimming — by design.** `MAX_FX_TAIL_SECONDS` (5s) bounds the cost. The mixer's `apad,atrim=0:${totalDuration}` provides the hard stop at composition length. `tailSeconds` flows through as an optional field on `AudioTrack`.

**Minor edge case:** `knobMax` ignores the static parameter value when an automation lane exists, returning only `max(automation points)`. A lane that only dips below the static value would compute a zero tail. Unlikely in practice but worth tracking.

---
_Review by Miga_

miguel-heygen (dismissed):

— Magi

**Verdict:** APPROVE

PR #3037, opened 2026-08-04
feat(producer): fail typed when a media source is a text document, not media

vanceingalls:
## What

Refuses to probe a media source that is actually a text document (HTML, XML, SVG, JSON), failing with a typed `NotMediaPayloadError` (`code: NOT_MEDIA_PAYLOAD`, `owner: "user"`, `retryable: false`) instead of letting ffprobe emit an inscrutable `[mov,mp4,m4a,3gp,3g2,mj2 @ ...] moov atom not found`.

## Why

STUDIO-5433 — enterprise Maximus customer (`erictburkey@maximus.com`, space `5457a018a47644fc8b57f9832f20cf15`) was blocked with "moov atom not found" on video `MTS - Cancel Work Order Training` (`57a488a2585f4c1fbed86d4671cbb4a4`). Root cause (via CloudTrail + Athena on the actual failing S3 object): the a-roll element with `type: "hyperframes"` had `content.src` pointing at a `streamed-preview.html` URL — a legit 6.5 KB `<!DOCTYPE html>` page, NOT an MP4. ffprobe's `[mov,mp4,m4a,...]` prefix in the error was just its default probe order, not the file's actual format. The same signature also fired on a second unrelated user Aug 4 17:40:35Z via `/v2/distributed/plan-s3` — cross-user pattern, not a one-off.

## Reachability: only a 2xx response gets here

Worth stating precisely, because it bounds what this can and cannot classify.

`downloadToTemp` rejects a 404/410 as `UrlDownloadError("http_not_found")` before writing a byte (`urlDownloader.ts` `classifyHttpFailure`), and every ffprobe input is local — `videoFrameExtractor.ts:1420-1427` downloads an http src first. So a genuine 404 surfaces as a download failure or a silent duration-0 drop, never as `moov atom not found`.

The payloads that do reach ffprobe are served with a **success status**: a soft-404 or interstitial HTML body, an S3/CloudFront error document, or a JSON API error body relayed with a 200. Those are what this classifies.

## Relationship to the EF fix

[experiment-framework#44667](https://github.com/heygen-com/experiment-framework/pull/44667) (merged) mirrors non-trusted external media URLs into S3 at compose time and fails fast on an unfetchable URL. It removes the authoring source of dead external URLs for new Zephyr compositions. This PR still covers what it leaves:

- **No backfill** — every composition zip already in S3 keeps its external URLs, including `c6f32d10…`. Renders of those still hit the class.
- **Soft-404s pass the mirror.** Its fail-fast triggers on a non-2xx; a URL answering 200 with an HTML body downloads fine and is mirrored into S3 as `<sha>.bin` under a trusted host — permanently. That is the shape of the original incident, and after the merge this PR is what catches it at render.
- **Trusted-suffix passthrough** — `.amazonaws.com`, `.heygen.ai`, `.cloudfront.net` are never mirrored.
- **Non-Zephyr authoring** — the scrub lives in `compose_addressable_artifact`; CLI projects, Studio, and hand-authored compositions don't pass through it.

Also complements #3033, which adds the src URL to *all* ffprobe errors; this PR fails before ffprobe for the non-media-payload subset.

## How

Detection lives in `packages/engine/src/utils/notMediaPayload.ts` so all three call sites share one classifier.

**Detection** — `isNotMediaPayload(filePath)` reads up to 512 bytes (looped, so a short read on NFS/FUSE can't truncate the window), skips a UTF-8/UTF-16 BOM plus leading whitespace and NUL bytes, and reports whether the first meaningful byte is `<`, `{`, or `[`.

Three bytes instead of a shape allowlist: `<!doctype`, `<html`, `<?xml`, a prolog-less `<svg`, an S3 `<Error>` body, and Replicate's `{"detail": "requested file not found"}` all start with one of them, and no supported container does — mp4/mov open with a box size, Matroska/WebM `1A 45 DF A3`, Ogg `OggS`, RIFF `RIFF`, MPEG-TS `0x47`, FLAC `fLaC`, ADTS `FF Fx`, MP3 `ID3`. An allowlist needs a new entry per payload shape observed in production.

It never throws. An unreadable file — EISDIR on a directory src, EACCES, a temp file evicted between `existsSync` and the read, EMFILE — reports "not a document" so the real probe still surfaces the real error. This is a classifier, not a gate.

**Call sites**

| Where | Element | Behavior |
| --- | --- | --- |
| `htmlCompiler.resolveMediaDuration` | video / audio needing a resolved duration | Inside the existing probe `try`, so `<video>` surfaces the typed error and `<audio>` still degrades to duration 0 (with a warning naming the element) — the documented audio/video split is preserved |
| `preflightCompositionAssetMediaTypes` | video | Terminal. This is the only place every local media src is seen regardless of authored timing, so it catches a `data-end` video the compiler never resolves |
| `audioMixer` prepare | audio | Per-element `source` / `invalid_media` / `owner: "user"` failure, replacing a `prepare` / `ffmpeg_failed` / `owner: "system"` that fired after every frame had been captured |

Image sources are exempt: an `` may legitimately be an SVG, which ffprobe reads through its `svg_pipe` demuxer.

**Message discipline** mirrors the sibling `AssetMediaTypeMismatchError`: bounded text, a sha256 element fingerprint for correlation, and never the authored src or the payload's own bytes — producer forwards `error.message` to API clients, and `redactTelemetryString` preserves host and path for HTTP srcs. It names both causes (an unresolved nested-composition URL, or an error page / API error body carrying a success status) rather than pointing on-call at an authoring bug when an expired signed URL is the actual cause.

**Routing** — `NOT_MEDIA_PAYLOAD` is registered in `SAFE_RENDER_ERROR_CODES`, the Lambda terminal-name map, the CDK and SAM non-retryable plan lists, and the Cloud Run non-retryable set, so a deterministic authoring bug does not burn the distributed retry budget.

Overhead: one 512-byte read per probed source. No new dependencies.

## Not covered

A **remote** (`https:`) video src carrying an authored `data-end`: the preflight skips remote sources and the compiler never resolves its duration, so nothing downloads it before extraction. Closing that gap means moving the download earlier and is left out deliberately.

## Test plan

- `packages/engine/src/utils/notMediaPayload.test.ts` — 30 cases: document shapes (doctype, bare `<html>`, XML prolog, prolog-less `<svg`, S3 `<Error>`, leading comment, JSON object, JSON array), UTF-8 and UTF-16 LE/BE BOMs, a leading NUL run, whitespace overrunning the sniff window, nine real container signatures, a container containing a document byte deeper in, an empty file, a directory, and a missing path. Plus the error's routing metadata, fingerprint correlation, and that neither the src nor a token in the payload reaches the message.
- `htmlCompiler.test.ts` — `<video>` aborts with the typed error before ffprobe; `<audio>` drops to duration 0 with a warning and the render continues.
- `assetMediaType.test.ts` — preflight rejects a document under video, leaves audio to the mixer, leaves an SVG image alone, and reports the document verdict ahead of the type mismatch the same file also produces.
- `audioMixer.test.ts` — a document audio source classified as `source` / `invalid_media` / `owner: "user"` without reaching ffmpeg.
- `server.notMediaErrorCode.test.ts` — the API emits `errorCode` / `errorOwner` / `retryable` rather than `undefined`.

Green: producer unit lane 38 files, engine 1406 tests, aws-lambda 140, gcp-cloud-run 101. Producer integration matches baseline — `crossWorkerIdempotency` fails 2 tests identically on the parent commit (local `captureMode` expects `beginframe`), verified by re-running against the stashed tree.

Correcting the earlier test plan on this PR: the `htmlCompiler.mediaType.test.ts` failure was **not** a missing ffmpeg binary. Both binaries are present; that file imports from `vitest` and fails under `bun test` with `vi.waitFor is not a function`. It is the only suite asserting the 4-wide `sharedMediaProbeSemaphore` invariant across `compileForRender` — it passes under vitest, including with the sniff now running inside a probe slot.

Signed-off-by: Via <vance@heygen.com>


jerrai-bot-heygen (commented):
Reviewed. Solid, well-tested fix for a real production incident (STUDIO-5433).

**Verified independently (not just the PR body):**
- Hand-traced `isNotMediaPayload`'s byte-sniff logic (`notMediaPayload.ts`) against every documented container signature (mp4/mov box-size prefix, Matroska EBML, Ogg, RIFF, MPEG-TS, FLAC, MP3/ID3, ADTS, MPEG-PS) — none share a first-meaningful-byte with `<`/`{`/`[`, so no false positives. The BOM-then-skip-then-break loop correctly handles interleaved-NUL UTF-16 (both endiannesses) and the 512-byte sniff-window-overrun case (padding > window → correctly "unknown", not a misread).
- Traced the `resolveMediaDuration` integration in `htmlCompiler.ts`: a video-tagged error propagates (hard fail before ffprobe), an audio-tagged error degrades to duration 0 + warns — matches the documented pre-existing audio/video split, both directions covered in `htmlCompiler.test.ts`.
- Traced `preflightCompositionAssetMediaTypes`'s `sniffableReferences` video-only filter — an audio-only reference at the same path correctly skips the new check and falls through to the existing (unchanged) probe-failure handling; confirmed against the `assetMediaType.test.ts` cases including the SVG-image exemption.
- The `github-advanced-security`/fallow `unused-class-member` finding on `HtmlNotVideoError.code` is from a stale commit (`8a11e977…`) that predates the current head (`349c066a…`) — that class doesn't exist at head anymore (confirmed via `search_code` + a direct file read), so the finding is stale and doesn't apply to what's actually being merged.

**Two things worth flagging, not blockers on the logic itself:**
- `mergeable_state: dirty` — will need a rebase against current `main` before merge; base is pinned at `a99caad…` and main has moved since (a couple of release/version-bump commits), most likely just a routine collision rather than a real conflict with this PR's own changes, but worth confirming at rebase time.
- No CI test-run checks are visible on this PR from where I sit (only `WIP` and a skipped `Mintlify Deployment`) — the PR body's claimed pass counts (producer 38 files, engine 1406 tests, aws-lambda 140, gcp-cloud-run 101) are self-reported; I couldn't independently confirm they ran green via GitHub checks.

— Review by Jerrai

PR #3038, opened 2026-08-04
feat(studio): curve, snap and type a value in an automation lane

vanceingalls:
The gestures that make an automation lane editable: curving a segment, snapping, and typing an exact value.

Typing matters more than it looks. A lane is a drawing surface, and a drawing surface cannot express "exactly -6 dB" — so every precise value would otherwise be approximated by hand until it looked right.


miga-heygen (commented):
### Review: feat(studio): curve, snap and type a value in an automation lane — #3038

**Verdict: LGTM with one finding**

**Curve representation is correct.** Exponential curve (not Bezier) — `applyCurve` raises progress to `2^(2*curve)`, matching DAW conventions. The solver (`curveForDrag`) has proper degenerate-case guards (flat segments, endpoints, extremes). One scalar per segment keeps the data model lean.

**Snap behavior is clean.** Snaps to beat grid times and neighbouring points, threshold `SNAP_SEC = 0.04`. Alt bypasses snap (free placement), Shift also bypasses (fine-tuning gesture). Beat grid flows cleanly through the component pipeline.

**Architecture is excellent.** Pure math in geometry module, gesture state machine in a dedicated hook, thin input component, lane as a rendering shell. DAW-convention modifier keys (Ableton muscle memory) are a smart UX call.

**Type-a-value works** — double-click opens inline input at point position, Enter/blur commits, Escape cancels, non-numeric silently dropped, values clamped to range.

**Finding: Enter+blur double-commit.** `AutomationValueInput` has both `onKeyDown(Enter) → onCommit()` and `onBlur → onCommit()`. Pressing Enter fires `commitEdit`, then React unmounts the input (since `editing` becomes null), which fires blur, which calls `commitEdit` again with the same non-null `editing` from the closure. Impact: duplicate undo entry, not data corruption. Fix: use a functional updater in `commitEdit` so the second call sees `prev === null` and bails.

Undo/redo correctly wired through `commitPoints(_, persist)` dual-path (false for transient drag, true for gesture end).

---
_Review by Miga_

miguel-heygen (dismissed):

— Magi

**Verdict:** APPROVE

PR #3039, opened 2026-08-04
fix(producer): retry screenshot calibration timeouts

miguel-heygen:
fixes reported:1785886305.173749; no related claim remains outstanding

Source claim: https://slack.com/archives/C0BGC335AQY/p1785886305173749

When Page.captureScreenshot times out during non-screenshot auto-worker calibration, classify it as a capture-protocol timeout. The existing fallback now closes the failed session, forces screenshot mode, and retries calibration instead of retaining drawElement with a conservative single-worker plan.

Verification:
- renderOrchestrator focused suite: 187/187 passed
- Producer typecheck passed
- oxlint and oxfmt checks passed

Scope note: Terra reproduced the same current 60-second drawElement no-progress class on a 720-frame fixture, not the reporter’s exact 2,406/6,989-frame position.

### DIFF
--- a/packages/producer/src/services/render/captureCost.ts
+++ b/packages/producer/src/services/render/captureCost.ts
@@ -339,7 +339,7 @@ export function createFailedCaptureCalibrationEstimate(reason: CaptureCalibratio
 
 export interface CaptureCalibrationOutcome {
   calibration: { estimate: CaptureCostEstimate; samples: CaptureCalibrationSample[] } | undefined;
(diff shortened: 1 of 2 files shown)


miguel-heygen:
Closing as stale. Current main already has capture-protocol fallback and session cleanup. The additional screenshot-timeout classification lacks a current reproduced case showing that the screenshot retry recovers, so this branch is not being recreated.

PR #3042, opened 2026-08-05
fix(cli): upgrade project wrapper pins

miguel-heygen:
fixes reported:1785895165.335319:upgrade-wrapper-pins; no related claim remains outstanding

Source claim: https://slack.com/archives/C0BGC335AQY/p1785895165335319

upgrade --project now rewrites strict hyperframes@semver pins in regular project-root .sh wrappers alongside package.json scripts. It does not recurse into dependencies or generated trees, excludes symlinks/non-files, preserves unrelated version text, and applies the same safe-version validation and --check behavior.

Verification:
- Focused upgrade/project-pin tests: 12/12 passed
- CLI typecheck, oxlint, and oxfmt passed
- Exact Terra fixture: package scripts, hyperframes-full-cpu.sh, and publish-wrapper.sh all resolve to 0.7.92

### DIFF
--- a/packages/cli/src/commands/upgrade.project.test.ts
+++ b/packages/cli/src/commands/upgrade.project.test.ts
@@ -78,6 +78,19 @@ describe("upgradeProjectPins", () => {
     expect(pkg.scripts.render).toBe("npx --yes hyperframes@0.7.55 render");
   });
 
+  it("rewrites project-root shell wrapper pins alongside package scripts", async () => {
+    const d = project({ render: "npx --yes hyperframes@0.7.48 render" });
+    const fullCpu = join(d, "hyperframes-full-cpu.sh");
+    const publish = join(d, "publish-wrapper.sh");
+    writeFileSync(fullCpu, "npx --yes hyperframes@0.7.48 render --workers 2\n");
+    writeFileSync(publish, "./publish.sh 0.7.48 hyperframes@0.7.48\n");
+
+    await upgradeProjectPins(d, { json: false, check: false });
+
+    expect(readFileSync(fullCpu, "utf8")).toContain("hyperframes@0.7.55");
+    expect(readFileSync(publish, "utf8")).toContain("hyperframes@0.7.55");
+  });
+
   it("--check reports without writing", async () => {
     const d = project({ render: "npx --yes hyperframes@0.7.48 render" });
     const before = readFileSync(join(d, "package.json"), "utf-8");
--- a/packages/cli/src/commands/upgrade.ts
+++ b/packages/cli/src/commands/upgrade.ts
@@ -3,10 +3,10 @@ import { defineCommand } from "citty";
 import type { Example } from "./_examples.js";
 import * as clack from "@clack/prompts";
 import { execFileSync } from "node:child_process";
-import { existsSync, readFileSync, writeFileSync, renameSync } from "node:fs";
+import { existsSync, readFileSync, writeFileSync, renameSync, readdirSync } from "node:fs";
 import { resolve } from "node:path";
 import { c } from "../ui/colors.js";
-import { rewriteProjectPinnedScripts } from "../utils/projectPin.js";
+import { rewritePinnedHyperframesText, rewriteProjectPinnedScripts } from "../utils/projectPin.js";
 
 export const examples: Example[] = [
   ["Check for updates interactively", "hyperframes upgrade"],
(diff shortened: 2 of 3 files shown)


jrusso1020 (approved):
Reviewed at head `adc1610d00cd8a6c554ce3614a49b374a27c2b19`.

**Strengths**

- Extracting `rewritePinnedHyperframesText` and re-pointing `rewriteProjectPinnedScripts` at it (`projectPin.ts:21-38`, `:58-62`) keeps one implementation of the rewrite. The `changed` contract survives the refactor: the old return was `fromVersions.size > 0` over the script strings, the new one is the same predicate over a superset.
- The safe-version gate is honoured on the new path too (`projectPin.ts:25`), so an unverified version can't be written into a wrapper the user will execute.
- Writing wrappers with an in-place `writeFileSync` rather than tmp+rename preserves the file mode, which matters because these wrappers are usually `chmod +x`. That looks incidental but it is the correct choice here - see the atomicity note below, which is why it is worth stating rather than assuming.

**important - the pin regex was tuned for package.json script strings and silently misses the common shell forms**

`HYPERFRAMES_PIN_RE` (`projectPin.ts:6-7`) ends in `(?=\s|$)`. In a package.json script a pin is always followed by a space or the end of the string, so that lookahead is exactly right there. Shell wrappers are not so uniform. I ran the exact regex against realistic wrapper lines:

```
  REWROTE  npx --yes hyperframes@0.7.48 render --workers 2
  REWROTE  npx --yes hyperframes@0.7.48 && echo ok
  REWROTE  npx --yes hyperframes@0.7.48<newline>
* MISSED   CMD="npx --yes hyperframes@0.7.48"
* MISSED   RUN='npx --yes hyperframes@0.7.48'
* MISSED   npx --yes hyperframes@0.7.48; echo ok
* MISSED   OUT=$(npx --yes hyperframes@0.7.48)
* MISSED   npx --yes hyperframes@0.7.48|tee log
```

A quoted pin is textbook in a wrapper script, and so is `;`. The consequence is not just "some pins are skipped" - it is that the output at `upgrade.ts:238` now reads `bumped project pins 0.7.48 -> 0.7.55` off the *union* of versions found, with no per-file breakdown. So when package.json bumps and a quoted wrapper does not, the user is told the project moved when part of it did not.

That reporting line is also newly widened by this PR: it used to say "project scripts", which was accurate about its scope, and now says "project pins", which implies the wrappers are included. Either widen the boundary or narrow the message.

The lookahead exists to avoid matching a longer package name (per the comment at `:3-5`). A character-class negation keeps that property while accepting the shell forms - e.g. `(?![0-9A-Za-z.\-+@/])` instead of `(?=\s|$)`. Worth one test row per form above.

**important - the wrapper writes are not atomic, while the package.json write immediately above is**

`upgrade.ts:206-211` writes package.json to `${pkgPath}.tmp` and `renameSync`s it into place. `upgrade.ts:212-216` writes each wrapper with a bare `writeFileSync`, so an interrupt mid-write leaves the user with a truncated shell script. That is a worse loss than a truncated package.json, which is regenerable from the registry.

If you move these to tmp+rename, carry the mode across (`statSync(path).mode`, or `chmodSync` after the rename): rename-from-tmp would silently drop the executable bit that the current in-place write preserves. The fix needs both halves, which is the only reason the present code is safe today.

**nit - the new test does not assert the claim its fixture was built for**

`upgrade.project.test.ts:86` writes `./publish.sh 0.7.48 hyperframes@0.7.48`, which is plainly there to check that the bare positional `0.7.48` survives while the pin moves. The assertion at `:91` is only `toContain("hyperframes@0.7.55")`, which passes whether or not the bare version was also rewritten. I verified the behaviour is right (only the `hyperframes@`-prefixed token is rewritten), so this is a coverage gap rather than a defect - assert the whole expected line instead of a substring.

**nit - discovery is `.sh` in the project root only**

`entry.name.endsWith(".sh")` at `upgrade.ts:201` misses extension-less wrappers with a shebang (`bin/run`), and the deliberate no-recursion rule also excludes `scripts/*.sh`, which is an ordinary project layout rather than a dependency or generated tree. There is also no way to exempt a wrapper that is *deliberately* pinned to an old version, such as a bug-repro script; `--check` at least surfaces it before the write.

**Checked, not a finding**

- "excludes symlinks/non-files" holds. `readdirSync(dir, { withFileTypes: true })` does not follow symlinks, so a symlinked wrapper reports `isFile() === false` and is filtered at `:201`.
- "preserves unrelated version text" holds - confirmed on the fixture's own bare `0.7.48` positional, which is left alone.
- `readdirSync` runs after the `isSafeVersion(latest)` early return (`upgrade.ts:198-200`), so a failed update check does not touch the directory.
- `rewriteProjectPinnedScripts` now carries two `isSafeVersion` checks, its own at `projectPin.ts:53` and the delegate's at `:25`. Redundant, not wrong, and the only callers are `upgrade.ts` and the unit test.
- CRLF wrappers are fine: `\r` satisfies `\s`, and the carriage return is preserved.

**Verdict:** APPROVE
**Reasoning:** The refactor is clean and nothing here regresses - before this PR wrappers were never touched at all. Both important findings are about a new capability being narrower than the message it prints, not about broken behaviour, and each has a concrete one-line fix.

— Rames Jusso

miga-heygen (approved):
Reviewed at exact head. Extends `upgradeProjectPins` to rewrite `hyperframes@X.Y.Z` pins in `.sh` wrapper scripts alongside `package.json` scripts. Extracts `rewritePinnedHyperframesText` for shared use, scans root-level shell files, and correctly aggregates versions across both sources. Test covers the wrapper rewrite path. Clean refactor. LGTM.

— Miga

PR #3046, opened 2026-08-05
feat(cli): add thread message stack

miguel-heygen:
## What

Adds the `thread-message-stack` registry block together with the CLI machinery it needs: a bounded caller-payload validator that materializes the block at install time, an OAuth-only install gate, a resumable install intent that survives the browser round-trip, and a privacy-safe funnel that spans catalog, auth, install, preview, and render.

Supersedes the closed mirror attempt in #3035.

## Why

Every existing registry block ships fixed content, so `hyperframes add` only ever had to copy files. This block's content comes from the caller, which opens three holes the current add path cannot cover:

1. No way to prove which upstream catalog artifact and version an installed copy came from.
2. No way to refuse an install that is backed only by an API key, where a real account session is required.
3. No way to finish an install that was interrupted mid-flow while the user completed OAuth in a browser, without either losing the intent or replaying it twice.

The block is the vehicle. The install path is the actual change.

## How

- **Materialization is one function.** `packages/cli/src/registry/threadMessageStack.ts` validates the entire payload before touching a file: at most 100 messages, `text` at most 8192 chars, `sender` at most 256, `side` restricted to `incoming` / `outgoing`, and `stagger` / `hold` finite numbers in 0 to 60. Only then does it rewrite the single `data-hf-primitive-data` island in the block HTML. Validate-then-mutate means a rejected payload can never leave a half-written file behind.
- **Install authorization is its own boundary.** `threadMessageStackAuthorization.ts` requires a persisted OAuth credential and verifies it against the current-user endpoint. Environment and file API keys deliberately do not satisfy it, and the outcome is a closed set (`authorized` / `api-key-only` / `cancelled` / `failed`) so callers cannot treat an ambiguous result as success.
- **Install intent is durable and claimed once.** `commands/catalog-resume.ts` persists a versioned record (item, artifact, version, funnel and install ids, status) through an atomic temp-file plus rename, with an explicit claim, so an install interrupted by the OAuth round-trip resumes exactly once instead of zero or twice.
- **Telemetry has a single side-effect contract.** `telemetry/primitive-funnel*.ts` owns the lifecycle events. Discovery works fully anonymously, and the funnel context is persisted under `.hyperframes/` so preview and render stay stitched to the same funnel after the user authenticates.
- **Provenance rides in the registry item.** The registry-item schemas (`docs/schema/registry-item.json`, `packages/core/schemas/registry-item.json`) gain the upstream artifact and version fields, and the resolver validates the recorded source digest, so an installed copy is traceable to an immutable upstream projection.
- **Preview generation covers the new block.** `scripts/generate-catalog-previews.ts` and `scripts/catalog-preview-temp.ts` extended so the block is rendered by the same pipeline as every other catalog entry.

## Test plan

- 134 focused tests covering validation, authorization, resume, funnel, and the registry item, including a real installer to Studio preview to producer compile chain that produces a deterministic 8-frame MP4.
- Full CLI suite at the reviewed head: 2,498 passed, 2 skipped.
- Catalog preview scripts: 2 passed.
- `npx hyperframes lint` and `npx hyperframes check` on the installed temp project: clean, zero runtime errors.
- CLI and core typechecks, schema sync, changed-file `oxlint` / `oxfmt`, fallow audit, CLI build.
- CI at this head: 61 successful checks, 1 expected Mintlify skip.
- Known and unchanged: aggregate registry lint still reports 38 pre-existing failures across older blocks. The new block passes with 0 warnings.

- [ ] Manual testing performed (the end-to-end installer, preview, and render path is exercised in-suite rather than by hand)
- [x] Documentation updated (if applicable) - registry-item schemas

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -282,6 +282,10 @@ jobs:
       - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
           node-version: 22
+      - name: Install FFmpeg for real CLI render integration
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install -y --no-install-recommends ffmpeg
       - uses: ./.github/actions/prepare-ffmpeg-bin
       - run: bash scripts/ci/install-workspace-dependencies.sh
       - run: bun run test:scripts
--- a/docs/schema/registry-item.json
+++ b/docs/schema/registry-item.json
@@ -128,6 +128,18 @@
     "relatedSkill": {
       "type": "string",
       "minLength": 1
+    },
+    "provenance": {
+      "type": "object",
+      "required": ["kind", "artifactId", "versionId", "canonicalUri", "sourceDigest"],
+      "additionalProperties": false,
+      "properties": {
+        "kind": { "const": "heygenverse-export" },
+        "artifactId": { "type": "string", "format": "uuid" },
+        "versionId": { "type": "string", "format": "uuid" },
+        "canonicalUri": { "type": "string", "pattern": "^heygenverse://app/" },
+        "sourceDigest": { "type": "string", "pattern": "^sha256:[a-f0-9]{64}$" }
+      }
     }
   },
   "allOf": [
--- a/packages/cli/src/auth/index.ts
+++ b/packages/cli/src/auth/index.ts
@@ -25,7 +25,7 @@ export {
 
 export { configDir, credentialPath } from "./paths.js";
 
-export { tryResolveCredential } from "./resolver.js";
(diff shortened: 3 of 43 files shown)


miga-heygen (approved):
Reviewed at exact head. Large feature PR that adds the thread-message-stack primitive: OAuth authorization gate, HeyGenVerse catalog resolver, registry provenance schema, catalog-resume command, and primitive funnel telemetry. Well-structured with extensive test coverage (8 test files, integration tests for both the OAuth boundary and the data pipeline). The authorization gate correctly blocks download/materialization before registry resolution when OAuth is missing.

— Miga

PR #3050, opened 2026-08-05
feat(studio): select a time range on an automation lane

vanceingalls:
Selecting a time range on an automation lane, so an edit can address a stretch of envelope rather than one point at a time.

The range operations are pure and separately tested — given points and a range, return new points — which keeps the interaction layer thin and the arithmetic checkable.

Two fixes worth naming:
- `replaceRange` budgeted its inner points after capping rather than before, so a dense selection lost detail it should have kept
- a range selection let Delete fall through to the clip, so deleting automation could delete the clip under it


miga-heygen (commented):
### Review: feat(studio): select a time range on an automation lane — #3050

**Verdict: LGTM**

**Selection primitive is well-designed.** `AutomationSelection` in a Zustand store slice with `elementKey`, `target`, `t0`, `t1` (clip-local, ordered so t0 ≤ t1). Visual feedback: semi-transparent accent rect at 15% opacity with edge lines at 50%.

**Two operations ship:** Delete/Backspace empties the range via `replaceRange(inner: [])` with anchor pins at edges; Escape clears. The `replaceRange` API composes cleanly with future operations (#3055 shapes, #3056 paste).

**Hotkey arbitration is the critical piece.** When `automationSelection` is active, Delete/Backspace falls through WITHOUT `preventDefault` at the window level, so the document-level handler sees it. Without this, the window-level handler would delete the entire clip. Well-tested with 4 arbitration cases.

**Integration across four layers:** store binding (selection filtered by elementKey), slot (stale-selection guard), gesture hook (background drags start range selection, point drags win), and hotkey arbitration. Clean architecture.

**Test coverage is strong:** 24+ cases across 7 test files covering store slice, range operations (envelope preservation, anchor pinning, edge dedup, point cap, even thinning), hotkey arbitration, keyboard handler, gesture integration, and stale-selection guard.

**Minor findings (non-blocking):** (1) SVG `<line>` elements use `key={t}` — if `t0 === t1` both lines share the same key (gesture threshold prevents this in practice). (2) Escape doesn't `preventDefault` — confirm no double-action in fullscreen mode.

---
_Review by Miga_

miguel-heygen (dismissed):

— Magi

**Verdict:** APPROVE

PR #3055, opened 2026-08-05
feat(studio): shape and simplify an automation selection

vanceingalls:
Shaping a selected range: ramp, swell and dip generators, plus simplification for a run that has more points than it needs.

Generators because the common envelope shapes are worth one gesture rather than ten placed points. Simplify because a hand-drawn or imported lane accumulates points that no longer describe anything — they make the lane slow to edit and impossible to read.


miga-heygen (commented):
### Review: feat(studio): shape and simplify an automation selection — #3055

**Verdict: LGTM**

**Four shapes** (ramp-up, ramp-down, swell, dip) operate in unit space via `toUnit`/`fromUnit`, so log-scaled knobs (frequency) behave correctly. Edge values sampled from existing envelope via `sampleAutomationLane`, so shapes splice cleanly into whatever is already there. Boundary preservation is handled by `replaceRange`'s anchor-pinning.

**RDP simplification** is standard and correct: stack-based (avoids stack overflow on dense envelopes), vertical deviation in unit space (correct for time-series — value deviation matters, not geometric distance), default epsilon 2% of parameter range. The test verifies all original points remain within epsilon of the simplified polyline via linear interpolation.

**Shapes tested:** ramp-up floor-to-envelope, ramp-down envelope-to-floor, swell peak at range.max with easing, dip quarter-duck in unit space, log-scale correctness, empty lane fallback. **Simplification tested:** collinear collapse, wave reduction >50% within epsilon, short input passthrough. **Integration:** right-click menu opens/closes, swell inserts and commits once.

**Observation:** simplification preserves `curve` values from surviving points, but the simplified polyline may no longer match the original curvature (a curve exponent over a 10-point run may misbehave when collapsed to 3). Acknowledged in module comment as accepted reshaping.

---
_Review by Miga_

miguel-heygen (dismissed):

— Magi

**Verdict:** APPROVE

PR #3056, opened 2026-08-05
feat(studio): copy and paste automation ranges across lanes

vanceingalls:
Copying and pasting automation ranges, including across lanes.

The clipboard is internal rather than the system one: these are typed points with times and values, and round-tripping them through text would lose the typing and invite paste of anything.

Across lanes is the useful half — the same swell applied to a second parameter is a common thing to want, and rebuilding it by hand is where envelopes stop being worth drawing.

Includes a repair to the paste path and the key arbitration that decides which surface owns a keystroke.


miga-heygen (commented):
### Review: feat(studio): copy and paste automation ranges across lanes — #3056

**Verdict: LGTM**

**Cross-lane paste through unit space is the key design.** `toUnit(sourceRange, p.v)` normalizes to [0,1], `fromUnit(targetRange, ...)` maps to the target parameter. A volume duck pasted onto reverb wet lands proportionally. Curve shape preserved as-is (curve describes time-domain interpolation, not value meaning).

**Hotkey arbitration solves the double-write problem.** Two clipboard systems (clip duplication on `window` capture, automation on `document` capture) would both fire on Cmd+C/V. `automationOwnsKey` predicate lets the central dispatcher stand down without `preventDefault`. Accepted residual: dead key in edge cases beats today's double write.

**`commitTargetKey` guard** prevents the async dom-edit selection race: clicking clip B then immediately pressing Cmd+V would write B's automation onto clip A (because `onCommit` targets whatever dom-edit has selected). The guard refuses when `binding.commitTargetKey !== elementKey`. Well-tested.

**Paste chaining** is smart: `markLastPaste` + `isLastPasteSpan` distinguishes paste-created selections from user-drawn ones. Second Cmd+V chains after instead of overwriting. Test confirms: first paste at [2,4], second at [4,6].

**Project-scoped clipboard** via `useProject()` — project A's clipboard invisible to project B, switching projects permanently drops the entry. Selection cleared on project switch via `createTimelineResetState()`.

Tests cover: clipboard operations (rebase, curve preservation, cross-parameter mapping, project isolation), keyboard integration (copy, paste at playhead, chaining, dom-edit target mismatch, empty-lane copy), hotkey arbitration (all 4 combinations), store reset.

---
_Review by Miga_

miguel-heygen (dismissed):

— Magi

**Verdict:** APPROVE

PR #3058, opened 2026-08-05
feat(studio): box-select automation, dynamic voiceover carve, live automated readouts (replaces edge-stretch)

vanceingalls:
Replaces edge-stretch with the automation work that grew out of it: box selection, a dynamic voiceover carve, live automated readouts — and, since the last review, **shared automation lane rows** plus the fixes that shipping them surfaced.

## Shared automation lane rows

Several clips can share a track row. The row was named after whichever of them was selected, carried its clip count as a badge, and listed only that clip's envelopes — so a lane belonging to one slice read as governing the whole row, and clicking a sibling silently swapped which envelopes existed.

A lane is a property over time, not a clip's private strip. Clips on one row now **share a lane row when it is the same property of the same effect**, keyed by `laneGroupKey` — the label, which is what identifies a parameter to a reader. The key cannot be the lane target: node ids are minted per chain, so `fx.n1.q` on one clip and `fx.n1.q` on another may be different effects entirely.

Each clip draws its own envelope over its own span. A clip that does not automate the property leaves that stretch **empty**, rather than a flat line that would claim an envelope exists.

Gestures stay per clip — every clip keeps its own binding, its own `useAutomationLaneGestures`, its own selection box. A shared row is a shared lane *track*, not a shared envelope, so two clips' curves can never drag as one thing. That is why the slot loops over components rather than over lanes.

Row height and the label column follow the grouped count, and the header names the track (`Track 12`) once the row holds several clips. Disclosure had the same selection-dependence — it was stored against the active clip, so expanding one slice and clicking another collapsed the row — so the caret now opens and closes every clip on the row together.

**Effects stay on the clip.** There is no track entity to own a chain: `data-track-index` is parsed in one place, only to choose a row, and both runtimes build audio per element. Track FX would mean inventing a storage location and a bus node in two runtimes.

## Fixes this surfaced

- **Clip bars overpainted the lanes.** Only the clip whose property lanes were showing had its bar capped to a track height; its siblings filled the whole expanded row and painted their waveforms over the envelopes below. The bar's height belongs to the row's disclosure, not to which clip owns the lanes.
- **Clips swallowed each other's envelopes.** Every clip mounts a lane band spanning the full row width (the separator needs it), so stacked bands made whichever clip rendered last intercept hover and drag over all its siblings. The band is a positioning device, not a hit target; only the drawn parts take pointer events.
- **Selecting a clip destroyed the lanes.** They hung off the *active* clip's property lanes, so pressing a lane to select its clip moved the subtree into a different element and remounted every lane — the handles vanished at the moment of the press. They are the row's lanes, so the row mounts them. `aria-controls` now names both regions, since a clip's keyframe lanes and the track's automation lanes cannot be one element.
- **A carve outlived its voice.** The filters and envelopes a carve produces are a measurement of specific tracks; delete one and the bed keeps ducking for a voice nobody can hear. Sources that have left the composition are pruned, which re-analyses the survivors together. Existence is asked of the timeline, not of `element.ownerDocument` — the preview DOM outlives a delete. Also stops auto-apply choosing off the picker's *fallback* list: with the narration deleted, a 200 ms explosion was the only candidate and the bed carved itself against it.
- **Deleting a clip refused its own save.** Unrelated to the lanes, found while testing them, and pre-existing (reproduced on the branch point). The flow reads the file, POSTs `remove-element` — which rewrites it server-side — then saves the duration shrink expecting the content it read at the start. The server had moved on, so it came back `409 file conflict`; the save queue pauses on a conflict, so the error surfaced and the clip stayed on the timeline until a reload. The undo baseline and "what is on disk" are now separable.

## Verification

- `packages/studio`: **3656 passing**, 18 todo, 1 file skipped. `tsc`, `oxlint`, `oxfmt` clean.
- Every behavioural change was verified by **deleting the fix and confirming the test fails** — including two cases where the first attempt passed under mutation and the test had to be rewritten to actually exercise the code.
- Exercised in Studio against a four-narration fixture: shared rows render one lane row per property with per-clip envelopes and a blank stretch where a clip does not automate; hover lights only the hovered clip's handles; press-to-select keeps them; deleting a voice prunes and re-measures (1.6 kHz band moved −3.2 → −3.18 dB); delete no longer errors and undo restores the clip.

## Known limit

The carve reconciles a deleted voice when the bed's **Audio FX section is opened**, not at delete time — that panel is only mounted while the section is open. A version that watches the timeline instead is built and working on `carve-reconcile-wip`, but it disables undo, so it is parked rather than included here.


vanceingalls:
Superseded — split into nine PRs (#3207 … #3215) to keep each under the review limit. Session handoffs and the execution plan are stripped; skills/ and the design specs stay. No code changed, only the boundaries.

Issue #2550, opened 2026-07-16
Non-deterministic render: sub-compositions double-mount when video metadata loads after capture starts (timeline_rebind_after_media_metadata)


jurczykpawel:
### Summary

When a composition mounts `data-composition-src` sub-compositions **over a full-length `<video>`**, the runtime can re-run the montage after the video's `loadedmetadata` arrives (diagnostic `timeline_rebind_after_media_metadata`). If that happens **after frame capture has started**, every sub-composition is mounted a **second time**: the `data-composition-id` collision is uniquified to `…-1` (+ `data-hf-original-composition-id`), the renamed instance's scoped GSAP selectors stop matching, and captured frames composite **two phase-offset copies** of the block. The same project renders correctly when the metadata wins the race — so the output is **non-deterministic across environments**.

### Evidence — same project, same CLI, different environment

Identical project directory (index.html + sub-compositions + local mp4/wav), rendered with stock `npx hyperframes@0.7.60 render`, frame at t=14s inside a 3 s `flowchart-vertical` window:

| macOS arm64 native (~34 s render) | Docker linux/amd64, same files (~5 min render) |
|---|---|
| ![clean](https://raw.githubusercontent.com/jurczykpawel/hyperframes/evidence/subcomposition-double-mount/evidence/frame-t14-macos-native-clean.png) | ![duplicated](https://raw.githubusercontent.com/jurczykpawel/hyperframes/evidence/subcomposition-double-mount/evidence/frame-t14-docker-amd64-duplicated.png) |
| ONE instance | TWO phase-offset instances |

Removing the `<video>` (changing nothing else) renders a single clean instance in **both** environments — the media-metadata event is the trigger.

### Mechanism (as observed via runtime diagnostics)

On `loadedmetadata`/`durationchange` the runtime compares the captured root-timeline duration `g` with the media duration floor `h` and, when `V(h) && (!V(g) || h >= g + T)`, emits `timeline_rebind_after_media_metadata` and re-runs the montage:

- metadata **before** capture (fast env): the `!e.capturedTimeline` early path re-binds before anything is mounted — harmless;
- metadata **after** capture starts (slow env — big DOM, large media, slow I/O): the guarded rebind re-mounts every `data-composition-src` host mid-capture → duplicate instance → id uniquify breaks one instance's selectors → both get composited at different timeline phases.

### Which blocks make it visible

[...]

PR #2521, opened 2026-07-16
fix(renderer): prevent stale SwiftShader layers

miguel-heygen:
## What

Prevents a transformed element from flashing twice for a few frames when a GSAP yoyo animation reverses during CPU/SwiftShader rendering.

The browser state was correct, but Chromium's GPU compositor could briefly retain the element's previous transformed layer. The stale layer was already present in the raw captured JPEG, before FFmpeg encoded the video.

## How

- Add `--disable-gpu-compositing` only when using SwiftShader with BeginFrame capture.
- Keep BeginFrame itself, WebGL, hardware-GPU rendering, and screenshot capture unchanged.
- Add a 25 fps visual regression that reproduces the reversal at the affected frame boundary.
- Let regression fixtures explicitly select BeginFrame so this path stays covered.

## Why

Switching all CPU renders to screenshots hid the problem but made rendering slower. Routing only compositing through Chrome's software path removes the stale layer while preserving deterministic BeginFrame capture.

On the reduced reproduction:

- screenshot capture: ~41s
- previous BeginFrame path: ~28s, visible duplicate
- fixed BeginFrame path: ~21–29s, no duplicate

Upstream Chromium tracker and removal condition: [issue 535256667](https://issues.chromium.org/issues/535256667).

## Verification

- Original project ZIP: clean at the previously failing 10.12–10.20s window with 1 and 6 workers
- Standalone sanitized Chromium repro: control duplicates frames 253–255; adding only `--disable-gpu-compositing` is clean
- Regression proof: fails without the fix at 10.1376s (PSNR 29.47), passes 100/100 checkpoints with the fix
- WebGL BeginFrame smoke: rendered the expected non-blank canvas color
- `bun test packages/engine/src/services/browserManager.test.ts` — 28 passed
- Engine + Producer typecheck — passed
- Full workspace build — passed
- Full Engine suite — 1079 passed; 2 unrelated VFR tests require a newer FFmpeg than the host's 4.2 (`-fps_mode` unsupported)
- Fallow `--gate all` — passed (default new-only attribution could not create its temporary worktree on this busy checkout)

PR #3062, opened 2026-08-06
fix(engine): disable GPU compositing on SwiftShader for all capture modes

### LINKED ISSUE #3049: Horizontal-band ghosting in transparent renders: content duplicated at exactly renderHeight/4 offsets; persists from 0.6.101/Chrome 131 through 0.7.82/Chrome 152
### Describe the bug

When rendering certain HTML compositions to transparent video (`--format mov`, alpha), the output contains **ghost copies of page content duplicated vertically at an offset of exactly `renderHeight / 4`** (e.g. 286 px bands for a 1144 px-high render). Static, non-animated DOM elements are duplicated as well, so this is not an animation/timeline issue. The defect reproduces deterministically on the same composition and has survived a major engine upgrade.

Evidence highlights:

- Ghost copies sit at a vertical offset of **exactly one band height** (renderHeight/4); cross-correlation between the original band and the ghost band is **~99.6%** — whole-band content displacement, not motion blur or timeline error.
- A frame-differencing metric over 959 frames scores **65.8% dirty frames** on 0.6.101/Chrome 131 and **78.8%** on 0.7.82/Chrome 152 (absolute numbers not directly comparable across engines because 0.7.82 also renders the composition's background grid differently, adding a new vertical-stripe component; the band-duplication signature itself is unchanged). A control band shows ~0.5%, so the signal is structural.
- **Static `<div>`/SVG rects with no animation are duplicated too.** Trigger appears to be structural (which elements/paints the page contains), not load-related: sibling compositions with far more primitives render clean.

### Link to reproduction

Self-contained single-HTML composition available on request (contains our channel content, so not published in a public repo; we can also run candidate builds on our side)

### Steps to reproduce

1. Single-page composition: one GSAP paused timeline registered on `window.__timelines`, seek-driven frame capture (standard hyperframes particle pattern). One `<template>` root, no external assets.
2. Render: `hyperframes render <projDir> -o out.mov --format mov --fps 60 --quality high --workers 8 --no-browser-gpu`
3. Inspect the alpha channel of the output frames (e.g. extract frame 300) — ghost copies of page content appear vertically offset by exactly renderHeight/4.

### Expected behavior

Each piece of page content appears exactly once per frame; the alpha channel contains no duplicated bands.

### Actual behavior

Ghost copies of page content appear in the alpha channel, duplicated vertically at exactly renderHeight/4 offsets. The same annotated text row appears twice, one band apart, in both frames below. Reproduces on both engine generations.

Alpha masks of frame 300 from both engines are attached below (order-independent identification): the frame WITHOUT wide vertical bars is 0.6.101/Chrome 131; the frame WITH the additional wide translucent vertical bars is 0.7.82/Chrome 152 — that stripe component is new in 0.7.82 (background grid lines rendered as wide bars), while the band-duplication signature itself is unchanged across both.

alpha mask frame 300, engine A
alpha mask frame 300, engine B

### Environment

```shell
run A: hyperframes CLI/producer 0.6.101, chrome-headless-shell 131.0.6778.85 (linux)
run B: hyperframes CLI/producer 0.7.82, chrome-headless-shell 152.0.7928.2 (linux)
host: Linux x86_64, software GL (SwiftShader), no GPU flags
command: hyperframes render <projDir> -o out.mov --format mov --fps 60 --quality high --workers 8 --no-browser-gpu
```

### Additional context

Things we already ruled out / tried:

- Injecting Chromium compositor flags (`--disable-partial-raster`, large raster tiles, `--disable-checker-imaging`): **no effect** on the dirty-frame rate, so the naive "partial raster invalidation" hypothesis appears falsified — the defect seems to live in the capture path rather than in cc tile reuse. (Side note: `buildChromeArgs` is a hard-coded array with no extra-args injection point; we had to wrap the chrome binary to test this. An escape hatch like `PRODUCER_EXTRA_CHROME_ARGS` would help future debugging.)
- Merging the affected SVG `line` primitives into a single `path` makes the ghosting disappear on that composition — we use this as a content-side authoring workaround, but it is guidance, not a fix.

Questions:

1. Is the frame-capture path expected to reuse any per-band/tile buffers between `seek` captures in screenshot mode?
2. Any known invalidation issue where band N's content can be re-presented at band N±1?
3. If you have a debug flag to force full-page re-raster per captured frame, we can A/B it immediately.

miga-heygen:
## Summary

- Move `--disable-gpu-compositing` from the BeginFrame-only block to all capture modes when `browserGpuMode === "software"`
- Add `PRODUCER_EXTRA_CHROME_ARGS` env var for passing additional Chrome flags during debugging

## Problem

SwiftShader's GPU compositor retains stale transformed layers across frames, producing phantom duplicate content at exactly `renderHeight/4` vertical offsets. The artefact appears in both BeginFrame and `Page.captureScreenshot` capture modes — the existing comment documents this ("both BeginFrame and Page.captureScreenshot read the stale surface"), but the `--disable-gpu-compositing` workaround was gated to BeginFrame mode only.

Software-GPU renders are clamped to screenshot mode by the `forceScreenshot` invariant (`applyConcreteGpuScreenshotClamp`), so the workaround never applied to the most common SwiftShader capture path.

## Root cause

`buildChromeArgs` placed `--disable-gpu-compositing` inside the `if (options.captureMode !== "screenshot")` block (line 914–926). This was originally scoped to BeginFrame because BeginFrame drives the compositor step-by-step (where stale retention is deterministic), while screenshot mode was considered "good enough" because the compositor runs freely. Issue #3049 proves that SwiftShader's compositor also retains stale layers in screenshot mode, with a different pattern: band duplication at `renderHeight/4` rather than sequential-frame yoyo persistence.

## Fix

Hoist `--disable-gpu-compositing` out of the BeginFrame block so it fires for any `browserGpuMode === "software"` session, regardless of capture mode. Hardware-GPU captures are unaffected.

The `PRODUCER_EXTRA_CHROME_ARGS` escape hatch (space-separated env var) lets users inject Chrome flags without wrapping the binary — the original issue reporter specifically called this out as needed for debugging.

## Validation

- `npx vitest run packages/engine/src/services/browserManager.test.ts` — 42/42 passing (8 `resolveHeadlessShellPath` tests require `bun`, pre-existing skip in this env)
- Test updated: "disables GPU compositing for all software capture modes (#2550, #3049)" now asserts `--disable-gpu-compositing` present for BOTH `softwareBeginFrame` and `softwareScreenshot`
- New test: "appends PRODUCER_EXTRA_CHROME_ARGS when set"
- `bunx oxlint packages/engine/src/services/browserManager.ts` — 0 warnings, 0 errors
- `bunx oxfmt --check` — clean

Closes #3049

### DIFF
--- a/packages/engine/src/services/browserManager.test.ts
+++ b/packages/engine/src/services/browserManager.test.ts
@@ -132,7 +132,7 @@ describe("buildChromeArgs browser GPU mode", () => {
     expect(args).not.toContain("--enable-gpu-rasterization");
   });
 
-  it("disables GPU compositing only for software BeginFrame capture", () => {
+  it("disables GPU compositing for all software capture modes (#2550, #3049)", () => {
     const softwareBeginFrame = buildChromeArgs(
       { ...base, captureMode: "beginframe" },
       { browserGpuMode: "software" },
@@ -145,10 +145,27 @@ describe("buildChromeArgs browser GPU mode", () => {
       { ...base, captureMode: "beginframe", platform: "linux" },
       { browserGpuMode: "hardware" },
     );
+    const hardwareScreenshot = buildChromeArgs(
+      { ...base, captureMode: "screenshot", platform: "linux" },
+      { browserGpuMode: "hardware" },
+    );
 
     expect(softwareBeginFrame).toContain("--disable-gpu-compositing");
-    expect(softwareScreenshot).not.toContain("--disable-gpu-compositing");
+    expect(softwareScreenshot).toContain("--disable-gpu-compositing");
     expect(hardwareBeginFrame).not.toContain("--disable-gpu-compositing");
+    expect(hardwareScreenshot).not.toContain("--disable-gpu-compositing");
+  });
+
+  it("appends PRODUCER_EXTRA_CHROME_ARGS when set", () => {
+    process.env.PRODUCER_EXTRA_CHROME_ARGS =
+      "--disable-partial-raster --force-gpu-mem-available-mb=4096";
+    try {
+      const args = buildChromeArgs(base);
+      expect(args).toContain("--disable-partial-raster");
+      expect(args).toContain("--force-gpu-mem-available-mb=4096");
+    } finally {
+      delete process.env.PRODUCER_EXTRA_CHROME_ARGS;
+    }
   });
 
   it("uses Metal-backed ANGLE for hardware browser GPU mode on macOS", () => {
--- a/packages/engine/src/services/browserManager.ts
+++ b/packages/engine/src/services/browserManager.ts
@@ -910,20 +910,21 @@ export function buildChromeArgs(
     chromeArgs.push(WEBGPU_FLAG);
   }
 
+  // SwiftShader's GPU compositor can retain a transformed layer across frames,
+  // producing phantom duplicate content at regular vertical offsets (HF#2550,
+  // HF#3049). The artefact appears in BOTH BeginFrame and Page.captureScreenshot
+  // capture modes: a band of page content is duplicated at exactly
+  // renderHeight/4 in the alpha channel, and static DOM elements (SVG lines,
+  // rects) are affected equally. Routing compositing through Chrome's software
+  // path eliminates the stale-layer retention. Hardware-GPU captures keep their
+  // existing compositor path. Remove this workaround once the pinned
+  // chrome-headless-shell includes https://issues.chromium.org/issues/535256667.
+  if (browserGpuMode === "software") {
+    chromeArgs.push("--disable-gpu-compositing");
+  }
+
   // BeginFrame flags — only when using chrome-headless-shell on Linux
   if (options.captureMode !== "screenshot") {
-    // SwiftShader's GPU compositor can retain a transformed layer for several
-    // sequential frames after a GSAP yoyo/reversal. The DOM and timeline are
-    // already at the requested time, but both BeginFrame and
-    // Page.captureScreenshot read the stale surface (the duplicate is present
-    // in the raw JPEG before encoding). Keep deterministic BeginFrame capture,
-    // but route compositing through Chrome's software path when the browser is
-    // already in software-GPU mode. Hardware-GPU and screenshot captures keep
-    // their existing compositor paths. Remove this workaround once the pinned
-    // chrome-headless-shell includes https://issues.chromium.org/issues/535256667.
-    if (browserGpuMode === "software") {
-      chromeArgs.push("--disable-gpu-compositing");
-    }
     chromeArgs.push(
       "--deterministic-mode",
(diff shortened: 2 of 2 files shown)


miguel-heygen:
Closing this. The premise does not hold, and the change measurably degrades software renders.

## The one true part

`--disable-gpu-compositing` really was unreachable in practice. `browserGpuMode === "software"` forces screenshot capture (`packages/engine/src/config.ts`), and the flag was gated on BeginFrame, so only the pinned `software-beginframe-yoyo-compositor` fixture ever exercised it. That observation is correct.

## Why the fix is wrong

**The stated root cause is falsified.** This PR claims SwiftShader retains stale layers in screenshot mode too. I rendered `packages/producer/tests/software-beginframe-yoyo-compositor/src` in software screenshot mode on `main` (25 fps, png-sequence, `--no-browser-gpu --workers 1`). Frames 254 and 255, the exact window where the #2550 ghost badge appears on the unfixed BeginFrame path, are clean. Two output badges, no third. That is the one composition where stale-layer retention is known to reproduce, and it does not reproduce on the screenshot path.

The existing comment does not say what the PR body says it does. It says both readback methods see the stale surface, which was measured inside a BeginFrame session. #2521's evidence was BeginFrame only.

**The flag makes software output worse, not better.** Same transparent composition (1280x1144, static SVG line primitives, alpha), rendered software vs a hardware-GPU render of the same file as ground truth. Frame 30:

| build | pixels differing | max channel delta | fully opaque pixels with wrong RGB |
|---|---|---|---|
| `main`, software | 1.61% | 12 | 0 |
| this PR, software | 2.53% | 255 | 2698 |

`main`'s software path already matches hardware to within antialiasing noise. Adding `--disable-gpu-compositing` changes how SwiftShader rasterizes thin strokes and glyph edges, and pushes the output away from ground truth. That is a fidelity regression on the alpha path, which is the path #3049 is about.

**CI caught the same thing.** `png-sequence` failed 60 of 60 frames. That suite is the strict RGBA byte identity gate for transparent renders. `main` was green on the same code base minutes earlier.

**No reproduction.** The validation section lists unit tests over `buildChromeArgs` and nothing that exercises #3049. We do not ship capture-path changes without a repro.

## What #3049 actually looks like

The reporter's own evidence points away from the compositor: compositor flags had no effect, static elements duplicate, it survives Chrome 131 through Chrome 152, and merging the SVG `<line>` primitives into a single `<path>` makes it disappear. That is a content-dependent paint or raster signature, not layer retention.

I built a matching composition (240 static SVG lines, transparent background, yoyo timeline) and could not reproduce the banding on either build. Follow-up is on the issue: we need the reporter's composition before guessing again.

## Salvageable

`PRODUCER_EXTRA_CHROME_ARGS` is a reasonable ask and the reporter called for it explicitly. Worth landing on its own, separate from any capture-path change.

PR #3064, opened 2026-08-06
docs(skills): fix anime.js v3 syntax in v4 adapter guidance

### LINKED ISSUE #3293: anime.js adapter and skill document v3 API while loading the v4 CDN bundle
The anime.js adapter and its skill document **v3 API** while loading the **v4 CDN bundle**, so any agent or user following the docs writes code that fails at runtime.

### Where

`packages/core/src/runtime/adapters/animejs.ts` — the docblock states *"Supports anime.js v4+ (the `.seek(timeMs)` API)"*, then the usage example right below it uses v3:

```js
<script src="https://cdn.jsdelivr.net/npm/animejs@4.0.2/lib/anime.iife.min.js"></script>
<script>
  const anim = anime({
    targets: '.box',
    translateX: 250,
    duration: 2000,
    autoplay: false,
  });
```

`skills/animejs/SKILL.md` has the same problem in both examples — the basic pattern uses `anime({ targets: ..., easing: "easeOutExpo" })` and the timeline pattern uses `anime.timeline({ ... })`.

In anime.js v4 the global callable `anime()` and `anime.timeline()` were removed in favour of `anime.animate(target, params)` and `anime.createTimeline(params)`; `targets` is now the first argument rather than a property, and `easing` was renamed to `ease`.

### Evidence that v4 is the intended target

The integration test under `packages/producer/tests/animejs-adapter/src/index.html` already uses the correct v4 API:

```js
const tl1 = anime.createTimeline({ autoplay: false });
tl1.add('#label', { opacity: [0, 1], translateY: [-30, 0], duration: 400, ease: 'out(4)' }, 200);
```

So the test and the docs disagree, and the test is the one that runs.

### Impact

Agents are explicitly told to read the skill before authoring, so the broken snippets are the most likely thing to get copied. The failure is silent at author time and only shows up as a dead composition at render.

### Suggested fix

Align the docblock in `adapters/animejs.ts` and both examples in `skills/animejs/SKILL.md` with the v4 API already exercised by the adapter test — or, if v3 is still meant to be supported, pin the CDN example to `animejs@3` so the snippet at least matches the bundle it loads.

youngeuibae:
## Problem

The Anime.js guidance loaded a v4 bundle but taught the incompatible v3 API:

```js
anime({ targets: ".mark", easing: "easeOutExpo" });
anime.timeline().add({ targets: ".mark" });
```

Anime.js v4 exposes a namespace object instead: `anime.animate(targets, params)` and `anime.createTimeline()`. It also renamed `easing` to `ease`, changed ease names, and moved timeline targets to the first `add()` argument. v4 also has no `anime.running`, so the documented auto-discovery fallback cannot discover v4 instances.

## Changes

- Rewrite the adapter and skill examples for the v4 namespace API.
- Require explicit `window.__hfAnime` registration and document deterministic v4 usage.
- Correct the keyframe reference's timeline example.
- Align the public and adapter-local `anime` global types with v4 while retaining optional v3 `running` discovery for compatibility.
- Add a source-level regression test that rejects executable v3 call sites and pins both global type shapes.
- Regenerate `skills-manifest.json`.

The producer fixtures remain on `animejs@4.0.2/lib/anime.iife.min.js`; that URL still resolves and those fixtures already use the correct v4 `createTimeline()` API.

## Attribution

The documentation fix remains authored by @youngeuibae in the first commit. The branch was rebased onto current `main`, and the maintainer follow-up only adds the missing type alignment and regression guard.

## Verification

- Published `animejs@4.0.2` and `animejs@4.5.0` exports inspected: both expose `animate` / `createTimeline`, neither exposes a callable default or `running`.
- `node --test scripts/animejs-v4-guidance.test.mjs` — 5/5 pass.
- `vitest run --environment jsdom packages/core/src/runtime/adapters/animejs.test.ts` — 15/15 pass.
- Pinned `oxfmt` check and `oxlint` on changed source/test files — clean.
- `bun packages/cli/scripts/gen-skills-manifest.ts --check` — in sync.

Full workspace typechecking is left to CI because this isolated review worktree intentionally has no installed/generated workspace dependency tree.

Fixes #3293


miguel-heygen (approved):

miguel-heygen (approved):
Reviewed at `37fab5a7fdbdba1e4754ed3ab0ac76c9a4f8a0a1`.

The v4 API claims match the published 4.0.2 and 4.5.0 packages and the official v4 API shape. Executable examples now use `animate(targets, params)` / `createTimeline()`, explicit adapter registration is accurately required, the public and local global types no longer teach v3, and the new source contract test protects both the examples and type shapes. The contributor's authored commit remains intact after the rebase.

No code-level blockers found. Required CI should remain the merge gate.

**Verdict:** APPROVE
**Reasoning:** The issue is reproduced, the fix covers every repository guidance/type surface found, regression coverage exercises the failure mode, and targeted tests plus manifest/format/lint checks pass.

— Magi

PR #1655, opened 2026-06-22
fix(compiler): skip CSS var() in font resolver

miga-heygen:
## Summary

- Skip `var()` expressions in `extractRequestedFontFamilies()` — they are CSS custom property references, not font family names
- Add fail-closed test verifying `var()` font-family values don't trigger `FONT_FETCH_FAILED`

Closes #1654

## Problem

The regex-based font scanner in `deterministicFonts.ts` treated `var(--ui-font)` as a literal font family name. When a composition uses CSS variables for `font-family` values, the distributed renderer's fail-closed mode threw `FONT_FETCH_FAILED` because `var(--ui-font)` cannot be resolved as a Google Font, bundled alias, or local font.

## Fix

One-line guard: skip any family that starts with `var(`. These expressions resolve at browser paint time via CSS custom property substitution. The distributed renderer uses headless Chrome, which handles variable resolution normally at paint time — the font just won't be pre-embedded as a `@font-face` rule, which is correct behavior.

## Test plan

- [ ] New test: `does NOT throw when font-family uses CSS var() references` — verifies compositions with `var(--ui-font)` and `var(--vowel-font)` pass through fail-closed mode without error
- [ ] Existing tests still pass (no regressions in font resolution pipeline)
- [ ] CI green

— Miga

Issue #1654, opened 2026-06-22
fix(compiler): font resolver treats CSS var() as literal font names, fails distributed renders


miga-heygen:
## Bug

The font resolver in `deterministicFonts.ts` treats CSS custom property references (`var(--my-font)`) as literal font family names. When a composition uses CSS variables for font-family values, the distributed renderer's fail-closed mode throws `FONT_FETCH_FAILED` because `var(--my-font)` cannot be resolved as a Google Font, bundled alias, or local font.

## Root Cause

`extractRequestedFontFamilies()` collects all font-family values via regex. `parseFontFamilyValue()` splits on commas and strips quotes, but has no awareness of CSS `var()` functions. The result: `var(--my-font)` passes through as a font name candidate, fails all resolution paths, and becomes "unresolved."

The pipeline is entirely regex-based (no DOM, no computed styles), so CSS custom properties cannot be resolved statically.

## Reproduction

Any composition that uses CSS custom properties for font-family values will fail distributed rendering:

```html
<style>
  :root { --ui-font: "Inter"; }
  body { font-family: var(--ui-font), sans-serif; }
</style>
```

Error:
```
FONT_FETCH_FAILED: [Compiler] Unresolved fonts in fail-closed mode: var(--ui-font). 
Distributed renders require all fonts to be resolvable.
```

## Fix

Skip `var()` expressions in `extractRequestedFontFamilies()`. These are not font family names — they resolve at browser render time via CSS custom property substitution. The distributed renderer uses headless Chrome, which resolves CSS variables normally at paint time.

If pre-embedding of variable-referenced fonts is desired later, a follow-up can parse `:root` definitions and substitute values before the font scan.

PR #3067, opened 2026-08-06
fix(compiler): split font-family only on top-level commas

### LINKED ISSUE #3066: fix(compiler): font resolver mis-parses var() fallback, yielding "inherit)" as a font name
## Describe the bug

`font-family: var(--brand-font, <fallback>)` breaks distributed rendering in fail-closed mode. The font resolver splits the family stack on commas without tracking parentheses, so the `var()` fallback argument is torn off as a separate token and treated as a requested font family. The render aborts before the first frame:

```
FontFetchError: [Compiler] Unresolved fonts in fail-closed mode: inherit).
Distributed renders require all fonts to be resolvable.
```

`inherit)` is not a font — it is the tail of the `var()` expression, with the closing paren still attached.

This is the sibling of #1654 / #1655. That fix handled the bare `var(--x)` form; the `var(--x, <fallback>)` form was left broken.

## Link to reproduction

No repo link — the reproduction is a single HTML file, inlined below, and the defect is visible statically in the source on `main` (details under Root cause). Happy to push a repo if that would still help triage.

```html
<!doctype html>
<html>
  <head>
    <style>
      /* --brand-font intentionally left undefined, as it would be when a
         theme layer supplies it at runtime; see note below for the
         "defined" case, which fails too. */
      .title {
        font-family: var(--brand-font, inherit);
      }
    </style>
  </head>
  <body>
    <div id="main-comp" data-composition-id="repro" data-width="640" data-height="360" data-start="0">
      <div class="title">Hello</div>
    </div>
  </body>
</html>
```

## Steps to reproduce

1. `npx hyperframes init repro --non-interactive --example blank`
2. Set the composition's `src/index.html` to the file above (the only change needed is `font-family: var(--brand-font, inherit)` on a styled element).
3. Run a distributed render with fail-closed font fetching enabled.

## Expected behavior

`var(--brand-font, inherit)` is one CSS value. It should either be skipped entirely by the resolver (the #1655 behaviour for `var()` expressions) or resolved through the custom property, but under no circumstance should the fallback argument be extracted as a font family name.

The render should proceed, with Chrome performing the custom-property substitution at paint time as it already does for the bare `var()` form.

## Actual behavior

The render aborts before the first frame with `FontFetchError` naming `inherit)` as an unresolved font.

### Root cause

Verified against `main` @ `88853f1`; the blob is byte-identical to `v0.7.94`, so the latest release is affected.

`parseFontFamilyValue()` — `packages/producer/src/services/deterministicFonts.ts:58-63` — splits on `,` with no parenthesis awareness:

```ts
export function parseFontFamilyValue(value: string): string[] {
  return value
    .split(",")
    .map((piece) => piece.trim().replace(/^['"]/, "").replace(/['"]$/, "").trim())
    .filter((piece) => piece.length > 0);
}
```

So the `var()` expression is torn in half:

| declaration | parsed families |
| --- | --- |
| `var(--ui-font), sans-serif` | `["var(--ui-font)", "sans-serif"]` — fine today |
| `var(--brand-font, inherit)` | `["var(--brand-font", "inherit)"]` |
| `var(--brand-font, "Some Font"), sans-serif` | `["var(--brand-font", "Some Font\")", "sans-serif"]` |
| `var(--brand-font, sans-serif)` | `["var(--brand-font", "sans-serif)"]` |

Then `extractRequestedFontFamilies()` — same file, lines 418-433 — filters the pieces:

```ts
const normalized = originalCase.toLowerCase();
if (!normalized || GENERIC_FAMILIES.has(normalized)) continue;
if (normalized.startsWith("var(")) continue;          // added by #1655
if (!requested.has(normalized)) requested.set(normalized, originalCase);
```

Both guards miss the second fragment. `"var(--brand-font"` is correctly skipped by the `startsWith("var(")` check, but `"inherit)"` is not: the trailing paren means it never matches `GENERIC_FAMILIES` (line 34) and it does not start with `var(`. It falls through as a requested family, fails every resolution path, and lands in `unresolved` at the fail-closed throw site (lines ~1132-1139).

Two things worth flagging:

1. **This is not limited to CSS-wide keywords.** Any fallback is affected. A quoted fallback yields `Some Font")` — the leading quote is stripped but the trailing one is not, since the last character is `)`, not a quote. Even a generic-family fallback yields `sans-serif)`, which misses the `GENERIC_FAMILIES` set for the same reason. Every one of these is equally unresolvable.

2. **Defining the custom property does not save you.** `resolveFontFamilyDeclarationFamilies()` (lines 263-274) handles the primary `var()` correctly — `primaryCssVariableName()` at line 238 *is* parenthesis-aware and extracts `--brand-font` properly — but the last line is:

   ```ts
   return [...parseFontFamilyValue(resolved), ...families.slice(1)];
   ```

   `families` came from the naive splitter, so `families.slice(1)` is `["inherit)"]`, not `[]`. The poisoned fragment is carried through even on the happy path where the property resolves. The fix therefore needs to be in the splitter (or upstream of it), not only in the skip list.

### Why the existing regression coverage did not catch it

Not a criticism of #1655, just an explanation of how the case survived — the fixture added there, `packages/producer/tests/distributed/css-var-fonts/src/index.html`, uses `var(--ui-font), sans-serif` and `var(--display-font), sans-serif`. There is no `var()` fallback argument anywhere in it, and the unit tests in `deterministicFonts-failClosed.test.ts` (the `var()` cases around lines 245-275) are all the same bare shape. Every `var()` in the suite is the one form that happens to survive a naive comma split. Adding a `var(--x, <fallback>)` line to that fixture would lock this down.

### Suggested fix

A suggestion, not a prescription — you'll know the codebase's preferences better.

Make the stack splitter parenthesis-aware so a `var()` expression stays a single token: walk the string tracking paren depth and only split on commas at depth 0. `primaryCssVariableName()` at line 242 already has exactly this depth-tracking loop, so the logic could likely be shared. `parseFontFamilyValue("var(--brand-font, inherit)")` would then yield `["var(--brand-font, inherit)"]`, which the existing `startsWith("var(")` guard from #1655 already skips correctly, and `families.slice(1)` would correctly be empty.

An alternative would be to strip whole `var(...)` expressions before the comma split, though that loses the sibling families in a stack like `var(--a, X), Helvetica`, so the depth-aware split seems cleaner.

## Environment

```shell
Version   0.7.94 (latest)
Node.js   v22.14.0 (darwin arm64)
```

(Hand-written rather than pasted `doctor` output, to avoid including local filesystem paths. The root cause above is verified by reading the source at `main` / `v0.7.94` rather than inferred from the local install.)

## Additional context

Prior art: #1654 (issue) and #1655 (fix) cover the same code path and the same function. This report is the fallback-argument case that the earlier fix did not reach.

`var(--x, <fallback>)` is a common authoring pattern — it is the standard way to write a themeable font stack that still degrades gracefully when the theme layer has not defined the property — so this is likely to be hit by anyone using CSS custom properties for typography. It fails closed, meaning distributed renders fail outright rather than degrading.

akzarma:
## What

- Split `font-family` values on **top-level** commas only, so a `var()` expression stays a single token
- Regression coverage: parser unit cases + a fail-closed case using a `var()` fallback, and the distributed `css-var-fonts` fixture now exercises the fallback form

Closes #3066

## Why

`parseFontFamilyValue()` in `packages/producer/src/services/deterministicFonts.ts` split the family stack on every comma with no parenthesis awareness, so:

```css
font-family: var(--brand-font, inherit);
```

parsed to **two** tokens — `var(--brand-font` and `inherit)`.

The guard added in #1655 (for #1654) skips only tokens that start with `var(`, so the orphan fragment `inherit)` survived as a "requested font family". It matches no bundled alias, no Google Fonts family and no system font, so fail-closed distributed renders abort before the first frame:

```
FontFetchError: [Compiler] Unresolved fonts in fail-closed mode: inherit).
Distributed renders require all fonts to be resolvable.
```

Any fallback argument is affected, not just CSS-wide keywords — `var(--brand-font, "Inter")` leaks `Inter")` the same way. #1655 fixed the bare `var(--x)` form and its regression fixture uses no fallback argument, which is why this shape survived.

## How

The splitter now walks the value and only cuts at commas seen at paren depth 0, which keeps `var(--x, fallback)` — and nested `var(--x, var(--y, "Inter"))` — as one token, so the existing `var(` guard and `resolveFontFamilyDeclarationFamilies()` both see the whole expression. No special-casing of `inherit` or any other keyword.

Quotes are tracked in the same pass for two reasons: a parenthesis inside a quoted family name would otherwise skew the depth counter, and a legal quoted comma (`"Display, Condensed"`) no longer splits into two junk names. Trimming, surrounding-quote stripping and empty-entry dropping are unchanged.

Adversarial notes, in case they matter to a reviewer:

- `resolveFontFamilyDeclarationFamilies()` is unaffected in shape — `families[0]` is now the complete `var()` expression it always meant to be, and `families.slice(1)` correctly holds only the *external* fallbacks, not the ones inside the parens.
- A stray unmatched `)` is harmless: depth is clamped at 0.
- Malformed input (unbalanced `(`, unterminated quote) consumes the rest of the value as one token. That is invalid CSS which the browser would also discard, and it fails *closed* toward "skip", not toward "request a bogus family". Left as-is rather than growing the change.
- Not addressed here (pre-existing, and out of scope): when a `var()` primary is undefined, a concrete font in its fallback argument still isn't pre-embedded — it is skipped like any other `var()`. That is #1655's behaviour, unchanged, and strictly better than today's hard failure.

The distributed fixture change (`var(--display-font)` → `var(--display-font, "Montserrat")`) is deliberately render-neutral: `--display-font` is defined as `"Montserrat"` in `:root`, so the computed value and the embedded faces are identical and the existing baseline mp4 stays valid. Reverting the source fix makes that fixture fail resolution again.

## Test plan

- [x] Unit tests added/updated — 3 parser cases in `planValidation.test.ts` (var() fallback, nested var() fallback, quoted comma) and `does NOT throw when font-family uses a CSS var() reference with a fallback` in `deterministicFonts-failClosed.test.ts`
- [x] All 4 new tests verified **failing** with the source change reverted and passing with it
- [x] `bun test` over the touched files: 49 pass / 0 fail
- [x] `bun test` over every `deterministicFonts*` + `planValidation` test file: 84 pass / 0 fail
- [x] `oxfmt --check` and `oxlint` clean on the touched files
- [ ] Distributed regression lane (`css-var-fonts` baseline) — not run locally, no render infrastructure; relying on CI

### DIFF
--- a/packages/producer/src/services/deterministicFonts-failClosed.test.ts
+++ b/packages/producer/src/services/deterministicFonts-failClosed.test.ts
@@ -253,6 +253,17 @@ describe("injectDeterministicFontFaces — failClosedFontFetch: true", () => {
     expect(result).toBe(html);
   });
 
+  it("does NOT throw when font-family uses a CSS var() reference with a fallback", async () => {
+    const html = `<!doctype html><html><head><style>
+      .title { font-family: var(--brand-font, inherit); }
+    </style></head><body><h1 class="title">hello</h1></body></html>`;
+    const result = await injectDeterministicFontFaces(html, {
+      failClosedFontFetch: true,
+      fetchImpl: makeFailingFetch(),
+    });
+    expect(result).toBe(html);
+  });
+
   it("resolves simple CSS var() font aliases when injecting deterministic fonts", async () => {
     const html = `<!doctype html><html><head><style>
       :root { --ui-font: "Inter"; --vowel-font: "Montserrat"; }
--- a/packages/producer/src/services/deterministicFonts.ts
+++ b/packages/producer/src/services/deterministicFonts.ts
@@ -54,10 +54,44 @@ export const GENERIC_FAMILIES: ReadonlySet<string> = new Set([
  * Whitespace and surrounding `"…"` / `'…'` quotes are stripped; case is
  * preserved. Pass each name through `normalizeFamilyName` for case-
  * insensitive comparisons.
+ *
+ * Only top-level commas split: a `var(--x, fallback)` expression stays one
+ * token, as does a comma inside a quoted family name.
  */
 export function parseFontFamilyValue(value: string): string[] {
-  return value
-    .split(",")
+  const pieces: string[] = [];
+  let start = 0;
+  let depth = 0;
+  let quote: "'" | '"' | null = null;
+  for (let index = 0; index < value.length; index += 1) {
+    const char = value[index];
+    if (char === "\\") {
+      index += 1;
+      continue;
+    }
+    if (quote) {
+      if (char === quote) quote = null;
+      continue;
+    }
+    if (char === "'" || char === '"') {
+      quote = char;
+      continue;
+    }
+    if (char === "(") {
+      depth += 1;
+      continue;
+    }
+    if (char === ")") {
+      depth = Math.max(0, depth - 1);
+      continue;
+    }
+    if (char !== "," || depth !== 0) continue;
+    pieces.push(value.slice(start, index));
+    start = index + 1;
+  }
+  pieces.push(value.slice(start));
+
+  return pieces
     .map((piece) => piece.trim().replace(/^['"]/, "").replace(/['"]$/, "").trim())
     .filter((piece) => piece.length > 0);
 }
--- a/packages/producer/src/services/render/planValidation.test.ts
+++ b/packages/producer/src/services/render/planValidation.test.ts
@@ -169,6 +169,26 @@ describe("parseFontFamilyValue", () => {
     expect(parseFontFamilyValue(`'My Custom Font', serif`)).toEqual(["My Custom Font", "serif"]);
   });
 
+  it("keeps a comma inside a quoted family name", () => {
+    expect(parseFontFamilyValue(`"Display, Condensed", serif`)).toEqual([
+      "Display, Condensed",
+      "serif",
+    ]);
+  });
+
+  it("keeps a var() fallback in a single token", () => {
+    expect(parseFontFamilyValue(`var(--brand-font, inherit), sans-serif`)).toEqual([
+      "var(--brand-font, inherit)",
+      "sans-serif",
+    ]);
+  });
+
+  it("keeps a nested var() fallback in a single token", () => {
+    expect(
+      parseFontFamilyValue(`var(--brand-font, var(--fallback-font, "Inter")), sans-serif`),
+    ).toEqual([`var(--brand-font, var(--fallback-font, "Inter"))`, "sans-serif"]);
+  });
+
   it("ignores empty entries (trailing commas)", () => {
     expect(parseFontFamilyValue(`Inter,,sans-serif`)).toEqual(["Inter", "sans-serif"]);
   });
--- a/packages/producer/tests/distributed/css-var-fonts/src/index.html
+++ b/packages/producer/tests/distributed/css-var-fonts/src/index.html
@@ -35,7 +35,7 @@
         top: 30%;
         left: 50%;
         transform: translateX(-50%);
-        font-family: var(--display-font), sans-serif;
+        font-family: var(--display-font, "Montserrat"), sans-serif;
         font-size: 48px;
         font-weight: 900;
         color: #e94560;


miga-heygen (commented):
## Review: font-family top-level comma splitting

### Is this the best solution?

Yes. The hand-written scanner is the right call for this specific problem. I considered three alternatives:

1. **Regex** (e.g. `,(?=(?:[^()]*\([^()]*\))*[^()]*$)`) — doesn't handle nesting or quotes reliably. The nested `var(--x, var(--y, "Inter"))` case would need a context-free grammar, not a regular expression.

2. **postcss-value-parser** — a real CSS value parser that handles the full grammar. But it's a new dependency for a single utility, and `parseFontFamilyValue` is a hot path (font extraction across all stylesheets in every compile). The scanner is ~30 lines, self-contained, and does exactly what's needed.

3. **Patch the downstream `var(` guard** to also match fragments like `inherit)` — wrong level of abstraction. The bug is in the parser, not the filter. Fixing the parser is SSOT-correct: every consumer gets the right tokens instead of each one learning to work around broken ones.

The scanner is minimal, correct, and fails closed on malformed input (unterminated quote / unbalanced paren → consumes the rest as one token, which is invalid CSS the browser would also discard, so it skips rather than requesting a bogus family). That's the right failure mode.

### What I verified

**Scanner correctness.** Walked all five state transitions:

| Character | Behavior |
|---|---|
| `\` | Skip next char (handles `\"` in quotes, `\,` in values) |
| `'` or `"` outside quotes | Enter quote mode |
| Matching quote | Exit quote mode |
| `(` outside quotes | `depth += 1` |
| `)` outside quotes | `depth = Math.max(0, depth - 1)` (clamp prevents negative on stray `)`) |
| `,` at depth 0, outside quotes | Split |
| Everything else | Continue |

Priority is correct: escape check first (so `\"` doesn't exit quotes), then quote check (so `(` inside quotes doesn't bump depth).

**Trailing backslash edge case.** If `\` is the last character: `index += 1` → `value.length`, `continue` → loop increment → `value.length + 1`, condition fails, loop exits. `value.slice(start)` captures the trailing backslash in the final piece. Correct — malformed CSS, fails closed.

**Post-processing chain.** The `.trim().replace(/^['"]/, "").replace(/['"]$/, "").trim().filter(nonEmpty)` is unchanged. Verified it doesn't corrupt `var()` tokens — outer quotes aren't present on a `var(--x, "Inter")` token, so nothing is stripped. The `var(` guard at line 462 correctly skips the whole expression.

**Downstream consumer: `resolveFontFamilyDeclarationFamilies`.** With the fix, `families[0]` for `var(--brand-font, inherit), sans-serif` is the complete `var(--brand-font, inherit)` token. `primaryCssVariableName` correctly extracts `--brand-font` via its own paren-aware walk. `families.slice(1)` holds only the external fallbacks (`["sans-serif"]`), not the var's internal fallback. Shape is correct.

**Downstream consumer: `extractRequestedFontFamilies`.** The `normalized.startsWith("var(")` guard at line 462 correctly skips the complete var token. No fragments like `inherit)` survive to the font-request map. Fail-closed mode no longer aborts.

### Test coverage

- **Parser unit tests** (3 new): `var()` fallback, nested `var()` fallback, quoted comma — all verify the split produces the expected token list.
- **Fail-closed integration test** (1 new): `var(--brand-font, inherit)` in a full HTML document through `injectDeterministicFontFaces` with `failClosedFontFetch: true` — confirms no `FontFetchError`.
- **Distributed fixture** (`css-var-fonts`): changed from `var(--display-font)` to `var(--display-font, "Montserrat")`. Render-neutral because `--display-font` is defined as `"Montserrat"` in `:root` — computed value and embedded faces are identical, existing baseline stays valid. Good coverage addition.

### CI note

Only the WIP check has run so far — full CI hasn't triggered yet. The distributed regression lane (`css-var-fonts` baseline) is the key check to watch.

No blocking concerns. Clean, correct fix at the right level of abstraction.

— Miga

miguel-heygen (approved):

akzarma:
hey @miguel-heygen can you check, it looks like the checks are stuck on this PR? I'm unable to merge.

PR #2720, opened 2026-07-21
perf(studio-server): coordinate cancelable thumbnail generation

miguel-heygen:
## What

Family G — Adaptive thumbnails.

Coordinate cancelable thumbnail generation in Studio Server and the browser bridge.

## Why

Client-side virtualization is insufficient if duplicate or obsolete thumbnail requests continue consuming browser and server resources.

## How

Deduplicate generation by request identity, own cancellation and cleanup in one server coordinator, make output writes atomic, expose the shared device-scale contract, and propagate aborts through the CLI/Vite bridge.

## Test plan

- [x] Unit tests added/updated
- [ ] Manual testing performed
- [ ] Documentation updated (not applicable)

Validated with 78 changed-surface Studio tests, 19 focused server tests, 3,026 full Studio tests, 406 full Studio Server tests, both typechecks, format, lint, file-size gate, and full workspace build.

PR #3075, opened 2026-08-06
chore: release v0.7.96

vanceingalls:
## What

Stable release **v0.7.96**, covering everything since **v0.7.94** — the last version actually on npm.

Headline is canary attribution: rollouts now record *why* each install landed in or out of a cohort, on both the CLI and Studio (#3074). Before this a forced `HF_CANARY_*` override and an ordinary cohort roll produced an identical assignment, so cohort changes could not be explained. Also ships identity-persistence classification on every telemetry event (#3065), a Studio fix for rooted timeline media in preview (#3061), and cancelable thumbnail generation (#2720).

## Why this is 0.7.96 and not 0.7.95

**v0.7.95 was never published.** Its release PR (#3068) merged correctly at 17:12 UTC and the publish workflow fired — but that run has been stuck `in_progress` for hours and produced no tag and no npm artifact. npm `latest` is still `0.7.94`, while main's `package.json` reads `0.7.95`.

The cause is the ongoing GitHub incident: **Actions is in `major_outage`** (started 15:22 UTC, still unresolved). Webhook triggers are throttled, so pushes and PRs are not scheduling runs, and runners are being assigned jobs that no longer exist.

Because `v0.7.95` has no tag, the generated notes correctly span `v0.7.94...v0.7.96`, so nothing from 0.7.95 is lost — its changelog entry is also already on main from #3068.

## ⚠️ Merge ordering matters

**Do not merge this until v0.7.95 is settled.** The publish workflow sets the npm `latest` dist-tag. If 0.7.96 publishes and the stuck 0.7.95 run later resumes, `latest` would move **backwards** to 0.7.95.

Suggested order once Actions recovers:
1. Let the 0.7.95 run finish, or re-run its merged-PR event (the documented recovery path), or cancel it if it is unrecoverable.
2. Confirm CI is green on `main` — **#3074 merged with no matrix coverage at all** (only `WIP` plus a skipped Mintlify), because of the same outage.
3. Then merge this.

## Verification

Local suites on the merged content: core 1674, cli 2475, studio 3480, studio-server 416. Lint clean. The canary-reason change was mutation-tested at all three layers — core helper, CLI binding, Studio binding.

This PR itself will show no checks until Actions recovers.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "hyperframes",
   "description": "HyperFrames by HeyGen. Write HTML, render video. Compositions, GSAP and runtime adapter animations, captions, voiceovers, audio-reactive visuals, and website capture for HyperFrames.",
-  "version": "0.7.95",
+  "version": "0.7.96",
   "author": {
     "name": "HeyGen",
     "email": "hyperframes@heygen.com",
--- a/.codex-plugin/plugin.json
+++ b/.codex-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "hyperframes",
   "description": "Write HTML, render video. Compositions, Tailwind v4 styles, GSAP and runtime adapter animations, captions, voiceovers, audio-reactive visuals, and website capture for HyperFrames.",
-  "version": "0.7.95",
+  "version": "0.7.96",
   "author": {
     "name": "HeyGen",
     "email": "hyperframes@heygen.com",
--- a/.cursor-plugin/plugin.json
+++ b/.cursor-plugin/plugin.json
@@ -3,7 +3,7 @@
   "name": "hyperframes",
   "displayName": "HyperFrames by HeyGen",
   "description": "Write HTML, render video. Compositions, Tailwind v4 styles, GSAP and runtime adapter animations, captions, voiceovers, audio-reactive visuals, and website capture for HyperFrames.",
-  "version": "0.7.95",
+  "version": "0.7.96",
   "author": {
     "name": "HeyGen",
     "email": "hyperframes@heygen.com"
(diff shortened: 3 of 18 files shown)


vanceingalls:
Ready for review — all checks terminal: **55 success, 0 failing**, 5 skipped (path-filtered). `mergeStateStatus` is `BLOCKED` only on the `pull_request` ruleset's approval requirement.

Context worth having before approving, since two of the last three releases went wrong:

**This ships 0.7.94 → 0.7.96. v0.7.95 is permanently skipped.** Its release PR (#3068) merged correctly at 17:12 UTC and fired the publish workflow, but that run was caught by the GitHub Actions incident (`major_outage`, 15:22–00:0x UTC). It is now a phantom: **zero jobs**, cancel and force-cancel both return HTTP 500, re-run refuses with "already running". It cannot execute anything, so it cannot publish — which also means it cannot move the npm `latest` tag backwards. `publish.yml` has no `concurrency` group, so it will not block this release either.

Nothing is lost: since `v0.7.95` was never tagged, the generated notes correctly span `v0.7.94...v0.7.96`, and 0.7.95's changelog entry is already on main from #3068.

**The empty commit `ae72e8c09` is deliberate.** The original `pull_request` webhook for this branch was dropped during the outage, and dropped webhooks are not replayed. `ci.yml` has no `workflow_dispatch`, so an empty commit was the only way to fire the event without closing the PR. The release commit `026e6941a` is unmodified beneath it, and no tag has been pushed — the workflow creates `v0.7.96` from the merge SHA itself.

**This is also the first CI the canary work has seen.** #3074 merged during the outage with only `WIP` and a skipped Mintlify — no matrix at all. Those 55 green checks are the first real verification of the canary-reason emission on all three layers (core helper, CLI binding, Studio binding).

Merging publishes to npm under `latest`.

PR #1123, opened 2026-05-29
fix(producer): cache Google Fonts woff2 per subset, preserve unicode-range

miguel-heygen:
## Symptom

Big Shoulders Display headlines render with the wrong **`A`** glyph — every other letter is the correct condensed display font, but `A` falls back to a plain sans (visible in produced videos: `THE C𝖠T𝖠LOG` etc.). Other Google-fetched fonts with a non-latin-first subset order are affected the same way.

## Root cause

Google Fonts' `css2` API returns **one `@font-face` per (weight × unicode-range subset)** — e.g. for one weight you get separate `vietnamese`, `latin-ext`, and `latin` faces, each a distinct woff2 whose glyph coverage matches its `unicode-range`.

`fetchGoogleFont` in `deterministicFonts.ts` cached woff2 files by `` `${weight}-${style}` `` only — **ignoring the subset**. So every subset of a weight collided on a single filename:

1. first face (`vietnamese`, for display families) → cache miss → downloaded → written to `900-normal.woff2`
2. `latin-ext` face → path exists → **skipped**, reads back the vietnamese file
3. `latin` face (the one with `A`–`Z`!) → path exists → **skipped**, reads back the vietnamese file

So the embedded font was the **vietnamese subset** — which happens to contain a base `A` but none of `B`–`Z`. Inspection of the cached file confirmed it: 114 glyphs, only `A` among ASCII letters, name table `Big Shoulders Display Thin`, and all 9 weight files byte-identical. The injected `@font-face` also dropped `unicode-range`, so it advertised full coverage it didn't have and the one Latin glyph it *did* carry (`A`) collided with the real font.

## Fix

- **Cache per subset:** key the woff2 cache by a hash of the subset-unique woff2 URL, so `latin` / `latin-ext` / `vietnamese` are cached separately.
- **Preserve `unicode-range`:** parse it from each Google face and carry it into the injected `@font-face`, so the browser selects the correct subset per codepoint (matching Google's own CSS semantics).
- **Bundled-font supplement:** add *every* subset of an uncovered weight instead of de-duping by weight (which dropped the extra subsets).
- Extracted the per-subset download/cache into a helper to keep `fetchGoogleFont` within complexity limits.

## Verification

- New hermetic regression test (`deterministicFonts-googleSubsets.test.ts`, injected `fetch` + temp cache dir): two subsets for one weight, vietnamese first — asserts both subsets' distinct bytes are embedded and each face keeps its `unicode-range`. **Fails on old code, passes on the fix.**
[...]

PR #3086, opened 2026-08-07
fix(fonts): declare the bundled subset's unicode-range

akzarma:
## What breaks

A pre-bundled family cannot render the scripts its embedded subset omits. The requested typeface is lost for those codepoints and the text falls through to whatever fonts the render host happens to have installed — the outcome a deterministic font injector exists to prevent.

Decoding the committed `fontData.generated.ts` and reading each font's `cmap`:

| embedded face | codepoints | notes |
|---|---|---|
| `@fontsource/inter:400:normal` | 230 | no `ł`, `č`, `ğ`, `ș`, `ạ`; no Cyrillic; no Greek |
| `@fontsource/roboto:400:normal` | 229 | same |
| `@fontsource/noto-sans-jp:400:normal` | **218** | **zero kana, zero kanji** |

`skills/hyperframes-creative/references/typography.md` lists `Noto Sans JP | 400 · 700 | CJK (Japanese)`, and the real `noto-sans-jp-japanese-400-normal.woff2` (~1 MB) sits unused in the same package directory.

This is not limited to CJK, and that is the more common case: `FONT_ALIAS_MAP` routes `arial`, `helvetica`, `segoe ui`, `verdana` and friends into Inter/Roboto, so ordinary Central European, Turkish, Vietnamese and Cyrillic text hits it.

## Root cause

`packages/producer/src/services/deterministicFonts.ts`, path 1 of `buildFontFaceCss`:

- `scripts/generate-font-data.ts` resolves `${slug}-latin-${weight}-${style}.woff2` — the latin subset, always.
- The embedded face was emitted with **no** `unicode-range`, so it advertises the full range.
- The supplementary Google fetch then skipped a weight entirely once the bundle covered it: *"A weight covered by the embedded bundle is already full-coverage"*. It is covered for latin only.

## The fix

1. Emit the latin `unicode-range` on embedded faces, so they stop claiming coverage they lack.
2. Skip only the fetched face that duplicates that subset, rather than the whole weight — every other subset is kept.

Merged PR #1123 made the same correction for Google-fetched faces (a face without `unicode-range` claiming full coverage), so this brings the bundled path in line with it.

**On download cost:** `fetchGoogleFont` already downloads the faces before the caller discards covered weights, so this does not add network requests to the existing fetch — it keeps subsets that were being fetched and then thrown away. It does add `@font-face` rules and embedded bytes to the generated HTML for bundled families whose weights Google serves in multiple subsets.

**Known imprecision, called out deliberately:** the range is a constant here because the generator does not record which subset it embedded. Deriving the true range per face at generation time would be more precise and is the natural follow-up; it would mean regenerating the committed font data, which felt out of scope for a fix. The constant matches Google's current `latin` subset range, so a handful of combining marks present in some embedded faces (U+0303, U+0309, U+0323) now resolve through a supplementary face instead of the bundled one.

## Test

`deterministicFonts-bundledSubset.test.ts` — hermetic, injects `fetchImpl`, no network. For `Noto Sans JP` weight 400 (a bundled weight) it asserts that every emitted face declares a `unicode-range`, that the embedded faces declare the latin subset, that the Japanese subset face **is** injected, and that the duplicate latin face is still skipped.

Verified the test bites: against unfixed source **0 pass / 1 fail**; with the fix **1 pass / 0 fail**.

## Test plan

- [x] `bun test packages/producer/src/services/deterministicFonts-bundledSubset.test.ts` → 1 pass, 0 fail
- [x] Same test against unfixed source → 1 fail, confirming it bites
- [x] `bun test` on `deterministicFonts.test.ts`, `deterministicFonts-failClosed.test.ts`, `deterministicFonts-systemCapture.test.ts` → 31 pass, 0 fail
- [x] `oxfmt --check` and `oxlint` on both changed files → clean
- [x] Codepoint counts above measured directly from the committed `fontData.generated.ts` with fontTools
- [ ] No end-to-end distributed render performed — relying on CI

### DIFF
--- a/packages/producer/src/services/deterministicFonts-bundledSubset.test.ts
+++ b/packages/producer/src/services/deterministicFonts-bundledSubset.test.ts
@@ -1,0 +1,100 @@
+/**
+ * Regression test for the coverage a bundled face claims.
+ *
+ * `scripts/generate-font-data.ts` embeds the `-latin-` subset file of every
+ * canonical family, so an embedded face carries only Google's `latin` subset.
+ * The emitted `@font-face` used to omit `unicode-range` — advertising full
+ * coverage — and the supplementary Google fetch then skipped every subset of a
+ * weight the bundle "covered". A bundled family therefore could not render the
+ * scripts its own subset omits: `Noto Sans JP` weight 400 carries 218
+ * codepoints with no kana and no kanji.
+ *
+ * These tests inject `fetchImpl` (no network) and a temp
+ * `HYPERFRAMES_FONT_CACHE_DIR` so they are hermetic.
+ */
+
+import { afterAll, beforeAll, describe, expect, it } from "bun:test";
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+let cacheDir: string;
+let prevCacheEnv: string | undefined;
+
+beforeAll(() => {
+  prevCacheEnv = process.env.HYPERFRAMES_FONT_CACHE_DIR;
+  cacheDir = mkdtempSync(join(tmpdir(), "hf-font-subset-"));
+  process.env.HYPERFRAMES_FONT_CACHE_DIR = cacheDir;
+});
+
+afterAll(() => {
+  if (prevCacheEnv === undefined) delete process.env.HYPERFRAMES_FONT_CACHE_DIR;
+  else process.env.HYPERFRAMES_FONT_CACHE_DIR = prevCacheEnv;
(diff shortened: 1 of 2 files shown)


jrusso1020:
Continuing this contribution in #3652, rebuilt on current main with regression coverage and attribution to this PR. Keeping this PR open until the successor lands. Thanks for the original fix.

jrusso1020:
Closing as implemented in #3652, now merged. The successor preserves this contribution while updating it for current main and adding regression coverage. Thanks for the original fix.

PR #3096, opened 2026-08-07
fix(engine): stop SwiftShader ghosting in software screenshot captures

### LINKED ISSUE #3049: Horizontal-band ghosting in transparent renders: content duplicated at exactly renderHeight/4 offsets; persists from 0.6.101/Chrome 131 through 0.7.82/Chrome 152

miguel-heygen:
## What

Apply the SwiftShader software-compositing workaround to **every** software capture, not just BeginFrame ones. One condition moves; no other behaviour changes.

Fixes #3049.

## Why

The reporter's composition duplicates content at a fixed vertical offset in transparent renders, on 68% of frames. #3062 guessed at this same flag and was closed because "the screenshot path does not retain stale layers". That premise was wrong, and I can now show why.

I reproduced the defect **outside hyperframes entirely** — 40 lines of puppeteer, chrome-headless-shell 152, the engine's own Chrome args, seek + `Page.captureScreenshot` in a loop. What the frame series shows is not layer retention, it is **accumulation**: a 1px vertical rule in a translating SVG group is 1px wide on frame 0, 4px on frame 8, 25px on frame 79, growing at exactly the group's own 0.29px/frame velocity. Raster from earlier seeks is never cleared, so each capture contains the union of everything painted before it. Static siblings then read as duplicated one band lower.

That also explains the report's other observations: content-shape dependence (`line` → `path` changes the invalidation rect), survival across Chrome 131→152, and why compositor flags did nothing.

In the real pipeline the accumulation resets per worker page, which is why the dirty frames come in ~108-frame runs separated by ~12 clean ones: each of the 8 workers starts clean and degrades through its own ~120-frame chunk.

## How

`--disable-gpu-compositing` was already applied for software BeginFrame capture. The defect is in the compositor, not in either capture implementation, so the gate is `browserGpuMode === "software"` alone. Alpha formats force `forceScreenshot`, which is why exactly the transparent renders were left unprotected.

Everything else I tried does **not** fix it (all measured, 25-40 frame series each):

| tried | result |
|---|---|
| `fromSurface: false` | still ghosts |
| `captureBeyondViewport: true` | still ghosts |
| capture twice per frame | still ghosts |
| two extra rAF ticks before capture | still ghosts |
| 250ms settle before capture | still ghosts |
| `--disable-partial-raster` | still ghosts |
| `--disable-checker-imaging` | still ghosts |
| `--disable-zero-copy` | still ghosts |
| forced tile sizes / `--max-untiled-layer-*` | still ghosts |
| `--disable-gpu-rasterization` | clean — output byte-identical to `--disable-gpu-compositing` |
| `--disable-gpu-compositing` | clean |

The two working flags produce the same bytes, so there is no "keep GPU raster" option to prefer.

### On the cost that closed #3062

Real, and smaller than it looked. Measured on the repro's frame 0 (clean under both settings, so this isolates rasterization): 96 opaque pixels differ, by 1/255. The rest of the delta is antialiased edge alpha on thin strokes and glyph edges. Weighed against duplicated content on 68% of frames, and against the fact that software BeginFrame renders — the Linux/Docker default — have already been on this exact raster path since #2550, so this makes software renders consistent rather than introducing a new path.

Not covered here: `PRODUCER_EXTRA_CHROME_ARGS`, which is agreed but lands separately as promised in the issue thread.

## Test plan

- [x] Unit tests added/updated — `buildChromeArgs` now asserts the flag for software screenshot capture and its absence for both hardware modes

End-to-end on Linux/SwiftShader (Ubuntu 24.04, chrome-headless-shell 152.0.7928.2), the reporter's exact command and composition:

```
hyperframes render <dir> -o out.mov --fps 60 --quality high --format mov --workers 8 --no-browser-gpu
```

Scoring the alpha channel of all 959 frames for the duplication signature:

| | dirty frames |
|---|---|
| before | 68.5% |
| after | 0.0% |

Frame 300, alpha channel, from the render above. Before, the annotated row appears twice, 286px apart, over smeared grid bars; after, it appears once over 1px grid lines.

**Before**

frame 300 alpha, before the fix: content duplicated 286px lower

**After**

frame 300 alpha, after the fix: each element painted once

Engine suite green (`vitest run packages/engine/src/services`, 733 passed).


james-russo-rames-d-jusso (commented):
_Reviewed at `3b1ac6544`._

Clean fix, and one of those PRs where the writeup carries as much weight as the code. `--disable-gpu-compositing` moves out of the `captureMode !== "screenshot"` block and gates only on `browserGpuMode === "software"`, which is the right shape for the defect the PR proves: the accumulation is in SwiftShader's compositor, not in either capture implementation. Alpha renders being forced onto `forceScreenshot` (per `renderOrchestrator`) explains exactly why the transparent reproducer hit this and the opaque BeginFrame path never did.

The evidence discipline here is above the bar HF#3062's rejection tried to hold. The 8-row "still ghosts" table + measured 96-pixel/1-of-255 raster cost on frame 0 does what the previous attempt didn't: forecloses the "have you tried X" review path and quantifies the trade. That matches the pattern the HF team already accepts on the software BeginFrame path (`packages/engine/src/services/browserManager.ts:934` — this compositing route has been Linux-default since #2550), so the change is *tightening a consistency invariant across capture modes*, not introducing a new raster path.

**What lands cleanly:**

- **The gate now matches the compositor topology, not the capture entry point.** `CaptureMode` is `"beginframe" | "screenshot" | "drawelement"` (`packages/engine/src/services/browserLeasePool.ts:3`); the compositor surface is shared by all three, so `browserGpuMode` alone is the correct axis. `browserGpuMode` is already resolved to `"software" | "hardware"` before `buildChromeArgs` runs (`browserManager.ts:525`), so no `"auto"` branch to worry about.
- **The runtime `chromeArgs.push` order is preserved.** The flag moves ABOVE the BeginFrame block rather than being duplicated inside both branches, so a software screenshot capture ends up with exactly `[…prior…, "--disable-gpu-compositing", …no BeginFrame flags…]` and a software BeginFrame ends up with `[…prior…, "--disable-gpu-compositing", "--deterministic-mode", "--enable-begin-frame-control", …]` — no observable ordering change for either path that was already receiving the flag.
- **The in-code comment (`:911-931`) is the PR body distilled**, including the exact upstream Chromium tracker (`https://issues.chromium.org/issues/535256667`) and the removal condition. Next reader of this hunk will understand why this is a workaround and when to drop it. That's the shape of comment I want to see on load-bearing workarounds.
- **`(hardware, screenshot)` gets an explicit negative assertion in the test** — pre-PR the test only pinned `(hardware, beginframe)`, and the natural anti-regression is a symmetric denial that the flag doesn't leak to the hardware screenshot path.

**One optional tightening (nit, not a blocker):** `CaptureMode` has a third value `"drawelement"`. The test title *"for every software capture, whatever the capture mode"* is stronger than what the four assertions actually cover — `(drawelement, software)` and `(drawelement, hardware)` aren't pinned. Pre-PR they were already correct (drawelement was on the `!== "screenshot"` branch, so `(drawelement, software)` already received the flag), and they remain correct post-PR by the same logic — but that's the sort of "already right for reasons unrelated to this fix" that a future refactor could accidentally unstick. Two extra `expect` lines would close the sweep, and the "T2b transparent + SwiftShader" note in `drawElementService.integration.test.ts:17` shows drawelement can escalate to screenshot mode, so the compositor path *is* live-relevant on this axis.

_Curiosity, not a review point:_ the `PRODUCER_EXTRA_CHROME_ARGS` follow-up you flag as "lands separately" — is there a risk the escape hatch lets someone override `--disable-gpu-compositing` back out (e.g., `--enable-gpu-compositing` or a competing raster flag) and re-introduce the ghost? Not to solve here, just for the follow-up unit — a small "reserved keys" filter or a warning if the extras contain `-compositing` would keep this workaround in force even when someone is experimenting.

LGTM from my side.

vanceingalls (approved):
**APPROVE** @ `3b1ac65445bd6e67ede7e3e3ec60ad2220c6a10c` — R1

Moves `--disable-gpu-compositing` out of the BeginFrame-only branch in `packages/engine/src/services/browserManager.ts` and applies it whenever `browserGpuMode === "software"`, so software `Page.captureScreenshot` captures (which every alpha render is forced onto via `forceScreenshot`) now route through the same Chrome-software-compositor path that software BeginFrame captures have been on since #2550. That takes SwiftShader's GPU compositor — where the stale-raster accumulation lives — off the paint path feeding the readback. Fixes #3049 (68.5% → 0.0% dirty frames on the reporter's composition, per author's end-to-end measurement).

### Findings

1. **[NOTE] Mechanism.** The condition moves from `if (options.captureMode !== "screenshot") { if (browserGpuMode === "software") { chromeArgs.push("--disable-gpu-compositing"); } ... }` to a top-level `if (browserGpuMode === "software") { chromeArgs.push("--disable-gpu-compositing"); }` at `packages/engine/src/services/browserManager.ts:934-936`. The BeginFrame block below is otherwise unchanged. The four-corner selector grid (`browserGpuMode` × `captureMode`) is now: software+beginframe → flag (unchanged), software+screenshot → **flag (new)**, hardware+beginframe → no flag (unchanged), hardware+screenshot → no flag (unchanged). All four corners are asserted in `packages/engine/src/services/browserManager.test.ts:132-160`.

2. **[NOTE] BeginFrame-probe-failure fallback is safe.** If `HeadlessExperimental.beginFrame` probing fails at launch, the browser is re-launched with `stripBeginFrameFlags(...)` (`browserManager.ts:730`). `BEGINFRAME_ONLY_FLAGS` (line 221) does not contain `--disable-gpu-compositing`, so the fallback screenshot re-launch still carries the workaround — the runtime-fallback screenshot path is protected, not just the statically-requested one.

3. **[NOTE] Hardware paths untouched.** `browserGpuMode` is resolved via `resolveBrowserGpuMode` (line 549) and the new push is gated on that resolved value; the `getBrowserGpuArgs` GL-flag branch (line 871) is separate. Hardware users see the same Chrome-args set as before.

4. **[NOTE] Cost trade-off is bounded.** Software BeginFrame captures have been on this exact raster path since #2550, so the change collapses software captures onto ONE consistent raster path rather than introducing a new one. Author's frame-0 measurement (both settings otherwise clean): 96 opaque pixels differ by 1/255; the rest is antialiased edge alpha on thin strokes and glyph edges. Every non-compositing mitigation the author enumerated (`fromSurface: false`, `captureBeyondViewport`, double-capture, extra rAF ticks, 250ms settle, `--disable-partial-raster`, `--disable-checker-imaging`, `--disable-zero-copy`, forced tile sizes, `--max-untiled-layer-*`) left the accumulation untouched; only `--disable-gpu-compositing` and `--disable-gpu-rasterization` fixed it, and they produced byte-identical output — so there is no "keep GPU raster" option to prefer.

5. **[NIT] `Distributed: png-sequence` baseline needs regenerating — required `regression` check is red for this mechanical reason.** Shard-4 reports 60/60 failed frames at psnr=0 with `maxFrameFailures: 0` on the `png-sequence` alpha fixture. `packages/producer/tests/distributed/png-sequence/meta.json` documents this exact failure mode in its `description`: "the upstream byte sources are Chrome's CDP screenshot output and libpng's deflate, so a Chromium or zlib bump in Dockerfile.test will produce identical pixels but different bytes — the failure mode on a Chrome version bump is 'regenerate baselines via docker:test:update', not 'investigate regression'." This PR is a Chrome-flag change on a byte-strict alpha fixture — the same class of drift the meta contemplates. Please run `docker:test:update` on the affected fixture (and any sibling alpha fixtures on the same gate), push the refreshed baselines, then merge. Approving the code because the mechanism is correct and well-tested; the required-check red is a mechanical baseline refresh, not a defect the diff introduces.

### CI (ruleset 14211637)

- Semantic PR title — pass
- Test: runtime contract — pass
- Typecheck — pass
- Build — pass
- Test — pass
- Render on windows-latest — pass
- Tests on windows-latest — in progress
- **regression — fail** (shard-4 `png-sequence` alpha baseline; see finding 5. Shards 1-3, 5-8 cancelled by fail-fast; shard-9 pass. The red is isolated to byte-strict alpha fixture drift.)

— Via

miguel-heygen:
Both review points addressed at `4cd69ad41`, plus the red `regression` check.

**drawelement sweep (@james-russo-rames-d-jusso).** Fair — the title claimed more than the assertions covered. Replaced the four ad-hoc expects with an `it.each` over the whole `CaptureMode` union, so `(drawelement, software)` and `(drawelement, hardware)` are now pinned alongside the other four corners. If a fourth mode is ever added, the union type forces it into the sweep rather than letting it silently opt out the way `screenshot` did.

**png-sequence baseline (@vanceingalls).** Regenerated in the Docker test image built from this branch (`png-sequence --update`), so it carries the flag. Two things I checked before accepting the new bytes, since "regenerate the baseline" is exactly how a real regression gets laundered into a golden file:

- **It is not a content change.** Old vs new baseline across all 60 frames: min PSNR 41.3 dB, median 44.4 dB, at most 7,367 of 230,400 pixels touched in any frame. Differences are confined to antialiased glyph and stroke edges — the same delta class quantified in the PR body. Frame 32 (the worst) is visually identical side by side.
- **`output/compiled.html` was deliberately not included.** Regenerating also rewrote it, but the only differences are the `@fontsource` woff2 base64 payloads — package byte drift that has nothing to do with a Chrome flag, and which `validateCompilation` does not compare anyway (it checks timed elements and composition structures, not bytes). Committing it would have buried an unrelated environment diff in a fix PR. Reverted; only the 60 frames moved.

Worth noting the psnr=0 in the shard log is a sentinel, not a measurement — `regression-harness.ts:1283` reports Infinity for byte-identical png-sequence frames and 0 otherwise. So it read as catastrophic when it was one LSB of edge alpha. Not asking for a change, just flagging it for whoever next reads that log at 3am.

**On the escape hatch.** Checked: `PRODUCER_EXTRA_CHROME_ARGS` does not exist anywhere in the tree yet, so there is nothing to guard from this PR — a reserved-keys filter here would be guarding a flag nobody can pass. The concern is real though, and it is a requirement on that PR rather than a nice-to-have: whatever lands has to refuse `--enable-gpu-compositing` (and any `-compositing` / `gpu-rasterization` variant) in software mode, or it hands users a one-env-var path back into #3049. I will carry that into the follow-up unit rather than leave it as a review thread.

PR #3101, opened 2026-08-08
test(repro): harness for PRINFRA-372 detached-Frame / Target-closed crash

vanceingalls:
Repro harness for [PRINFRA-372](https://linear.app/heygen/issue/PRINFRA-372). No product code changes — one script under `scripts/repro/`.

## Why

The crash did not reproduce on a 14-core / 48 GB darwin/arm64 workstation across two attempts. The reporter's box was **10-core / 16 GB**. 30 days of `render_error` telemetry says the signature is:

- enriched **2.6×** on darwin/arm64 (66.5% / 62.0% vs 25.8% / 24.7% baseline)
- concentrated on **`os_release 25.5.0`** (369 of ~700 darwin cases)
- spread across CLI **0.7.10 → 0.7.94** — not a release regression
- **depleted** on multi-worker (23.0% vs 42.8% baseline)

That last one contradicts the ticket title. Worker count reads as a *correlate* of memory pressure, not the cause — so the missing variable is memory headroom, which a dev workstation can't exercise. This harness makes that condition reproducible so the hypothesis is falsifiable either way.

## What it does

Generates the reporter's composition shape — long 1080p, 457 word-pop caption tweens, 14 media segments, optionally 2 videos — runs N iterations, and greps for the four Puppeteer-family signatures.

```bash
node scripts/repro/prinfra-372-detached-frame.mjs                  # defaults: 176s, auto workers, 3 runs
node scripts/repro/prinfra-372-detached-frame.mjs --docker --memory 16g   # emulate the reporter's box
node scripts/repro/prinfra-372-detached-frame.mjs --workers 4 --clip path/to.mp4
```

**Exits 1 when the crash reproduces**, so it can gate a fix.

## Two things it does deliberately

**Warns when the host has too much RAM** to make a negative result meaningful. A clean run on a 48 GB machine is not evidence of absence, and the script says so rather than letting the next person conclude "cannot reproduce".

**Only emits video segments when `--clip` is supplied.** An unresolvable `src` would fail the render for the wrong reason and read as a repro.

## Verification

Smoke-tested end-to-end at reduced scale (600 frames) — clean run, correct reporting. Signature matchers verified against real Puppeteer strings:

| input | detected |
|---|---|
| `Attempted to use detached Frame '2A9F...'` | detached Frame |
| `ProtocolError: Protocol error (Runtime.callFunctionOn): Target closed` | Target closed, Protocol error |
| `Session closed. Most likely the page has been closed.` | Session closed |
| clean render output | *(no match)* |

Lint and format clean.

## What this does not do

It does not fix anything. I could not reproduce the failure, and the telemetry says the mechanism in the ticket's description is not the one to patch — so a patch would be speculative. Someone needs to run this on a ≤16 GB darwin/arm64 host, or via `--docker --memory 16g`.

### DIFF
--- a/scripts/repro/prinfra-372-detached-frame.mjs
+++ b/scripts/repro/prinfra-372-detached-frame.mjs
@@ -1,0 +1,177 @@
+#!/usr/bin/env node
+/**
+ * Repro harness for PRINFRA-372 — `Attempted to use detached Frame` /
+ * `Target closed` during parallel capture.
+ *
+ * WHY THIS EXISTS
+ *
+ * The crash did not reproduce on a 14-core / 48 GB darwin/arm64 workstation
+ * across two attempts. The reporter's machine was 10-core / 16 GB. Fleet
+ * telemetry (30d) says the signature is:
+ *
+ *   - enriched 2.6x on darwin/arm64 (66.5% / 62.0% vs 25.8% / 24.7% baseline)
+ *   - concentrated on os_release 25.5.0 (369 of ~700 darwin cases)
+ *   - spread across CLI 0.7.10 -> 0.7.94, so not a release regression
+ *   - *depleted* on multi-worker (23.0% vs 42.8% baseline) — i.e. worker count
+ *     looks like a correlate of memory pressure, not the cause
+ *
+ * So the missing variable is almost certainly memory headroom, which a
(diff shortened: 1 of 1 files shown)


vanceingalls:
Closing — the harness should not live as a repo PR. Attaching it to PRINFRA-372 directly instead.

PR #3103, opened 2026-08-08
feat(cli): measure a composition against a reference with compare --against

miguel-heygen:
## What

Adds `hyperframes compare --against <reference>`: measure one composition against an external reference video or still, at up to 8 sampled times.

Each run emits three things:

- a reference-over-replica contact sheet (`--out`),
- a red/cyan deviation overlay per sampled time (`<out>-overlay-NN.png`): agreement grey, reference-only ink red, replica-only ink cyan,
- numbers per time: `ssim`, `meanAbsDiff`, `meanSignedDiff`, and ink bounding-box deltas `dw` / `dh` / `dcx` / `dcy` / `scale`.

`--fail-under <ssim>` turns it into a gate (exit 1 when the worst sampled SSIM is below the threshold). Existing multi-variant `compare` behaviour is unchanged.

## Why

Every gate we ship is self-referential. `lint` and `check` audit a composition against its own rules, so a scene that is inverted, mis-scaled, or missing its hero element passes both under a green check. When the brief comes with a reference artifact (a video being rebuilt or recut, an approved cut, a design still, the previous accepted render), nothing in the CLI compares the build to it, and agents end up hand-rolling a one-off ffmpeg measuring rig, or eyeballing composites.

## How

- `packages/cli/src/utils/referenceDiff.ts`: pure frame math (ink bounding box from a median-luma threshold, bounds deviation, mean absolute and mean signed difference, red/cyan overlay buffer, ffmpeg SSIM args + log parsing). No I/O, so it unit-tests without a browser or a built workspace.
- `packages/cli/src/capture/compareAgainstReference.ts`: orchestration. Reference frames via ffmpeg, replica frames from one browser session seeked with `AUDIT_SEEK_OPTIONS` (the producer bridge, the same seek target `render` drives), replicas normalized to reference dimensions so the sheet, the overlay and SSIM all read the same pixels.
- SSIM comes from ffmpeg's own `ssim` filter rather than a reimplementation. No new dependencies: `sharp` and ffmpeg are already required.
- No default threshold, deliberately. The floor depends on content (below), so a guessed threshold either passes everything or fails every video-backed build.
- Skills: `/hyperframes-cli` gains a "ground the gate in a reference" section, `references/compare-and-batch.md` gains the full contract with a read-the-numbers guide, `/remotion-to-hyperframes` points at it for mid-build iteration between full renders, and the `/hyperframes` router now says to read only the routed workflow's `SKILL.md` and none of its siblings.

### The floor, measured rather than assumed

A replica is a live browser paint; a reference is a decoded compressed video. That decode gap is a floor no correction goes below, and it depends on content. Compared against their own renders:

| Content                    | Self-comparison SSIM                        |
| -------------------------- | ------------------------------------------- |
| Flat graphics and type     | 0.998 to 0.999                              |
| Photographic video layers  | ~0.93 at `--quality high`, ~0.89 at `draft` |

`meanSignedDiff` exists to make that legible: on the video-backed case, 1.6 of a 2.1% mean absolute difference is uniform bias, not structural error. The CLI prints it as `diff 2.1% (bias +1.6%)`.

An earlier revision of this branch documented the video-backed floor as seek drift. That was wrong and is corrected here: probing which reference frame the live seek lands on shows it lands exactly on the right one (SSIM peaks at frame 210 for t=7 and falls off on both sides, 0.881 / **0.893** / 0.888). The residual is encode loss plus browser-versus-FFmpeg colour conversion.

## Test plan

Unit tests cover the pure math and the argument parsing (ink bounds, bounds deviation, mean absolute and signed difference, overlay channel mapping, SSIM log parsing, `--at` / `--fail-under` validation and caps).

End to end, against real renders:

- `registry/examples/nyt-graph` (graphics only) rendered at `--quality high`, compared against its own render at `--at 0,3,7,11,14`: SSIM 0.9987 to 0.9999, ink deltas all zero.
- Same project against an unrelated video: SSIM 0.2892, `meanAbsDiff` 62.5%.
- Gate exit codes verified directly: mismatch + `--fail-under 0.9` exits 1, match + `--fail-under 0.9` exits 0.
- `registry/examples/kinetic-type` (video-backed) against its own high-quality render: 0.9353 / 0.9316 / 0.9301 with bias +1.6% / +1.1% / +1.0%, and against the draft render 0.8901 / 0.8930 / 0.8955. Repeat runs are bit-identical, so the number is stable.
- Frame-landing probe (above) run against neighbouring reference frames 200 through 220.
- `packages/cli` suite: 2501 passed, 2 failed. Both failures are the pre-existing `transcribe.test.ts` cases that also fail on `main`.

### DIFF
--- a/docs/developers/cli.mdx
+++ b/docs/developers/cli.mdx
@@ -37,6 +37,7 @@ remains authoritative: run `npx hyperframes <command> --help`.
 | List compositions | `npx hyperframes compositions` |
 | Inspect keyframe behavior | `npx hyperframes keyframes` |
 | Compare two or more versions | `npx hyperframes compare v1/ v2/` |
+| Measure against a reference video | `npx hyperframes compare . --against reference.mp4 --at 0,4,10` |
 
 ## Rendering and automation
 
--- a/docs/packages/cli.mdx
+++ b/docs/packages/cli.mdx
@@ -729,6 +729,29 @@ npx hyperframes compare ./variant-a ./variant-b --labels "A,B"
 | `--timeout` | Render-ready timeout per variant (default 5000 ms) |
 | `--json`    | Machine-readable results                           |
 
+With `--against`, `compare` measures one composition against an external
+reference video or still instead of against sibling variants. `lint` and
+`check` only ever audit a composition against its own rules, so this is the
+gate that catches a scene which passes everything and still looks nothing like
+the thing it reproduces.
+
+```bash
+npx hyperframes compare . --against reference.mp4 --at 0,4,10,21 --fail-under 0.95
+```
+
+| Flag           | Description                                                                    |
+| -------------- | ------------------------------------------------------------------------------ |
+| `--against`    | Reference video or image to measure against (requires FFmpeg)                  |
+| `--at`         | Up to 8 comma-separated sample times, sampled in both reference and replica    |
+| `--fail-under` | Exit non-zero when the worst sampled SSIM falls below this threshold           |
+| `--out`        | Contact sheet path; overlays are written next to it as `<out>-overlay-NN.png`  |
+
+Each run writes a reference-over-replica contact sheet, a red/cyan deviation
+overlay per sampled time (agreement grey, reference-only ink red, replica-only
+ink cyan), and per-time numbers: `ssim`, `meanAbsDiff`, and ink bounding-box
+deltas `dw` / `dh` / `dcx` / `dcy` / `scale`. There is no default threshold:
+measure a known-good state first, then gate slightly below it.
+
 `grade-compare` does the same for colour: candidate grades or LUTs applied to
 one reference frame.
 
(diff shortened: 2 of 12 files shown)


miga-heygen (approved):
Reviewed at exact head. Adds `compare --against <reference>` to measure a composition against an external video or still. Produces three instruments per sampled time: SSIM via ffmpeg's ssim filter, ink bounding-box deltas (dw/dh/dcx/dcy/scale), and a red/cyan deviation overlay. --fail-under gates on worst SSIM. Well-structured: referenceDiff.ts for the pure math (inkBounds, meanAbsDiff, meanSignedDiff, redCyanOverlay, parseSsimAll), compareAgainstReference.ts for the orchestration, with thorough unit tests on both. Docs and skill references updated with the floor-depends-on-content guidance. Clean and well-tested addition.

— Miga

miguel-heygen:
Closing in favor of the current reference workflow: snapshot --against already creates render/reference pairs, and the Remotion render_diff script supplies an all-frame SSIM gate. The remaining convenience metrics do not justify reviving a separate capture path. The proposed threshold gate also needs redesign because unmeasured samples are silently excluded.

Audit baseline: `723cd0d7853e2ef5cc4f1b2fba12521ed5d9d5bc`. Closing at the owner’s request as part of the pre-September backlog audit; branch history is preserved.

PR #3105, opened 2026-08-08
fix(core): let a composition declare its caption colour states

vanceingalls:
## Problem

`applyCaptionOverrides` classifies a word's colour tweens as "dim" or "active" by **colour equality**: it takes the first colour tween's value as the dim baseline and treats every tween matching it as dim, everything else as active.

That inference fails outright when a composition's two states share a colour. Every tween then matches the baseline, all of them are classified dim, and the caller's `activeColor` is **silently discarded** — no error, no warning, the override simply does nothing.

## Fix

A tween may now declare which state it belongs to:

```js
gsap.to(word, { color: "#888", data: { captionState: "active" } });
```

GSAP passes unknown vars through untouched, so declaring costs a composition nothing at runtime.

Resolution is **per tween**: a declared tween uses its declaration, an undeclared one falls back to the existing colour-equality classification. Nothing changes for compositions that declare nothing.

The distinction is that the composition tells us what it built, rather than us inferring it from what it happens to look like.

## Testing

Three cases added to `packages/core/src/runtime/captionOverrides.test.ts`:

- identical dim/active colours, both tweens declared — each gets the right override (this is the bug; it fails without the change)
- nothing declared — falls back to colour classification, unchanged behaviour
- declaration deliberately contradicting what colour-equality would infer — the declaration wins

Verified against `main`'s applier: **2 failed, 4 passed**. With the fix: **6 passed**.

### DIFF
--- a/packages/core/src/runtime/captionOverrides.test.ts
+++ b/packages/core/src/runtime/captionOverrides.test.ts
@@ -34,6 +34,19 @@ function installGsapMock() {
   return { setCalls };
 }
 
+/** A gsap mock whose `getTweensOf` returns the supplied colour tweens, so classification is testable. */
+function installGsapMockWithTweens(tweens: Array<Record<string, unknown>>) {
+  const gsap = {
+    set() {},
+    killTweensOf() {},
+    getTweensOf() {
+      return tweens.map((vars, i) => ({ vars, startTime: () => i }));
+    },
+  };
+  Object.defineProperty(window, "gsap", { configurable: true, value: gsap });
+  return tweens;
+}
+
 async function flushCaptionOverrides() {
   for (let i = 0; i < 4; i++) {
     await Promise.resolve();
@@ -120,5 +133,54 @@ describe("applyCaptionOverrides", () => {
     expect(setCalls).toHaveLength(1);
     expect(setCalls[0]?.target).toBe(wrapper);
     expect(setCalls[0]?.vars).toEqual({ x: 16 });
+  });
+});
+
+describe("caption state declaration", () => {
+  it("honours a declared state when dim and active colours are IDENTICAL", async () => {
+    // The failure the declaration exists for. Classification by colour equality takes the first
+    // tween's colour as the dim baseline, so a composition whose two states share a colour has
+    // every tween classified dim — and `activeColor` is silently dropped.
+    const tweens = installGsapMockWithTweens([
+      { color: "#888", data: { captionState: "dim" } },
+      { color: "#888", data: { captionState: "active" } },
+    ]);
+    installCaptionOverrideFetch([{ wordIndex: 0, dimColor: "#111", activeColor: "#eee" }]);
+    document.body.innerHTML = `<div class="caption-group"><span id="w0">Hi</span></div>`;
+
+    applyCaptionOverrides();
+    await flushCaptionOverrides();
+
+    expect(tweens[0].color).toBe("#111");
+    expect(tweens[1].color).toBe("#eee");
+  });
+
+  it("falls back to colour classification when nothing is declared", async () => {
+    // Undeclared compositions must keep working exactly as before — the declaration is additive.
+    const tweens = installGsapMockWithTweens([{ color: "#222" }, { color: "#fff" }]);
+    installCaptionOverrideFetch([{ wordIndex: 0, dimColor: "#111", activeColor: "#eee" }]);
+    document.body.innerHTML = `<div class="caption-group"><span id="w0">Hi</span></div>`;
+
+    applyCaptionOverrides();
+    await flushCaptionOverrides();
+
+    expect(tweens[0].color).toBe("#111");
+    expect(tweens[1].color).toBe("#eee");
+  });
+
+  it("prefers the declaration over the colour heuristic when they disagree", async () => {
+    // Declared order is deliberately the reverse of what colour-equality would infer.
+    const tweens = installGsapMockWithTweens([
+      { color: "#222", data: { captionState: "active" } },
+      { color: "#fff", data: { captionState: "dim" } },
+    ]);
+    installCaptionOverrideFetch([{ wordIndex: 0, dimColor: "#111", activeColor: "#eee" }]);
+    document.body.innerHTML = `<div class="caption-group"><span id="w0">Hi</span></div>`;
+
+    applyCaptionOverrides();
+    await flushCaptionOverrides();
+
+    expect(tweens[0].color).toBe("#eee");
+    expect(tweens[1].color).toBe("#111");
   });
 });
--- a/packages/core/src/runtime/captionOverrides.test.ts
+++ b/packages/core/src/runtime/captionOverrides.test.ts
@@ -183,4 +183,55 @@ describe("caption state declaration", () => {
     expect(tweens[0].color).toBe("#eee");
     expect(tweens[1].color).toBe("#111");
   });
+
+  it("does not let a declared tween poison the baseline for undeclared ones", async () => {
+    // The baseline exists to guess about tweens the declaration has NOT spoken to. Deriving it from
+    // a tween declared "active" makes every undeclared same-state tween compare against a colour
+    // that has explicitly said it is not the dim reference — so the one genuinely dim tween here
+    // gets classified active and receives the wrong override.
+    const tweens = installGsapMockWithTweens([
+      { color: "#eee", data: { captionState: "active" } },
+      { color: "#111" },
+    ]);
+    installCaptionOverrideFetch([{ wordIndex: 0, dimColor: "#dim", activeColor: "#active" }]);
+    document.body.innerHTML = `<div class="caption-group"><span id="w0">Hi</span></div>`;
+
+    applyCaptionOverrides();
+    await flushCaptionOverrides();
+
+    expect(tweens[0].color).toBe("#active");
+    expect(tweens[1].color).toBe("#dim");
+  });
+
+  it("prefers a declared dim tween as the baseline over an undeclared one", async () => {
+    const tweens = installGsapMockWithTweens([
+      { color: "#aaa" },
+      { color: "#bbb", data: { captionState: "dim" } },
+    ]);
+    installCaptionOverrideFetch([{ wordIndex: 0, dimColor: "#dim", activeColor: "#active" }]);
+    document.body.innerHTML = `<div class="caption-group"><span id="w0">Hi</span></div>`;
+
+    applyCaptionOverrides();
+    await flushCaptionOverrides();
+
+    // #bbb is the declared dim reference, so the undeclared #aaa is not dim.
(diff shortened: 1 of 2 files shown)


miga-heygen (commented):
Clean fix — the declaration mechanism is the right approach. Opt-in, backwards compatible, per-tween granular, and the implementation is minimal.

**The bug is real and well-explained.** When both states share a colour, every tween matches the dim baseline, `activeColor` is silently discarded, and the override has no visible effect. The declaration breaks the ambiguity by letting the composition say what it built.

**`declaredCaptionState` is well-guarded.** The `as { captionState?: unknown } | undefined` → string comparison pattern correctly handles `data` being absent, `null`, a non-object, or containing an unexpected value — all fall through to `undefined` and the heuristic takes over.

**Per-tween fallback is the right granularity.** A composition can declare some tweens and leave others to colour classification. This makes adoption incremental.

**Tests cover the three critical scenarios:** the bug case (identical colours), backward compat (no declarations), and precedence (declaration vs heuristic). The mock design — `getTweensOf` returning objects whose `vars` are the test-level references — is clean and lets assertions inspect mutations directly.

**One observation (non-blocking):** `dimBaseline` is still computed from the first colour tween's colour regardless of whether that tween has a declaration. If a composition declares its *first* tween as `"active"` and leaves a later same-coloured tween undeclared, the fallback classifies the undeclared tween as `"dim"` (matches baseline) even though the first tween was declared active — the baseline's semantic meaning shifted under the heuristic. In practice this only matters for mixed declared/undeclared with non-standard ordering, and the fix is just "declare all your tweens" — but worth being aware of if this convention spreads.

LGTM — ship it.

_Review by Miga_

james-russo-rames-d-jusso (commented):
_Reviewed at `1e7799bfb`._

Small, well-scoped fix. Names the exact class of composition the pre-PR classifier silently drops overrides on (two states sharing a colour → every tween matches the dim baseline → `activeColor` becomes a no-op), and picks the right shape for the repair: a composition declares what it built, per tween, with a colour-equality fallback for anything undeclared. Docstring on `declaredCaptionState` earns its keep — states the bug the declaration exists to close and the "cost nothing at runtime" invariant that makes it safe to land ahead of any writer. One concern about baseline pollution in mixed-declaration compositions, plus a few smaller notes; nothing blocking.

## Concern

**`dimBaseline` is computed from `colorTweens[0]` unconditionally, so a declared "active" tween at position 0 poisons the classifier for its undeclared siblings.** `captionOverrides.ts:155-162` — the pre-PR heuristic ("first tween is dim") was implicit but self-consistent: `dimBaseline = colorTweens[0].vars.color` and same-baseline-tweens got the dim override, everything else got the active override. Post-PR the per-tween declaration correctly overrides that guess for declared tweens, but `dimBaseline` is still derived from `colorTweens[0]` even when THAT tween is declared "active".

Concrete failure — mixed composition where the timeline emits an active tween first:

```
tweens    = [{ color: "#eee", data: { captionState: "active" } }, { color: "#111" }]
override  = { dimColor: "#dim-override", activeColor: "#active-override" }
dimBaseline = "#eee"                            // colorTweens[0].vars.color

tw[0]: declared "active" → color = "#active-override"                       ✓
tw[1]: undeclared → "#111" !== "#eee" → classified "active" → "#active-override"   ✗
                                                                              ^ tw[1] is actually the dim tween;
                                                                                should get "#dim-override"
```

The PR body's "an undeclared one falls back to the existing colour-equality classification" is technically true — the classifier ran, colour equality was evaluated — but the "existing" classifier assumed nothing before it had already been resolved to a specific state. Now something can be, and if that something happens to sit at index 0, the fallback baseline is derived from a tween that has explicitly said "I am not the dim baseline". `[[feedback_baseline_first_sequencing_no_canary]]` and `[[feedback_assertion_hardening_vs_fallback_introduction]]` both point at the same shape: when you introduce a stronger source of truth on top of a heuristic, the heuristic's inputs should stop drawing from records the new truth has already spoken to.

The tightening is small — derive the baseline from tweens the classifier still has to guess about, or from a declared "dim" tween if one is present:

```ts
const dimReference =
  colorTweens.find((tw) => declaredCaptionState(tw) === "dim") ??
  colorTweens.find((tw) => declaredCaptionState(tw) === undefined);
const dimBaseline = dimReference ? String(dimReference.vars.color) : "";
```

Additive over the current shape. All three current tests still pass by construction (test 1 has both declared; tests 2 has none declared, `find(undefined)` returns `colorTweens[0]`; test 3 has both declared and finds the declared "dim" one). The bug this PR exists to close for the fully-declared case is closed either way; this just keeps a partial-migration composition from ending up more broken than it was before it started declaring.

Not a blocker on its own — the mixed case only bites during a composition-by-composition rollout, which is presumably brief. But the guard is cheap enough that "we'll fix it in the next round" isn't a great trade against "it lands with the same PR that introduces the declaration".

## Nits

**Stale classification comment at `:141-146`.** *"Instead of relying on timeline position order (fragile if custom color tweens exist), we classify each tween by comparing its target color to the current computed color of the element."* — that stopped being the full story at this PR: the classification is now declaration-first with colour-equality as fallback, and the declaration exists precisely because "comparing target color to current computed color" is itself fragile when the two states share a colour. Small doc drift worth naming both layers.

**Producer/writer side lands separately.** `grep -rn "captionState" packages/ --include='*.ts*' | grep -v test | grep -v captionOverrides.ts` returns zero — the six unrelated hits I found are all a local variable named `captionState` inside `usePlaybackKeyboard.ts` / `playbackShortcuts.ts` under `packages/studio/src/player/`, which are the studio's playback-shortcut state, not the tween data field. So on Day 1 after this PR merges, no shipped composition benefits from the reader — every composition still hits the colour-equality fallback. Reader-first sequencing is the right ordering (writer would fail closed without it), and the PR body's "GSAP passes unknown vars through untouched" makes clear this is a contract for future writers. Just worth naming where and when the writer PR lands (`packages/studio/src/captions/generator.ts` looks like the natural producer for the caption-generator path) so the fix isn't dormant longer than necessary. Not scope of THIS PR — but the point of naming the follow-up on the reader PR is that it makes the fix reviewable as a two-part change rather than a "we added a codepath and nothing uses it yet" report a year from now.

**Test coverage gaps worth adding to lock in the contract.**
- Invalid `captionState` value (e.g. `data: { captionState: "typo" }`, `null`, `42`) — the code correctly falls through via the `state === "dim" || state === "active"` narrowing, and the `??` operator, but there's no unit test that pins that fallback path. A single case would prevent a future contributor from tightening the union without noticing.
- Mixed declared + undeclared, so whichever baseline decision you land on stops being an emergent behaviour and becomes a stated invariant. The concern above is exactly the class of case a mixed-declaration test would surface.
- Non-object `data` (e.g. `data: "someString"`) — safe today because `("someString")?.captionState` returns `undefined` in JS, but the `as { captionState?: unknown } | undefined` cast reads as if `data` is expected to be an object. One case would document the actual robustness.

**`tween.vars.data as { captionState?: unknown } | undefined`** — reads slightly loose against a `Record<string, unknown>`. A `typeof data === "object" && data !== null` narrow would be a tighter shape, but the JS-level fall-through on primitives is already safe (`(42)?.captionState === undefined`), so this is stylistic — the current shape doesn't produce a bug.

## What lands cleanly

- **The docstring on `declaredCaptionState`.** Names the bug (colour equality breaks on shared colours), the trade-off (composition declares vs runtime infers), and the "declaring costs nothing" invariant that makes the reader-first landing safe. Reads exactly like the sort of "why this exists" comment `[[project_hf_ponytail_comment_idiom]]` picks out as load-bearing.
- **The `??` operator is deliberate and correct.** A declared-but-invalid state (`"disabled"`, `null`, `42`) returns `undefined` from `declaredCaptionState`, which then falls through to the colour-equality fallback — the composition doesn't ship broken because someone typoed a state name.
- **`data` is a real GSAP-reserved passthrough.** GSAP's docs mark `data` as one of the "special" vars it stores on the tween but doesn't animate, so the PR body's "GSAP passes unknown vars through untouched" isn't just true, it's specifically the field that carries this contract without a runtime cost. Well chosen.
- **Test 1 documents the exact bug the declaration exists for.** The comment at `:138-140` states the failure mode (identical colours → dropped active override) as the reason the test exists — reader can find the "why this test exists" without hunting git history. Same for test 3's "deliberately contradicting" note. `[[feedback_test_name_flips_signal_premise_inversion]]` doesn't fire here because the test *names* aren't inverted — the tests exercise a new codepath and the assertions match the codepath's intent.
- **`main`'s applier vs PR applier claim (`2 failed, 4 passed` → `6 passed`) is falsifiable.** Two tests exist that would fail against pre-PR `main` (test 1's identical-colour case, and test 3's contradicting-heuristic case), plus test 2 as a regression pin. Consistent with the PR body's arithmetic.
- **Scope is exactly right.** Two files, +87/-6, one class of failure, one repair, one fallback preserved. No drive-by refactors, no seam disruption.

Vance-authored, so this is a different track from Miguel's HF PR arc this evening (7 open PRs). CI is green through this SHA (Producer: unit tests, Studio load smoke, Studio timeline viewport gate, runtime contract all pass; Windows-latest tests and regression shards still in progress at review time). LGTM from my side; would take the mixed-case baseline tightening but wouldn't block on it.

vanceingalls:
Thanks both — you independently found the same real bug, and it was worth fixing here rather than deferring.

**Baseline pollution — fixed in `cef0dde5f`.** Taking the suggested shape verbatim:

```ts
const dimReference =
  colorTweens.find((tw) => declaredCaptionState(tw) === "dim") ??
  colorTweens.find((tw) => declaredCaptionState(tw) === undefined);
```

Reproduced first as a failing test using the exact case from the review (`[{ color: "#eee", declared active }, { color: "#111" }]`) — the undeclared dim tween was receiving the active override. Two tests now pin it: the poisoned-baseline case, and a declared-`"dim"` tween being preferred as the reference over an undeclared one. Reverting to `colorTweens[0]` fails both.

The framing that a partial migration could leave a composition *worse off than declaring nothing* is what made this worth blocking on rather than following up.

**Malformed declaration — test added.** Covers `{ captionState: "typo" }`, a number, `null`, and a non-object `data`. All fall through to the fallback, so a future tightening of the accepted union can't quietly turn an unrecognised value into a broken composition.

**Stale classification comment — rewritten.** It now names both layers (declaration first, colour equality as fallback) and why the reference excludes declared tweens.

**Not taken:** the `as { captionState?: unknown }` cast. Agreed it reads loose, but as you both note the primitive fall-through is already safe, so a `typeof` narrow would be churn without a behaviour change.

**Writer side:** correct that nothing emits `captionState` yet, so every shipped composition still takes the fallback on day one. `packages/studio/src/captions/generator.ts` is the natural producer and is where I'd land it. Reader-first was deliberate — a writer without this merged would fail closed.

Tests: 9 passed, up from 6.

miguel-heygen (approved):

PR #3107, opened 2026-08-08
fix(cli): count font axes as motion in the sweep fingerprint

miguel-heygen:
## What

Fold the computed `font-variation-settings` into the layout audit's per-element sweep fingerprint.

## Why

The fingerprint is rect plus opacity, with an 8x8 pixel hash for `<canvas>` and `<video>`. A composition whose only change across the timeline is a font axis touches none of those, so all six samples hash identically and `check` reports `sweep_static` on a timeline that is visibly animating.

Duplexed variable fonts make this unavoidable rather than unlikely. Recursive holds a single advance width across its entire weight axis by design, so even the element rect stays fixed while the letterforms change underneath it.

This is the third instance of the same failure class in this function. The canvas-repaint and opacity-reveal cases already have regression tests sitting right where these two go, which is itself the argument that the fingerprint is the correct place to fix it rather than the compositions.

## How

`font-variation-settings` reads `"normal"` on every element that does not use it, so no existing composition's fingerprint changes.

## Test plan

`packages/cli` full suite: 175 files passed, 1 skipped, 2489 tests passed, 2 skipped, 0 failed. `oxfmt --check` and `oxlint` clean on both files.

Two regression tests, both mutation-checked rather than trusted green:

| Sabotage | Failing test |
| --- | --- |
| Drop the axis string from the fingerprint | `changes the sweep fingerprint when only font-variation-settings moves` |
| Add `Math.random()` to the axis read | `keeps the sweep fingerprint identical when nothing moves, font axes included` |

Exactly one test failed in each case (1 failed, 82 passed). The second is the load-bearing one: without it a fingerprint that varied on its own would make `sweep_static` unfireable and every green layout verdict meaningless, and the suite would still be green.

## Worth knowing

A fresh worktree needs `bun run build` before `vitest`, or 68 files fail to collect on `Cannot find package '@hyperframes/parsers/ff-binaries'`. That is missing workspace build output, not a real failure, and it looks alarming.

`feat/primitive-weight-wave` depends on this landing first. Demonstrated on the same composition: the CLI without this fix reports `sweep_static` and `check` fails; with it, layout reports 0 issues across 9 samples and `check` passes.

### DIFF
--- a/packages/cli/src/commands/layout-audit.browser.js
+++ b/packages/cli/src/commands/layout-audit.browser.js
@@ -1517,7 +1517,18 @@
     const parts = elements.map((element) => {
       const rect = toRect(element.getBoundingClientRect());
       const opacity = round(opacityChain(element));
-      return `${rect.left},${rect.top},${rect.width},${rect.height},${opacity}`;
+      // Variable-font axis animation (font-variation-settings) is a real,
+      // visible motion channel that moves no geometry and no opacity, so a
+      // box+opacity fingerprint reads it as a frozen timeline. Worse in a
+      // DUPLEXED face (Recursive holds an identical advance width at every
+      // weight by design), where not even the line width shifts — the whole
+      // run then false-positives sweep_static. Fold the computed axis string
+      // in; it is "normal" for every element that does not use it, so this
+      // adds nothing to the fingerprint of an ordinary composition.
+      const axes = getComputedStyle(element).fontVariationSettings;
+      return `${rect.left},${rect.top},${rect.width},${rect.height},${opacity},${
+        axes && axes !== "normal" ? axes : ""
+      }`;
     });
     for (const media of root.querySelectorAll("canvas, video")) {
       if (!isVisibleElement(media)) continue;
--- a/packages/cli/src/commands/layout-audit.browser.test.ts
+++ b/packages/cli/src/commands/layout-audit.browser.test.ts
@@ -106,6 +106,79 @@ describe("layout-audit.browser", () => {
     expect(revealed).not.toBe(fading);
   });
 
+  // Variable-font axis animation (registry block `weight-wave`): a crest of
+  // weight travels along a headline by rewriting each character's
+  // font-variation-settings, and NOTHING else changes — no geometry, no
+  // opacity, no canvas. A duplexed face makes it total: Recursive holds one
+  // advance width at every weight by design, so not even the line width
+  // shifts and all six sweep samples hashed identically until the axis string
+  // joined the fingerprint. `check` then failed a working composition with
+  // sweep_static, and the documented remedies (spread the reveal, keep an
+  // element animating) cannot help — the motion is real, the fingerprint was
+  // just blind to it.
+  it("changes the sweep fingerprint when only font-variation-settings moves", () => {
+    document.body.innerHTML = `
+      <div id="root" data-composition-id="main" data-width="640" data-height="360">
+        <div id="line"><span id="char">P</span></div>
+      </div>
+    `;
+
+    let axes = '"wght" 400, "slnt" 0';
+    installGeometry(
+      {
+        root: rect({ left: 0, top: 0, width: 640, height: 360 }),
+        line: rect({ left: 40, top: 40, width: 560, height: 48 }),
+        char: rect({ left: 40, top: 40, width: 18, height: 48 }),
+      },
+      {
+        char: {
+          get fontVariationSettings() {
+            return axes;
+          },
+        } as Partial<CSSStyleDeclaration>,
+      },
+    );
+
+    installAuditScript();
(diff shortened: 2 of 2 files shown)


jrusso1020 (approved):
Reviewed at head `2ed1ab40022620b72114b7ed3070a61a92e1502e`.

**Strengths**

- The body is mutation-checked rather than trusted-green, and it publishes which sabotage each test catches. That is the difference between a test and a diff-presence assertion, and the second test is the one that matters: a fingerprint that varied on its own would make `sweep_static` unfireable and every green layout verdict meaningless, while the suite stayed green. Naming that as the load-bearing direction is exactly right.
- The diagnosis is specific rather than generic. Duplexed faces are the case that makes this unavoidable instead of merely unlikely - a face holding one advance width across its whole weight axis freezes the rect too, so no amount of "keep an element animating" advice can rescue the composition. That belongs in the fingerprint, and the comment at `layout-audit.browser.js:1520-1527` says so at the right length.
- Appending the axis field *last* in the record is load-bearing and correct. `font-variation-settings` values contain commas (`"wght" 400, "slnt" 0`), and the record's own field separator is a comma - putting it anywhere else would shift every later field's position. Records are joined with `|` (`:1537`), which the axis grammar does not produce in practice, so there is no cross-record aliasing either.

**important - the failure class is not closed, and the PR's own argument generalises to the siblings**

The body says this is the third instance of the same class in this function, which is the strongest thing in it - and the fix still handles one channel. Other properties animate visibly while touching neither the rect, nor the opacity chain, nor a canvas/video pixel hash:

- `stroke-dashoffset` / `stroke-dasharray` - SVG line-drawing. Path bbox is fixed while the stroke draws on. This one is a headline motion technique, so it is the likely fourth instance.
- `color` / `background-color` / `border-color` - a pure colour fade.
- `filter` (blur, brightness, hue-rotate) and `backdrop-filter`.
- `clip-path` and `mask-position` - a wipe reveal.
- `background-position` / `background-size` - a travelling gradient.
- `box-shadow`, `text-shadow`.

The justification at `:1526-1527` - it reads a constant for every element that does not use it, so it adds nothing to an ordinary composition's fingerprint - is not specific to font axes. It is true of every computed property in that list, which means one pass folding a small fixed set costs the same single template slot and closes the class, instead of queueing a fourth PR shaped exactly like this one.

The counterweight is the one your second test already guards: each added property must resolve to a value that is stable unless animated. `filter`, `clip-path`, `color` and `stroke-dashoffset` all do. `background-position` is the one to be careful with, since percentage values resolve to pixels in some engines and can move with layout - but layout is already in the rect, so it would be redundant rather than unstable.

**nit - "no existing composition's fingerprint changes" is not what the code does**

`axes && axes !== "normal" ? axes : ""` at `:1530` collapses the ordinary case to an empty string, but the slot is still emitted, so every element's record gains a trailing comma:

```
before:  40,40,18,48,1
after:   40,40,18,48,1,
```

So every existing composition's fingerprint string does change. It is inert, but not for the reason the body gives - it is inert because the fingerprint is only ever compared against other samples from the same run. Worth stating precisely, because the claim as written would license persisting one of these across CLI versions.

**Checked, not a finding**

- An ancestor-level animation is caught. `font-variation-settings` inherits, so even when the composition animates the headline rather than each character, `getComputedStyle(child).fontVariationSettings` reflects it - and `elements` at `:1514-1516` is `root.querySelectorAll("*")` filtered only for visibility, so the animated ancestor is usually in the set as well. Both paths cover it.
- Nothing persists this fingerprint across runs. The only consumer is `checkBrowser.ts`, within a single sweep, and `sweep_static` is explicitly never persistence-tiered (`layoutAudit.ts:33`). So the format change above cannot invalidate a stored baseline, because there is no stored baseline.
- The second test's stub returns a constant, so it cannot catch an engine that reports a *varying* computed value for a static scene. That is inherent to a DOM-level test rather than a gap in this one, and the `Math.random()` sabotage covers the version of it that lives in our code.
- An element that explicitly authors `font-variation-settings: "wght" 400` hashes differently from one left at `normal`, even though they render identically. Harmless - both are stable across the sweep, so neither direction of the check is affected.

**Verdict:** APPROVE
**Reasoning:** Right fix in the right place, with the strongest test story of anything in this batch. The important finding is about scope rather than correctness: the body correctly identifies this as the third instance of a recurring class, and the same one-line argument would have closed the class instead of advancing it by one.

— Rames Jusso

miga-heygen (approved):
Reviewed at exact head. Variable-font axis animation (font-variation-settings) is a real motion channel invisible to the box+opacity fingerprint. This folds the computed axis string into the sweep fingerprint so duplexed faces like Recursive don't false-positive `sweep_static`. Both directions tested: axis change produces different fingerprint, no-change produces identical fingerprint. Clean and minimal. LGTM.

— Miga

PR #3108, opened 2026-08-08
feat(registry): add the lt-neon-border lower third

miguel-heygen:
## What

`lt-neon-border`: two arcs of light travel the edge of a rounded panel and snap corner to corner, trailing a three-layer bloom.

## Why

We ship eleven lower thirds and not one of them has a travelling light. This is the same slot in a different register: precise and expensive rather than loud.

## How

The traversal is closed form rather than accumulated, which is what makes it seekable. `total = t / beat`, `corner = floor(total)`, and the arc head is the eased interpolation between two corner positions, each a pure function of an integer. No accumulator, no clock, no dt.

Three implementation choices worth flagging, since none is obvious from the diff:

**Arcs are SVG dash offsets, not sampled polylines.** Sampling exists in canvas implementations because canvas has no path-relative drawing; along an SVG path `stroke-dasharray` is exact. That also makes a minimum-arc-length clamp meaningless here, so there isn't one. The segmentation that remains (20 glow, 26 core) is for the opacity taper only: at 8 segments the filament stair-steps visibly.

**The bloom widens the stroke before blurring.** Blur alone reads as haze. Widening then blurring is what makes it read as emitted light.

**Easing uses fixed 32-iteration bisection, not Newton**, so the curve is bit-reproducible rather than convergence-dependent.

Nine variables, every number carrying explicit `min`, `max` and `step`.

## Test plan

`hyperframes check` passes: 0 errors, layout 0 issues across 9 samples, motion 0, contrast 0. Registry item lint 0 errors. `oxfmt --check` and `check-large-files.sh` clean.

Determinism verified by hash rather than by eye. Frame 71 at an off-beat time, PNG bytes SHA-256:

```
run1 seek     b2e2340f...   separate browser launch
run2 seek     b2e2340f...   separate browser launch
run3 stepped  b2e2340f...   timeline stepped 0..71
control: frame 78 differs from frame 71
```

The control matters: without it, three identical hashes would also be consistent with a static image. Eight frames at off-beat times were rendered and looked at; both arcs relocate frame to frame and stay diametrically opposed throughout.

A full `render --strict-variables` pass with all nine variables overridden produced 144 validated frames with cyan bloom, rounder corners, shorter arcs and faster travel, confirming every variable is wired.

## Not covered

The contrast auditor reports 0/0 text checks, meaning WCAG contrast here is unverified rather than verified: it found no text nodes to audit. Legibility rests on the panel fill and text shadows, the same as the sibling `lt-*` blocks.

The file trips the 300-line advisory warning at 353 lines. Splitting the geometry, easing, taper and timeline into sub-compositions would make it harder to follow, and eight existing blocks run 1100 to 2900 lines.

### DIFF
--- a/registry/blocks/lt-neon-border/lt-neon-border.html
+++ b/registry/blocks/lt-neon-border/lt-neon-border.html
@@ -1,0 +1,390 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <script src="https://cdn.jsdelivr.net/npm/gsap@3.14.2/dist/gsap.min.js"></script>
+    <style>
+      * {
+        margin: 0;
+        padding: 0;
+        box-sizing: border-box;
+      }
+      html,
+      body {
+        background: transparent;
+        overflow: hidden;
+      }
+      body {
+        width: 1920px;
+        height: 1080px;
+        font-family: "Inter", sans-serif;
+      }
+
+      #root .nb-wrap {
+        position: absolute;
+        left: 50px;
+        bottom: 56px;
+        width: 1120px;
+        height: 510px;
+      }
+      #root .nb-svg {
+        position: absolute;
+        inset: 0;
+        display: block;
+      }
+      /* The frame's inner box, inset by the bloom padding so the copy lands
+         inside the traced rectangle. */
+      #root .nb-panel {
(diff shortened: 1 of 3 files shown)


miguel-heygen:

PR #3109, opened 2026-08-08
feat(registry): add the cosmic-orb hero object

miguel-heygen:
## What

`cosmic-orb`: an iridescent sphere of nebula bands, dust lanes and discrete stars behind a refracting glass limb.

## Why

The catalog is passes, wipes, type and lower thirds. It has almost no hero objects. An orb like this carries a title card by itself, works as a logo pedestal, and runs indefinitely.

## How

The sphere is analytic rather than raymarched, so cost is flat per pixel: `z = sqrt(1 - r^2)` gives the normal directly. One hash drives every noise layer. The limb samples a back layer at three refraction indices, which is what produces the chromatic fringing at the edge.

**The repaint is driven from a property setter on the tweened driver, not from a timeline `onUpdate` callback.** This is the part worth reviewing. GSAP's `seek(pos, suppressEvents)` defaults `suppressEvents` to `true`, so the callback pattern leaves the canvas frozen on frame 0 under any consumer that seeks without opting out. In a bare page, `t=0`, `t=6.0` and `t=6.2` produced byte-identical pixels. Tweened values are always written during render, suppressed or not, so the setter is the reliable hook and it receives the frame time directly.

The beat hook is closed form: `pulseEnvelope` is comma-separated samples spread across the clip and lerp-sampled at `t`. No audio analysis, no state.

Nine variables, every number carrying `min`, `max` and `step`.

## Test plan

`hyperframes check` passes with 0 errors: lint 0 errors, runtime 0 errors, layout 0 issues across 9 samples, motion 0, contrast 0. The four runtime warnings are the checker's own ReadPixels GPU-stall notices. `oxfmt --check` clean.

WebGL surviving seek-capture was the main risk, so it was tested directly rather than assumed. Six frames captured by `snapshot` and looked at: the orb is present in every one, the starfield and nebula rotate between frames, and the glass limb and chromatic fringing hold throughout. A full 300-frame 1920x1080 render also completed.

Determinism, all four cases identical:

```
direct seek(6.0)                    ==
playback 0 -> 6.0 in 0.1s steps     ==
out-of-order scrub 9.9,1.2,7.4,0,6.0 ==
repeated 6.0,3.0,6.0
```

Seven sampled times produced seven distinct hashes, which rules out the identical-because-static failure. Two separate CLI processes produced the same md5 at `t=4`.

## Not covered

The catalog preview assets (`docs/images/catalog/blocks/cosmic-orb.{png,mp4}`) are not generated; the mdx points at those URLs and they will 404 until `scripts/upload-docs-images.sh` runs with the publishing profile. That is a maintainer step.

Not a seamless loop: the noise is not periodic in rotation, so it runs indefinitely but does not cut back to frame 0.

### DIFF
--- a/docs/catalog/blocks/cosmic-orb.mdx
+++ b/docs/catalog/blocks/cosmic-orb.mdx
@@ -1,0 +1,57 @@
+---
+title: "Cosmic Orb"
+description: "An iridescent soap-bubble planet: a slowly spinning sphere of nebula bands, dust lanes and twinkling stars wrapped in a refracting glass limb with chromatic aberration at the edge."
+---
+
+<video className="w-full aspect-video rounded-xl object-cover bg-zinc-100 dark:bg-zinc-800" src="https://static.heygen.ai/hyperframes-oss/docs/images/catalog/blocks/cosmic-orb.mp4" poster="https://static.heygen.ai/hyperframes-oss/docs/images/catalog/blocks/cosmic-orb.png" autoPlay muted loop playsInline />
+
+## Install
+
+```bash Terminal
+npx hyperframes add cosmic-orb
+```
+
+That writes one file: `compositions/cosmic-orb.html`.
+
+## Add it to your video
+
+It runs for 10 seconds at 1920×1080. Paste this into your composition:
+
+```html index.html
+<div
+  data-composition-id="cosmic-orb"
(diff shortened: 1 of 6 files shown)


miguel-heygen:

PR #3110, opened 2026-08-08
feat(registry): add the weight-wave kinetic type block

miguel-heygen:
## What

`weight-wave`: a wave of thickness travels along a headline. Letters swell from light to black and lean into the crest, then settle.

Depends on #3107. Without it the layout audit reports this timeline as static, for the reason that PR explains.

## Why

Every kinetic type move in the catalog animates transform, opacity or a mask. This animates the letterforms themselves, which is a different kind of motion and one almost nobody ships.

## How

Closed form by construction. Each character's weight is a function of its distance from a crest that sweeps the index range, so per-character stagger falls out of the geometry rather than needing a separate mechanism. One proxy tween drives the paint; no accumulator, no clock.

Ships Recursive (SIL OFL-1.1), subset to the two axes it uses: 1.79 MB to 76 KB, 217 glyphs, three unused axes pinned. Vendored the same way every existing font-carrying registry item does it, so there is no render-time network fetch.

**Three findings that are not guessable from the diff:**

**Recursive is duplexed.** `HAMBURGEFONSTIV` measures 925.00px at weight 300, 650 and 1000, and identically at slant 0 and -15. It holds one advance width across the whole axis by design. A width-delta liveness probe therefore reports a completely healthy font as dead, which is what the first implementation did. The guard measures rendered ink instead: 15,149 px at weight 300 to 49,020 px at weight 1000.

**An ink-only guard would have been a false pass.** With the font file deleted, the fallback family still varies its own ink with weight (21,699 to 38,057), so ink alone reports the axis live on a completely broken composition. The guard needs `document.fonts.check()` as well. A falsifier proved this rather than reasoning about it, and the failure mode it protects against is the silent one: every character renders at the default instance while the page looks fine.

**`CRSV` defaults to 0.5, which is Recursive's "auto".** Cursive single-storey letterforms swap in the moment the slant axis goes negative. Under a travelling crest that is a per-character glyph substitution popping on and off, not an interpolation. Pinning it to 0 is a correctness fix, and reads like a routine `varLib.instancer` argument in the diff.

Axis ranges were read from the binary rather than assumed: `wght 300..1000`, `slnt -15..0`. Variable bounds match exactly, so no slider offers a value the font cannot render. `slant` defaults to -12 rather than the -15 axis end, which leaves the control room to travel in both directions.

## Test plan

`hyperframes check` (with #3107 applied): 0 errors, layout 0 issues across 9 samples, motion 0, contrast 120/120 AA. Registry item lint 0 errors. `oxfmt --check` clean.

Per-character rendered `font-variation-settings` read from `getComputedStyle` at eight seek positions. The crest at t=0.75s: `400 400 400 400 400 438 588 738 888 963 813 663 513 400 400` — a clean triangle, moving one group per sample. Eight frames were rendered and looked at.

Determinism: stepping 40 frames to t=1.333 is byte-identical to a direct seek there, and an out-of-order seek (5.5, 2.2, 1.333) matches. Two independent capture runs produced identical SHA-256 for all frames.

## Worth knowing

Recursive was chosen over Roboto Flex deliberately, and the duplexing is why. Roboto Flex's line breathed about 10px horizontally as the crest passed, so a centred headline wobbled frame to frame. Recursive's line is dimensionally locked, so crest letters get denser in place rather than shouldering their neighbours aside, which reads as pressure applied to the letterforms rather than as a bold toggle.

Honest caveat: at weight 1000 the counters tighten noticeably to hold the fixed advance. At 132px display that reads as intentional and dense; at small sizes it would clot.

## Not covered

The catalog preview assets are not generated; the mdx points at URLs that will 404 until `scripts/upload-docs-images.sh` runs with the publishing profile. Maintainer step.


miguel-heygen:

PR #3116, opened 2026-08-08
feat(registry): add five video primitives

miguel-heygen:
## What

Five video primitives in one PR. Supersedes #3108, #3109, #3110, #3112 and #3113, which are closed in favour of this.

| Block | What it is |
| --- | --- |
| `cosmic-orb` | A nebula sphere behind a refracting glass limb |
| `gallery-tunnel` | A corridor of panels flying toward camera |
| `lt-neon-border` | Two arcs of light travelling a rounded panel, snapping corner to corner |
| `weight-wave` | A crest of thickness moving along a headline |
| `halftone-field` | A dot grid tracking a flowing noise field |

19 files, +3104/-235.

## Why these five

Chosen against the catalog's gaps rather than by what was cheapest to build. We are thick in passes, wipes, kinetic type, lower thirds and captions, and thin in hero objects, depth and texture.

- **`cosmic-orb`** is a hero object, a shape we had almost none of. It carries a title card by itself and loops indefinitely.
- **`gallery-tunnel`** is the cheapest real depth win available, and uniquely its panels are **content slots** — a caller supplies their own screenshots — so it carries a shot rather than decorating one.
- **`lt-neon-border`**: we ship eleven lower thirds and not one had a travelling light.
- **`weight-wave`**: every other kinetic type move we own animates transform, opacity or a mask. This animates the letterforms themselves.
- **`halftone-field`**: we had no texture register at all, and this runs *under* type without competing with it.

## How

All five are closed form: **state at frame N is computed from N, never from frame N-1.** That is what makes them seekable, and it is the reason each is a rewrite rather than a port — the reference implementations are all `requestAnimationFrame` accumulators, which render frame 0 forever under a seeking renderer.

The rewrites, concretely: `lt-neon-border`'s corner traversal is `floor` and `frac` of `t/beat`, bit-exact. `cosmic-orb`'s spin collapses to a constant rate. `halftone-field` is `uTime = frame/fps` with no state at all. `gallery-tunnel` keys every choice off a hash of the absolute segment number. `weight-wave`'s per-character weight is a function of distance from a sweeping crest, so stagger falls out of the geometry.

## Two framework defects found while building these

Both are fixed in the blocks that hit them, and both are worth knowing about beyond this PR:

**A timeline `onUpdate` callback does not fire on `seek()`.** GSAP defaults `suppressEvents` to true, so a canvas driven that way freezes on frame 0 under any consumer that seeks. Proven directly: `t=0`, `t=6.0` and `t=6.2` produced byte-identical pixels. The repaint hangs off a property setter on the tweened driver instead, since tweened values are always written during render. Note this is the pattern our own registry contributing guide currently recommends.

**A CSS-sized image uploads a 1x1 texture.** The browser samples the laid-out size, not `naturalWidth`, so every image panel sampled one dark pixel — invisible against a black backdrop. Isolated with a minimal repro that passed, then bisected against the composition.

## Test plan

`bun run lint` exit 0. Registry item lint passes for all five, 0 errors. `hyperframes check` passes per block.

**`check` is not a visual gate** — a composition can pass it and render nothing — so every block was verified by rendering frames and looking at them, and determinism was verified by hashing rather than by eye. Per block: a direct seek to frame N against stepped playback against an out-of-order scrub, all identical, with a control frame proving the hashes are not just of a static image.

`weight-wave`'s font-axis liveness has a falsifier rather than an assertion: with the font file removed the guard fires and `check` fails. That matters because the failure mode is silent — every character renders at the default instance while the page looks fine. An ink-only guard would have been a **false pass**, since the fallback family varies its own ink with weight.

## Dependency

`weight-wave` needs #3107 to land first. Without it the layout audit reports its timeline as static: the sweep fingerprint is rect plus opacity, a font axis touches neither, and Recursive is duplexed so even the element rect stays fixed.

## Not covered

Catalog preview assets are not generated for the new blocks; their mdx points at URLs that 404 until `scripts/upload-docs-images.sh` runs with the publishing profile. Maintainer step.

All WebGL blocks were verified only under software GL; the build box has no hardware GPU.

`gallery-tunnel`'s wall cells are portrait at the default grid, so a 16:9 screenshot letterboxes rather than filling. Deliberate — contain-fit beats squashing — but a caller wanting edge-to-edge artwork will see gutters.


miguel-heygen:
Folded into #3090 along with eight further primitives, per the request to keep everything catalog-facing in one PR. Same compositions, same verification; nothing dropped.

PR #2176, opened 2026-07-10
fix(cli): make skills update converge on a skill retired upstream

miguel-heygen:
## Summary

- `hyperframes skills update` now converges when a skill has been retired/renamed upstream while still installed locally — instead of failing hard or looping forever.
- Target selection resolves the canonical published manifest, so a retired skill is never (re)installed from a repo that no longer ships it.
- Orphaned lock entries with no on-disk bundle are pruned directly, so `check || update` reconciles instead of re-flagging them on every run.
- The stale-skills nudge no longer fires from `skills` / `skills update` itself, and its count now includes removed-upstream skills.

## Reproduction

Both surfaced after `hyperframes-media` was folded into `media-use` (and earlier `hyperframes-captions`/`-compose`/`-tts` were consolidated), for users who still had the old skills installed:

- Inside a repo checkout: `npx hyperframes skills update` → `Update failed: Skill(s) still missing after install: hyperframes-media`.
- Outside any repo: `npx hyperframes skills update` → `No matching skills found for: hyperframes-captions, hyperframes-compose, hyperframes-tts`, and the same nudge reappears on every rerun.

## Root cause

`skills update` had no notion of "gone upstream". A retired skill reached the engine two ways, and both dead-ended:

- **Install path** — `resolveLatestManifest` could resolve `latest` from a stale local `skills-manifest.json` (`findRepoManifest`), while `skills add` always installs from the canonical repo. `isCoreSkill` matches by the `hyperframes-` prefix, so a retired skill was forced into the target set; `skills add` silently declined it (exit 0), and strict `verifyInstalled` threw.
- **Prune path** — upstream `skills remove` scans on-disk directories, so a lock entry retired before it ever shipped a bundle has nothing to match. The call is a silent exit-0 no-op that never touches the lock, so `detectRemoved` re-flags the same entry forever.

The nudge compounded it: it fired even from `skills update`, pointing users back at the command that just failed, and its count ignored the `removed` bucket (the "2 vs 3" mismatch).

## Fix

- `updateSkills` resolves targets against the canonical manifest (`checkSkills({ canonical: true })`), so a retired skill never becomes a target.
- New `pruneOrphanedLockEntries` clears the orphaned lock entries the upstream remover can't, reusing the pinned lock path (`SKILLS_CLI_LOCK_PATHS_VERIFIED_AT`). Idempotent — a second run is a clean no-op.
- Exclude `skills` from the update-nudge gate, and thread the `removed` count through the nudge total.

## Test plan

[...]

PR #3118, opened 2026-08-08
fix(cli): stop skills update deleting skills the manifest never covered

### LINKED ISSUE #3111: Global skill-mirroring step deletes unrelated user-authored skills, then erases evidence of doing so
## Summary

`hyperframes init`/`skills update` deleted two unrelated, pre-existing, user-authored skills from `~/.claude/skills/` while linking into global agent directories. Root cause looks like the "link into N other agent directories" step writes lock entries in `~/.agents/.skill-lock.json` for **every skill folder already present** in a target directory — not just the ones it just installed — attributing them to `heygen-com/hyperframes`. The very next removed-skill check then finds those names absent from HyperFrames' own manifest and deletes them as "no longer published," then prunes the now-orphaned lock entries — which erases the evidence of the false attribution.

## Environment

- `hyperframes` CLI via `npx` (cached at `~/.npm/_npx/2d24f54ded6cd194/node_modules/hyperframes`)
- macOS, Node via Homebrew
- Multiple AI-tool skill directories present on the machine (`~/.claude/skills`, `~/.codex/skills`, `~/.cursor/skills`, `~/.copilot/skills`, `~/.gemini/skills`, `~/.continue/skills`, `~/.agents/skills`, etc.) — some pre-populated with unrelated, hand-authored skills unconnected to HyperFrames.

## Steps to reproduce

1. Have at least one unrelated skill already present in a global skills directory HyperFrames will target, e.g. `~/.claude/skills/my-own-skill/SKILL.md`, authored independently and never installed via HyperFrames.
2. In a project, run `npx hyperframes skills update <workflow-name>` (project-scoped) or `npx hyperframes init <dir> --skill=<workflow-name>` (which links into every detected global agent directory).
3. Observe the CLI's own output: `Linked skills into N other agent directories`, followed on a subsequent `check`/`update` pass by `Removing M skill(s) no longer published: ...` — where the removed list can include names that were never HyperFrames skills.
4. Check `~/.claude/skills/my-own-skill/` — deleted from disk.
5. Check `~/.agents/.skill-lock.json` — no trace of `my-own-skill` (pruned as an "orphaned lock entry" in the same pass), making the false attribution hard to spot after the fact.

## What actually happened to me

Running `npx hyperframes skills update general-video` (project-scoped) followed shortly by `npx hyperframes init "Video/.../hyperframes-final" --skill=general-video` (which reported `Linked skills into 9 other agent directories`) resulted in two of my own, user-authored skills being deleted from `~/.claude/skills/` — both git-tracked in my personal skills repo and completely unrelated to HyperFrames, no shared name, no shared source. I was able to restore them with `git checkout` since that directory happens to be under version control; on a machine without that safety net this would be permanent, silent data loss.

## Expected behavior

The "link into other agent directories" step should only ever write/attribute lock entries for skills it is *actually installing in this operation*. It should never treat a target directory's pre-existing, unrelated contents as something it now owns and is therefore responsible for pruning.

## Suggested fix direction

In the compiled CLI (`dist/cli.js`), the relevant functions are `detectRemoved`, `skillsAttributedToSource`, and whatever performs the initial "link into N other agent directories" step (upstream of `detectRemoved`, since by the time `detectRemoved`/`pruneOrphanedLockEntries` run, `~/.agents/.skill-lock.json` already (incorrectly) attributed the unrelated skills to `heygen-com/hyperframes`). The link step should record only skills it created/copied in that run, not enumerate and claim everything already on disk in the target directory.

Happy to provide more detail or test a fix — this cost real (recoverable, in my case) data loss and I'd guess it's silent for most users since the orphan-prune step deletes the evidence.

jrusso1020:
## What

Scopes `skills update`'s removed-skill prune to skills the published manifest is actually authoritative for, so it stops deleting skills the same command just installed.

Fixes #3111.

## Why

`hyperframes skills update` deleted skills from every agent directory on the machine and reported them as "no longer published" — while they were, in fact, published.

The chain:

1. `skills add --skill '*'` installs **every** skill in the repo. That includes the repo-native ones under `.claude/skills/` and `.agents/skills/`, not just the 19 under `skills/`.
2. The upstream lock attributes all of them to `heygen-com/hyperframes` — `source` is identical for every entry.
3. The published manifest is generated from `<repoRoot>/skills` **only** (`packages/cli/scripts/gen-skills-manifest.ts:18`), so it never lists the repo-native ones.
4. `detectRemoved` read "absent from the manifest" as "removed upstream" and pruned them.

So `check || update` could never converge: `add` reinstalls those skills, the next `update` deletes them again, forever.

The category error is in step 4. The manifest answers *"is this skill's content current?"* for the set it covers. The prune reuses it to answer *"does this skill still exist upstream?"* — a different question, and for anything outside `skills/` the manifest is simply silent. Silence was being read as a delete instruction.

## How

**1. Scope the prune to the manifest's coverage root** (`skillsManifest.ts`). The lock's `skillPath` records where in the repo each skill came from — `skills/general-video/SKILL.md` vs `.agents/skills/seam-craft/SKILL.md` — and it is the *only* field that separates them, since `source` is the same for both. `detectRemoved` now only considers entries under `skills/`.

An entry with **no** `skillPath` (older upstream lock format) is treated as **not** covered. This is a delete path, so unknown provenance fails safe.

**2. Resolve the prune's manifest canonically** (`skills.ts`). Without it, `resolveLatestManifest` takes the `findRepoManifest` shortcut: any `skills-manifest.json` within 16 parent directories of cwd becomes "latest". Since HyperFrames' own manifest declares `source: heygen-com/hyperframes`, such a file matches lock attribution and drives deletion. The install-side check already resolved canonically (#2176); the deleting path did not — and the comment asserting that was deliberate and "tested separately" had no such test. An explicit `--source` still wins, because `canonical` only decides what "latest" means when no source was given.

**Scope note for reviewers:** only change 1 is backed by a reproduction. Change 2 fixes a contract the code documents but does not enforce; I found it while investigating and it sits directly in this deletion path, but I could not produce a user-visible failure from it at `main` (there, local and canonical manifests are identical). Easy to split out if you'd rather.

## Test plan

**End to end against the real CLI**, sandboxed `HOME`, run from a checkout at `b08cefea6`:

| | before | after |
|---|---|---|
| `skills add --skill '*'` | 25 skills installed | 25 installed |
| `skills update` | `Removing 6 skill(s) no longer published: captions-overlay, changelog-video, cut-the-curve, motion-doctrine, oversized-cursor, seam-craft` | no removal line |
| `~/.claude/skills` | 27 → **21** | 27 → **27** |

Two hand-authored, unrelated skills were present in the store throughout; they survived both runs (see the issue thread — they are *not* what this bug deletes).

**Regression tests**, all three fail on the pre-fix source and pass after:

- `never prunes a skill installed outside the manifest's coverage root` — the bug itself.
- `leaves an entry with no skillPath alone rather than guessing` — the fail-safe.
- `resolves the prune's manifest canonically, so a local manifest can never drive deletion` — asserts on *every* `checkSkills` call so a future caller can't reintroduce a non-canonical delete path.

Plus `still prunes a manifest-covered skill that was genuinely dropped upstream`, which passes both before and after — the guard against over-correcting and blunting the retired-skill convergence from #2176.

Existing lock fixtures were widened to carry `skillPath`. They previously omitted a field upstream writes on **every** entry (verified against a real `~/.agents/.skill-lock.json` written by `skills@1.5.22`), so they under-modelled the real lock; the helper defaults to the covered root, which keeps each existing assertion meaning exactly what it meant before.

`oxfmt`, `oxlint`, and `tsc --noEmit` clean; 98 tests pass across both affected files.

— Rames Jusso

### DIFF
--- a/packages/cli/src/commands/skills.test.ts
+++ b/packages/cli/src/commands/skills.test.ts
@@ -425,9 +425,38 @@ describe("hyperframes skills", () => {
 
     await runSkillsUpdate();
 
-    // The update engine's own check (first call) must ask for canonical;
-    // the prune's check (last call, tested separately) intentionally doesn't.
+    // The update engine's own check (first call) must ask for canonical. So
+    // must the prune's (see the GH #3111 regression below) — every caller that
+    // decides what is "still published" resolves the same way.
     expect(checkSkills).toHaveBeenNthCalledWith(1, expect.objectContaining({ canonical: true }));
+  });
+
+  // GH #3111 — silent, permanent data loss. The prune deletes; its notion of
+  // "no longer published" must therefore come from the canonical repo, never
+  // from resolveLatestManifest's findRepoManifest shortcut, which accepts any
+  // `skills-manifest.json` within 16 parent directories of cwd. HyperFrames'
+  // own manifest declares `source: heygen-com/hyperframes`, so such a file
+  // matches lock attribution, and every published skill missing from it is
+  // removed from every agent dir on the machine.
+  //
+  // Reproduced on the pre-fix build: running `skills update` from a hyperframes
+  // checkout whose manifest listed 19 of the 25 published skills printed
+  // "Removing 6 skill(s) no longer published: captions-overlay, changelog-video,
+  // cut-the-curve, motion-doctrine, oversized-cursor, seam-craft" and deleted
+  // all six — every one of them currently published.
+  it("resolves the prune's manifest canonically, so a local manifest can never drive deletion", async () => {
+    setPlatform("linux");
+    const { checkSkills } = await import("../utils/skillsManifest.js");
+
+    await runSkillsUpdate();
+
+    // The prune's check is the LAST call; assert on every call so a future
+    // caller can't reintroduce a non-canonical deletion path.
+    const calls = vi.mocked(checkSkills).mock.calls;
+    expect(calls.length).toBeGreaterThan(1);
+    for (const [arg] of calls) {
+      expect(arg).toEqual(expect.objectContaining({ canonical: true }));
+    }
   });
 
   // Retired-skill regression (variant 2): `skills remove` is a silent no-op
@@ -496,9 +525,14 @@ describe("hyperframes skills", () => {
     await runSkillsUpdate({ source: "owner/repo", dir: "/custom/skills" });
 
     // The last checkSkills call is the prune's — the update engine's own check
-    // (first call) intentionally uses default detection, matching where the
-    // install actually lands.
-    expect(checkSkills).toHaveBeenLastCalledWith({ source: "owner/repo", dir: "/custom/skills" });
+    // (first call) doesn't take --source/--dir, matching where the install
+    // actually lands. `canonical` rides along on every call (GH #3111); an
+    // explicit --source still wins over it inside resolveLatestManifest.
+    expect(checkSkills).toHaveBeenLastCalledWith({
+      source: "owner/repo",
+      dir: "/custom/skills",
+      canonical: true,
+    });
   });
 
   // Skill names come from lock-file JSON keys; a flag-like / shell-special name
--- a/packages/cli/src/commands/skills.ts
+++ b/packages/cli/src/commands/skills.ts
@@ -727,7 +727,17 @@ const updateCommand = defineCommand({
     // failure doesn't fail the update — the install the CI contract gates on
     // already succeeded.
     try {
-      const { skills, scope } = await checkSkills({ dir, source });
+      // `canonical: true` for the same reason the install's target selection
+      // uses it (see updateSkills) — and more urgently, because this branch
+      // DELETES. Without it, resolveLatestManifest takes the findRepoManifest
+      // shortcut: any `skills-manifest.json` within 16 parent dirs of cwd
+      // becomes "latest". HyperFrames' own repo manifest declares
+      // `source: heygen-com/hyperframes`, so a checkout (or any project
+      // carrying a copy) matches attribution and every published skill absent
+      // from that local file is deleted globally as "no longer published".
+      // An explicit `--source` still wins — canonical only decides what
+      // "latest" means when no source was given. GH #3111.
+      const { skills, scope } = await checkSkills({ dir, source, canonical: true });
       const removed = skills.filter((s) => s.status === "removed").map((s) => s.name);
       if (removed.length) {
         console.log();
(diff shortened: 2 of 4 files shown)


miguel-heygen (approved):
No blocking findings.

1. **`skillPath` is the right discriminator.** The manifest authority is explicitly the repo `skills/` root (`packages/cli/src/utils/skillsManifest.ts:518-539`), and removed detection now requires both source attribution and a path under that root (`packages/cli/src/utils/skillsManifest.ts:554-565`). I checked this against `vercel-labs/skills@1.5.22`: GitHub installs write repo-relative, forward-slash-normalized `skillPath` values to both lock formats, so `skills/<name>/SKILL.md` and `.agents/.claude` roots are distinguishable exactly this way. I also checked the generated manifest: it walks only `<repo>/skills`.

2. **Failing safe when `skillPath` is absent is the correct trade.** A destructive reconciliation cannot infer coverage from `source` alone. The regression test at `packages/cli/src/utils/skillsManifest.test.ts:582-598` pins that behavior, while `:566-580` proves a covered retired skill still converges. This also matches upstream’s own update delete path, which refuses removal when a lock entry has no `skillPath`. An old-format stale entry may persist, but that is preferable to another false-positive delete.

3. **The canonical-manifest change belongs in this fix.** The prune is a delete decision, so a nearby/stale checkout manifest must not define “published.” The call at `packages/cli/src/commands/skills.ts:729-740` now uses the same canonical source as installation; `resolveLatestManifest` still honors any explicit source override (`packages/cli/src/utils/skillsManifest.ts:706-735`). The tests at `packages/cli/src/utils/skillsManifest.test.ts:740-804` pin both the bypass and override behavior. This closes a separate route to the same destructive misclassification even though it did not reproduce visibly at current main.

4. **The fixture widening is legitimate.** Current upstream GitHub lock entries do carry `skillPath`; defaulting existing fixtures to `skills/<name>/SKILL.md` preserves their original semantic (“manifest-covered published skill”). The new tests explicitly opt out for the repo-native root (`packages/cli/src/utils/skillsManifest.test.ts:536-564`) and bypass the helper entirely for the legacy no-path case, so the suite is not hiding either boundary.

Locally verified the two touched suites: **98/98 tests passed**. Exact-head CI at `611dbd237` is terminal-green; I also confirmed the new tests ran in the general `Test` job despite the narrower skills jobs being path-skipped. The PR correctly does not claim to explain the reporter’s separate unrelated-skill loss, so keeping #3111 open is appropriate.

**Verdict: APPROVE.** The delete authority is now scoped to what the manifest can actually speak for, unknown provenance fails safe, and retired manifest-covered skills still converge.

— Magi

PR #3128, opened 2026-08-09
fix(studio): stop a Studio edit from reloading the preview

miguel-heygen:
## What

A write made by Studio itself no longer reloads the preview iframe. Writes from outside Studio still do.

## Why

Every Studio edit went to disk, the file watcher saw the change, and the preview reloaded as if an external editor had touched the file. The result was a visible flash after every edit, and any in-flight gesture lost its state.

## How

The client mints a write token, marks it before issuing the request, and sends it as a header. The server records a receipt against the file version and attaches it to the watcher event, so the client recognises its own write and suppresses the reload. Marked before the request because the watcher event can beat the response.

Every write path claims its writes, not just the DOM ones: batch, patch-element, group, timeline and caption. A rollback write deliberately does not claim, so the preview does reload.

## Test plan

- 68 server route tests, 1189 studio hook/util tests
- Editing an element in Studio does not flash the preview; touching the file in an external editor still reloads it

---

First of eight stacked PRs re-cutting #3077, which stays open as the reference for the whole tree.

### DIFF
--- a/packages/studio-server/src/routes/files.test.ts
+++ b/packages/studio-server/src/routes/files.test.ts
@@ -425,6 +425,38 @@ describe("registerFileRoutes", () => {
     expect(readFileSync(join(projectDir, "index.html"), "utf-8")).toContain("After");
   });
 
+  // Without the receipt the client cannot recognise its own edit in the watcher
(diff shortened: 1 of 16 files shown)


miguel-heygen:

PR #3129, opened 2026-08-09
feat(core): sanitize rich text on the way into a composition

miguel-heygen:
## What

A rich-text patch operation, guarded by one sanitizer, with stable element ids. No UI.

## Why

Studio's patch vocabulary was inline-style, attribute, html-attribute and text-content. text-content assigns textContent, and the text-field model escapes markup on the way out and refuses a change in child structure, so a styled span had no route into a composition file.

## How

One sanitizer in packages/core, called on both ends of the trip: in the browser so the preview shows what will be saved, and on the server because that is where the file is written. Tags and style properties are a small allowlist, and an unexpected tag loses its formatting rather than its words. Spans an edit adds get their ids in the same write, so no follow-up write can race it.

## Test plan

- 45 sanitizer tests, 84 source-mutation tests
- Sanitizing is idempotent; disallowed tags are stripped while their text survives

---

Second of eight stacked PRs re-cutting #3077. Base is the PR below it.


miguel-heygen:

PR #3130, opened 2026-08-09
feat(studio): apply a style to a run of characters

miguel-heygen:
## What

The engine that applies a style to selected characters inside an element. Pure logic, no UI.

## Why

Wrapping a DOM range in a span is three lines, and then every interesting case is a special case: recolouring nests spans that shadow each other, removing a style cannot reach the ancestor that set it, and styling across an existing run's boundary has to split it. Each fix is a new branch and the branches interact.

## How

The element is read into a flat list of styled runs, the style is applied to a span of characters in that list, and the element is rebuilt from it. Replacing, removing, splitting and merging stop being cases: the rebuild emits one span per distinct run and cannot nest or duplicate.

Selection offsets count UTF-16 units, so a boundary can land between the halves of an emoji; the applied range widens to whole characters. A colour an ancestor overpaints is mirrored into the fill, because a colour that does not paint reads as a colour that did not save.

## Test plan

- 44 tests covering replace, remove, split, merge, emoji boundaries and the overpainted-fill case

---

Third of eight stacked PRs re-cutting #3077. The toolbar that drives this arrives in the next PR.

### DIFF
--- a/packages/studio/src/components/editor/inlineTextStyleRange.test.ts
+++ b/packages/studio/src/components/editor/inlineTextStyleRange.test.ts
@@ -1,0 +1,557 @@
+// @vitest-environment happy-dom
+
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { applyInlineStyle, readInlineStyle } from "./inlineTextStyleRange";
+
+afterEach(() => {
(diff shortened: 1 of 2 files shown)


miguel-heygen:

PR #3131, opened 2026-08-09
feat(studio): edit and style text in the preview

miguel-heygen:
## What

Double-press a text element in the canvas and the caret opens where you pressed, in the element itself rather than in a panel. Select characters and a small toolbar offers colour, bold, italic and underline, applied to exactly those characters.

## Why

Editing text meant finding it in a panel and retyping it there. The panel could not express styling at all.

## How

The toolbar lives in Studio's document rather than the composition's. Putting it in the preview would inject Studio's chrome into the user's composition, where a render would capture it and the composition's own styling would inherit into it. It is positioned over the selection in viewport coordinates, so it does not have to know which of the canvas' nested coordinate systems it was mounted into.

In a flex or grid container the rebuilt runs go inside one wrapper, so a coloured word cannot reflow the element it sits in.

Also fixes the keyboard: the shortcut guards matched contenteditable=true only, so playback shortcuts ate letters typed into the composition.

## Test plan

- 3264 studio tests green
- Open an edit, type, click away, reload: the text persisted. Bold half a word and the other half stays unbolded through a save.

---

Fourth of eight stacked PRs re-cutting #3077. This one is over the usual size budget: the editor and the toolbar cannot be separated, since the editing layer is what mounts the toolbar. About 40% of the diff is tests.


miguel-heygen:

PR #3132, opened 2026-08-09
feat(studio): show every colour of a mixed selection in the swatch

miguel-heygen:
## What

The text colour swatch shows every colour in the selection, blended, instead of falling back to white.

## Why

The toolbar reports a property only when the whole selection agrees on it. That is right for bold and italic, where a toggle is on or off, but wrong for a swatch: with nothing to report it fell back to the default, so a red-and-green selection claimed to be white.

## How

The swatch reads the colours as they run through the selection and blends them left to right, each sitting at the middle of the share of characters that carry it. Whitespace paints no glyph, so it contributes no colour — otherwise recolouring one word inside coloured text drew a stray band of the surrounding colour at the edge.

Painting is anchored to the border box. A gradient maps to the padding box and then repeats to fill the border box, so the 1px border was showing the gradient's opposite end on each side.

## Test plan

- 65 tests across the toolbar and the style reader
- Verified by reading rendered pixels: the border pixel is now white over the adjacent colour, not over its complement

---

Fifth of eight stacked PRs re-cutting #3077.

### DIFF
--- a/packages/studio/src/components/editor/InlineTextToolbar.test.tsx
+++ b/packages/studio/src/components/editor/InlineTextToolbar.test.tsx
@@ -200,9 +200,7 @@ describe("InlineTextToolbar", () => {
     selectAll(element);
 
     const swatch = toolbarIn(host)!.querySelector<HTMLElement>("span[aria-hidden]")!;
-    expect(swatch.style.background).toBe(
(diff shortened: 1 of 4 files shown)


miguel-heygen:

PR #3140, opened 2026-08-09
fix(studio): hold a resized element's size while the timeline is rebuilt

miguel-heygen:
## What

A resized element keeps its new size across a seek instead of snapping back to its stylesheet size.

## Why

Size does not compose the way the position offset does: both channels write width and height, so the later write wins. Standing aside meant nothing held the size while a soft reload reverted the old timeline, since GSAP hands back each tween's recorded starting width. The element sat at its stylesheet size until the new timeline rendered, and the next gesture then started from a box disagreeing with the committed size.

## How

The size reapply no longer skips GSAP-sized elements. Only elements on which Studio has committed a size carry the reapply variables. Those variables intentionally remain authoritative across seeks; when the same element also has authored width or height GSAP tweens, the Studio-committed size wins until a later Studio resize changes it.

## Test plan

- Resize an element, scrub the timeline, and verify the committed size holds
- Verify an untouched element keeps its authored width and height behavior
- Verify a Studio-resized element with width or height GSAP tweens keeps the documented Studio-committed precedence

### DIFF
--- a/packages/studio/src/components/editor/DomEditOverlay.tsx
+++ b/packages/studio/src/components/editor/DomEditOverlay.tsx
@@ -519,6 +519,7 @@ export const DomEditOverlay = memo(function DomEditOverlay({
               top: cr.top,
               width: cr.width,
               height: cr.height,
+              transform: cr.angle ? `rotate(${cr.angle}deg)` : undefined,
             }}
           />
         ))}
--- a/packages/studio/src/components/editor/domEditOverlayGeometry.test.ts
+++ b/packages/studio/src/components/editor/domEditOverlayGeometry.test.ts
@@ -5,6 +5,7 @@ import {
   orientedGroupAwareOverlayRect,
   overlayCornersCentroid,
   selectionCacheKey,
+  orientedVisibleOverlayRect,
 } from "./domEditOverlayGeometry";
 
 describe("overlayCornersCentroid", () => {
@@ -157,6 +158,30 @@ describe("orientedOverlayRect — rotation gate (perf fix, V15 18a/18b)", () => 
     expect(rect!.width).toBeCloseTo(200, 5);
     expect(rect!.height).toBeCloseTo(100, 5);
     expect(rect!.angle ?? 0).toBe(0);
+  });
+
+  /**
+   * A child outline is drawn ON the child, not around it. Measured axis-aligned,
+   * a text layer inside a rotated card got an upright dashed box sitting across
+   * the rotated glyphs — the parent's chrome rotated and its children's did not.
+   */
+  it("child outlines carry the element's angle, so they can co-rotate with it", () => {
+    const { overlayEl, iframe, el } = buildHarness();
+    el.style.transform = ROTATE_30DEG_MATRIX;
+    const rect = orientedVisibleOverlayRect(overlayEl, iframe, el);
+    expect(rect).not.toBeNull();
+    expect(rect!.angle).toBeCloseTo(30, 3);
+  });
+
+  it("an unrotated child outline is unchanged — no angle, same box as before", () => {
+    const { overlayEl, iframe, el } = buildHarness();
+    const rect = orientedVisibleOverlayRect(overlayEl, iframe, el);
+    expect(rect).not.toBeNull();
+    expect(rect!.angle ?? 0).toBe(0);
+    expect(rect!.left).toBeCloseTo(400, 5);
+    expect(rect!.top).toBeCloseTo(450, 5);
+    expect(rect!.width).toBeCloseTo(200, 5);
+    expect(rect!.height).toBeCloseTo(100, 5);
   });
 
   it("rotated element takes the corner-geometry path — reports the live angle", () => {
--- a/packages/studio/src/components/editor/domEditOverlayGeometry.ts
+++ b/packages/studio/src/components/editor/domEditOverlayGeometry.ts
@@ -430,6 +430,24 @@ export function orientedOverlayRect(
   };
 }
 
+/**
+ * `toVisibleOverlayRect`'s oriented twin: the element's crop-hugged box plus its
+ * live rotation, for chrome that has to sit on a rotated element rather than
+ * around it. Rendering the result with `transform: rotate(angle)` about its
+ * centre lands it on the element's real corners.
+ *
+ * At angle 0 `orientedOverlayRect` returns the plain AABB, so an unrotated
+ * element measures exactly as it did before.
+ */
+export function orientedVisibleOverlayRect(
+  overlayEl: HTMLDivElement,
+  iframe: HTMLIFrameElement,
+  element: HTMLElement,
+): OverlayRect | null {
+  const rect = orientedOverlayRect(overlayEl, iframe, element);
+  return rect ? { ...rect, ...hugRectForElement(rect, element) } : null;
+}
+
 const OVERLAY_RECT_EPSILON_PX = 0.5;
 const OVERLAY_RECT_ANGLE_EPSILON_DEG = 0.1;
 
--- a/packages/studio/src/components/editor/manualEditsDom.ts
+++ b/packages/studio/src/components/editor/manualEditsDom.ts
@@ -36,6 +36,7 @@ import { roundRotationAngle } from "./manualEditsParsing";
 import { applyStudioMotionFromDom } from "./studioMotion";
 import { gsapAnimatesProperty } from "./gsapAnimatesProperty";
 import { splitTopLevelWhitespace } from "./manualEditsStyleHelpers";
+import { logResize } from "../../utils/resizeDebug";
 
 /* ── Gesture tracking ─────────────────────────────────────────────── */
 let studioManualEditGestureId = 0;
@@ -454,6 +455,11 @@ export function applyStudioBoxSize(
   element: HTMLElement,
   size: { width: number; height: number },
 ): void {
+  logResize("apply-box-size", {
+    to: `${Math.round(size.width)} x ${Math.round(size.height)}`,
+    from: `${element.style.width || "-"} x ${element.style.height || "-"}`,
+    stack: new Error("apply-box-size").stack?.split("\n").slice(1, 6).join(" < "),
+  });
   promoteInlineForTransform(element);
   applyStudioBoxSizeDimensions(element, size);
 }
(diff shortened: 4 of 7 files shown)


james-russo-rames-d-jusso (commented):
_Reviewed at `3eae95d23`._

Right fix for the class. The size channel is last-writer-wins on the same numeric properties (unlike offset, which composes), and standing aside for GSAP-sized elements meant the seek reapply left nothing holding the studio size across the soft-reload window — element sat at the stylesheet size for the flash. Dropping the `gsapAnimatesProperty(el, "width", "height")` skip closes the window. The "only elements the studio has committed a size on carry these vars" invariant is what makes it safe to unconditionally reapply.

Test at `reapplyBoxSizeAfterSeek.test.ts:35-43` is a direct contract lock (GSAP put the tween's starting size back → we put the studio's committed size back on top). The paired negative test at `:45-53` proves the "leave GSAP alone once the studio props are cleared" reciprocal — this is the reassuring half, because it pins the reason the fix doesn't affect elements the user never resized.

Nice small refactor tucked in alongside: `reapplyPathOffsets` at `manualEditsDom.ts:557-568` gets an early continue and a leading comment `// Unlike size below, the offset channels COMPOSE — applying both doubles the move.` — makes the two functions read as intentionally different, which the previous shape didn't quite convey.

Clean, no findings from my side. LGTM.

vanceingalls (approved):
## R1 review

**Verdict: APPROVE (high confidence).** Small, well-diagnosed fix. `reapplyBoxSizes` used to skip GSAP-sized elements, which left a hole during the soft-reload teardown/rebuild window: GSAP would restore each tween's *recorded* starting width on the way out, nothing held the committed size, and the element sat at its stylesheet size until the new timeline rendered — the visible "jump after resize." Removing the skip pins width/height to `STUDIO_WIDTH_PROP` / `STUDIO_HEIGHT_PROP` at every seek so the resized size holds. Diagnosis is corroborated by the preceding debug commit (`chore(studio): name whoever puts the pre-resize size back`) — Miguel first proved the mechanism with logging, then fixed the right end.

The `reapplyPathOffsets` refactor in the same hunk is behavior-preserving (moves the GSAP skip earlier, extracts the offset object literal). Old `if (x || y)` and new `if (!x && !y) continue;` are inverses of the same check; `Number.parseFloat(x) || 0` is unchanged.

### P1 blockers

None.

### P2 non-blockers

- **Docstring / PR-body claim is imprecise.** `manualEditsDom.ts:568-575` and the PR body say *"Only an element mid-edit carries the vars, so this cannot affect anything the user is not actively resizing."* That's not what the code does. `STUDIO_BOX_SIZE_ATTR` + `STUDIO_WIDTH_PROP` are written into the source HTML by `buildBoxSizePatches` (`manualEditsDomPatches.ts:152-163`) and only cleared by `clearStudioBoxSize` (from "Reset layer edits" or from `finalizeScaleResizeCommit`). They persist across sessions on any element that has *ever* been resized in the studio. The fix is still correct — the intended scope is "any element the user has resized" — but the phrase "actively resizing" reads as "during the drag gesture only," which underspecifies the blast radius. Suggest: *"Only an element the studio has resized carries the vars; untouched elements are unaffected."*
- **Trade-off with user-authored width/height GSAP tweens.** The studio's animation panel exposes `width` and `height` as animatable properties (`gsapAnimationConstants.ts:20-21, 53-54, 88-89, 162-163`), and the seek reapply runs on `seek()` AND every `rAF` tick during playback (via `startStudioManualEditPlaybackReapply` in `manualEdits.ts:168-182`). An element that is *both* (a) resized via the studio gesture and (b) targeted by a user-authored `width`/`height` GSAP tween with different start/end values now has that tween pinned to `STUDIO_WIDTH_PROP` on every frame — the animation freezes at the committed base size. Old code side-stepped this by skipping GSAP-sized elements entirely. Trade seems worth it (resize→scrub jump is common; resize + width-tween is exotic), but consider a `TODO`/follow-up note in the comment, or a test that documents the intended precedence.

### P3 nits

- **Test 2 has a fixture-initial tautology.** `reapplyBoxSizeAfterSeek.test.ts:56-64` sets `el.style.width = "395px"`, clears the studio vars, calls `reapplyPositionEditsAfterSeek`, and asserts `el.style.width === "395px"`. The fixture initial value equals the expected outcome, so a mutation that made the reapply a full no-op would still pass. Per the mutation-resistance principle, consider additionally asserting that `STUDIO_BOX_SIZE_ATTR` is still `"true"` and that `applyStudioBoxSize` was *not* invoked (e.g. via a spy on `setProperty("width", …)`), so the test discriminates "did nothing because vars are empty" from "did nothing at all."
- **Comment symmetry.** The comment above `reapplyBoxSizes` (L568-575) references "the offset above"; the two comments read as a pair, but `reapplyPathOffsets` is only one line (`manualEditsDom.ts:558`) while `reapplyBoxSizes` explains itself in seven. Consider symmetrizing.

### Per-lens findings

1. **What was the bug** — resize → soft-reload window → GSAP hands back tween's recorded starting width → element sits at stylesheet size until new timeline renders → visible jump. Mechanism traced in the second commit body against real logs.
2. **Persistence mechanism** — CSS custom properties (`STUDIO_WIDTH_PROP`, `STUDIO_HEIGHT_PROP`) plus a data-attribute marker (`STUDIO_BOX_SIZE_ATTR`). Written to the source HTML via `buildBoxSizePatches`. Not keyframe-based; base-size only.
3. **Seek behavior** — all seek entry points route through `__hfStudioManualEditsApply` → `reapplyPositionEditsAfterSeek`. Wrapped seek/totalTime/play/pause on `__hf`, `__player`, `__timeline`, and every entry in `__timelines`. Includes a `Proxy` on `__timelines` so late-registered timelines get wrapped. Coverage complete.
4. **Multi-keyframe elements** — the fix pins to the *base* studio size at every seek/frame; no keyframe-aware routing. Elements with legitimate multi-keyframe width/height animations are the P2 concern.
5. **Undo/redo** — snapshotted through `manualEditsSnapshot.ts:126-128`. Not touched by this diff.
6. **Instant-hold interaction (HF #3092)** — that fix lives in `gsapResizeIntercept.ts` (scale-route commit path), disjoint surface. No overlap.
7. **`useEffect` / state syncing** — none added. Pure DOM helpers.
8. **Element-identity** — marker + vars live on the DOM node itself; not keyed on id.
9. **Test coverage** — first test proves the regression. Second test guards the "vars cleared" no-op path with the tautology caveat above. No test exercises playback rAF loop or animating (not constant-value) width tween.
10. **CI** — at HEAD `3eae95d`: no failures. Regression shards 2/3/5/7/8 + Windows still IN_PROGRESS at review time; every finished check green.

— Via

vanceingalls (approved):
## R2 re-verify @ `1377e60cb`

**Verdict: APPROVE.** Miguel's "unchanged code" claim is correct — `manualEditsDom.ts` and `reapplyBoxSizeAfterSeek.test.ts` are byte-identical vs prior head `3eae95d`. PR description now precisely states the resize scope and the width/height precedence, closing the R1 P2 items at the doc level.

**Per-finding delta:**

- **P2-1 (imprecise docstring, precise scope): FIXED via PR body.** The imprecise sentence R1 quoted (*"Only an element mid-edit carries the vars…"*) had already been folded out of the docstring by the prior head; R1 was actually quoting an earlier commit. PR body now states verbatim: *"Only elements on which Studio has committed a size carry the reapply variables. Those variables intentionally remain authoritative across seeks; when the same element also has authored width or height GSAP tweens, the Studio-committed size wins until a later Studio resize changes it."* Matches the R1 suggested rewrite semantically.
- **P2-2 (precedence with user-authored width/height GSAP tweens): PARTIAL.** FIXED in PR body ("Studio-committed size wins…") and in test plan (manual checklist item). NOT ADDRESSED in code comment (`manualEditsDom.ts:568-575` docstring still doesn't mention the tween-precedence trade-off) or in an automated test (`reapplyBoxSizeAfterSeek.test.ts` unchanged). Non-blocking — the trade-off was flagged as non-blocking in R1, and the PR-body + test-plan resolution covers the immediate need — but PR bodies aren't durable, so a one-line note in the comment would close the doc drift for future readers.

**P3 nits: UNCHANGED (informational).**
- Test 2 fixture-initial tautology at `reapplyBoxSizeAfterSeek.test.ts:56-64` — still present; the guard still parses NaN and skips, but no positive discriminator/spy added.
- Comment symmetry between `reapplyBoxSizes` (7 lines) and `reapplyPathOffsets` (1 line) — asymmetry preserved.

**CI:** Green on completed checks (Build, CLI smoke, CLI npx macos/ubuntu/windows, Format, Lint, Typecheck, Producer unit + integration, Test: runtime contract, Preview parity, Perf: drift/fps/load/scrub, Preflight, SDK: unit/contract/smoke, Fallow audit, File size, Studio: load smoke + timeline viewport gate, preview-regression, Analyze actions/python, Codex plugin package). In-progress: Perf: parity, Render+Tests on windows-latest, Test, Smoke: global install, regression-shards 1-9.

— Via

james-russo-rames-d-jusso (commented):
_Re-verified at `1377e60cb` against prior R1 (`3eae95d23`)._

The two files in this PR's own scope (`manualEditsDom.ts` + `reapplyBoxSizeAfterSeek.test.ts`) are byte-identical to the prior R1 head — the only change between the two is a docblock tightening on `reapplyBoxSizes` at `manualEditsDom.ts:568-574` (folded the pre-write jump narrative into a shorter form without losing the load-bearing "size does not compose, later write wins on the same number; standing aside meant nothing held the size across the soft-reload window" invariant). Docblock still names why size differs from offset and why the reapply is safe.

Prior R1 LGTM stands. No new findings.

james-russo-rames-d-jusso (commented):
_Re-verified `0df3d7ee5` against prior R2 (`1377e60cb`)._

Both files in this PR's own 2-file scope (`manualEditsDom.ts` + `reapplyBoxSizeAfterSeek.test.ts`) are byte-identical to the prior R2 head — the delta at this SHA is stack drift from the fileWatcher fix in `#3137`. Prior R1/R2 LGTM stands. Review at `4892125186`.

No new findings.

vanceingalls (approved):
R3 re-stamp @ 0df3d7ee5 — Byte-clean rebase on #3137 R3 (f1de808e7). Own-delta against parent (`f1de808e7...0df3d7ee5`) = 47 files with blob SHAs and +/- counts *identical* to R2 own-delta (`4d2a48ae6...1377e60cb`); no code change to #3140 introduced by the rebase, only the fileWatcher burst-fix propagated up from #3137. Prior R2 approval stands. — Via

miguel-heygen:

PR #3144, opened 2026-08-09
feat(studio): show every colour of a mixed selection in the swatch

miguel-heygen:
## What

The inline text toolbar shows every distinct colour in a mixed selection instead of falling back to white.

## Why

The toolbar reports a property only when the whole selection agrees on it. That is right for toggles such as bold and italic, but a colour swatch can represent several values at once. Falling back to the default made mixed text report a colour that was not selected.

## How

- Read distinct inline colours in document order while excluding whitespace, which paints no glyph.
- Render an evenly spaced 135-degree sweep so even a colour used by one character remains visible in the compact swatch.
- Anchor the gradient to the border box so it does not repeat beneath the ring.
- Normalize shorthand, functional, transparent, and named CSS colours for the native colour input.
- Clear selections owned by a closing inline-edit session, including selections dragged across its boundary, while preserving unrelated selections.

## Test plan

- 114 focused toolbar, style-range, and edit-session tests
- Full Studio suite: 3,791 passed, 18 todo
- Studio typecheck, oxlint, and oxfmt checks
- Codex in-app-browser E2E with `rich-text-demo`: mixed diagonal swatch, border-box paint, colour-picker value, selection preserved through formatting, source persistence, selection cleanup, and no preview reload

### DIFF
--- a/packages/studio/src/components/editor/InlineTextToolbar.test.tsx
+++ b/packages/studio/src/components/editor/InlineTextToolbar.test.tsx
@@ -249,7 +249,10 @@ describe("InlineTextToolbar", () => {
     selectAll(element);
 
     const swatch = toolbarIn(host)!.querySelector<HTMLElement>("span[aria-hidden]")!;
-    expect(swatch.style.background).toBe("linear-gradient(90deg, red 25.00%, lime 75.00%)");
+    expect(swatch.style.backgroundImage).toBe("linear-gradient(90deg, red 25.00%, lime 75.00%)");
+    // Without this the gradient repeats under the border, painting the end
+    // colour along the leading edge and the start colour along the trailing one.
+    expect(swatch.style.backgroundOrigin).toBe("border-box");
   });
 
   it("shows a plain swatch when the whole selection is one colour", () => {
@@ -259,7 +262,7 @@ describe("InlineTextToolbar", () => {
     selectAll(element);
 
     const swatch = toolbarIn(host)!.querySelector<HTMLElement>("span[aria-hidden]")!;
-    expect(swatch.style.background).toBe("red");
+    expect(swatch.style.backgroundColor).toBe("red");
   });
 });
 
--- a/packages/studio/src/components/editor/InlineTextToolbar.test.tsx
+++ b/packages/studio/src/components/editor/InlineTextToolbar.test.tsx
@@ -3,7 +3,7 @@
 import React, { act } from "react";
 import { createRoot } from "react-dom/client";
 import { afterEach, describe, expect, it } from "vitest";
-import { InlineTextToolbar } from "./InlineTextToolbar";
+import { InlineTextToolbar, swatchBackground } from "./InlineTextToolbar";
 import type { InlineTextEditSession } from "../../hooks/useInlineTextEdit";
 
 (globalThis as unknown as { IS_REACT_ACT_ENVIRONMENT: boolean }).IS_REACT_ACT_ENVIRONMENT = true;
@@ -219,7 +219,6 @@ describe("InlineTextToolbar", () => {
     expect(toolbar.style.left).toBe(`${100 + (20 + 50) * scale}px`);
     expect(toolbar.style.top).toBe(`${50 + 40 * scale - 10}px`);
   });
-
   it("moves below a selection that leaves no room above the viewport", () => {
     const { element, session, iframe } = scene("hello world");
     iframe.getBoundingClientRect = () => ({ left: 0, top: 0, width: 400 }) as DOMRect;
@@ -239,5 +238,52 @@ describe("InlineTextToolbar", () => {
     const scale = 400 / window.innerWidth;
     expect(toolbar.style.top).toBe(`${10 * scale + 10}px`);
     expect(toolbar.style.transform).toBe("translate(-50%, 0)");
+  });
+
+  it("shows the selection's colours in the swatch when they differ", () => {
+    const { element, session, iframe } = scene(
+      '<span style="color: red">Hello</span><span style="color: lime">world</span>',
+    );
+    const { host } = render(session, iframe);
+
+    selectAll(element);
+
+    const swatch = toolbarIn(host)!.querySelector<HTMLElement>("span[aria-hidden]")!;
+    expect(swatch.style.background).toBe(
+      "linear-gradient(90deg, red 0.00% 50.00%, lime 50.00% 100.00%)",
+    );
+  });
+
+  it("shows a plain swatch when the whole selection is one colour", () => {
+    const { element, session, iframe } = scene('<span style="color: red">Hello world</span>');
+    const { host } = render(session, iframe);
+
+    selectAll(element);
+
+    const swatch = toolbarIn(host)!.querySelector<HTMLElement>("span[aria-hidden]")!;
+    expect(swatch.style.background).toBe("red");
+  });
+});
+
+describe("swatchBackground", () => {
+  it("shows every colour in the selection, sized by how much text carries it", () => {
+    expect(
+      swatchBackground(
+        [
+          { value: "red", chars: 5 },
+          { value: "lime", chars: 15 },
+        ],
+        undefined,
+      ),
+    ).toBe("linear-gradient(90deg, red 0.00% 25.00%, lime 25.00% 100.00%)");
+  });
+
+  it("stays a plain swatch when the selection is one colour", () => {
+    expect(swatchBackground([{ value: "red", chars: 5 }], "red")).toBe("red");
+  });
+
+  it("falls back to the agreed colour when the characters carry none", () => {
+    expect(swatchBackground([], "rgb(1, 2, 3)")).toBe("rgb(1, 2, 3)");
+    expect(swatchBackground([], undefined)).toBe("#ffffff");
   });
 });
--- a/packages/studio/src/components/editor/InlineTextToolbar.test.tsx
+++ b/packages/studio/src/components/editor/InlineTextToolbar.test.tsx
@@ -249,7 +249,7 @@ describe("InlineTextToolbar", () => {
     selectAll(element);
 
     const swatch = toolbarIn(host)!.querySelector<HTMLElement>("span[aria-hidden]")!;
-    expect(swatch.style.backgroundImage).toBe("linear-gradient(90deg, red 25.00%, lime 75.00%)");
+    expect(swatch.style.backgroundImage).toBe("linear-gradient(135deg, red 0.00%, lime 100.00%)");
     // Without this the gradient repeats under the border, painting the end
     // colour along the leading edge and the start colour along the trailing one.
     expect(swatch.style.backgroundOrigin).toBe("border-box");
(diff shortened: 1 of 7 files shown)


james-russo-rames-d-jusso (commented):
_Reviewed at `a9452d5e`._

Small, focused close on D5 of the rich-text stack. All four PR-body promises deliver and hold under adversarial pressure. The `03e8f6d0 → a9452d5e` amend that followed is a structural-only split (file-size gate: the read-only style interpretation moved to `inlineTextStyleRead.ts`, DOM flattening stays in `inlineTextStyleRange.ts`) — `readInlineStyle` / `readInlineStyleSpread` bodies byte-identical, `coveredChars` renamed and exported as `readCoveredInlineStyleChars`, no behavior change. All my adversarial findings against `03e8f6d0` survive the move unchanged.

## PR-body promises → verified

- **"One owner for distinct-colour ordering."** Dedup lives entirely in `readInlineStyleSpread` (`inlineTextStyleRead.ts:30-47`) via `Set`-with-insertion-order. `swatchBackground` in the renderer just consumes the array. Grepped for any secondary dedup — none. So a future refactor renaming/moving the reader can't silently double-dedup or drop the ordering guarantee.
- **"Shorthand / functional / transparent / named CSS colours normalize for the native picker."** `toPickerColour` tries `parseCssColor` first (handles `#rgb`, `#rrggbb`, `rgb()`, `rgba()`), then falls to canvas `fillStyle` for full CSS-colour grammar (named, `transparent`, `hsl()`). Canvas path is try/catch'd with DEFAULT_COLOR fallback, guarded on `!context`, and preflighted with a known-good value (`context.fillStyle = DEFAULT_COLOR` before assigning the tested value) so an invalid input rejected silently by the browser doesn't inherit stale state. Docstring `:265-272` calls out that DOM-only test envs (happy-dom) miss the canvas normalisation path — E2E covers it. Fine tradeoff.
- **"Closing an edit clears a selection that crosses the element boundary while leaving unrelated selections alone."** `clearSelectionWithin(element)` at `useInlineTextEdit.ts:75-81`: clears iff EITHER endpoint is inside `element`. Three cases verified by tests — inside-only cleared (`.test.tsx:63-84`), outside-only preserved (`:86-104`), cross-boundary (start-in-end-out) cleared (`:106-125`). Gated behind `open.element.isConnected` in teardown so a detached-element teardown doesn't touch the selection.
- **"Authored `outline-offset` restoration remains intact."** Not re-modified; the R2 fix from `a1b1d11b` (session captures `outlineOffset`, restores in teardown) is preserved through the rebase. Verified `useInlineTextEdit.ts:44,97,113-114` unchanged in shape.

## Rich-text stack integration → verified

- Drag/drop plain-text rail from #3143 R3 (`onDragOver` + `onDrop` mirror paste via `insertPlainText`) preserved verbatim through the rebase.
- RAF unmount cleanup effect + inline `openRef.current?.element !== element` guard preserved.
- `ownsCurrentPreviewElement` + `element === selection.element` + `innerHTML === appliedHtml` revert guard preserved.
- `PressMark.element` same-element pairing preserved.
- Single canonical `isTypingTarget` still routed through — no shortcut caller re-imports `isEditableTarget` from anywhere but `timelineDiscovery.ts`'s 3-line re-export.

## Delta-focused adversarial pass — nothing blocking

- **`background` shorthand vs `background-origin: border-box` ordering.** Correct — object property order determines React's assignment order, `background` first (resets everything including origin), `backgroundOrigin` second overrides. Comment at `InlineTextToolbar.tsx:110-111` calls this out explicitly and the test at `.test.tsx:255` pins `backgroundOrigin === "border-box"`. If a future refactor swaps the keys' declaration order the swatch quietly loses its border strip — worth watching but not blocking.
- **`swatchBackground` division-by-zero on single-colour.** Guarded — length-1 early-returns the raw colour string, length-0 returns `agreed || DEFAULT_COLOR`. `(index / (length-1))` only runs when `length >= 2`.
- **`clearSelectionWithin` with a detached range.** `element.contains(...)` on a detached-tree node returns `false`, so both endpoint-inside checks fail, function skips. Selection object stays valid. No throw path.
- **`readInlineStyleSpread` "" value.** Values come from `ownStyle` (`inlineTextStyleRange.ts:334-343`) which filters via `isRichTextFormattingStyle` (empty rejected). So the distinct list can't contain `""`, which would produce an invalid `linear-gradient(..., "" 50%, ...)`.
- **`styles.color ?? colours[0]` coherence.** Traced the four states (agreed / mixed / all-whitespace / no-colour) and picker + swatch stay coherent in each. Notably: all-whitespace-with-agreed-colour uses `styles.color` (not `colours[0]`, which is empty) — correct.

## Nits — none blocking, none new-in-this-PR that I hadn't already parked

- Uniform-scale assumption (`box.width / view.innerWidth` applied to both axes at `InlineTextToolbar.tsx:236`) still open as the standing follow-up from #3143 R1. Currently correct.
- `char.trim()` whitespace filter in `readInlineStyleSpread` misses zero-width joiners (U+200D), variation selectors (U+FE0F), zero-width spaces (U+200B). Text authoring rarely colours individual ZW characters, and even if it did, in practice the ZW char sits inside a run whose neighbours contribute the same colour, so the distinct set wouldn't spuriously widen. Very edge case; leaving as a future follow-up.
- The unit-test path for `toPickerColour` doesn't exercise the canvas fallback because happy-dom's canvas doesn't normalise CSS colour strings. Verified by Codex E2E per Miguel's note; the code comment explicitly documents the tradeoff.
- `readInlineStyle` + `readInlineStyleSpread` both call `readCoveredInlineStyleChars` → `readRuns` per `selectionchange`. On a normal composition element this is negligible; on the pathological 2000-nested-span element (the regression test at `inlineTextStyleRange.test.ts:406-420`) it's two full iterative walks per selectionchange during a native-picker drag. Not a bug — the walker is iterative and cheap in the common case.

## What I didn't verify

- Did not re-run the 3,791 Studio tests locally; trusting Miguel's green + the extended coverage for each fix.
- Did not construct a live browser session against a real canvas to verify `toPickerColour("transparent")` and `toPickerColour("currentColor")` — the fallback path handles both via canvas normalisation, but the resulting picker colour for `transparent` in production browsers is not asserted anywhere. Fine to defer.
- Did not run the Codex `rich-text-demo` E2E — trusting Miguel's report of the mixed-swatch, border-box, picker-value, selection-retention, source-persistence, highlight-cleared, and no-preview-reload assertions.

## Verdict

LGTM from my side. Six commits, one concern each, tests to match. The rich-text stack lands cleanly with this. Not stamping.

vanceingalls (approved):
_R1 review of `03e8f6d0d1` → `a9452d5e7b` — the split-refactor amendment landed while I was reading. Adversarial pass on the 4 review fixes + rebase-across-merge preservation._

## What holds up

- *Distinct-color ordering, one owner.* `readInlineStyleSpread` (now `inlineTextStyleRead.ts:22-40`) emits deduped in document order via `Set<string>`. Renderer at `InlineTextToolbar.tsx:163-173` (`swatchBackground`) consumes it as-is — no `new Set()`, no `filter((c,i,a) => a.indexOf(c) === i)`, no second sort. Direction of authority matches the claim; can't silently drift.

- *CSS-color normalization for native picker.* `toPickerColour` at `InlineTextToolbar.tsx:258-280` delegates to `parseCssColor` (`colorValue.ts`), then falls back to a canvas `fillStyle` round-trip for anything the parser rejects:
  - shorthand `#f00` → `parseCssColor:39-48` → `#ff0000`
  - functional `rgb(...)` → `colorValue.ts:60-70`
  - `transparent` → `:35-37` (α=0 → `toHexColor` drops alpha → `#000000`; picker has no alpha, degrades honestly)
  - named (`red`, `lime`) → parser returns null → canvas normalises to `rgb(r,g,b)` → parser re-runs → hex
  Defensively guarded (`if (!context) return DEFAULT_COLOR;`).

- *Selection-crossing-boundary cleanup.* `clearSelectionWithin` at `useInlineTextEdit.ts:75-81`:
  ```
  if (!element.contains(range.startContainer) && !element.contains(range.endContainer)) return;
  selection.removeAllRanges();
  ```
  Truth table checks out — both-outside → no-op, either-inside → clear. Test coverage complete for the three shapes: highlight-inside drops (`.test.tsx` diff L59-77), outside-only preserved (L79-95), boundary-crossing drops (L97-118).

- *Rebase-across-`main`-merge — #3143 hardening preserved.* Grepped each construct at HEAD:
  - Drag/drop plain-text rail — `useInlineTextEdit.ts:216-223` (`onDragOver` preventDefault + `onDrop` reads `text/plain`)
  - RAF cancel on teardown — `useInlineTextEdit.ts:102-105`; unmount fires teardown via `:242-247`
  - Zero-range guard — `InlineTextToolbar.tsx:63` (`if (!selection || selection.rangeCount === 0) return;`)
  - `outlineOffset` snapshot/restore round-trip — `:44` (snapshot), `:132` (open save), `:139` (edit set), `:114` (restore in teardown)
  Nothing silently dropped by conflict resolution.

- *135° sweep with evenly-spaced stops.* `InlineTextToolbar.tsx:169-172`: `((index / (distinctColours.length - 1)) * 100).toFixed(2)%`. Every distinct color gets equal weight regardless of char count — a single-char color is as visible as a 30-char one. Length-1 short-circuits to plain swatch (`:168`); length-0 falls back to `agreed || DEFAULT_COLOR` (`:167`). Tests cover all three arms.

- *Border-box gradient anchoring.* `InlineTextToolbar.tsx:111-114` sets `background` (shorthand — resets `background-origin` to `padding-box`) then `backgroundOrigin: "border-box"` in the *next* key so it wins the reset. Insertion-order matters here because React applies inline-style keys in declaration order — the code comment cites the specific "green edge, red edge" regression this prevents. Clean.

## CI drama (both resolved at `a9452d5e7b`)

- `File size check` at `03e8f6d0d1`: `inlineTextStyleRange.ts` was 635 lines vs the 600 gate. Fixed by `refactor(studio): split inline text style readers` — readers moved to `inlineTextStyleRead.ts` (47 lines) via a new exported `readCoveredInlineStyleChars` primitive. `inlineTextStyleRange.ts` is now 593. Verified byte-identical bodies for `readInlineStyle` + `readInlineStyleSpread` — pure carve, no behavior change.
- `Test` at `03e8f6d0d1` was supersession: Producer unit/integration jobs on run 31471744814 CANCELLED when `a9452d5e7b` was pushed, so the "Producer unit/integration tests did not succeed" gate fired. Fresh CI run in progress at HEAD; no re-fire in current rollup.

## Non-blocker follow-ups

1. **ZWSP U+200B not stripped.** `inlineTextStyleRead.ts:32` uses `if (!char.trim()) continue;` — strips SP/TAB/LF/NBSP (all no-ink), but `String.prototype.trim()` does NOT strip zero-width space. A colored ZWSP would contribute to the swatch though it paints no glyph. Very rare authored char; follow-up.

2. **Firefox multi-range selection edge case.** `clearSelectionWithin` inspects only `getRangeAt(0)`. A Firefox multi-range selection with one range inside and another outside would nuke both. Mainstream browsers keep to one range so I'd file this as a follow-up.

3. **Test coverage on the color-normalization contract.** The picker path has a hex-only test; the `rgb(...)`, `transparent`, and named-color paths (where the canvas fallback lives) go through `toPickerColour` unexercised. Not blocking — happy-dom's canvas is unreliable enough that the code degrades rather than asserts — but a `describe("toPickerColour")` block with one case per input shape would fingerprint any future normaliser regression.

## Cross-reviewer

Rames posted `COMMENTED` at 08:14:38Z on this same head with independent verification of the same 4 promises + confirmation that his adversarial findings against `03e8f6d0` survive the split move unchanged. Two-reviewer convergence.

## Verdict

APPROVE at `a9452d5e7b`. Rebase-across-`main`-merge was clean: all four review fixes verified end-to-end (owner-of-distinct-order, color normalization, selection-crossing cleanup, `outline-offset` unchanged), and every #3143 hardening construct is still at its post — no silent drops from conflict resolution. The 635-line file-size fault was the single real CI blocker and is already fixed in the split commit; `Test` FAILURE was push-cancellation. Rich-text stack 5/5 ships. Merge on the remaining IN_PROGRESS required checks landing green.

— Via

PR #3149, opened 2026-08-10
feat(scripts): typecheck the scripts directory

miguel-heygen:
## What

Adds a TypeScript project for `scripts/`, wires it into CI and the pre-commit typecheck, and fixes the errors that turning it on found. Also lifts a capture setup the two preview generators had drifted into sharing.

Stacked on #3148.

## Why

`scripts/` was the one TypeScript surface nothing typechecked. Its errors only surfaced when a script was run, which for the preview generators means during a render, long after the change that caused them.

Turning it on found real defects rather than style:

- a preview generator passing `string | undefined` where a `string` was required
- a `readdirSync` result indexed without a bound
- two non-null assertions standing in for a filter that could have narrowed the type instead

## How

The project uses the same `strict` and `noUncheckedIndexedAccess` settings as the packages, so the bar is the repo's rather than a lower one invented for scripts. It covers every `.ts` under `scripts/` with no exclusions.

`@hyperframes/producer` becomes a dev dependency because `generate-template-previews.ts` imports it by package name. Without that the project resolves on a machine with a warm `node_modules` and fails in CI, which is how it failed the first time this was attempted.

The two preview generators had converged on the same capture setup, down to the comment explaining why the capture is opaque rather than transparent. That lifts into `scripts/preview-capture.ts`, so the reason is written once and both callers still own the handles they have to close.

## Test plan

`bun run typecheck:scripts` exits 0, `bun run test:scripts` exits 0, `bun install --frozen-lockfile` exits 0, lint and format are clean. The gate runs in the CI Typecheck job and in the lefthook pre-commit typecheck, so it fails before a push as well as after one.

**Correcting an earlier version of this PR.** It claimed two files were excluded from the project, each for its own reason. There are no exclusions, `tsc --listFiles` shows both files in the program, and the "Not covered" note built on that claim was wrong in the opposite direction. It also added `@hyperframes/core` as a dev dependency to justify an import that this same PR rewrites to a source path; nothing imports it by package name, so it has been dropped.

**Not covered.** No test asserts the gate fails on a type error; that is the CI job's own behaviour. The extraction into `preview-capture.ts` normalises `generate-template-previews.ts` from `fps: 30` to the rational form the session options require, which is a behaviour change inside what is otherwise a pure move.


vanceingalls (approved):
Adversarial review of PR #3149 @ `692b6158de70af200c16b1673893b8b58ddc69d6`.

### Verdict: APPROVE

Genuine coverage upgrade. `scripts/` was the last TS surface with no typecheck; turning it on found real defects (unbounded `readdirSync` index, `string | undefined` where `string` was required, non-null assertions covering for missing type-guards), and each is fixed at the type level rather than suppressed. **Zero** `@ts-ignore` / `@ts-expect-error` / `@ts-nocheck` / `as any` introduced in the diff — grep confirms. The one place `unknown` appears is a callback param that lets `.filter((item: unknown): item is string => …)` narrow, which is stricter, not looser.

### Wiring checks

- `scripts/tsconfig.json` uses the same `strict: true` + `noUncheckedIndexedAccess: true` as `packages/producer/tsconfig.json` — no lower bar invented for scripts.
- CI: `bun run typecheck:scripts` is appended to the `Typecheck` job in `.github/workflows/ci.yml`, and `Typecheck` is one of the 8 required contexts in the branch ruleset. Gate is real.
- Pre-commit: `bunx tsc --noEmit -p scripts/tsconfig.json` appended to the `lefthook.yml` typecheck hook. Fails locally before push.
- `@hyperframes/producer` added as a dev dep — legitimate; `generate-template-previews.ts` imports it by package name and the PR body explains the previous attempt failed CI because of warm-`node_modules` masking.
- `moduleResolution: "bundler"` + `allowImportingTsExtensions: true` are consistent with `.js`-suffixed source imports (`../packages/producer/src/index.js`); safe because `noEmit: true`.
- Source-path imports (`packages/producer/src/index.js`, `packages/core/src/index.js`) match the workspace-linking rationale already used by other scripts.

### Removals

The -109 lines are cleanup, not lost coverage. Both preview generators had drifted into duplicating the same opaque-capture setup down to the comment; `scripts/preview-capture.ts` lifts it once and both callers still own their handles. The one behavior change — `generate-template-previews.ts` moves from `fps: 30` to `{ num: 30, den: 1 }` — is called out in the PR body and matches what `createCaptureSession`'s options require.

### Nits (non-blocking)

- `scripts/tsconfig.json` sets `allowJs: true` but the include (`**/*.ts`) picks up no JS. `.test.mjs` files stay outside the typecheck program by design (they're plain JS). Flag is dead; remove or add a `.js`/`.mjs` include if you actually meant to cover them.
- `set-version.ts` uses `const [path = artifact, marker] = artifact.split("#")` — `split` always returns at least one element, so the `= artifact` default is dead at runtime but appeases `noUncheckedIndexedAccess`. Fine as a pattern; worth a one-line comment.
- `lint-skills.ts:229` `parseFrontmatterYaml(match[1] ?? "")` — the regex has a capture group and `!match` is guarded above, so `match[1]` cannot actually be undefined; `?? ""` is type-only appeasement. Equivalent to a `!` here but safer style. No change requested.

### Positives

- CI gate + local hook gate wired together; can't slip past one by hitting the other.
- Refactor extracts a real, named seam (`openOpaqueCapture`) with the historical why (opaque vs `format: "png"` transparent mode) written down once.
- PR body is self-correcting — explicitly rescinds an earlier claim about excluded files and dropped an incorrect `@hyperframes/core` dev dep from that earlier attempt.

— Via

james-russo-rames-d-jusso (commented):
_Reviewed at `692b6158d`._

The core move — a `scripts/tsconfig.json` with `strict: true` + `noUncheckedIndexedAccess: true` at `scripts/tsconfig.json:1-16`, wired to `bun run typecheck:scripts` at `package.json:52`, run in CI at `.github/workflows/ci.yml:262`, and enforced pre-commit via `lefthook.yml:22` — is the right shape. This closes a real gap: scripts/ was silently exempt from typechecking, so contract-critical entry points (release, changelog, catalog generators) drifted at their unchecked-array-index seams. Extracting the shared `openOpaqueCapture` helper into `scripts/preview-capture.ts` collapses the two-copy duplication in `generate-catalog-previews.ts` + `generate-template-previews.ts` cleanly — Miga's docblock names the reason directly ("They diverged only in how they picked width and height, so keeping two copies meant the opaque-capture rationale below had to be re-derived, and re-read, every time either script changed.").

The individual `noUncheckedIndexedAccess` fixes read correctly at each site: `value.charAt(0)` swaps for `value[0]` in `changelog-weekly.ts:534` and `draft-changelog.ts:443` (both `capitalize` helpers) since `charAt` always returns a string; `parseConventionalSubject` at `draft-changelog.ts:276-277` destructures with explicit defaults; `cli-options.ts:30` uses `args[index] ?? config.fail(...)` — fail-fast is right for the caller's contract. `generate-catalog-pages.ts:406` swaps `.filter(Boolean)` for a proper `is string` type predicate, which is what silences the strict-index error. `lint-skills.ts:228` `match[1] ?? ""` is a defensive default — fine since an empty string parses as empty frontmatter, which the downstream `parsedDataError` catches.

Two concerns and a couple of small notes.

## Concerns

**Inconsistent script→package imports.** Half the scripts import `@hyperframes/producer` by package name (`generate-template-previews.ts:6-11`), half go through source paths (`generate-catalog-previews.ts:9-13` after this PR now imports `openOpaqueCapture` from `./preview-capture.js`, which in turn imports `../packages/producer/src/index.js` at `preview-capture.ts:9-15`). And `generate-registry-items.ts:24-30` deliberately switched FROM `@hyperframes/core` TO `../packages/core/src/index.js` with the docstring "bun workspace linking does not resolve for scripts outside packages/, so the package name typechecks on a machine with a warm node_modules and fails in CI." That reasoning cuts both ways: if package-name resolution fails on cold CI, why does `generate-template-previews.ts` still use the package-name form for `@hyperframes/producer`? The root `package.json` was updated at `:65` to add `@hyperframes/producer` as a workspace dep, which is presumably meant to make that package-name path work — but then the same fix should let `generate-registry-items.ts` keep using `@hyperframes/core` without switching to source. Pick one direction and apply it uniformly; the current split ("package-name for producer, source-path for core") reads like drift, not design. Cheap follow-up.

**`bun.lock` carries 12 unrelated `version: 0.7.78 → 0.7.104` bumps across every workspace.** `bun.lock:26-360` — every workspace package version is bumped by the same delta. This isn't in the PR title ("typecheck the scripts directory") and doesn't obviously relate to the typecheck change. Two possibilities: (a) a version-stamp propagated by the release process and accidentally staged into this PR; (b) an intentional rebase-in of main's version bump that should be a separate commit. Worth naming in the PR body or splitting into its own commit — a reader looking at the diff shouldn't have to figure out whether these bumps are load-bearing for the typecheck change.

## Nits

- `scripts/tsconfig.json:12` has `"allowJs": true` but `include: ["**/*.ts"]` at `:14`. The `.js` allowance is unused for the include glob. Not harmful; small.
- `scripts/tsconfig.json` doesn't set `types` — will pick up global `@types/node` from the root. Fine for now, but if a subset of scripts want a narrower type env (e.g. no Node globals for a pure-string transform), they'll need per-file overrides. Follow-up if it comes up.
- `lefthook.yml:22` chains three `cd`+`tsc` invocations with `&&`. Reads fine for now, but if a fourth typecheck lands (e.g. per-package), consider a script file to keep the hook readable.
- `parseConventionalSubject` at `draft-changelog.ts:276-277` — the destructuring is `[, type = "other", scope, breaking, summary = subject]`. The `= "other"` fallback for `match[1]` fires only if the regex matched but the capture group is empty, which for `/^(feat|fix|...)/` shouldn't be reachable. Harmless; naming it so a future reader doesn't wonder whether "other" is a real subject type flowing downstream.

## What lands cleanly

- `scripts/tsconfig.json` with `noUncheckedIndexedAccess: true` — the load-bearing win.
- CI + lefthook + package.json script wiring, all three levels covered.
- `preview-capture.ts` extraction — one place for the capture setup, one place for the opaque-capture rationale.
- Type-predicate at `generate-catalog-pages.ts:406` — proper `is string` narrowing rather than a `.filter(Boolean)` that TS can't refine.
- All the small `charAt(0)`/`??`/destructure-with-default swaps read as the correct minimum change per site — no over-refactor, no downstream signature widening.

LGTM from my side; the `bun.lock` bumps and the producer-vs-core import inconsistency are the two things worth naming before landing.

vanceingalls (dismissed):
R2 delta-verify of PR #3149 @ `bf0fa1915843` (was `692b6158de` at R1). Delta commit: `fix(scripts): use source imports consistently`.

### Verdict: APPROVE

Clean follow-up. The one delta commit does exactly what its subject says and no more, and it retires two of my earlier "wiring" observations without introducing new surface.

### Claims verified

- *"uses the source-import convention consistently"* — Verified. `scripts/generate-template-previews.ts` now imports from `../packages/producer/src/index.js` with the "Import from source — scripts are typechecked before workspace packages are built." comment, matching what `scripts/generate-catalog-previews.ts` was already doing. Both preview generators are now on the same convention; no `@hyperframes/producer` package-name imports remain in `scripts/`.
- *"removes unrelated lockfile churn"* — Verified at the net-PR level. `git diff 4db0a34e84ee...bf0fa1915` (base…head) touches 15 files, none of them `bun.lock`. The R1 head carried a `bun.lock` delta (14+, 13−) because `@hyperframes/producer` had been added as a devDep to make the package-name import resolve; the source-import switch retires that devDep (net `package.json` delta is just the added `typecheck:scripts` script) and the lockfile change falls away with it. Intermediate commits touch `bun.lock`, but the final PR diff does not.

### R1 nits (all non-blocking, all still standing — file contents unchanged between the two heads)

- *P3-1 `scripts/tsconfig.json` `allowJs: true` with `.ts`-only include* — **UNCHANGED**. `diff -u` of the file at `692b6158de` vs `bf0fa1915` is empty. Flag still present; include still `**/*.ts`. Worth noting on re-read: `.ts` scripts import `./registry-target-paths.mjs` with no `.d.mts`, so `allowJs` is arguably load-bearing there for the referenced-module resolution path; my R1 "dead flag" phrasing was too strong. Either way, non-blocking and unchanged.
- *P3-2 `scripts/set-version.ts:270,282` `= artifact` default on split destructure* — **UNCHANGED**. Identical at both heads. Original take stands: the default is `noUncheckedIndexedAccess` appeasement, fine as a pattern.
- *P3-3 `scripts/lint-skills.ts:228` `match[1] ?? ""`* — **UNCHANGED**. Identical at both heads. Original take stands: type-only appeasement; safe.

### New findings

None. Delta touches only `scripts/generate-template-previews.ts`, `package.json` (single devDep removal), and `bun.lock` (natural consequence of the devDep removal, cancels out at PR net).

### Positives

- Two convergent script entry points now share both the extracted `openOpaqueCapture` helper *and* the source-import convention — the "same-shape callers, same imports" invariant holds.
- Retiring the `@hyperframes/producer` devDep removes the warm-`node_modules` gotcha the PR body originally called out — the source path resolves without any package linking.

— Via

james-russo-rames-d-jusso (commented):
_Reviewed at `bf0fa19158` (delta from `692b6158d`)._

Both R1 concerns closed cleanly.

**Import convention now uniform** — every `scripts/*.ts` reaches into packages via source paths: `../packages/core/src/*.js` and `../packages/producer/src/*.js`, no package-name imports left. `generate-template-previews.ts:37` swapped from `@hyperframes/producer` to `../packages/producer/src/index.js`, matching what `generate-registry-items.ts` already did in R1 and what `preview-capture.ts` established in the new helper. The `@hyperframes/producer` workspace dep added at root in R1 is dropped from `package.json`, so the direction is: scripts read source, they don't participate in workspace resolution. Cheaper and matches the reason `generate-registry-items.ts` cited on R1 ("bun workspace linking does not resolve for scripts outside packages/").

**bun.lock churn reverted** — the 12 `0.7.78 → 0.7.104` version bumps that were in R1 across every workspace are gone; `bun.lock` now only carries the `@hyperframes/producer` workspace-dep removal (one hunk, ~10 lines). Own-scope of `#3149` (against `pr-3148`) touches no `bun.lock` at all — clean carve-out.

Everything else that landed in R1 (the `scripts/tsconfig.json` with `noUncheckedIndexedAccess: true`, the CI + lefthook + package.json wiring, the `preview-capture.ts` extraction, the per-site strict-index fixes) is unchanged and correct as reviewed before.

PR #3154, opened 2026-08-10
docs(changelog): weekly digest 2026-08-03–2026-08-10

jrusso1020:
## What

Adds the weekly digest entry for August 3 to August 10, 2026 to `docs/weekly-updates.mdx`.

Generated with `bun run changelog:weekly --from 2026-08-03 --to 2026-08-10 --write`, then rewritten to publish quality: grouped under Features, Fixes, Under the hood, and Docs, prose rewritten around user impact, and the `<!-- TODO: review -->` marker removed.

## Why

Keeps the public weekly-updates page current. The digest is the RSS-backed summary readers get between versioned releases.

This was a 199-commit week across 15 releases (v0.7.91 to v0.7.105), and the generator's own draft surfaced 10 items, so the rewrite is substantially wider. The draft also put two internal tooling commits in Highlights, split the router story across three bullets that supersede one another, and covered none of the week's 70 fixes or 76 docs commits. It was used as a commit index, not a shortlist.

## How

- Derived the real shape from `git log --no-merges` grouped by type and scope, then wrote from that.
- New `<Update>` block: 7 Features, 10 Fixes, 4 Under the hood, 5 Docs bullets, citing 109 commits and 45 pull requests.
- Headline items: the parallel drawElement router back on for every install; catalog search that can rank by meaning on-device; the ffprobe filename bug class closed at all nine call sites; path redaction widened from an allowlist to any path; the Studio canvas resize correctness run; and the docs site restructure.
- No SHA or PR number was hand-typed. Every link was expanded from a short-SHA token by a script that refuses to write unless each one resolves to a 40-character commit, prefix-matches, is an ancestor of `origin/main`, and lands inside the window.

### Four claims checked at HEAD rather than taken from a commit body

A `feat:` subject is not evidence that a capability is on, so each of these was read in the merged file at `origin/main`, and on a multi-commit branch the last commit was treated as authoritative.

- **The DE parallel router is genuinely default-on for everyone.** The `de-parallel-router` entry is gone from `packages/core/src/canaryRegistry.ts` (only two inert calibration entries remain) and `packages/cli/src/commands/render.ts:1228` reads "Nothing left to gate: the router is a shipped default for every install." It is described as a **restoration**, not a new speedup: the canary had cut fleet exposure from 3.13-4.25% of non-CI renders to 0.13%, so three commits in the window read as a progression that the last one reverses.
- **Catalog meaning search is opt-in, not default.** `applySearch` gates on `status.status === "ready"`, and that status requires recorded consent plus a one-time 33 MB download (`LOCAL_MODEL_SIZE_MB`). Writing it as "catalog search now matches by meaning" would have implied default behaviour, so the opt-in and the download are stated.
- **The Reference Project was added and removed inside the same window.** `edfe66a95` shipped it and `b5509602f` deleted it along with `examples/docs-reference-project/`. It survives at HEAD only as a historical line in `docs/changelog.mdx`. Announcing it as new would have pointed readers at a page that no longer exists, so the digest reports both events in one clause.
- **Counts taken from files, not subjects.** `docs/thirty-days.mdx` really does carry 30 distinct film sources, so "all thirty films" is published. `docs/examples.mdx` now has 23, not the 19 its commit subject claims, because later commits in the same window added more, so no count is published for Examples.

## Test plan

Docs-only change. No runtime code touched.

Verification performed:

- `npx mint validate` passes locally from `docs/` (`success build validation passed`), which is the check the `Validate docs` job runs.
- `git diff --stat origin/main` shows only `docs/weekly-updates.mdx`, 50 insertions and 0 deletions. Purely additive. The three prior entries are untouched and the file still holds 4 balanced `<Update>` blocks.
- All 109 commit links resolve to 40-character commits that prefix-match, are ancestors of `origin/main`, and sit inside the window. The expander refuses to write otherwise.
- All 45 cited pull requests were confirmed through the API to exist, be merged, and carry a title matching the claim they are cited for. Every number was taken from an inline `(#N)` in a real in-window commit subject rather than from merge ancestry, which is unreliable for squash merges.
- Confirmed the block contains no em-dashes, no TODO marker, no unexpanded tokens, balanced inline-code backticks, and no JSX tag other than `Update`.

The generator also writes `updates/weekly/` and `updates/social/`. Those are left uncommitted, matching the precedent set by the previous digests.

*Digest compiled by Rames Jusso.*


miga-heygen (commented):
Docs-only, single file. Scanned the digest against the PRs I reviewed this cycle — the claims hold up:

- Router restoration framing is correct (canary cut exposure, not ramped it; removal is a restoration, not a new feature).
- Catalog meaning search correctly described as opt-in behind a consented download.
- Caption colour state declaration, font-family comma splitting, canary reason emission, capture mode mislabelling, silent drawElement refusals — all match what shipped.
- Reference Project in-and-out within the window is reported as both, not announced as new. Good editorial call.

Video is embedded, CDN-verified per the thread. CI green.

LGTM.

_Review by Miga_

miguel-heygen (approved):
Exact-head docs review at `a11fc9da12f96d66c180a9194bd2897fcd9cecaf`:

- Audited all 109 linked commits: every object exists, is on current `main`, and has a committer date inside the August 3–10 window.
- Rechecked the four counterintuitive summaries against source: drawElement is default-on again after the 5% canary cut exposure to 0.13%; semantic catalog ranking runs only after the consented 33 MB local model is ready; the Reference Project was both added and removed in-window; and the docs contain 30 Thirty Days films versus 23 Examples films, so omitting the latter count is correct.
- Verified the release boundary from remote tags: `v0.7.90`, then `v0.7.92` through `v0.7.105` — 15 tags, with unpublished `v0.7.91` correctly excluded.
- Verified the embedded CDN asset directly: 41,930,129-byte MP4, H.264 1080×1080/30 fps plus AAC stereo, 45.1 s.

Miga’s cycle-specific pass and Rames’s source reconciliation converge with this independent audit. No blocking findings.

**Verdict:** APPROVE
**Reasoning:** The digest is internally coherent, its deliberately non-obvious claims match the shipped code/history, its references resolve to the stated weekly window, and exact-head CI is terminal-green.

— Magi

PR #2190, opened 2026-07-11
fix(studio): debounce flat slider drag commits to prevent rapid-fire writes

vanceingalls:
## Summary

`FlatSlider`'s drag support (added in the prior stacked PR) called `onCommit` synchronously on every `pointermove`/`pointerdown`, with no debouncing — unlike the legacy `SliderControl`, which coalesces rapid drag input into a single debounced commit. This stack rewires `FlatSlider` to match that established pattern:

- Local `draft` state drives instant knob position / `aria-valuenow` feedback on every pointer event (no lag).
- Actual `onCommit` calls are debounced 40ms during drag and flushed immediately on `pointerup`/release, deduped against the current committed value (skips a no-op commit).
- Fixed a real bug found while wiring this up: `onPointerUp` previously read the `draft` closure, which can be stale if `pointerdown`+`pointerup` land in the same React batch (e.g. a very fast click) — it now recomputes the release value fresh from the event's own `clientX`.

This is a defensive fix for a reported issue ("live preview shows the wrong value while dragging; the correct value only appears after deselecting," reproduced by the reporter on Layer Blur). I was not able to reproduce the exact symptom myself despite extensive scripted browser testing, but the un-debounced commit-per-pointermove is a plausible root cause and this brings `FlatSlider` in line with the codebase's existing debounce pattern for exactly this class of bug.

## Test plan

- [x] `bunx vitest run` — full studio suite passes (1585 passed)
- [x] `bunx oxlint` / `bunx oxfmt --check` on all touched files — clean
- [ ] Please re-verify the original repro (Layer Blur slider drag on an image/video clip) live — this fix is well-reasoned but not confirmed against your exact repro

PR #3156, opened 2026-08-10
fix(audio): code-review fixes — clip-before-duck, phaser lanes, rate reschedule

vanceingalls:
Code-review fixes for the audio FX stack. 35 commits, and **several change
rendered audio** — see below before reviewing this as a cleanup pass.

## What changes what you hear

- **`55f852a72` the FX output is ducked before it is quantised, not after.**
  The old order clipped the chain's output to ±1 and *then* applied the volume
  lane, so a hot chain lost everything above full scale before the duck brought
  it back down. Measured on real narration: the level-matched difference is
  audible distortion, not rounding. Written up with the numbers in
  `plans/` and reproducible from `~/audio-fx-clip-ab/`.
- **`accac9984` the phaser's trim lanes aim at the trims.** They were pointed at
  a wet/dry pair the builder pins to 1, so an envelope modulated a constant and
  the trim it was meant to move stayed frozen — "fade the phaser out" left the
  dry leg at full level.
- **`5e166ad47` FX automation reschedules when the transport changes rate.**
  Lanes are committed to absolute context times, so a lowpass sweeping over ten
  clip-seconds kept its wall-clock plan while the audio underneath ran at 2×.
- **`f6f0e4cf2` the volume lane is sampled at clip-local time**, the way the
  render does it.

## Correctness fixes with no audible change

`eb60133e3` retires a worklet processor on dispose instead of only
disconnecting it — they live until `process()` returns false, so every chain
rebuild that dropped a limiter left one running on the audio thread for the
session. `05d081c61` disposes a clip's graph when it ends. `85257571c` lets a
failed worklet registration be retried. `b754e72d0` moves a node's id with its
slot on an in-place update, so a lane cannot end up addressing its neighbour.
`eb0080dec` stops a blank parameter reading as zero. `c9448aec0` moves a
track's PCM across the wire in bounded chunks (a long track was ~184 MB in one
`page.evaluate`). `43c88961b`, `714f662ed`, `bfa27381e`, `b1faacb3a` are engine
robustness: an empty track, the browser lease, the ffmpeg abort signal, and the
audio failure that used to take the whole mix down silently.

## Studio

`fddd596e8` keys FX rows by node id — on `${type}-${index}` two effects of the
same type keep their keys through a reorder, so a half-typed number or an
in-flight drag lands on whichever effect moved into that slot. `436db2506` lets
a parameter be typed and stops a bare blur from re-persisting. `371920695`
locks the rack while the carve measures.

## Cleanups

Four of the nine listed in the review: `ac0e37247`, `ce7c9e486`, `5a2247435`.
The rest have written reasons in `plans/audio-review-fixes-handoff.md` §4c —
the largest is blocked behind the FlatSlider throttle question in #2190.
Lining up the three enabled-node call sites was not cosmetic: it exposed that
`buildFxChain`'s filter had **no test**, and its failure is silent, because a
bypassed node shifts every later node's parameters into its neighbour with an
identical shape either way. Now covered.

## Tests

core 1721 (110 files) · studio 3674 + 18 todo · engine services 736 + 3 skipped
· producer 534 vitest + 203 bun.

Commit bodies carry the falsification notes — each test was broken on purpose
and confirmed to fail.

## Reviewing

`plans/audio-review-fixes-handoff.md` is the map: §4b the two items the
reviewer's cap cut, §4c the cleanup ledger with reasons, §8 the fixture.

One pre-existing failure, **not from this branch**:
`producer src/services/coreRuntimeBrowser.test.ts > removes the control bridge
during teardown` times out at 5 s and does so identically with core and engine
checked out before this work.

### DIFF
--- a/.fallowrc.jsonc
+++ b/.fallowrc.jsonc
@@ -149,13 +149,6 @@
       "file": "packages/studio/src/utils/studioHelpers.ts",
       "exports": ["resolveDroppedAssetDimensions"],
     },
-    // Audio FX worklets sit near the bottom of the audio stack: the worklet
-    // source and its test reset are consumed by the runtime and engine PRs
-    // upstack, so a per-PR audit against the merge base sees them as unused.
-    {
-      "file": "packages/core/src/audio/audioFxWorklets.ts",
-      "exports": ["AUDIO_FX_WORKLET_SOURCE", "__resetAudioFxWorkletsForTests"],
-    },
     {
       "file": "packages/core/src/audio/audioFxGraph.ts",
       "exports": ["ensureAudioFxWorklets"],
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -549,7 +549,7 @@
   },
   "scripts": {
     "build": "bun run build:hyperframes-runtime && bun run build:position-edits-render && bun run build:audio-fx-runtime && tsc && tsx scripts/rewrite-esm-extensions.ts",
-    "test": "bun run check:position-edits-render && vitest run",
+    "test": "bun run check:position-edits-render && bun run build:audio-fx-runtime && vitest run",
     "test:watch": "vitest",
     "test:coverage": "vitest run --coverage",
     "test:runtime-coverage": "vitest run --coverage src/runtime",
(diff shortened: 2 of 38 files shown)


vanceingalls:

PR #3157, opened 2026-08-10
feat(audio): FX presets, a multi-band Tone EQ, and "Even Out Levels"

vanceingalls:
Presets, a multi-band EQ and a levelling script for the audio FX rack.

The starting point was an observation, not a review finding: *voice carve is a
tool that uses the effects to give the author something they can actually
understand — find the other features like it, and ship presets so nobody has to
learn what a peaking filter is.* Colour grading's preset set was the named
precedent.

## What ships

**18 presets in 4 families** — voice 3, repair 4, character 6, space 5
(`packages/core/src/audioFxPresets.ts`). Applying one **appends** by default:
stacking a character preset onto an already-cleaned voice is a real thing to
want, and replacing silently throws work away, so the destructive option is a
separate gesture. Re-applying a preset that is already present replaces its own
nodes **in place**, because position in the chain is signal order and signal
order is audible.

**Every preset node is named for the job it does** — `"Remove Rumble"`, not
`highpass`. A chain that cuts mud and then lifts clarity must not show the same
name twice.

**Tone (EQ)** — `HF_AUDIO_EQ_3` (Bass/Middle/Treble) and `HF_AUDIO_EQ_5`
(Bass/Warmth/Middle/Clarity/Air), built as a **composite over existing
filters** rather than a new effect type, the same device the carve already
uses. Partly a design choice — users know what an EQ is, they don't know they
want five peaking filters — and partly forced: the effect registry's params are
flat key/value and cannot express a band array. The chain is authoritative;
bands are read back out of the tagged nodes on every render, so there is no
cached list to fall out of sync.

**"Even Out Levels"** — an adaptive script, the same shape as carve: measure in
an `OfflineAudioContext`, then generate an automation lane. Two things the
tests forced, both of which matter:

- **The target is the 80th percentile of the track's own speaking windows**, not
  an absolute level. The first version derived it from the peak, which pulled an
  already-even track bodily up or down — a volume change wearing a levelling
  label.
- **The lane rides a `gain` node, not the track volume**, because `VOLUME_RANGE`
  is 0..1 and can only attenuate; levelling has to be able to lift.

## Two things worth a reviewer's attention

**The leveller goes in FRONT of a trailing limiter** (`insertAt`, not append).
The likely sequence is "apply Clean Voice, then even out the levels", and Clean
Voice ends in a Peak Ceiling — appending put up to 12 dB of lift *after* the
ceiling that exists to bound the chain. A ceiling with something after it is not
a ceiling. Same bug class as the clip-before-duck fix below this PR.

**`withLane` merges by target.** A script hands back a whole `HfAutomation`
describing only its own lane; writing that to the attribute takes the carve's
per-band lanes and the track's volume lane with it — silently, and noticed only
when the mix has lost its ducking. This had no test until the mutation survived
one.

## Invariant added here

> Any new `HfAudioFxNode` field must be added to **both** `parseAudioFxChain`
> **and** `serializeAudioFxChain`.

Not hypothetical: `a533d1677` added `fromPreset` to the type and the writer but
not the parser, so the tag was silently dropped on every reload and a preset
could no longer find its own nodes. The round-trip test passed the whole time
because it compared only node *types*. Fixed in `29f53f935`; that test now
compares type, id, `fromPreset` and `label`.

## Design docs

Six commits of design work ride along, including the review page generated from
the real registry (`bun plans/audio-fx-ux/build-preview.mts /tmp/rack-ux.html`).
That build **fails** if any shipped effect, parameter or preset lacks copy, so
the page cannot quietly omit something real.

## Deliberately not built

De-esser and tone match, with reasons in `plans/audio-fx-presets.md` §6c. The
de-esser has two hard constraints — `analyseCarveDynamics`' 85–150 ms hop cannot
land on a 50–150 ms sibilant, and `MAX_AUTOMATION_POINTS` is 512 where a long
voiceover needs hundreds of events.

## Tests

core 1721 (110 files) · studio 3674 + 18 todo. Falsification notes in the
commit bodies.

### DIFF
--- a/packages/core/package-subpaths.json
+++ b/packages/core/package-subpaths.json
@@ -92,6 +92,12 @@
       "types": "./dist/audioFx.d.ts",
       "environments": ["browser", "bun", "node"]
     },
+    "./audio-fx-presets": {
+      "source": "./src/audioFxPresets.ts",
+      "runtime": "./dist/audioFxPresets.js",
+      "types": "./dist/audioFxPresets.d.ts",
+      "environments": ["browser", "bun", "node"]
+    },
     "./audio-fx-tail": {
       "source": "./src/audio/audioFxTail.ts",
       "runtime": "./dist/audio/audioFxTail.js",
--- a/packages/core/package-subpaths.json
+++ b/packages/core/package-subpaths.json
@@ -92,6 +92,12 @@
       "types": "./dist/audioFx.d.ts",
       "environments": ["browser", "bun", "node"]
     },
+    "./audio-fx-eq": {
+      "source": "./src/audioFxEq.ts",
+      "runtime": "./dist/audioFxEq.js",
+      "types": "./dist/audioFxEq.d.ts",
+      "environments": ["browser", "bun", "node"]
+    },
     "./audio-fx-presets": {
       "source": "./src/audioFxPresets.ts",
       "runtime": "./dist/audioFxPresets.js",
--- a/packages/core/package-subpaths.json
+++ b/packages/core/package-subpaths.json
@@ -98,6 +98,12 @@
       "types": "./dist/audioFxEq.d.ts",
       "environments": ["browser", "bun", "node"]
     },
+    "./audio-leveller": {
+      "source": "./src/audioLeveller.ts",
+      "runtime": "./dist/audioLeveller.js",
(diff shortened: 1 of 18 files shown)


vanceingalls:

PR #3158, opened 2026-08-10
fix(core): LFO phase, an LFO leak, one RIFF walk, and the copy layer in core

vanceingalls:
Follow-ups on the preset/EQ/levelling work: two refactors, two real bugs, and
the plain-language layer landing in core.

## The one that changes rendered audio

**`9b6c0e1bd` the chorus and phaser LFOs get a phase.**

`lfo.start()` with no argument puts an OscillatorNode at phase zero at *attach*
time. Offline that is clip-relative — the graph is built at ctx time 0 and the
source starts at 0 — but preview rebuilds the graph whenever the shape changes,
and a seek or a scrub does the same. So a chorus attached 3.6 s into a clip
started its sweep from the top there: preview disagreed with the render, and
with itself across an edit.

An OscillatorNode's phase cannot be set and `start(when)` clamps a past `when`
to now, so the modulator is now **one cycle of the waveform in a looping
`AudioBufferSourceNode`**, where `start(when, offset)` *is* a phase control.
`buildFxChain` takes the clip position it is being built at; the runtime passes
the playhead it already computes for the automation scheduler, and the render
passes nothing and gets 0 — which is what it always effectively used.

Two consequences worth naming:

- The buffer is exactly one second, so it runs at 1 Hz at the default rate and
  `playbackRate` reads directly in Hz. That is what the `speed` knob is in, so
  its automation target moves from `frequency` to `playbackRate` with no
  mapping either side.
- The phaser's waveform is baked into the buffer, so switching Triangular to
  Sinusoidal is a **shape change** now — `shapeOf` carries it, the same way it
  already carries a one-pole filter's fixed cutoff. Pushed into the running
  graph it would have been a no-op.

**It got the before/after listen**, through the real engine path, both sides,
swapping only the runtime the browser is given. Chorus differs by 76.5 dB down,
the Triangular phaser by 63.7, the Sinusoidal one by 88.1, and a bare 440 Hz
tone with nothing to mask it by 87.6. Worst single sample anywhere: 0.00027.
Inaudible, as expected — a render builds at position 0, so only the modulator's
generation changed. Numbers and harness in `~/audio-fx-lfo-ab/README.md`.

The half that motivated the fix cannot be rendered at all, since a render never
builds mid-clip, so it is asserted in tests instead.

## A leak the cleanup list pointed at

`e105b2f59` — the chorus and phaser **stopped their LFO and never disconnected
it**, so every chain rebuild that dropped a modulated effect left a modulator
still wired to the delay or the allpass bank it drove. Same shape as the worklet
leak fixed one PR down.

The §4c item also asked for the four wet/dry shells to be folded together. Not
done, deliberately: they are wired differently enough (a feedback loop, a
modulated delay, six allpass stages between two trims, a convolver) that one
factory over them needs a config surface bigger than the four dispose lines it
replaces. The leak was the part with a defect behind it.

## A parser with no test

`3fa558769` — `readWavChunks` and `parseWavLayout` each carried the same
word-aligned RIFF walk. They are **not** the same function (one breaks at the
first `data` and returns a slice, the other scans every chunk and returns
offsets; one lets the decoder judge the format, the other refuses anything but
16-bit PCM), so only the walk moves. `riffChunks` yields chunks and holds no
policy at all.

It also had **no test in either copy**. A real WAV out of the mixer has an
even-sized `fmt ` first and `data` last, so neither thing the walk exists for —
the pad byte after an odd chunk, and not assuming an ordering — ever came up.
Both mutations survived a full engine run before `wavChunks.test.ts`.

## The plain-language layer

`1c214abd9` moves `copy.mts` out of `plans/` into
`packages/core/src/audioFxCopy.ts`, and turns its completeness check into a
test: every shipped effect, every one of its parameters and every preset must
have copy. That check was a step in a build script, which means it only fired
when somebody remembered to rebuild the review page; now it fires on the commit
that adds an effect without a plain name for it.

Four assertions the build step never made, each a real hole: copy for an effect
the registry no longer ships, a missing `SUMMARY`, a summary that renders
`undefined`/`NaN` at the effect's own defaults — the first thing an author reads
after adding one — and a gap or overlap in the shared frequency ruler.

`PROFILES` deliberately did **not** come along: its figures are proposed, not
measured, and nothing derives from them yet.

## Refactors

`93cba50dc` and `f2aca8344` take `propertyPanelFxSection.tsx` from **992 lines
to 401**, under the studio's 600-line cap. Pure moves, no behaviour change. It
took two extractions rather than the one the handoff predicted — pulling
`FxCarveModule` out only got it to 647, because the estimate counted the
component and not the four helpers that exist solely to serve it.

## Tests

core 1745 (111 files) · engine services 739 + 3 skipped · studio 3674 + 18 todo.

## Note for whoever runs this

`packages/core/src/generated/audio-fx-runtime-inline.ts` is **gitignored and
generated** by `bun run build:audio-fx-runtime`. It is what the engine injects
into the headless browser, so a stale one hides exactly the preview/render
disagreement fixed here with every test still green.


vanceingalls:

PR #3159, opened 2026-08-10
feat(studio): make the FX rack speak the author's language

vanceingalls:
Wires the plain-language layer that landed in core into the FX rack, to three
UX decisions recorded in `plans/audio-fx-ux/README.md` §Decided.

## What the rack says now

**The plain name replaces the DSP name; the DSP name moves inside.** A module
reads "Remove Rumble" with "Cutting everything below 80 Hz" under it, and
opening it shows "Details — High-pass" above the knobs. Every knob is renamed
too — `plainDef` writes the copy over the registry's labels and hints while
leaving range, step, unit and automatability alone, because those are facts and
only the words were ever the problem. A preset's own job name still wins where
it has one: it is the most specific truth available.

**Presets read as the complaint they answer.** "My voice sounds boomy", with
`Boom Tame` underneath in small type — which is also how the name gets learned.
The four shelves stay: eighteen sentences in a column is a wall, and
Voice/Fix/Character/Space is already the author's grouping rather than the
registry's. The effect list is untouched and stays in signal order.

**`da5403026`** fixes what that first pass broke: the rack was renamed and the
add menu was not, so an author picked "High-pass" and a module called "Remove
Rumble" appeared — the exact confusion this layer exists to remove,
reintroduced one gesture upstream of it.

## Hover-audition

Hovering a preset or an offered effect plays it through the same preview channel
a slider drag already uses, so it costs no new machinery and never touches the
document. **Four things have to hold, and each has a test that fails without
it:**

1. Leaving puts the stored chain back — captured when the audition starts,
   because the preview channel does not persist and the prop never changes.
2. Applying does **not** put it back. The audition *was* the preset, and
   reverting over the write that just landed is a race the author hears as it
   arriving and then leaving again.
3. The panel going away mid-audition reverts — but the teardown must have
   **empty deps**.
4. Moving between two entries in a shelf is not leaving it.

**3 and 4 were bugs in the first commit here, both found by review rather than
by tests, and `8639f3076` fixes them.** The teardown was keyed on
`onChainPreview`, an inline arrow from a component that re-renders on every
playhead tick — so the audition reverted itself about thirty times a second,
during playback, which is the only time there is anything to audition. And
sliding from "Even Out Levels" to the effect beside it kept a measurement
running that then landed on top of the audition next to it. Both are the kind a
static mount cannot see.

**Even Out Levels auditions by measuring first** — it decodes the track and
analyses it before there is anything to play, so it says "measuring…" while it
works, caches the decode per `src`, and a run counter drops any result that
arrives after the pointer has left. Reverting puts back **both** the chain and
the automation, because levelling is a node AND the lane that drives it.

Tone deliberately does not audition: its bands arrive at 0 dB, so there is
nothing to hear, and a hover that changes nothing teaches that hovering does
nothing.

## Tests

studio 3685 + 18 todo. Tests that addressed rows by their registry labels now
look them up through `EFFECT_COPY`, and preset buttons by preset id — those
tests are about which row carries the automate button, not about how it reads,
and hard-coding the words would make every copy edit a test edit.

### DIFF
--- a/packages/studio/src/components/editor/propertyPanelAudioFxGroup.test.tsx
+++ b/packages/studio/src/components/editor/propertyPanelAudioFxGroup.test.tsx
@@ -366,23 +366,21 @@ describe("AudioFxGroup dynamic carve", () => {
     );
   }
 
-  afterEach(() => vi.unstubAllGlobals());
-
   /**
-   * Hover-auditioning the leveller has to measure before there is anything to
-   * hear, and measuring a long voiceover takes seconds — by which time the
-   * pointer has usually moved on. Applying then would put levelling on a track
-   * nobody asked to level, through a channel that does not persist: audible,
-   * absent from the document, and gone on the next reload.
+   * The same voice, but the decode does not finish until it is let go.
+   *
+   * Hover-auditioning the leveller is the one path where the result can arrive
+   * after the author has moved on, so the tests that cover that need to hold the
+   * decode open across a second gesture.
    */
-  it("drops a levelling measurement that lands after the pointer has gone", async () => {
+  function stubGatedDecode(): { release: () => void; decoded: Promise<void> } {
     const sampleRate = 48000;
     const data = new Float32Array(sampleRate * 4);
     for (let i = 0; i < data.length; i++) {
       const t = i / sampleRate;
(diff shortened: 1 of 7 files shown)


vanceingalls:
Superseded — replaced by a slice on the new chain to keep the stack linear after the split.

PR #3160, opened 2026-08-10
feat(studio): the rack redesign — named jobs, two faces, ruler, schematic

vanceingalls:
The rack redesign the review page had been showing all along. The PR below this
wired the *language*; this is the drawing and the structural rules it serves.

Prompted by a good question — *is everything designed as seen in the preview?*
It was not: half of `EFFECT_COPY` was still unread and none of the visual
direction had been built.

## The one that changes the catalogue

**`d7d36de35` — the range IS the module.**

"Shape One Range" has three controls — where, how much, how wide — and no honest
way to nominate one as the knob that matters. The copy nominated *how much*,
which is incoherent: boosting an unspecified frequency means nothing. The range
is the first decision, not the second.

So the add menu offers the decision instead of the machine: **Tame Boominess,
Reduce Mud, Reduce Boxiness, Add Clarity, Soften Harshness**, each a peaking
filter with its frequency already chosen, ordered low to high because that is
the order an author hears them in. `peaking` is no longer offered as itself —
two doors to the same effect, one of them the incoherent one, is worse than
either alone.

**Every job is one the preset catalogue already ships, at the settings it ships
it with, and a test asserts that.** The list names the vocabulary the presets
were written in rather than inventing a second one; a job nobody's preset does
is a guess about what authors want.

It needs no new node field — `label` already names a node for the work it does,
so a job node and a preset node are indistinguishable afterwards, which is
right. It also dissolves the duplicate-name problem at the root rather than
papering it with a role label: Clean Voice reads *Remove Rumble · Reduce Mud ·
Even Out Loudness · Add Clarity · Peak Ceiling*, and nothing repeats.

## The three rules

**Two faces** (`736597572`) — a module opens on its name, a line saying what it
is for, and the single control that carries it; everything else is behind a
Details disclosure, which is also where the DSP name now lives. Nothing is
hidden, it is ordered.

Ten of fifteen effects get this from a real parameter. The other five get a
**derived** one — see below.

Worth naming: `peaking`'s primary is "how much", and that only became coherent
in the commit before this one. **Two faces is honest for it *because* the jobs
landed first**, which is why the order in this PR is the order it is.

**The shared ruler** (`04a28135a`) — `BANDS` has named seven ranges since the
copy layer landed and nothing showed them; the rack spoke entirely in Hz, which
mean nothing to somebody who has not been taught them. Every spectral module now
shows where it acts on one ruler, with the range named underneath. Log-spaced,
because hearing is: rumble is 20–80 Hz, three tenths of one percent of the range
linearly and about a fifth of it by ear, so a linear layout collapses the bottom
six bands into a sliver and teaches nothing. A test asserts that, and a linear
layout fails it.

**Family lettering and the tint step** (`05c09053f`) — a rack of eight modules
is eight lines of text, and reading it should not mean reading eight names. The
sans carries four families apart by weight, case and tracking; the serif is
spent on the one family that generates signal rather than measuring or shaping
what is there; monospace goes to the measuring modules, because what they show
IS a readout. Alongside it a tint step per module *within* its family, derived
from registry position so adding an effect never re-colours its siblings. It
sits on the card's left edge rather than the text, which already carries the
family in its lettering.

## One knob for the five that cannot have one

`29d1cff42`. A compressor has seven controls and an author wants one, but
unlike a filter or a delay no single one of them can be its face: threshold
means nothing without ratio, ratio means nothing without make-up gain. So the
knob is derived, exactly as `carveProfile` already turns one number into six.

Continuous, not the three-point tables the design proposed — a table makes
gentle/middle/strong three settings to pick between, and the thing being
modelled is one axis.

**Three of the five sets of figures were wrong, and only rendering showed it.**
Measured through the real engine path on 20 s of narration; full numbers and
method in `~/audio-fx-profiles-ab/README.md`.

- **Compressor make-up was too small, and not linear.** The proposed 1/3/7 dB
  left the track **2.5 dB quieter at full evenness** — an evenness knob that
  turns the track down as it goes up. Gain reduction accelerates as threshold
  and ratio move together, so linear was wrong too. Now `s² × 9.5`: spread
  falls 19.1 → 8.8 dB with the level unmoved.
- **Saturation's trim went the wrong way.** A soft clipper at −18 dB threshold
  *is* a limiter at −18 dB: the proposed −3 dB trim took the peak from 0.496 to
  0.089 and nearly halved RMS, so "Warmth" mostly meant "much quieter".
  Reversed; RMS now holds within 0.4% across the knob while the peak comes
  down, which is saturation rather than attenuation.
- **The gate did essentially nothing.** The design derived threshold and range
  only; at the effect's 100 ms default release it moved the gaps **0.1 dB**
  against a range asking for 30. Swept, release dominates everything else —
  0.2 dB at 140 ms, 13.4 dB at 10 ms — so it joins the profile. Gaps now fall
  29.6 dB with speech unmoved to a tenth of a dB.

Reverb and bitcrush measured correctly as proposed and are unchanged.

Worth one line for a reviewer: **the first measurement reported the gate as
working when it was doing nothing at all.** A speaking-window measure excludes
exactly the windows a gate acts on, so the gaps are measured separately.

In the panel the derived knob writes mechanism, never itself: the chain stores
what renders, and `audioFxProfileStrength` reads the knob back by inverting the
curve — the same contract `normalizeCarveSettings` has. It renders through the
ordinary `FxParamRow` rather than `FxNodeParams`, because it has no AudioParam
behind it and nothing to automate; the parameters it *sets* are automatable
individually under Details.

## The schematic

`9ae8f03fd`, translated to one column. **IN and OUT terminals** — two lines,
and they change what the rack is: without them the order reads as a list, and a
list is the one reading that makes "move up" look cosmetic when it is the most
consequential control here. **Every step numbered**, counted over what the rack
shows rather than the chain, because the carve's filters and an EQ's bands live
inside their own modules and counting raw nodes would leave the rack jumping
from 02 to 07. **A preset draws as one thing** — consecutive nodes only, since a
preset pulled apart by a reorder is no longer a unit and a bracket around the
gap would claim an adjacency the signal path does not have.

## Tests

core 1762 (113 files) · studio 3695 + 18 todo. Every non-trivial change was
falsified — the notes are in the commit bodies.

One flaky test seen once in eight full-suite runs and **not from this branch**:
`timeline performance fixture > generates an identical 50k keyframe-heavy-expanded
fixture` times out at 5.7 s under full-suite load, and passes 5/5 in isolation.
Its file imports nothing from the audio code.

## Known debt, and one thing not done

`propertyPanelFxSection.tsx` is back over the 600-line cap at 670. Extracting
its add menu was **tried and abandoned**: the shelf needs about fourteen props
to lift out, and pulling it apart splits the hover-audition wiring across two
files — which is exactly where both audition bugs in the PR below came from. If
it is done, do it with a context rather than props.

**None of this has been looked at in a running studio yet.** Everything is
unit-tested and falsified, but tests do not tell you whether the tint step is
too subtle at 32% saturation or the ruler is noise under a filter you are
dragging.

### DIFF
--- a/packages/core/package-subpaths.json
+++ b/packages/core/package-subpaths.json
@@ -98,6 +98,12 @@
       "types": "./dist/audioFxCopy.d.ts",
       "environments": ["browser", "bun", "node"]
     },
+    "./audio-fx-jobs": {
+      "source": "./src/audioFxJobs.ts",
+      "runtime": "./dist/audioFxJobs.js",
+      "types": "./dist/audioFxJobs.d.ts",
+      "environments": ["browser", "bun", "node"]
+    },
     "./audio-fx-eq": {
       "source": "./src/audioFxEq.ts",
       "runtime": "./dist/audioFxEq.js",
--- a/packages/core/package-subpaths.json
+++ b/packages/core/package-subpaths.json
@@ -98,6 +98,12 @@
       "types": "./dist/audioFxCopy.d.ts",
       "environments": ["browser", "bun", "node"]
     },
+    "./audio-fx-profiles": {
+      "source": "./src/audioFxProfiles.ts",
+      "runtime": "./dist/audioFxProfiles.js",
+      "types": "./dist/audioFxProfiles.d.ts",
+      "environments": ["browser", "bun", "node"]
+    },
     "./audio-fx-jobs": {
       "source": "./src/audioFxJobs.ts",
       "runtime": "./dist/audioFxJobs.js",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -112,6 +112,12 @@
       "import": "./src/audioFxCopy.ts",
       "types": "./src/audioFxCopy.ts"
     },
+    "./audio-fx-jobs": {
+      "bun": "./src/audioFxJobs.ts",
+      "node": "./dist/audioFxJobs.js",
+      "import": "./src/audioFxJobs.ts",
+      "types": "./src/audioFxJobs.ts"
+    },
     "./audio-fx-eq": {
       "bun": "./src/audioFxEq.ts",
       "node": "./dist/audioFxEq.js",
@@ -419,6 +425,10 @@
       "./audio-fx-copy": {
         "import": "./dist/audioFxCopy.js",
         "types": "./dist/audioFxCopy.d.ts"
+      },
+      "./audio-fx-jobs": {
+        "import": "./dist/audioFxJobs.js",
+        "types": "./dist/audioFxJobs.d.ts"
       },
       "./audio-fx-eq": {
         "import": "./dist/audioFxEq.js",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -112,6 +112,12 @@
       "import": "./src/audioFxCopy.ts",
       "types": "./src/audioFxCopy.ts"
     },
+    "./audio-fx-profiles": {
+      "bun": "./src/audioFxProfiles.ts",
+      "node": "./dist/audioFxProfiles.js",
+      "import": "./src/audioFxProfiles.ts",
+      "types": "./src/audioFxProfiles.ts"
+    },
     "./audio-fx-jobs": {
       "bun": "./src/audioFxJobs.ts",
       "node": "./dist/audioFxJobs.js",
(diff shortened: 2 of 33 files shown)


vanceingalls:

PR #3164, opened 2026-08-10
chore(docs): probe whether a new docs/public subtree deploys

miguel-heygen:
Throwaway probe, do not merge.

#3090 shipped 42 MB across 1,278 files under `docs/public/catalog` and the docs host published none of it, while `docs/public/images` at 200 KB published fine. That leaves two explanations and they need different fixes: a size or file-count limit, or new subtrees under `docs/public` never publishing at all.

This adds three small files in a new subdirectory. If they serve on the preview deployment, it is a limit and the fix is to get under it. If they do not, `docs/public` is the wrong home for previews entirely.

Closing either way once the preview answers.


miguel-heygen:
Probe complete. Result: Mintlify publishes only .json and image files from docs/public; .html, .css, .js and .txt all 404 on the deployed site. Not a size limit. Closing without merging.

PR #3166, opened 2026-08-10
chore(docs): probe live composition embedding on a docs page

miguel-heygen:
Throwaway probe. Tests which live-composition mounting strategy survives a Mintlify deploy. Will be closed without merging.


miguel-heygen:
Probe complete. Findings: (1) an external <script src> from jsDelivr executes on a docs page and registers custom elements; (2) <iframe srcDoc> renders arbitrary HTML; (3) a custom element written directly in MDX is stripped, so the player must be mounted inside the iframe; (4) with that shape a live player runs with transport controls and no hosted files. Closing without merging.

PR #3172, opened 2026-08-10
fix(engine): bound drawElement frames so a wedged renderer falls back instead of failing

vanceingalls:
**Stack 3/3** — based on #3232. Review #3231 and #3232 first; this PR's own diff is the last four files.

A single drawElement frame could kill an entire render.

On one comp (`registry/components/caption-editorial-emphasis`) `drawElementImage` returns normally and the renderer then **stops draining its task queue**: the `setTimeout(…, 0)` that schedules `toDataURL` never fires, the capture `page.evaluate` never settles, and 60 s later the stage watchdog fails the whole render. Deterministic on 152.0.7977.30, always the same frame. Full root-cause trace and six falsified hypotheses on PRINFRA-488.

### Two gaps, both closed here

**1. Nothing below the 60 s stage watchdog bounded a frame.** Adds a per-frame deadline (`HF_DE_FRAME_TIMEOUT_MS`, default 15000, 0 disables) around the **whole frame operation**, not just the `drawElementImage` call. That distinction is load-bearing: the stall surfaces at whichever page round-trip comes next, and was observed at both the capture *and* the seek's background-image decode. An earlier revision wrapped only the capture call and did not fire.

Deliberately **no per-frame screenshot fallback** for this error. Once the renderer stops scheduling it is wedged for every subsequent round-trip on that page — measured, the screenshot fallback blew the same deadline. The frame fails fast so the recovery that actually works can run.

**2. That recovery was ineligible — the real defect.** `shouldRetryViaPinnedFallback` only retried a generic capture failure when the worker count was **pinned** by the inversion or the router. A comp that engaged drawElement on the ordinary single-worker path had **no whole-render fallback at all**, so one wedged frame failed the render. A renderer stall is now retryable on any routing: the failure is a property of drawElement itself, and the retry re-renders on a fresh page via screenshot.

Reported as `de_fallback_reason: "de_renderer_stall"` so it is distinguishable from `capture_error` in telemetry, and counted per session as `deFrameTimeouts`.

### Verification

End to end on the repro:

- **before:** exit 1, `Sequential drawElement capture stalled: no frame progress for 60000ms (stuck at frame 50/240)`
- **after:** deadline fires → `re-rendering via screenshot` → `RENDER_OK`, valid 1920×1080 / 240-frame / 8.0 s MP4

Tests cover the predicate both ways (stall retries with no pinned routing; a plain capture failure still does not) plus the cross-package error match and the cancellation case. Engine 1,481 and producer 587 green — `audioPadTrim.integration` flakes only under full-suite parallel load, passes in isolation, and is green on clean HEAD too, so it is not from this change.

### Why this matters beyond one comp

This is the blocker for lifting the `backface-visibility` over-gate: five shipping registry caption components use it as a 2D flicker hack on otherwise plain 2D compositions and are clamped off fast capture for it (~1.7% of local renders). One of the four renderable ones hit this stall, which would have turned a slow render into a failed one. With this in place the failure mode becomes a fallback.

Still open on PRINFRA-488, neither blocking: the upstream Chromium filing, and the unguarded double-`drawAndEncode` in the serial paint wait (real defect, tested — it does **not** cause this stall).

Refs PRINFRA-488

### DIFF
--- a/packages/cli/src/commands/render.ts
+++ b/packages/cli/src/commands/render.ts
@@ -1514,6 +1514,7 @@ function trackRenderMetrics(
     deBlankRecaptures: perf?.drawElement?.blankRecaptures,
     deBoundaryFrames: perf?.drawElement?.boundaryFrames,
     deNcprFallbacks: perf?.drawElement?.ncprFallbacks,
+    deFrameTimeouts: perf?.drawElement?.frameTimeouts,
     compositionDurationMs,
     compositionWidth: perf?.resolution.width,
     compositionHeight: perf?.resolution.height,
--- a/packages/cli/src/telemetry/events.ts
+++ b/packages/cli/src/telemetry/events.ts
@@ -235,6 +235,7 @@ export function trackRenderComplete(
     deBlankRecaptures?: number;
     deBoundaryFrames?: number;
     deNcprFallbacks?: number;
+    deFrameTimeouts?: number;
     // "cli" when triggered by `hyperframes render` (default), "studio" when
     // triggered by a studio preview-server render (POST /api/projects/:id/render).
     source?: "cli" | "studio";
@@ -336,6 +337,7 @@ export function trackRenderComplete(
       de_blank_recaptures: props.deBlankRecaptures,
       de_boundary_frames: props.deBoundaryFrames,
       de_ncpr_fallbacks: props.deNcprFallbacks,
+      de_frame_timeouts: props.deFrameTimeouts,
       ...powerStateFields(),
       source: props.source ?? "cli",
       composition_duration_ms: props.compositionDurationMs,
--- a/packages/engine/src/services/frameCapture-frameDeadline.test.ts
+++ b/packages/engine/src/services/frameCapture-frameDeadline.test.ts
@@ -1,0 +1,45 @@
+/**
+ * Tests for the per-frame drawElement deadline (`withFrameDeadline`, PRINFRA-488).
+ *
+ * The deadline races the capture round-trip from OUTSIDE `captureFrameCore`,
+ * because puppeteer cannot abort an in-flight `page.evaluate`. That is exactly
+ * why the stall counter has to live in the `onTimeout` hook: a wedged renderer
+ * never returns, so no catch block inside the work promise ever runs. The first
+ * shipped version incremented `session.deFrameTimeouts` in that unreachable
+ * catch, so the counter — and the `CapturePerfSummary` field it feeds — read 0
+ * on every stalled render.
+ */
+
+import { describe, expect, it, vi } from "vitest";
+import { withFrameDeadline } from "./frameCapture.js";
+
+describe("withFrameDeadline", () => {
+  it("rejects with DeFrameTimeoutError and fires onTimeout when work outlives the deadline", async () => {
+    vi.useFakeTimers();
+    try {
+      const onTimeout = vi.fn();
+      // Never settles — the wedged-renderer shape.
+      const raced = withFrameDeadline(new Promise<string>(() => {}), "frame 7", 15_000, onTimeout);
+      const assertion = expect(raced).rejects.toThrow(/frame 7 exceeded 15000ms/);
+      await vi.advanceTimersByTimeAsync(15_000);
+      await assertion;
+      expect(onTimeout).toHaveBeenCalledTimes(1);
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
+  it("passes the value through and leaves onTimeout alone when work wins", async () => {
+    vi.useFakeTimers();
+    try {
+      const onTimeout = vi.fn();
+      const raced = withFrameDeadline(Promise.resolve("buffer"), "frame 7", 15_000, onTimeout);
+      await expect(raced).resolves.toBe("buffer");
+      // Past the deadline: the cleared timer must not fire late.
+      await vi.advanceTimersByTimeAsync(30_000);
+      expect(onTimeout).not.toHaveBeenCalled();
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+});
--- a/packages/engine/src/services/frameCapture.ts
+++ b/packages/engine/src/services/frameCapture.ts
@@ -3163,21 +3163,34 @@ class DeFrameTimeoutError extends Error {
   }
 }
 
-function isDeFrameTimeoutError(err: unknown): boolean {
-  return err instanceof DeFrameTimeoutError;
-}
-
 /**
  * Race `work` against a deadline. The losing promise is NOT cancellable —
  * puppeteer cannot abort an in-flight `page.evaluate` — so its rejection is
  * swallowed to avoid an unhandled rejection when it eventually settles (or
- * never does).
+ * never does). The orphaned round-trip keeps running in its Chrome worker;
+ * that worker is reclaimed by the outer retry rebuilding the page
+ * (`closeOrphanedProbeForRetry`), not by anything here.
+ *
+ * `onTimeout` fires exactly when the deadline wins, and is the ONLY place the
+ * stall is observable: because the deadline races `work` from outside, nothing
+ * inside `work` — including its own catch blocks — ever sees this error.
+ *
+ * Exported for the deadline unit test; `captureFrameToBuffer` is the only
+ * production caller.
  */
-async function withFrameDeadline<T>(work: Promise<T>, label: string, ms: number): Promise<T> {
+export async function withFrameDeadline<T>(
+  work: Promise<T>,
+  label: string,
+  ms: number,
+  onTimeout?: () => void,
+): Promise<T> {
   if (!(ms > 0)) return work;
   let timer: ReturnType<typeof setTimeout> | undefined;
   const guard = new Promise<never>((_resolve, reject) => {
-    timer = setTimeout(() => reject(new DeFrameTimeoutError(label, ms)), ms);
+    timer = setTimeout(() => {
+      onTimeout?.();
+      reject(new DeFrameTimeoutError(label, ms));
+    }, ms);
   });
   try {
     return await Promise.race([work, guard]);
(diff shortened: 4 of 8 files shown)


vanceingalls (commented):
**R1 (COMMENT)** — Per-frame drawElement deadline + producer-side widened retry. Solid diagnosis of PRINFRA-488 and a clean split between "engine bounds it" and "producer retries it". One high-severity coverage-gap non-blocker (borderline blocker depending on how much of the fleet uses disk / pipelined paths), plus smaller nits.

### High-severity non-blocker — deadline coverage gap across sibling capture entry points

`withFrameDeadline` at `packages/engine/src/services/frameCapture.ts:3170` is only applied inside `captureFrameToBuffer` (`frameCapture.ts:3419-3425`). The other two drawElement-mode entry points still call `captureFrameCore` unbounded:

- `captureFrame` (`frameCapture.ts:3377-3389`, disk-write path)
  - called at `packages/producer/src/services/render/stages/captureStage.ts:335`
  - called at `packages/engine/src/services/parallelCoordinator.ts:548`
- `captureFrameToBufferPipelined` (`frameCapture.ts:3446`, worker-encode path)
  - called at `captureStage.ts:325`
  - called at `packages/producer/src/services/render/stages/captureStreamingStage.ts:480, 501`
  - called at `parallelCoordinator.ts:501`

A stall on either the disk-write path (opt-in via `useDrawElement` / `HF_FAST_CAPTURE_3D`) or the pipelined worker-encode path (macOS-only per the header comment on `captureFrameToBufferPipelined`) still hits the 60 s render-level watchdog and fails the whole render — the exact failure mode this PR is trying to eliminate.

Compounding: the producer-side disk-capture catch at `packages/producer/src/services/renderOrchestrator.ts:3767-3775` handles only `isDrawElementVerificationError` and rethrows everything else. Even if a `DeFrameTimeoutError` did fire on the disk path, the `isDeStall → de_renderer_stall` branch (wired at `renderOrchestrator.ts:3583-3611`) is only reachable through the streaming catch. The disk-path retry contract is a subset of the streaming-path one for this class of failure.

Failure scenario: user runs a fast-capture opt-in render on Linux (disk path) OR a normal render on macOS (pipelined worker-encode path), hits the same `drawElementImage`-then-idle stall the PR body reproduces, and gets the pre-fix behavior — one wedged frame → whole render fails at 60 s. No `de_renderer_stall` telemetry, no screenshot fallback.

Recommendation: wrap `captureFrameCore` inside `captureFrame` and inside the produce-phase of `captureFrameToBufferPipelined` the same way. If the pipelined path's split (produce phase returns a `Promise<Buffer>` that the caller races with N+1) makes the wrap ergonomically painful, at minimum wire `isDeStall` into the disk-capture catch and let the pipelined path stay unbounded pending a follow-up. Either way, this is a *sibling-path retry* / *safety-net coverage-gap* case that will re-surface as PRINFRA-488 recurrences on unpaved paths.

### Non-blockers

1. Telemetry gap on the new affordance. `session.deFrameTimeouts` is added to `CapturePerfSummary` (`packages/engine/src/types.ts:348-355`) and populated at `frameCapture.ts:3325`, but not threaded into the CLI `render_completed` payload at `packages/cli/src/telemetry/events.ts:337`. The `de_renderer_stall` value is queryable through `de_fallback_reason`, but the per-session count is not — for a defect the PR is deliberately turning into a fallback, being able to graph the count separately from other `capture_error` fallbacks is high-value. Suggest adding `de_frame_timeouts: props.deFrameTimeouts` next to `de_ncpr_fallbacks` at `events.ts:337`.

2. `renderOrchestrator.ts:47-56` — the `deFallbackReason` docstring still declares values ∈ `{oom, capture_error}` for the generic-failure retry. This PR adds a third value. Stale doc.

3. `withFrameDeadline` (`frameCapture.ts:3170-3187`) — `void work.catch(() => {})` correctly avoids the unhandled-rejection when the losing promise settles, but the `page.evaluate` still runs to completion in the orphaned Chrome worker. Retry does spawn a fresh probe (via `closeOrphanedProbeForRetry` a bit further down `renderOrchestrator.ts`), which reclaims the worker — worth a one-liner in the deadline function noting that recovery depends on the outer retry rebuilding the page, not this cleanup.

4. Test coverage. The four new producer-side tests (`renderOrchestrator.test.ts:2461-2513`) exercise `shouldRetryViaPinnedFallback` in the new dimension and the cross-package error match. There is no engine-side unit test for `withFrameDeadline` itself. It is a small pure function that races two promises — a `vi.useFakeTimers` + resolvable-inner-promise test at engine level (2 cases: fires-on-timeout, does-not-fire-when-inner-resolves) would lock the behavior against future edits. Non-blocker, cheap follow-up.

### Verified
- The `session.captureMode === "drawelement"` guard on the deadline (`frameCapture.ts:3420`) is a consistent literal — grep of the file shows the same string at four other sites.
- The retry ends up mode-switched to screenshot via `cfg.useDrawElement = false` (`renderOrchestrator.ts:3799`), so the retry can't re-trigger the stall (no retry loop).
- No adjacent knob widening: `shouldRetryViaPinnedFallback` only accepts the new `isDeRendererStall` branch; existing "unpinned generic capture failure does not retry" invariant is preserved and tested at `renderOrchestrator.test.ts:2472-2483`.

### CI status
`mergeStateStatus: UNSTABLE`. Grouping check runs by name and taking the latest run per name: three earlier failures on `Preview parity`, `preview-regression`, and `regression` at `~22:26-22:27` on 2026-08-11; only `regression` has a later SUCCESS at `22:40`. The two `preview-regression`-workflow failures have no superseding SUCCESS in the rollup — either they need a re-run or the failure is real. Please push a retry / triage before this lands.

— Via

terencecho (commented):
**tai concur with Via's R1** — deadline coverage gap is real; adding one wiring-error finding + a small CI note.

### Concur (Via's R1 findings verified at head `f7a297b3`)

`withFrameDeadline` is applied only at `captureFrameToBuffer` (`packages/engine/src/services/frameCapture.ts:3419-3425`). Independent grep confirms the two sibling entry points are unbounded:

- `captureFrame` (`frameCapture.ts:3382`) — disk path; called from `captureStage.ts:335` and `parallelCoordinator.ts:548`.
- `captureFrameToBufferPipelined` (`frameCapture.ts:3446`) — worker-encode path; does not go through `captureFrameCore` at all, calls `produceDrawElementFrame` directly at `frameCapture.ts:3525` — so wrapping `captureFrameCore` alone would still leave this path uncovered. Called from `captureStreamingStage.ts:480,501`, `captureStage.ts:325`, `parallelCoordinator.ts:501`.

And `isDeStall` only reaches the producer-side widened retry through the streaming catch at `renderOrchestrator.ts:3583-3611`; the disk-capture catch at `renderOrchestrator.ts:3767-3775` handles only `isDrawElementVerificationError`. Even if a `DeFrameTimeoutError` did fire on the disk path, `de_renderer_stall` retry wouldn't kick in from that catch.

Failure envelope Via described is accurate: a wedged renderer on Linux disk fast-capture (`useDrawElement` / `HF_FAST_CAPTURE_3D`) or macOS pipelined worker-encode still hits the 60s render-level watchdog and fails the whole render.

### New differentiating finding — `session.deFrameTimeouts` counter is dead code as wired

The `isDeFrameTimeoutError(err)` branch at `frameCapture.ts:3319-3334` sits inside a catch around `captureDrawElementFrame` (line 3277), i.e. INSIDE `captureFrameCore`. But `withFrameDeadline` (line 3176) is applied by `captureFrameToBuffer` — OUTSIDE `captureFrameCore`. When the deadline fires, `Promise.race` rejects at the outer await; `captureDrawElementFrame` is not interrupted (and it can't be — puppeteer can't abort `page.evaluate`, as `withFrameDeadline`'s own docstring notes). Nothing inside the inner catch runs.

Verified: `DeFrameTimeoutError` is only ever instantiated at `frameCapture.ts:3180` inside `withFrameDeadline`; `captureDrawElementFrame` cannot raise it, so the `if (isDeFrameTimeoutError(err))` branch at line 3319 is unreachable.

Concretely:
- `session.deFrameTimeouts` (line 3326) never gets incremented.
- The engine-side "capture exceeded {ms}ms; renderer stalled after drawElementImage (PRINFRA-488). Failing the drawElement attempt so the whole render retries via screenshot." console.log never prints.
- The new `CapturePerfSummary.deFrameTimeouts` field (`types.ts:348-355`) is always 0 in `getCapturePerfSummary`.

End-to-end recovery is intact — `withFrameDeadline` throws `DeFrameTimeoutError` up to `captureFrameToBuffer`, out to `executeRenderPipeline`, and `isDeRendererStallError` at `renderOrchestrator.ts:1830` matches it by `.name === "DeFrameTimeoutError"`, so `shouldRetryViaPinnedFallback({isDeRendererStall: true})` fires and the retry runs. Only the engine-side observability path is broken.

Fix: increment `session.deFrameTimeouts` + emit the console.log inside `withFrameDeadline` (either via a session parameter, or move the counter to the setTimeout callback). This also composes cleanly with Via's telemetry recommendation (routing `de_frame_timeouts` into `render_completed` at `events.ts:337`): if the counter never increments, wiring the telemetry is pointless.

### CI

Concurring with Via's rollup read at head `f7a297b3`: dedupe by name-latest-completed shows two unresolved FAILUREs — `Preview parity` and `preview-regression` at 2026-08-11 22:27, no superseding run. The earlier `regression` FAILURE at 22:26 was superseded by a SUCCESS at 22:40. Please rerun / triage before landing.

— Review by tai (pr-review)

miga-heygen (approved):

`withFrameDeadline` correctly races `captureFrameCore` against a configurable timeout, re-throws on stall (no per-frame screenshot fallback — correct, the renderer is wedged for all subsequent frames), and swallows the orphaned promise rejection. `shouldRetryViaPinnedFallback` gains `isDeRendererStall` so the stall is retryable on ANY routing, not just pinned — the gap that left single-worker comps with no whole-render fallback. Cross-package error detection via name+message is the right approach. Tests pin all four behaviors.

**Known limitation (non-blocking, echoing Via):** `withFrameDeadline` only wraps `captureFrameToBuffer` (streaming path). The disk path (`captureFrame`) and pipelined worker-encode path (`captureFrameToBufferPipelined`) still call `captureFrameCore` unbounded — sibling-path retry gap that could re-surface PRINFRA-488 recurrences on those paths.

— Miga

PR #3173, opened 2026-08-10
fix(audio): eleven small correctness fixes across engine, core and the panel

vanceingalls:
First of four slices of the old #3156, cut to stay under 1000 lines.

Small, independent fixes with no audible change:

- an empty audio track is handed back instead of failing the mix
- the browser lease is taken inside the try that releases it
- the phaser's trim lanes aim at the trims, not the pinned wet/dry pair — an envelope was modulating a constant while the trim it was meant to move stayed frozen
- a node's id moves with its slot on an in-place chain update, so a lane cannot end up addressing its neighbour
- the volume lane is sampled at clip-local time, the way the render does it
- an FX parameter can be typed, and a bare blur no longer re-persists
- a clip's FX graph is disposed when it ends, not only when the element goes
- the rack locks while the carve measures

Each commit body carries its falsification note — the test was broken on purpose and confirmed to fail.


miga-heygen (commented):
### Review: fix(audio): eleven small correctness fixes — #3173

**Verdict: LGTM**

Eleven fixes, each independently falsified — the commit notes prove each test broke when the old behavior was restored. Highlights:

**The signal routing fix is the biggest silent one.** `effectiveSignal` now correctly propagates to `downloadToTemp`, the video extract, and the trim. The old code passed the caller's `signal` to all three, so `internalController.abort()` cancelled *nothing* — and the `rmSync(workDir)` on the next line ran while their ffmpeg children were still writing into it. This is a genuine race that could corrupt renders under cancellation.

**Browser lease ordering** — moving `acquireBrowser` inside the try that releases it is correct. A `mkdtempSync` failure between the old acquire and the try leaked a pooled browser, and a pool with no leases left hangs every subsequent render.

**Phaser trim targets** — asserting by VALUE rather than node identity is the right call. The trims are internal; the honest question is whether the param the lane drives is the one the knob sets. Clean.

**FxParamRow typing fix** — the `typing` state as TEXT (not number) is exactly right: a half-typed "-" or "" is a valid intermediate state, not a value to clamp. The `edited` ref preventing no-op commits is a nice touch — prevents a full source-patch/selection-resync/preview-reload/audio-restart cycle for a gesture that moved nothing.

**Lint rule regex** — `[^;)]` instead of `[^;]` to stop the match from crossing `)` into a later `.to()` call in a chained timeline. The false-negative direction (missing a case where a call result contains `)`) is the safe one for a rule that guesses.

No issues found.

---
_Review by Miga_

PR #3174, opened 2026-08-10
fix(engine): duck before quantising, chunk the PCM, reschedule on rate change

vanceingalls:
The three fixes in the old #3156 that **change what you hear**.

**The FX output is ducked before it is quantised, not after.** The old order clipped the chain's output to ±1 and *then* applied the volume lane, so a hot chain lost everything above full scale before the duck brought it back down. Measured on real narration: the level-matched difference is audible distortion, not rounding.

**FX automation reschedules when the transport changes rate.** Lanes are committed to absolute context times, so a lowpass sweeping over ten clip-seconds kept its wall-clock plan while the audio underneath ran at 2×.

**A track's PCM crosses the wire in bounded chunks** — a long track was ~184 MB in a single `page.evaluate`.


miga-heygen (commented):
### Review: fix(engine): duck before quantising, chunk the PCM, reschedule on rate change — #3174

**Verdict: LGTM**

Three fixes that each change what you hear, and each backed by a test that listens.

**Duck before quantising** — the architectural change. The volume envelope now travels into `applyAudioFxChain` and is multiplied onto the float samples the chain produces, BEFORE `writeWav` quantises them. The old path: FX → clamp ±1 → duck in mixer. A chain at +6 dB into a 0.5 duck: old path clips 1.8 to 1.0 then halves to 0.5 with the tops sheared off; new path ducks 1.8 to 0.9 intact, then writeWav clamps nothing. The return type change from `string` to `{ path, envelopeBaked }` is breaking but all callers are updated in this PR.

The `createEnvelopeWalker` extraction from `applyVolumeEnvelopeToWav` is clean — same O(N+M) cursor walk, same logic, just lifted for reuse. The walker is frame-outer (not plane-outer) because the cursor only advances — restarting per-channel would hand the second one the tail gain for its entire length.

The test is especially thorough: stereo with a step partway through proves (a) the envelope is walked once across all channels, not restarted per plane, (b) the crest factor confirms no clipping distortion (1.35+ vs ≤1.15 for squared-off), and (c) the step landed, so the envelope was sampled over time, not once.

**Chunked PCM transfer** — 8 MiB chunks, multiple of 4 (Float32 boundary), so chunk boundaries never split a float. Input is stored as arrays of Uint8Array page-side and reassembled before the render, then dropped before the render allocates — so the page doesn't hold two copies at once. Output is read back in the same chunks. The 45s stereo round-trip test with a monotonic ramp catches reordered/duplicated/over-read chunks anywhere in the file.

**Rate change reschedule** — `ElementFxHandle.setRate(rate)` rebases the frame at the current playhead and reschedules all params. The frame becomes mutable to prevent `elapsed` from advancing at the old rate after the change. Three tests: re-aiming, measuring later edits from the new rate, and ignoring invalid rates (0, NaN). The transport's `setRate` properly calls `source.fx?.setRate(safeRate)` on each active source.

No issues found.

---
_Review by Miga_

PR #3175, opened 2026-08-10
fix(core): retire worklet processors, retry failed registration, reuse FFT scratch

vanceingalls:
Worklet lifecycle and one perf fix.

A processor lives until `process()` returns false, so every chain rebuild that dropped a limiter, compressor, gate or bitcrush left one running on the audio thread for the rest of the session. They are now told to retire on dispose.

Also: a failed worklet registration can be retried rather than poisoning the context, a blank FX parameter no longer reads as zero, the audio failure that took the whole mix down silently is now classified, and the carve's FFT scratch arrays are reused across windows.


miga-heygen (commented):
### Review: fix(core): retire worklet processors, retry failed registration, reuse FFT scratch — #3175

**Verdict: LGTM**

Five fixes, headlined by a worklet lifecycle bug that left processors running on the audio thread forever.

**Processor retirement** — the core issue: disconnecting an AudioWorkletNode is not enough. `process()` must return `false` for the processor to be garbage-collected. All four processors (compressor, limiter, gate, bitcrush) returned `true` unconditionally, so every chain rebuild that dropped a worklet effect left it running. After a few edits to a carved bed, you'd accumulate a stack of them, each still answering `data-fx-chain` edits by rebuilding a whole graph around a dead source.

The fix is simple and correct: `{ __hfDispose: true }` message, `this.dead = true`, `process()` returns `false`. The test in `audioFxWorklets.test.ts` is creative — it loads the actual base64-encoded module source, evaluates it with a mock `AudioWorkletProcessor` base class, and tests each processor directly: keeps running before dispose, retires after, stays retired even after a later parameter update. This is the right level of test — it proves the actual processor code that runs on the audio thread, not a mock of it.

**Registration retry** — the rejected promise stayed in `registered`, so every later `ensureAudioFxWorklets` replayed the same rejection. One transient failure (slow module load, context still warming up) permanently disabled half the rack. The `.catch` now deletes from `registered` and re-throws. `readyContexts.add(ctx)` only runs on success, so callers correctly keep asking.

**Blank FX param** — `Number(null)`, `Number("")`, `Number(false)`, `Number([])` are all 0 and all pass `Number.isFinite`, so a missing or blanked value clamped to 0 instead of falling back to the declared default. A compressor whose threshold arrived as `null` sat at 0 dB and never engaged, silently. The fix: only parse strings that actually spell a number. Tests cover all the blank variants plus the string-number case (`"-30"` → -30).

**Classified audio failure** — `audioStage.ts` now synthesizes an `audioFailures` array for fatal errors (stage "internal", owner "system", retryable false, detail bounded to 2000 chars). Previously undefined, so the warning policy described a FATAL failure with less classification than a single dropped track gets.

**FFT scratch reuse** — moves `re`/`im` Float64Array allocation outside the hop loop. ~7000 hops for a 5-minute clip was ~460 MB of transient allocation. `re` is fully overwritten each window; only `im` needs clearing (`im.fill(0)`) since the FFT writes into it. The test verifies identical results for short (0.5s, few windows) and long (12s, many windows) clips with the same steady tone, plus a dynamics envelope flatness check to catch contamination between windows.

No issues found.

---
_Review by Miga_

PR #3176, opened 2026-08-10
refactor(audio): the review's cleanup list, and the design record behind the rack

vanceingalls:
The last slice of the old #3156.

Four of the nine cleanups the review listed: the non-null assertion and two casts the guards already covered, one enabled-node helper instead of three restatements, and dataset keys derived from their attributes.

Lining up the three enabled-node call sites was not cosmetic — it exposed that `buildFxChain`'s filter had **no test**, and its failure is silent: a bypassed node shifts every later node's parameters into its neighbour with an identical shape either way. Now covered.

Also carries the design record for the rack work that follows: the preset research and catalogue rationale, and the schematic direction.


miga-heygen (commented):
### Review: refactor(audio): the review's cleanup list, and the design record behind the rack — #3176

**Verdict: LGTM**

Four cleanups from the review's list plus the missing test that aligning the call sites exposed, plus a thorough design record.

**enabledAudioFxNodes unification** — three restatements of `.filter((node) => node.enabled !== false)` replaced with the canonical `enabledAudioFxNodes(chain)`. This is the SSOT fix the skill flags: same predicate in two helpers, one owner. But lining up the call sites was not cosmetic — it exposed that `buildFxChain`'s update filter had *no test*. The failure is silent and shape-preserving: a bypassed node shifts every later node's parameters into its neighbour, and the shape string is identical either way (bypassed nodes don't contribute), so nothing forces a rebuild.

The new test creates [highpass enabled, peaking disabled, lowpass enabled], updates the lowpass frequency, and verifies the highpass didn't move. Clean proof of the filter's correctness.

**Dataset key derivation** — `HF_AUDIO_FX_DATA_KEY = HF_AUDIO_FX_ATTR.slice("data-".length)` and `HF_AUDIO_AUTOMATION_DATA_KEY` likewise. The studio writes through the attribute and reads through the dataset key; the read side was carrying its own `"fx-chain"` literal and a rename would half-land. Both spellings now derive from one constant, with tests proving they match. Applied to `audioFxSummary.ts`, `propertyPanelAudioFxGroup.tsx`, `useVolumeAutomation.ts`, and `propertyPanelAutomation.ts`.

**Shape reassignment removal** — `shape = shapeOf(next)` was recomputed on every observer tick to write back the string that was already there (the early return proved it equal). Now `const`.

**Narrowing vs assertion** — `resolveAutomationRange` uses `param` directly after the `param.kind !== "number"` guard instead of casting to `HfAudioFxNumberParam`. The type system already narrowed it.

**FxNodeRow keying** — `key={node.id ?? `${node.type}-${i}`}` instead of `${node.type}-${i}`. Two effects of the same type kept their keys through a reorder, so React reused each row in place. Controls hold real state (a half-typed number, an in-flight drag), which stayed at the position and landed on whichever effect moved into that slot.

**Cross-realm instanceof fix in analyse** — `el?.tagName !== "AUDIO"` instead of `instanceof HTMLAudioElement`. The composition's elements belong to an iframe's document, so the constructor is from a different realm and `instanceof` is false for every one. Reading to plain values (`{ src, start }`) earlier eliminates the non-null assertions downstream.

**Design record** — `plans/audio-fx-presets.md` is a well-researched catalog proposal. The honesty about what the 15 effects *can't* do (no noise reduction, no pitch shift, no spectral repair) is the most valuable part — it prevents scope creep into Tier 1 territory that would need ML. The carve-as-precedent argument is strong: these scripts reuse analysis code that already ships, no new process, no new dependency, no new render-path failure mode.

No issues found.

---
_Review by Miga_

PR #3177, opened 2026-08-10
feat(core): the audio FX preset catalogue, and applying one from the rack

vanceingalls:
18 presets in 4 families — voice 3, repair 4, character 6, space 5.

Applying one **appends** by default: stacking a character preset onto an already-cleaned voice is a real thing to want, and replacing silently throws work away. Re-applying a preset already present replaces its own nodes **in place**, because position in the chain is signal order and signal order is audible.

Carries the design docs for the rack UX that follows.


miga-heygen (commented):
### Review: feat(core): the audio FX preset catalogue — #3177

**Verdict: LGTM (as load-bearing pair with #3178)**

Catalog integrity is solid: 18 presets across 4 families, all with unique ids, valid effect types, and param values checked against declared min/max ranges. The test suite is unusually thorough for preset data — checking domain invariants (limiters always last, boost-bounded presets, shelves stocked), not just structure.

**Apply/re-apply mechanics are correct.** Append-by-default is the right call for the primary workflow (stacking character on voice). Re-apply uses `findIndex` to locate the first `fromPreset` match, then reconstructs with `kept.slice(0, existing)` + new nodes + `kept.slice(existing)`. The index arithmetic is subtle but sound — everything before the first match has `fromPreset !== preset.id`, so the `kept` filter doesn't shift indices below `existing`. The ordering test (telephone → hall → re-apply telephone) proves the position-preservation.

The `steep()` helper (two cascaded 2-pole filters for 24 dB/oct) is well-documented — worth encoding once here rather than having every author rediscover the registry's 12 dB/oct-per-node limit.

**Concur with tai's load-bearing pair finding.** At #3177's isolated head, `fromPreset` is on the `HfAudioFxNode` interface but NOT in `parseAudioFxChain` or `serializeAudioFxChain` — those additions are in #3178's diff. So at this PR's head, `fromPreset` drops on save/reload, and the next re-apply silently doubles the preset. The round-trip test at line 171-176 only checks `n.type`, so it wouldn't catch this. #3178 is precisely the fix (with an upgraded round-trip test that checks `fromPreset` and `label`). **These two must merge together.**

**Nit:** `activeAudioFxPresetIds` uses `Array.includes` for a linear scan — O(N*M) for N nodes × M distinct presets. Harmless at current scale but a `Set` would be trivially better.

No blocking issues.

---
_Review by Miga_

PR #3179, opened 2026-08-10
feat(studio): the Tone module — a multi-band EQ on faders

vanceingalls:
Faders rather than the rack's horizontal sliders, because a row of them around a centre detent is what an equaliser looks like to everybody who has met one. Recognising the control is most of the value.

The chain is authoritative: bands are read back out of the tagged nodes on every render, so there is no cached list to fall out of sync.


miga-heygen (commented):
### Review: feat(studio): the Tone module — a multi-band EQ on faders — #3179

**Verdict: LGTM**

**Chain-authoritative invariant holds by design.** Bands are never cached — `readAudioEqBands(chain, eqId)` is called inline in JSX, recomputed every render. The only local state is `openEq` (UI-only) and the per-Fader `local`/`dragging` pair, which correctly resets to `band.gain` via `useEffect` when not dragging. This is the right pattern: the chain is the source of truth, the fader is a view of it.

**Fader UX is clean.** Center-detent math maps gain=0 → 50%, positive → top/boost, negative → bottom/cut. Fill bar tracks `value` (not `band.gain`) so dragging across zero flips the fill direction live. The `<input type="range">` rotated -90° is smart — preserves keyboard control, focus management, and platform pointer handling that a div-with-pointer-events would reimplement badly.

**Drag/commit split is correct.** `move()` sets `dragging=true`, updates local state, calls `onPreview`. `settle()` sets `dragging=false`, calls `onCommit(local)`. The `if (!dragging) return` guard in `settle()` prevents double-commits from `pointerup` + `blur` racing.

**Integration with #3178's EQ API** uses all the right functions with correct signatures. The `handBuilt` filter excludes `fromEq` nodes so bands don't appear twice in the rack. `removeEq` correctly iterates `chain.nodes.filter(n => n.fromEq === eqId)` to clean up per-node automation before removing.

Tests exercise real chain manipulation (not mocked): `addAudioEq` creates the right node types, band nodes are hidden from the rack list, and the preview-during-drag / commit-on-release split is verified.

No blocking issues.

---
_Review by Miga_

PR #3180, opened 2026-08-10
feat(core): the levelling script — "Even Out Levels"

vanceingalls:
An adaptive script, the same shape as carve: measure in an `OfflineAudioContext`, then generate an automation lane. Two things the tests forced, and both matter:

**The target is the 80th percentile of the track's own speaking windows**, not an absolute level. The first version derived it from the peak, which pulled an already-even track bodily up or down — a volume change wearing a levelling label.

**The lane rides a `gain` node, not the track volume**, because `VOLUME_RANGE` is 0..1 and can only attenuate; levelling has to be able to lift.

**It goes in FRONT of a trailing limiter** (`insertAt`, not append). Clean Voice ends in a Peak Ceiling, and appending put up to 12 dB of lift *after* the ceiling that exists to bound the chain.


miga-heygen (commented):
### Review: feat(core): the levelling script — "Even Out Levels" — #3180

**Verdict: LGTM**

All three PR-body invariants are enforced in code AND tested with counterfactual assertions that would fail under the old assumption.

**Invariant 1: 80th percentile of own speaking windows.** `sorted[Math.floor(sorted.length * 0.8)]` on `speaking` (which filters out -Infinity silence windows). The test creates a mostly-quiet track (6s at 0.08, 1s burst at 0.8, 5s at 0.08) — the 80th percentile sits at the body level (~-25 dB), NOT the burst (~-5 dB), so correction at t=3 asserts < 3 dB. Under a peak-based target, the quiet body would get ~17 dB of correction (well over 3). Clean proof.

**Invariant 2: Lane rides a gain node, not volume.** `type: "gain"` with target `fxAutomationTarget(id, "gain")`. VOLUME_RANGE is 0..1 and `normaliseEnvelope` clamps to that range, so a volume lane literally cannot express a value > 1 (unity) — lifts are impossible. The gain effect's range is -60..+12 dB, and MAX_LIFT_DB = 12 matches the effect's ceiling exactly. The test verifies `node.type === "gain"` and the lane target format.

**Invariant 3: Insert FRONT of trailing limiter.** `chain.nodes[chain.nodes.length - 1]?.type === "limiter"` → insert at `length - 1`. The test applies `voice-clean` (ending in "Peak Ceiling" limiter), runs the leveller, and asserts `types[N-1] === "limiter"` and `types[N-2] === "gain"`. The `!existing` guard on re-run uses replace-in-place (preserving position), not the insert path. Correct.

**Idempotence** via `fromLeveller` replace-in-place: reuses the existing node's `id` so the lane target remains valid across re-runs. The map `chain.nodes.map(n => n.fromLeveller ? node : n)` guarantees exactly one. Round-trip test verifies `fromLeveller` persists through serialize/parse.

**Edge cases covered:** silent/empty tracks return `[]` immediately, all-silence (no finite windows) filters to empty `speaking`, room tone below `FLOOR_BELOW_PEAK_DB = 42` is excluded from correction, already-even tracks return `[]` with "Already even" summary.

**One observation (non-blocking):** at `strength = 0`, minimum correction is 25% (`levellerProfile(0).correction = 0.25`). "Leave it alone" still corrects a quarter of the difference. This is intentional (the gentlest setting should still be perceptible) and the test asserts `> 0` at strength 0, so it's by design.

No blocking issues. Consistent with the carve pattern and the design record from #3176.

---
_Review by Miga_

PR #3181, opened 2026-08-10
refactor(studio): lift the carve out of the FX section file

vanceingalls:
`propertyPanelFxSection.tsx` was 992 lines against the studio's 600-line cap. The carve is the one part not about the chain: it owns a source picker, a strength knob and a read-only list of what the analysis produced, none of it shared with an ordinary effect row.

Pure move — no behaviour change, no rendered-audio change. Section is 647 lines after this; the effect-row extraction is the next PR.


terencecho (commented):
LGTM. Empirical MOVE verified at `881608dd39b8` — two independent `diff exit 0` runs (carve region + Section tail). Byte-identical extraction; only delta is the `export` keyword. No hook / dep-array / prop changes.

BC preserved via `export type { AudioTrackOption }` re-export in Section (consumer at `propertyPanelAudioFxGroup.tsx:62` still compiles).

Tests untouched — carve is exercised via CSS class selectors (`.hf-fx-carve*`) and text strings ("Voiceover carve") which are preserved byte-identically. No new pin test needed for a byte-identical move.

**OLD-assumption counterfactuals:** (a) hook reorder / dep-array change / prop rename — pinned by `diff exit 0`. (b) `AudioTrackOption` consumer compiles — pinned by grep + re-export. (c) DOM contract (selectors + strings) — pinned by byte-identity of the extracted region.

**CI:** `Preflight (lint + format)` red on 9 pre-existing .md files (README, `plans/*.md`, `skills/hyperframes-audio/*.md`) — NONE touched by this PR; base-branch drift. `preview-regression` red downstream of preflight. `regression` + `player-perf` + `Detect changes` green.

— Review by tai (pr-review)

miga-heygen (commented):
### Review: refactor(studio): lift the carve out of the FX section file — #3181

**Verdict: LGTM**

Pure file-extraction refactor with zero behavioral changes. Verified line-by-line:

- `carveMemberName()`, `formatParamValue()`, `paramValueWidthCh()`, `FxCarveMember`, `FxCarveModule` — all verbatim moves, only change is the required `export` keyword on `FxCarveModule`.
- `AudioTrackOption` interface re-exported from the original module to preserve the public API for external consumers.
- Import housekeeping is correct: `HfAudioFxParam`, `FxParamRow`, `formatHz` removed from old file (only used by moved code), added to new file. `defaultAudioFxParams` correctly kept in both (used independently).
- No hooks exist in the moved code — `FxCarveModule` and `FxCarveMember` are pure render functions receiving all state via props. The extraction cannot change lifecycle behavior.

Ship it.

---
_Review by Miga_

PR #3185, opened 2026-08-10
feat: offer the job, not the machine — the range IS the module

vanceingalls:
"Shape One Range" has three controls and no honest way to nominate one as the knob that matters. The copy nominated *how much*, which is incoherent: boosting an unspecified frequency means nothing. **The range is the first decision, not the second.**

So the add menu offers the decision instead of the machine: Tame Boominess, Reduce Mud, Reduce Boxiness, Add Clarity, Soften Harshness — each a peaking filter with its frequency already chosen. `peaking` is no longer offered as itself: two doors to the same effect, one of them the incoherent one, is worse than either alone.

Every job is one the preset catalogue already ships, at the settings it ships it with, and a test asserts that.

Also fixes the add menu calling an effect something different from what the rack calls it.


terencecho (commented):
**Verdict: LGTM_WITH_NITS**

Small, well-scoped PR. No new `HfAudioFxNode` field — jobs reuse the existing `label`, which sidesteps the #3177/#3212 parse+serialize allow-list class entirely. Tests directly falsify the OLD behaviors named in the commit.

**Verified PR-body claims (head `ee878f6`)**

- "Each a peaking filter with its frequency already chosen." — `audioFxJobs.ts:51-92` — all 5 jobs `type: "peaking"` with explicit `{frequency, gain, q}`.
- "`peaking` no longer offered as itself." — `propertyPanelFxSection.tsx:153` — `!HF_AUDIO_FX_JOB_TYPES.has(d.id)` excludes it from `defs`. `HF_AUDIO_FX_JOB_TYPES` = `{"peaking"}` (`audioFxJobs.ts:94-96`).
- "Every job is one the preset catalogue already ships, at the settings it ships with." — cross-checked head `audioFxPresets.ts`: `boom-tame`, `voice-clean` (Reduce Mud, Add Clarity), `voice-broadcast` (Reduce Boxiness), `harsh-tame` — every `{label, frequency, gain, q}` triple in `HF_AUDIO_FX_JOBS` matches an existing preset node byte-for-byte. Test `audioFxJobs.test.ts:50-60` pins this.
- "Needs no new node field." — verified: `HfAudioFxNode` at `audioFx.ts:793-847` unchanged; parse/serialize (`:862`, `:926`) untouched. Round-trip inherits from #3178 (`label` already whitelisted).
- "Add menu calls an effect the same thing the rack does." — `propertyPanelFxSection.tsx:566` uses `EFFECT_COPY[d.id]?.title ?? d.label`; tooltip at `:551-553` keeps registry name in brackets.

**OLD-assumption counterfactuals (new tests)**

- `audioFxJobs.test.ts` "makes an ordinary effect node" — falsified by any job that didn't run through `normalizeAudioFxParams`+registry (missing `q`, wrong label).
- "gives every job a value for every parameter" — pins the `defaultAudioFxParams(job.type) ⊕ job.params` merge; a spread that dropped defaults fails.
- "mints an id that does not collide" — falsified by a no-op `mintAudioFxNodeId` OR building the second node against `EMPTY`.
- "names jobs the preset catalogue already ships" — falsified by any job label that isn't a preset node label (guards drift when tuning either side).
- `propertyPanelFxSection.test.tsx` menu-count + `not.toContain("Shape One Range")` — falsified by OLD (peaking present, count = `HF_AUDIO_FX.length`, labels = `d.label`).
- "adds a job as an ordinary effect" — OLD had no `Reduce Mud` button; `click(null)` no-ops, `onChainChange.mock.calls[0]` undefined.

**Weak-test flag.** "shows two jobs over the same effect" mounts a chain with labels already set and asserts the rack renders each. That property is already true from #3178 (`node.label ?? def.label`), independent of this PR. Still a useful regression seatbelt for the pair, but doesn't uniquely pin THIS PR's code.

**Cross-PR coupling verdict**

- Parse/serialize allow-list: N/A — no new node field.
- `node.label ?? def.label` chain (#3178): REUSED. Jobs write to `label`; rack renders through the existing chain. No parallel path.
- `EFFECT_COPY`/`SUMMARY` coverage (#3184): unaffected — jobs are NOT new effects; they reuse `peaking`, which already has full copy. Job copy (`label`, `does`) lives in `HF_AUDIO_FX_JOBS` directly. #3184's runtime enumeration over `HF_AUDIO_FX` still passes.

**Nits**

- Job tooltip is `job.does` only; effect tooltip includes registry name in `(...)`. Consistent with "jobs are the exposed name," but worth noting.
- `HF_AUDIO_FX_JOB_TYPES` currently degenerates to `{"peaking"}`; comment could name that explicitly.

**CI status**

- Preflight FAIL: `oxfmt --check` on 7-9 `.md` files under `plans/*` and `skills/hyperframes-audio/*`. Zero new-in-diff `.ts`/`.tsx` files in the failing list. Standard stack-wide drift — not this PR's problem.
- `player-perf`, `preview-regression`, `regression` FAIL: outer aggregators marking themselves failed because their shards were SKIPPED by `Detect changes`. Stack-wide reporting artifact.

— Review by tai (pr-review)

miga-heygen (commented):
### Review: feat: offer the job, not the machine — the range IS the module — #3185

**Verdict: LGTM**

The core design insight: a peaking EQ filter has three interdependent controls (frequency, gain, Q) and none works as a meaningful single knob. This PR replaces the raw "peaking" entry in the add-effect menu with five intent-named jobs (Tame Boominess, Reduce Mud, Reduce Boxiness, Add Clarity, Soften Harshness), each pre-seeded with the right frequency/gain/Q. Picking the module IS picking the range — Details still opens to all three controls.

Implementation follows every existing pattern: `audioFxJobNode` uses the same `mintAudioFxNodeId`/`normalizeAudioFxParams`/`defaultAudioFxParams` pipeline as preset application. Menu construction puts jobs before defs within each group (recommended choices first). The `withJob`/`addJob` callbacks mirror `withEffect`/`addEffect` exactly.

The integrity test is clever: every job label must appear in the preset catalogue, preventing the job list from diverging from the vocabulary presets actually use. Range bounds test ensures every job has a value for every parameter its effect declares.

No API breakage — a job node is just a regular peaking node with a `label`, indistinguishable from what presets already produce.

---
_Review by Miga_

PR #3204, opened 2026-08-11
feat(cli): forward a reported catalog gap to the feedback channel

miguel-heygen:
## What

`hyperframes feedback --search-miss` now also POSTs the gap to the backend, which forwards it to the same Slack channel a CLI rating goes to. PostHog still records the event unchanged.

## Why

A reported gap is the one search report that is meant to leave the machine, and it names a move the catalog does not have yet. Until now it only became a PostHog event, which nobody watches, so a gap sat unread until someone thought to query for it. Ratings already reach a channel by exactly this path, and gaps are read by the same people.

## How

`submitCatalogSearchMiss` mirrors `submitFeedback`: same base URL, same `heygen_route: canary` header, same 5s timeout, same field caps matched to the backend DTO, same swallowed failure. The ack prints before the forward, so a gap report cannot be delayed or failed by it.

Needs the matching backend endpoint (`POST /v1/hyperframes/catalog_search_miss`) to be deployed before anything reaches Slack. Until then the forward 404s and is swallowed, so this is safe to merge in either order.

## Test plan

- `bunx vitest run packages/cli/src/utils/submitFeedback.test.ts` - 9 passed, covering the posted URL and body, field truncation, and that a failed forward never throws.
- `bunx oxlint` and `bunx oxfmt --check` clean on the three changed files.
- Not covered: no live end-to-end post, since the backend endpoint is not deployed yet. `packages/cli/src/commands/feedback.telemetryJoinKeys.test.ts` fails to resolve `@hyperframes/core` in a fresh worktree, which reproduces identically on untouched `main` and is unrelated to this change.

### DIFF
--- a/packages/cli/src/commands/feedback.ts
+++ b/packages/cli/src/commands/feedback.ts
@@ -10,7 +10,7 @@ import { shouldTrack, flush } from "../telemetry/client.js";
 import { getDoctorSummary } from "../telemetry/feedback.js";
 import { readConfig, type RecentRenderRecord } from "../telemetry/config.js";
 import { publishProjectArchive } from "../utils/publishProject.js";
-import { submitFeedback } from "../utils/submitFeedback.js";
+import { submitCatalogSearchMiss, submitFeedback } from "../utils/submitFeedback.js";
 import { buildIssueUrl, HYPERFRAMES_REPO_URL } from "../utils/feedbackIssue.js";
 import { VERSION } from "../version.js";
 import { c } from "../ui/colors.js";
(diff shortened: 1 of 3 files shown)


miga-heygen (approved):
Reviewed at exact head. Catalog search misses are now forwarded to the publish API endpoint as a best-effort POST, bounded by a 5s abort signal and wrapped in a catch so it never fails the command. Fields are capped to backend limits (500/500/50 chars). The ack prints before the forward fires so the reporter never waits on it. Three tests cover the happy path, truncation, and offline resilience.

— Miga

PR #3207, opened 2026-08-11
feat(studio): retime and edge-stretch an automation selection

vanceingalls:
First of nine slices of the old #3058, cut to keep each PR reviewable.

Drag a selection to retime it, or take its edge to stretch the points inside. Edge-stretch gets the same gesture contract the other four already follow — pointer capture, Escape-revert, a commit on release — so it stops being the odd one out.

Retiming works from a fixed points snapshot rather than re-reading mid-gesture, which is what stopped a stretch from compounding on itself.


miga-heygen (commented):
Clean. The snapshot approach prevents compounding, the gesture lifecycle is properly wired, and SSOT is maintained.

**Snapshot is correct.** `retimeRange` works from the frozen `origin` passed at press time. Each pointermove recomputes `dt` from that origin rather than accumulating deltas, so whether the edge arrives in one move or four the result is identical — the anti-compounding test pins this with a two-instance comparison.

**Gesture lifecycle matches the contract.** Pointer capture retargeted from `e.target` to `e.currentTarget` (SVG root), fixing a real bug where a child unmounting mid-drag silently released the capture. Escape-revert and commit-on-release are wired through the existing `cancelDrag`/`finish` flow.

**Edge clamping order is non-obvious but correct.** For `t0`: `Math.max(0, Math.min(raw, origin.t1 - MIN_SELECTION_SEC))` applies the 0-floor last, so a narrow selection near t=0 can't push t0 negative. The test "never yields a negative time" pins this explicitly.

**Edge-over-point priority** is the right call — the "stretches the same edge twice in a row" test uses `mountRerenderable` to verify that a breakpoint left on the edge by the first stretch is still edge-stretchable on the second, not treated as a point drag.

No blocking issues. LGTM.

_Review by Miga_

miguel-heygen (dismissed):
Exact-head review at `30c05475` — independently audited the retime/edge-stretch transform, gesture lifecycle, pointer capture, boundary clamping, and the regression coverage Miga highlighted.

- `useAutomationEdgeStretch.ts:136-208` retimes every preview from the fixed arm-time snapshot, so multi-move drags cannot compound.
- `useAutomationLaneGestures.ts:258-338` gives edge handles priority and cleanly separates release/commit from pointer-cancel/revert.
- `automationLaneSelection.ts:82-116` scales the selected interior through the shared range replacement invariant; the endpoint behavior and degenerate spans are pinned.
- Exact head is open, mergeable/CLEAN, and every check run is terminal green. No blockers found.

Miga already covered the same strengths; this approval confirms them independently.

**Verdict:** APPROVE
**Reasoning:** The transform is snapshot-based, gesture exits own the correct commit/revert semantics, and the boundary/repeatability regressions are directly tested.

— Magi

PR #3208, opened 2026-08-11
fix(core): make audio automation survive being rescheduled

vanceingalls:
Lanes are committed to absolute context times, so anything that re-aims them mid-playback has to rebase from the live playhead rather than from where the source was originally scheduled.

683 lines of code; the rest is the test suite that pins the rescheduling behaviour across carve, automation and the runtime.


miga-heygen (commented):
Strong core fix — the rebase-from-playhead math is correct, the worklet-generation guard prevents stale resolutions, and the test suite models Chrome's real scheduling rules. One math issue at the clamp boundary needs attention.

**Rebase math is correct.** `timingNow()` computes `elapsed + (now - scheduledAt) * rate`, and `toCtx(t)` maps back as `scheduledAt + (t - elapsed) / rate`. Traced both directions: a point at clip time `t` is correctly scheduled at context time `now + (t - elapsed) / rate`. No off-by-one at boundaries; the seed value samples the lane at `elapsed` and writes at `scheduledAt = now`.

**Worklet generation guard is sound.** The `workletGeneration` counter plus `disposed` flag prevent stale promise resolutions. Multiple rapid chain edits during worklet loading are handled: each edit bumps the generation, only the latest resolution proceeds.

**Potential issue: NaN from `viaConic` at the clamp boundary.**

`clampVia` maps out-of-range via coordinates to exactly 0.999. `viaConic` uses `edge = 0.999`. When `viaX` is clamped to 0.999 and `dx > 0`:

```
needX = dx / (edge - viaX) = 0.499 / (0.999 - 0.999) = Infinity
w = Math.max(1, Infinity, ...) = Infinity
```

Then in `shapeVia`:
```
spread = 2 - 2 * Infinity = -Infinity
a = x * (-Infinity) - 1 + 2 * Infinity * 0.999 = NaN  (Infinity - Infinity)
```

NaN propagates through `sampleAutomationLane` → `setValueCurveAtTime`. Reachable when a user drags a via point past the segment edge (the test case `{ viaX: 5, viaY: -3 }` clamps to `(0.999, 0.001)`).

This is asymmetric in floating point: `0.999 - 0.999 = 0.0` exactly, while `0.001 - (1 - 0.999) = -8.67e-19` (not zero). So only coordinates clamped to 0.999 trigger it.

*Fix options:* (a) tighter clamp (e.g. `0.998`) so `edge - viaX` always has headroom, (b) guard `w` against Infinity in `viaConic`, or (c) add `if (!Number.isFinite(denominator))` to `shapeVia`'s existing guard.

**Non-blocking observations:**

- The `emit` fallback (line ~155 in `audioFxAutomation.ts`) silently degrades from a shaped curve to a linear ramp on `setValueCurveAtTime` failure. The degradation is correct; a debug trace would help field diagnostics.
- `isStraight` and `shapeProgress` have slightly different validity criteria for via points — a point like `{viaX: 0.5, viaY: 0.5}` makes `isStraight` return false (triggering sampling) but `shapeProgress` falls through to `applyCurve` identity. Result is correct but wastes work.
- The `carveProfile` NaN default of 0.5 differs from `DEFAULT_CARVE.strength` (0.25). Not reachable through `normalizeCarveSettings` but inconsistent if called directly.

_Review by Miga_

PR #3209, opened 2026-08-11
feat(studio): select automation points with a box, and stop them crossing

vanceingalls:
Box-select over a lane, and a guard that stops dragged points crossing their neighbours.

This also consolidates five separate gesture hooks into one — `useAutomationEdgeStretch` is deleted and folded into `useAutomationLaneGestures`, which is why the diff shows 530 deleted lines against 931 added code lines.


miga-heygen (commented):
Solid consolidation — five gesture hooks folded into one with no lost behavior. The crossing guard, box-select geometry, and store sync are all well-tested.

**Crossing guard is correct.** Single-point drags use `MIN_POINT_GAP_SEC` clamping against neighbours. Group drags compute `gapTo(-1)` / `gapTo(1)` to determine available travel in each direction, and the clamp correctly handles the case where existing points are already closer than the gap minimum (returns 0, preventing further closing but allowing separation). Boundary clamping at `[0, duration]` is applied independently.

**Box-select geometry is clean.** Endpoint-inclusive on both axes, same rule for rendering, deletion, and group drag initiation. The `indices.length > 1` threshold correctly falls through to single-point drag when a box contains exactly one point.

**Gesture consolidation preserves all behaviors.** Edge-stretch was the only gesture with a distinct `pointercancel` path (revert). The new code commits on cancel, using undo as the escape — same behavior the single-point drag already had. The "stretches the same edge twice" test from #3207 is deleted because the feature itself (edge-stretch) is absorbed.

**One consistency note (non-blocking):** The group-drag shift-axis lock recomputes the axis from scratch on every pointermove, while the single-point path locks it on first travel via a persistent `shiftAxis` ref. Near the diagonal, the group path could flip axes between moves. Since `dt`/`dv` are measured from the gesture's original anchor (not deltas), this won't accumulate errors, but it would feel inconsistent with single-point behavior. The test passes because its pointer path is unambiguously sideways. Consider sharing the `shiftAxis` ref for consistency.

**Lane height increase** from 48px to 72px is a 50% user-facing layout change — worth a release note.

**`automationStoreSync`** correctly converts `getAttribute` null to undefined, matching `TimelineElement` optionality. The "keeps the same array when nothing changed" test pins referential stability.

No blocking issues. LGTM.

_Review by Miga_

miguel-heygen (changes requested):
Additive exact-head review at `eff0cdea`. Miga covered the new box-selection, crossing clamps, store sync, and group-axis consistency note. One blocker remains in the stacked-feature contract.

**blocker — #3209 deletes #3207 edge-stretch instead of folding it into the unified hook.** The PR body says `useAutomationEdgeStretch` is “deleted and folded into `useAutomationLaneGestures`,” but `useAutomationLaneGestures.ts:234-506` only arms/moves box selection, point/group drag, and segment bend; there is no edge hit-test, arm/move/finish path, or retime transform. `TimelineAutomationLane.tsx:333-344` likewise has no resize cursor/edge gesture and routes `pointercancel` directly to ordinary `endDrag`. `automationLaneSelection.ts:58-97` no longer exports `retimeRange`. The commit also deletes every #3207 stretch regression, including repeat-stretch, snapshot anti-compounding, clamp, and cancel-revert tests.

Because #3209 is stacked on #3208/#3207, merging the stack would ship #3207 and then silently remove its user-facing retime/edge-stretch feature. Restore that behavior inside the consolidated hook (including the #3207 regressions), or restructure the stack so #3209 does not revert it.

Exact head is otherwise open, mergeable/CLEAN, and terminal green; green CI cannot cover a deleted feature whose tests were deleted with it.

**Verdict:** REQUEST CHANGES
**Reasoning:** The stated consolidation drops a feature introduced two PRs downstack, violating the stack contract and removing its regression coverage.

— Magi

PR #3210, opened 2026-08-11
feat(studio): show every automated knob at the playhead

vanceingalls:
An automated parameter has two values: the number in the chain, which is only the seed a lane replaced, and the number the envelope is on right now. The second is the true one, and a rack showing the first reads as broken during playback — the carve is visibly working and the readouts do not move.

Sampled while paused too, because a scrub has a playhead and the envelope has a value there.


miga-heygen (commented):
Live readouts are a real improvement — showing the envelope value at the playhead instead of the seed makes the rack match what the user hears. Carve module grouping and auto-apply are clean restructurings.

**Stack imports (`syncStoredAutomationFromPreview`, `sampleAutomationLane`, `pointInSelection`)** — these are imported from earlier PRs in the stack (#3207-#3209). Verified that the stack base provides them. Not an issue for stacked PRs.

**Performance observation (non-blocking):** `liveAutomationValues` creates a new `Map` identity on every render via the IIFE. During playback, `useLivePlayheadTime` drives ~30 re-renders/sec. The new map reference cascades through the 6-7 level prop chain (`FxSection` → `FxCarveModule` → `FxCarveMember` → `FxNodeRow` → `FxNodeParams` → `FxParams` → `FxParamRow`). Consider `useMemo` keyed on `playhead` + `automation.lanes` + `chain` to avoid re-renders when sampled values haven't changed. Same applies to the `carvedAgainstBy` IIFE that calls `querySelectorAll('[data-fx-carve]')` on every render.

**`setAttributeWhileDragging` applies `previewOnly` to all `onSetAttributeLive` consumers** — the PR asserts every consumer previews-on-move and commits-on-release. This is a contract that isn't enforced by the type system, so any new consumer calling `onSetAttributeLive` as its only write path would silently lose writes. Worth a comment on the contract.

**Delete rework (non-blocking):** The switch from anchor-pinning (`replaceRange` with `inner: []`) to point-removal (`pointInSelection` filter) is cleaner. Tests verify endpoint-inclusive behavior (t=2 and t=4 removed). `pointInSelection` itself isn't in this diff but is from #3209 in the stack.

**`paramValueWidthCh` edge case:** `Math.floor(Math.abs(-0.5))` = 0, giving 0 integer digits for parameters with range [-0.5, 0.5]. The formatted value "-0.50" has 1 digit before the decimal. Latent but may not apply to current audio FX parameters.

LGTM — no blocking issues.

_Review by Miga_

miguel-heygen (dismissed):
Additive to @miga-heygen at exact head `72fbcb8d425143202af5f3aacd956fc5cc05e508`.

**Audited:** `useLivePlayheadTime.ts` end-to-end; the live-value flow through `propertyPanelAudioFxGroup.tsx:130-139` and `propertyPanelFxSection.tsx`; the paused/playback/edge-value coverage in `propertyPanelAudioFxGroup.test.tsx:848-970`; and the preview-only drag/persist-on-release boundary in `StudioRightPanel.tsx` + `useDomEditAttributeCommits.ts`.

**Trusting:** the broader carve-analysis/module restructuring and the remaining 18-file stacked diff, after checking its integration points and relying on Miga's independent read. CI has no required contexts configured on this stack branch; every exact-head check-run that did execute is terminal success/skipped, and GitHub reports `CLEAN` / mergeable.

The clock split is correct: paused reads use store time, playback follows `liveTime` at a bounded 33 ms cadence, and cleanup unsubscribes plus clears the pending timer. The regressions distinguish stored seeds from sampled automation values, cover paused scrub and live playback, pin off-clip edge holding, and verify hand-built FX sliders move too. The preview-only pointermove path also preserves a single persisted release/undo boundary.

*Non-blocking:* I agree with Miga's Map-identity and repeated `querySelectorAll` performance note. It is worth memoizing/following up, but it does not change the sampled values or persistence contract.

— Magi

**Verdict:** APPROVE
**Reasoning:** The live-readout clock, value propagation, cleanup, and gesture persistence contracts are correct and regression-covered at the exact head; the remaining concern is bounded performance follow-up, not a ship blocker.

PR #3211, opened 2026-08-11
docs(skills): add /hyperframes-audio, and key the waveform cache by file

vanceingalls:
The audio skill, with a headless carve script so a chain can be built from the command line.

Also fixes the studio-server waveform cache, which was keyed loosely enough that two files could collide.


miga-heygen (commented):
Good skill documentation and a real cache-collision fix. One incomplete path in the fix needs attention.

**Cache key divergence: `generateWaveformCache` not updated.**

The route at `waveform.ts:22` now calls `buildWaveformCacheKey(assetPath, stats)`, producing a fingerprinted key like `v2_music-bed.m4a_4187869-1000.json`. But `generateWaveformCache` at `waveform.ts:98` still calls `buildWaveformCacheKey(assetPath)` without a fingerprint, producing `v2_music-bed.m4a.json`.

Two consequences:
1. Pre-generated cache from file upload is never found by the route (different key) — the optimization path is silently dead.
2. The proactive path still has the collision bug this PR exists to fix — an asset rebuilt in place keeps serving stale peaks because the path-only key hasn't changed.

`generateWaveformCache` already calls `existsSync(audioPath)`, so it can trivially `statSync` and pass the result.

**Skill documentation** is excellent — the architecture diagrams, signal-flow explanation, and "reach for a family by the problem" section are strong. The fx-registry and attribute references are thorough.

**Skill ordering inconsistency (non-blocking):** `/hyperframes-audio` appears in different positions relative to neighbours across CLAUDE.md, README.md, and templates. CLAUDE.md + Router agree (`...media-use, audio, cli...`); README and templates differ.

**`String.replace` dollar-sign injection (non-blocking):** `carve.mjs:1105` uses `html.replace(bedTag, nextTag)` where `nextTag` contains user-derived content. `$&`, `$'`, `` $` `` in element IDs would be interpreted as replacement patterns. `html.split(bedTag).join(nextTag)` avoids this.

**Stale cache accumulation (non-blocking):** New fingerprinted keys create a new file per asset change. Old entries are never cleaned up. Space leak over repeated edits.

The cache key fix needs to flow through to `generateWaveformCache` — otherwise the fix is only half-applied.

_Review by Miga_

PR #3212, opened 2026-08-11
feat(core): carve against every voice over a bed, always

vanceingalls:
Carve against every voice over a bed, always — not just the one the author happened to pick.

A bed usually runs under a whole sequence: a narrator, an answer, a second presenter. They are analysed together, so the cuts follow whoever is speaking, which makes the source a set of tracks to include rather than a choice between them.


miga-heygen (commented):
Clean design — summing voices then feeding one combined signal to the existing analysis pipeline avoids spreading multi-voice awareness through the whole stack. Backward compat migration in `normalizeCarveSettings` handles the `source` → `sources` and `dynamic` removal correctly.

**Two items worth addressing:**

**1. `clipsOverlap` has zero test coverage.** The function is exported, has non-trivial logic (unbounded-duration via `POSITIVE_INFINITY`, half-open interval comparison), and gates whether a voice enters a carve. Negative duration produces an inverted interval where `end < start`:

```
a = {start: 10, duration: -5}  →  end = 5
b = {start: 3, duration: 10}   →  end = 13
a.start < end(b) = true, b.start < end(a) = true → overlap detected
```

...for an interval that doesn't exist. Add a `describe("clipsOverlap")` block covering: normal overlap, no overlap, adjacent clips, null/undefined duration, zero duration, and negative duration. Guard or clamp negative duration if the contract disallows it.

**2. `\b` word-boundary regex misses underscore-separated names.** `\b` treats `_` as a word character, so `\bbed\b` does NOT match `bed_01`, `my_bed`, or `music_bed_loop`. Same for `\bvo\b`, `\bvox\b`, `\bsong\b`. A music bed named `bed_01` classifies as `"unknown"` and `couldBeCarveSource("bed_01")` returns `true` — the bed could end up in its own carve source list. Replace `\b` with a separator-aware pattern that treats `_` and `-` as boundaries.

**Non-blocking:**

- `mixCarveSources` output can exceed `[-1, 1]` (summing). Correct if downstream handles unnormalized input, but worth a comment confirming.
- Stale comment at line ~222 references `dynamic` which no longer exists in `DEFAULT_CARVE`.
- `normalizeCarveSettings` null guard (`raw === null`) is untested and not reachable through the type signature. Either test the defensive path or drop the check.
- Empty-samples voice allocates a zero-filled buffer in the reduce — a `continue` on `samples.length === 0` is tidier.

Tests for `mixCarveSources` (offset placement, gap silence, overlap summing, negative-offset clipping) and backward compat are strong.

_Review by Miga_

PR #3213, opened 2026-08-11
feat(studio): the carve is one module in the rack

vanceingalls:
A carve is one thing the author switched on; the peaking filters and the level stage are how it is built. Listed individually they read as hand-built effects — removable one at a time, reorderable, each with knobs the next strength change silently overwrites.

So the rack shows the unit, and the unit owns everything that means anything for it: which voice it listens to, how hard it works, and what the analysis made of it. Grouped is not hidden — opening it lists every effect inside with all of its settings, as readouts rather than controls, since strength is what sets them.

Also carries the CLI-side carve documentation.


miga-heygen (commented):
This PR provides the UI consumer port for #3212's `sources`/`enabled` shape — confirming the prior concern is addressed. The carve module grouping, auto-apply, and strength-driven readouts are well-designed.

**One issue: duplicate `useEffect` for single-candidate beds.**

Two auto-apply effects both fire when `sourceOptions.length === 1`:
- Effect 1 guards `sourceOptions.length === 0` (passes for length 1), calls `setCarve({ ...DEFAULT_CARVE, sources: all })`
- Effect 2 guards `sourceOptions.length !== 1` (passes for length 1), calls `setCarve({ ...DEFAULT_CARVE, sources: [only.id] })`

Both produce identical arguments (`all` and `[only.id]` are the same list), so the most common case (one narrator, one bed) triggers double analysis: two audio decodes, two FFT runs, two concurrent attribute writes. Fix: either remove the second effect or partition the space with `if (sourceOptions.length === 1) return;` in the first.

**Non-blocking:**

- `moveNode` `first={index === 0}` doesn't account for carve nodes preceding hand-built nodes — move-up is incorrectly enabled on the first hand-built node. Pre-existing but exposed by the architectural split.
- Brief flash of "pick a voice" on initial render before the auto-apply effect fires on the next tick.
- Tests are thorough: auto-apply (single/multi-voice), off-stays-off, module toggle, expand/collapse, slider snap-back, analysing spinner, voice-track exclusion.
- Skill docs and CLI script correctly updated to `sources`/`enabled`/repeatable `--voice`.

LGTM after the duplicate effect fix.

_Review by Miga_

PR #3214, opened 2026-08-11
feat(studio): give the clips on a track one automation lane row

vanceingalls:
Every clip on a track shares one automation lane row, rather than each clip growing its own. A track has one volume, and a row per clip made that look like several.

Includes the design record for why effects stay on the clip while lanes belong to the track.


miga-heygen (commented):
Clean architecture. The per-track lane grouping is well-designed — SSOT in `groupAutomationLanes`, stable React keys via `laneGroupKey`, correct multi-clip SVG positioning.

**No blocking issues.**

**How the key concerns are handled:**

- **Multi-clip resolution**: `groupAutomationLanes` unions by label. First clip fixes row index; later clips append new properties. Entries per group are in element order.
- **Overlapping clips**: Each clip draws its own SVG at its own `leftPx`/`widthPx` — overlapping visuals in the same row, correct for per-clip gestures.
- **Lane identity stability**: React keys use `group.key` (the label string), stable under selection changes. No remount/flicker.
- **SSOT**: One lane per track is encoded in exactly one place (`groupAutomationLanes`). All three call sites (header, slot, height) flow through it.

**Non-blocking observations:**

- Design record status is stale: says "Not yet implemented" but this PR IS the implementation.
- `groupAutomationLanes` is computed independently at 3 sites without memoization. Cost is Map iteration (not parsing — attribute text is cached), acceptable for typical track sizes but a `useMemo` in the Slot component would help if many-clip tracks become common.
- `laneGroupKey` delegates to `automationLaneLabel` — the display label IS the React key. Elegant but means a label format change remounts every row. Documented and intentional.
- The `holdingOpen` trick in `useTimelineRowHeights` is well-commented. Safe today since `trackHeights` only uses `clipId` for the `expandedClipIds.has()` check.
- No test for clip deletion from a multi-clip track — grouping/height shrinkage after removal. May be covered by integration tests elsewhere.

LGTM.

_Review by Miga_

PR #3215, opened 2026-08-11
fix(studio): the shared-row follow-ups

vanceingalls:
The follow-up fixes shared lane rows needed once they were real:

- cap every clip bar on an expanded row, so a long clip cannot paint past it
- stop a shared row's clips swallowing the pointer events the lane needs
- mount the lanes on the row itself, so they scroll and collapse with it
- re-analyse a carve when a voice it was listening to changes
- expect what deleting a clip already wrote, rather than a stale snapshot


miga-heygen (commented):
Solid final slice. All five fixes are well-scoped, individually tested, and I found no correctness bugs.

**No blocking issues.**

**The five fixes:**

1. **Clip bar capping**: `clipBarHeight` computed per-row, applied to ALL clips on expanded rows. Prevents waveform painting over envelopes.
2. **Pointer-event forwarding**: Lane wrapper gets `pointer-events-none`; only the SVG and value inputs opt back in. Prevents clip bands from swallowing lane interactions.
3. **Lane mounting on the row**: `TimelineAutomationLaneSlot` extracted from `footer` prop into a sibling div at row level. Fixes lane remount on selection change (which destroyed hover state and in-flight gestures).
4. **Carve re-analysis on voice deletion**: `survivingSources` effect watches the timeline, prunes deleted voices from `carve.sources`, triggers re-analysis. Confirmed safe against infinite loops — the filter is idempotent, and analysis can't modify the voice list.
5. **`diskContent` for optimistic-concurrency**: `saveProjectFilesWithHistory` gains `diskContent` so the write expectation matches what `remove-element` put on disk, not the pre-mutation baseline.

**Non-blocking:**

- `survivingSources.join(" ")` uses space separator while `candidateIds` three lines below uses `"\u0000"`. If an element ID ever contained a space, the effect could miss a real change. Element IDs are space-free by convention so near-zero risk, but inconsistent.
- SPEC.md says "Status: SPEC — not implemented" but the system IS implemented across the preceding 8 PRs.
- `aria-controls` multi-ID is correctly handled per WAI-ARIA spec.

LGTM — ship it.

_Review by Miga_

PR #3218, opened 2026-08-11
fix(skills): remove duplicated media-cleanup block in slideshow

imprimisxo:
## What

Deletes a 30-line duplicated block from `skills/slideshow/SKILL.md`. Pure deletion — **+0 −30**, no content added or reworded.

## Why

Lines **431–461** are repeated at **463–491**: the presenter-notes paragraph, then the whole of `### Media cleanup on slide exit` and `### Global nav mute`. Eleven paragraphs, 2,843 bytes — about 7.7% of the file.

**The copies are not identical.** The second is a strict subset that drops exactly one normative paragraph:

> If the source page has custom controls or visualizations attached to media, those controls must listen to the same native element the slideshow player stops and mutes. A pause caused by slide exit, presenter sync, native controls, custom controls, or the global mute button should all update the visible custom UI through media events, not through parallel state.

That matters because of ordering. An agent reading the file top-to-bottom hits the complete version first and the truncated one second — and the natural reading of a repeated section is that the later one supersedes. So the rule most likely to be lost is the one about custom controls syncing through media events, which is exactly the kind of thing that produces a deck where the play button and the actual `<video>` disagree after a slide change.

Everything else in the two copies is verbatim identical, so copy 1 is a strict superset and the later copy can go without merging anything.

## How

Removed lines 463–492 (the duplicate plus its trailing blank), keeping the `---` separator that follows. Nothing else touched — the diff is `+0 −30`.

Verified on the result:

- duplicated paragraphs: **11 → 0**
- the custom-controls rule: present, exactly once
- the presenter-notes paragraph: present, exactly once
- `### Media cleanup on slide exit`, `### Global nav mute`, `## Wrapping component`, `## Running a slideshow standalone`: one heading each
- both `references/standalone-harness.md` pointers retained (the presenter-notes paragraph and the standalone section — two distinct, legitimate mentions)
- code fences still balanced

## One separate observation — no change made

Lines **242–396** are a 154-line, 4.8 KB fenced block: the "Scene HTML (skeleton)" for the 3-slide worked example. It is by far the largest code block in the file; the next largest is under 25 lines.

It could ship as an asset the way `references/standalone-harness.md` already does, which would remove ~13% of the file and give the agent something to copy rather than retype. But an illustrative skeleton is a weaker case for extraction than a complete runnable document, and it may genuinely be more useful inline where the surrounding prose annotates it. Raising it rather than assuming — happy to send it as a follow-up if you want it.

## Test plan

- [ ] Unit tests added/updated — n/a, documentation only
- [x] Manual testing performed — verified against HEAD that the block is duplicated and that the second copy is a strict subset; verified the result has zero duplicated paragraphs, one of each affected heading, and retains the paragraph the duplicate dropped
- [x] Documentation updated (if applicable) — this PR *is* the documentation change

Found while auditing skill files for context-budget cost.

### DIFF
--- a/skills-manifest.json
+++ b/skills-manifest.json
@@ -70,7 +70,7 @@
       "files": 70
     },
     "slideshow": {
-      "hash": "71368acf61074198",
+      "hash": "6a24a84b0c1a75f9",
       "files": 2
     },
     "talking-head-recut": {
--- a/skills/slideshow/SKILL.md
+++ b/skills/slideshow/SKILL.md
@@ -460,36 +460,6 @@ The same cross-realm rule applies here: global mute must reach iframe `<video>` 
 
 `hyperframes present` serves built bundles from `packages/player/dist`. After changing player or slideshow chrome behavior, run `bun run build` in `packages/player` and restart the present server before testing in a browser.
 
-Presenter notes are editable in the presenter view. Edits are stored in `localStorage` per deck and slide, layered over the manifest notes without rewriting the composition file. Do not add one-off note-editing scripts to decks; rely on the shared slideshow player behavior. If a standalone/custom wrapper truly needs to implement this outside the shared player, use the deterministic storage snippet in `skills/slideshow/references/standalone-harness.md`.
-
-### Media cleanup on slide exit
-
-The slideshow controller owns slide-exit media cleanup. When navigation changes slide or sequence, it calls `hyperframes-player.stopMedia()` before entering the next slide. That command:
-
-- posts `stop-media` to the iframe runtime, which stops WebAudio and pauses native `<video>` / `<audio>` elements;
-- pauses same-origin iframe media directly as a fallback; and
-- pauses parent-frame proxies adopted from iframe media.
-
-Same-slide fragment navigation does **not** stop media. Global/deck-level parent audio, such as a background track wired through `audio-src`, is not treated as slide media.
-
-Do not add per-slide cleanup scripts for normal media players. Keep slide video/audio as normal media in the composition; use `data-has-audio="true"` only when the player should preserve audible native video audio instead of treating it as silent visual media.
-
-When implementing direct iframe fallback cleanup, treat iframe media as cross-realm DOM. Do not test iframe nodes with the parent page's `el instanceof HTMLMediaElement`; that returns false in real browsers. Use `el.ownerDocument.defaultView.HTMLMediaElement` (or an equivalent tag/duck-type guard) before setting `muted` or calling `pause()`.
-
-### Global nav mute
-
-When `<hyperframes-slideshow sound>` renders the nav mute button, that button is the global mute control for the page. It must mute:
-
-- child `<hyperframes-player>` instances, including same-origin iframe media;
-- top-level page `<audio>` / `<video>` elements; and
-- wrapper-owned SFX/global `Audio` objects via the `hf-sound` event.
-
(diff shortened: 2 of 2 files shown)


jrusso1020 (approved):
Approving. I verified the dedup keeps the *complete* copy rather than the truncated one, which is the load-bearing question on a "remove the duplicate block" change.

The two copies were not identical: the earlier one carried a paragraph requiring custom media controls to sync through media events, and the later one did not. Counts at this head vs `main`:

```
                             main   head
custom-controls paragraph      1      1   <- survives
Media cleanup on slide exit    2      1
Global nav mute                2      1
presenter-notes paragraph      2      1
file length                  559    529   <- exactly 30
```

Code fences stay balanced (28, even). The only lines still repeated are `{` / `}` / `</p>` inside the example HTML, which is expected.

Worth saying the payoff is not tidiness: a skill file is read into context on every invocation, so a duplicated instruction block is paid for on each run, and two copies of the same rule that disagree is a correctness hazard for whatever reads it.

The 154-line skeleton extraction noted in the description looks like a reasonable follow-up but is rightly not in this PR.

— Rames Jusso

miga-heygen (commented):
Clean dedup. The deleted block (lines 463-498) is the incomplete copy — missing the custom-controls paragraph that the surviving copy at lines 420-460 retains. Manifest hash update is generator output. LGTM.

_Review by Miga_

miguel-heygen (approved):
Exact-head review at `f5098b281a3bd599320707842791ca2eea31b47b`.

`skills/slideshow/SKILL.md:420-461` retains the complete media contract, including the sole custom-controls/native-event synchronization paragraph. The later 30-line copy is the incomplete duplicate: after deletion, the presenter-notes paragraph and both affected headings each occur exactly once, the custom-controls rule remains exactly once, all 28 fence markers remain balanced, and the section flows directly into the existing separator and standalone guidance.

`skills-manifest.json:66` is the only generated delta and updates the slideshow hash; exact-head `Skills: manifest in sync`, project-native skill lint/mirror, skills tests, Format, and every required check are terminal green or skipped as intended. Existing reviews contain no unresolved correctness finding.

**Verdict:** APPROVE
**Reasoning:** The current head removes only the truncated duplicate, preserves the stricter surviving contract, keeps the generated manifest coherent, and is green on all required gates.

— Magi

jrusso1020:
Heads up @imprimisxo, and thank you for this one.

Your change is blocked here by something that has nothing to do with its quality: the org enforces signed commits on all branches, and `a91aa031` is unsigned, so this cannot merge as-is even though it is approved.

Rather than ask you to re-sign and force-push, I carried it into #3225 byte-identical, with `--author` set to you so the authorship stays yours. There was a second reason to combine them: that PR also edits skill content, and both regenerate `skills-manifest.json`, so as separate PRs they would have conflicted on that file whichever merged second.

I re-verified your dedup at the new tree rather than taking it on trust: 559 to 529 lines, 28 code fences balanced, each repeated section down to one, and the custom-controls paragraph still present exactly once at line 445. Keeping the complete copy and deleting the incomplete one was the right call, and it is the part of a dedup that is easy to get backwards.

Your #3219 report is fixed in the same PR. Both files you flagged as orphaned are deleted.

jrusso1020:
Closing this out: your change is now on `main`.

It merged as part of #3225 in commit d3a4e903, carried over byte-identical with your authorship preserved on the commit. The reason for folding it in rather than merging this PR directly is in my earlier comment above, and it was not a reflection on the change: the org ruleset requires signed commits, and a second PR touching skills would have collided with this one on `skills-manifest.json` regardless.

Thanks for catching this. The two copies of that block had drifted apart, and the one you deleted was the incomplete one, which is the part that actually mattered and the part that is easy to get backwards.

PR #3222, opened 2026-08-11
fix(hyperframes-animation): correct the iOS system default spring register label

jbernard077:
Small factual correction in `skills/hyperframes-animation/adapters/gsap-easing-and-stagger.md`.

The file labels dampingFraction **0.80–0.85** as "the iOS system register" (springEase
comment) and "the iOS system default register" (dampingFraction table). Per Apple, the
iOS system default — [`Animation.default`](https://developer.apple.com/documentation/swiftui/animation/default),
since iOS 17 — is a spring with **response 0.55, dampingFraction 1.0**: critically
damped, no bounce. 0.80–0.85 corresponds to SwiftUI's `snappy` preset
([`Spring`](https://developer.apple.com/documentation/swiftui/spring) — "a small amount
of bounce"), an optional livelier register rather than the platform default.

## Changes

- springEase comment: the "iOS system default" fact moves to the `1.0` line
  (`Animation.default` = response 0.55, dampingFraction 1.0); `0.80–0.85` is relabeled
  ≈ SwiftUI's `snappy` preset.
- dampingFraction table: same relabel on the `0.80–0.85` row; the `1.0` row gains the
  true-default note.

No behavioral change — `springEase` itself is untouched, and the file's surrounding
doctrine ("Well-made system animations are critically damped or close to it", default
ζ=1) already agrees with the corrected labels.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3223, opened 2026-08-11
feat(hyperframes-animation): add platform-motion-idioms rule (Android/M3 vs iOS dialects)

jbernard077:
Adds `rules/platform-motion-idioms.md` + its `rules-index.md` row: platform motion
dialects, applied when a mocked surface resolves to a platform device frame (Pixel
frame → Android/M3 dialect, iPhone/iPad frame → iOS dialect; neutral browser/desktop
chrome stays in the house register). The point is that platform-flavored motion should
be chosen, not produced unknowingly — an Android frame animated with parallax pushes
and z-scale pops reads wrong to anyone who owns the device being mocked.

## What's in it

- **Dialect table** — forward/back/modal/expand/elevation/shared-element/easing rows
  for both platforms, with every claim quoted from and cited to Google's M3/Android or
  Apple's HIG/API documentation (22 sources listed in the file; M3 content is CC-BY,
  quotes verbatim).
- **M3 tokens, verbatim** — the Emphasized/Standard easing beziers, the short/medium/
  long/extra-long duration scale, suggested pairs, plus the legacy note: M3 marks this
  system legacy in favor of motion physics, so the rule also carries Google's published
  spring-to-curve web conversions for expressive briefs in a baked deterministic
  timeline.
- **Android back, disambiguated** — the predictive-back gesture spec (35% commit
  threshold, 100→90 / 110→100 scales, STANDARD_DECELERATE progress) applies only to a
  depicted edge swipe; button/programmatic back plays the forward transition reversed.
- **iOS register** — the system default is `Animation.default`: response 0.55,
  dampingFraction 1.0, critically damped (i.e. `springEase({ response: 0.55,
  dampingFraction: 1.0 })` in `adapters/gsap-easing-and-stagger.md`); 0.80–0.85 ≈
  SwiftUI's `snappy`, an optional livelier register. Pairs with #3222, which corrects
  the adapter's own label to match.
- **Recipe** — M3 tokens as GSAP/CSS-ready constants, and the two-segment
  `easing.emphasized` path baked as a stateless pure-function ease (piecewise cubic
  bezier + bisection), seek-safe, no CustomEase plugin.
- **Parallax caveat** — the iOS push parallax is real but undocumented by Apple; the
  rule marks any replicated ratio as reverse-engineered and cites the parallax's
  existence to Google's cross-platform note instead.

## Provenance

Every quoted value was pulled from the cited primary sources (m3.material.io,
developer.android.com, developer.apple.com) during a 2026-08 platform-idioms research
pass; anchors and cross-references re-verified against `main` @ 896bc336a before
cutting the branch.


miga-heygen:

PR #3225, opened 2026-08-11
fix(skills): stop embedded-captions shipping author-only paths; dedupe slideshow

### LINKED ISSUE #3219: embedded-captions: shipped files point at author-only ~/Downloads paths, and references/test-set.md is orphaned
Filing under the feature form because blank issues are disabled and the bug form requires a minimal-reproduction repo — there is nothing to reproduce here, the problem is in files the skill ships. Happy to refile if you'd rather have it as a bug.

## Problem

`skills/embedded-captions` is distributed to other machines via `npx hyperframes skills`, but parts of it point at directories that only exist on the author's machine. **12 references to `~/Downloads`, across 8 files, spanning 10 distinct paths. None of them exist on a clean install.**

One of those is fine and I'm not proposing to change it: `~/Downloads/hyperframes` (`SKILL.md:251`, `scripts/matte.cjs:26`, `scripts/measure-layout.cjs:22`, `scripts/render-and-composite.sh:17`) is the documented last resort in a real resolution chain — `HYPERFRAMES_ROOT` env → in-repo path → `~/Downloads/hyperframes` — and `render-and-composite.sh` implements it properly. That's a sensible default, not a leak.

The other 8 references have no fallback of any kind.

### `themes/PORTING.md` — the porting procedure is unrunnable off the author's machine

Six author-only directories, five of them load-bearing steps in the procedure:

| Line | Path |
|---|---|
| 7 | `~/Downloads/cap_fx3/<tNN_name>/` — the demo each port is modelled on |
| 26 | `~/Downloads/port_<name>` — scaffold target |
| 28 | `~/Downloads/cap_multi/rooftop/`, `~/Downloads/cap_fx/_frames_bg` — source frames |
| 35 | `~/Downloads/cap_anchor` — the regression project step 5 tells you to recompile |
| 45 | `~/Downloads/cap_fx3/CONTRACT.md` — the scheduling contract |

Line 45 is the one I'd flag hardest: it points at a **normative contract document** (`≤ DUR scheduling`) that ships nowhere. A reader following PORTING.md is told to consult a spec they cannot obtain.

### `SKILL.md:177` and `dna/README.md:17` — the retired template library

> The retired 54-template library lives outside the skill at `~/Downloads/embedded-captions-archive/standard-templates-54/`

Stated as a fact about where something lives. For anyone but the author, it doesn't.

### `references/test-set.md` — orphaned *and* pointing at a corpus nobody has

Two problems at once:

1. **Referenced from nowhere.** I grepped all 140 shipped files: zero references to `test-set` outside the file itself. Nothing in `SKILL.md`, no other reference, no script. (Worth noting I checked the whole tree rather than just `SKILL.md` — `typographic-moves.md` looked orphaned too until I found it linked twice from `references/composition-craft.md`. `test-set.md` really is unreachable.)
2. Line 9: `All in ~/Downloads/heygen_relevant_videos/` — so even when found, the test set can't be run.

## Proposed solution

Different answers for different items, and the right destinations are yours to pick — I don't know what these directories contain:

- **`themes/PORTING.md`** — either vendor the demo/contract material into the skill (at minimum `cap_fx3/CONTRACT.md`, since it's normative), or replace the paths with a documented env var + resolution chain like the one `~/Downloads/hyperframes` already has. Failing both, mark the file explicitly as internal/author-only so readers don't follow a dead procedure.
- **The retired 54-template library** — if it's genuinely retired and not needed, say so without the path. If it's still consultable, it needs a real home (release asset, separate repo, or in-tree).
- **`references/test-set.md`** — delete it, wire it up from `SKILL.md`, or move it out of the distributed tree. As-is it's dead weight that also can't be executed.

## Alternatives considered

Sending a PR instead. I didn't, because every fix here needs a decision only you can make — where the archive should live, whether the demos are vendorable, whether the test set should exist in the distribution at all. Mechanical substitution on my side would just be guessing. Happy to send a PR once you've picked a direction.

## Additional context

Found while auditing skill files for the Claude Code skill-listing budget, not while using the skill — so I haven't hit a runtime failure from this; it's a read of the shipped tree.

Two things that came up in the same pass and are **not** problems, recorded so they don't get re-flagged:

- `SKILL.md:177` says "`_motion.md` remains in-skill." It does — at `modes/standard/_motion.md`. Not a broken reference.
- `packages/cli/dist/cli.js` resolves correctly via the chain described above.

Credit where due: `embedded-captions` was the best-factored skill in the set I looked at — 140 files, 16 references, 7 scripts, and zero duplicated content, which is more than I can say for a couple of its siblings ([#3218](https://github.com/heygen-com/hyperframes/pull/3218), [#3117](https://github.com/heygen-com/hyperframes/pull/3117)). These paths are the one thing that doesn't travel.

jrusso1020:
## What

Skill-content hygiene across two skills, plus the manifest regen both require.

Closes #3219. Supersedes #3218 (carried here re-signed, authorship preserved).

| File | Change |
| --- | --- |
| `embedded-captions/references/test-set.md` | deleted (orphaned) |
| `embedded-captions/themes/PORTING.md` | deleted (orphaned) |
| `embedded-captions/SKILL.md` | archive location claim -> "not distributed with the skill" |
| `embedded-captions/dna/README.md` | same |
| `slideshow/SKILL.md` | duplicated media-cleanup block removed (559 -> 529) |
| `skills-manifest.json` | regenerated for both skills |

## Why

**embedded-captions (#3219).** The skill is distributed via `hyperframes skills`, which installs with `--copy` so the installed bundle matches the published tree. There is no per-file filter, and the skill's `.gitignore` only keeps render artifacts out of commits. So these files land on every install, where the directories they name do not exist.

1. `references/test-set.md` was unreachable and unrunnable: zero inbound references across all 140 files, corpus at `~/Downloads/heygen_relevant_videos/`. As a control, the same grep finds `typographic-moves.md` linked 4 times, so the method discriminates rather than just returning nothing.
2. `themes/PORTING.md` was equally unreachable, also zero inbound references. It authored *new* theme DNAs rather than using the skill, five of its steps took author-only inputs, and its normative `CONTRACT.md` ships nowhere in this repo (checked by content, not filename).
3. The retired 54-template archive was stated as a fact about where something lives. True for one machine.

Not touched, so it does not get re-flagged: the 10 references to `~/Downloads/hyperframes` are the documented last resort in a real chain (`HYPERFRAMES_ROOT` -> in-repo -> that default). Every file carrying the path also carries the env var, and the chain ends in a clear `set HYPERFRAMES_ROOT` error.

**slideshow (#3218).** The media-cleanup and global-nav-mute sections appeared twice, and the copies were not identical: the later one dropped the paragraph requiring custom media controls to sync through media events. The incomplete copy is the one deleted.

## How

Both orphaned files are deleted rather than annotated. Nothing referenced either, git history keeps them recoverable, and marking a dead procedure still ships the dead procedure.

The slideshow change is @imprimisxo's work from #3218, carried here byte-identical with `--author` preserving their authorship. Two reasons it moved: the org requires signed commits and the original was unsigned, and both changes regenerate `skills-manifest.json`, so as separate PRs they would conflict on that file no matter which merged first.

## Test plan

- `bun packages/cli/scripts/gen-skills-manifest.ts --check` -> `in sync (19 skills)`
- slideshow hash resolves to `6a24a84b0c1a75f9`, matching what CI computed on #3218
- Re-verified the dedup at this tree: 559 -> 529 lines, 28 code fences balanced, each repeated section 2 -> 1, and the custom-controls rule still present once at line 445
- Confirmed both deleted files had zero inbound references before removal
- Post-change audit grep: no author-only `~/Downloads` paths remain

- [ ] Unit tests added/updated (docs + manifest only, no code paths touched)

---

#3219 reported by @imprimisxo, who also authored the slideshow fix in #3218.

_Opened by Rames Jusso (James's assistant)_


miga-heygen (commented):
Two skill hygiene fixes combined to share one manifest regen.

**Slideshow dedup** — same change I reviewed in #3218, carried here re-signed. Surviving copy is the complete one (custom-controls paragraph intact, 559→529 lines, 28 fences balanced). Already verified.

**embedded-captions cleanup (#3219):**
- `test-set.md` deleted — zero inbound references across all 140 files, corpus at `~/Downloads/heygen_relevant_videos/` which doesn't exist on any install. Correct removal.
- `PORTING.md` deleted — zero inbound references, every step requires author-only inputs (`cap_fx3` demos, `CONTRACT.md`), none of which are distributed. Correct removal.
- Archive location claims in `SKILL.md` and `dna/README.md` changed from `~/Downloads/...` to "not distributed with the skill" — factual correction.
- The 10 remaining `~/Downloads/hyperframes` references in the skill are deliberately untouched — they're the documented last resort in a real env-var chain, not orphaned claims.

**Allowed-deletions entries** in `check-no-main-deletions.mjs` have clear explanations. Manifest regenerated for both skills.

All four commits are signed (`G`). James's authorship preserved on the slideshow commit.

LGTM.

_Review by Miga_

miguel-heygen (approved):
Audited: the complete 7-file diff at `1d0e18ca870b4966ff80b6b6eff1c06227a53afa`, including both skill sources, the two intentional deletions, deletion-guard exceptions, and regenerated manifest. The slideshow carry is byte-identical to the reviewed #3218 source; its complete custom-controls guidance remains, line count drops 559→529, and all 28 fences stay balanced. The deleted embedded-captions files have zero inbound references, while the surviving text correctly stops promising author-local assets that are not distributed. Manifest hashes/counts match the resulting tree, and all five commits are GitHub-verified while preserving Lance’s authorship on the carried change.

Trusting: GitHub Actions runner execution; every required exact-head check is terminal green, including the full test suite and both Windows render/test lanes. No unresolved review threads or unaddressed blockers remain.

**Verdict:** APPROVE

**Reasoning:** The combined PR removes unreachable/duplicated skill material without weakening the surviving guidance, keeps generated state synchronized, preserves provenance, and passes the repository’s complete required gate set.

— Magi

PR #3230, opened 2026-08-11
fix(fonts): supplement alias faces from the canonical family (carries #3085)

### LINKED ISSUE #3083: fix(fonts): alias weight/style supplementation fetches the authored family, mixing two typefaces under one font-family

jrusso1020:
## What

Carries @akzarma's font-alias fix from #3085, plus the re-recorded regression baselines it needs in order to go green.

**The fix is entirely @akzarma's work and is preserved as their commit, authored to them.** The only things added here are the baseline re-records, which they could not do themselves (see Why).

Closes #3083. Supersedes #3085.

## Why

A family resolving through `FONT_ALIAS_MAP` could emit `@font-face` rules drawn from **two unrelated typefaces under one `font-family` name**, split by weight and style. The embedded faces came from the canonical bundle, but the supplementation fetch was passed the *authored* name — so for a cross-typeface alias (`helvetica` → `inter`) it asked Google for the very typeface the alias exists to replace.

@akzarma's issue (#3083) is worth reading in full. They enumerated all 69 alias entries, identified the 51 that map to a different canonical, and measured which 19 actively mixed against Google's catalogue on the day of filing.

**This PR exists only because of a mechanical limitation.** The affected fixtures' baselines encode the bug, so the corrected render legitimately fails against them and they have to be re-recorded in the same change. @akzarma has no render infrastructure and asked for the command on 10 Aug; the re-record is a maintainer job. We cannot push it to their branch either, because the baselines are LFS objects and `maintainer_can_modify` grants ref push but **not** write to a fork's LFS store. Taking the branch over was the only way to land it without handing them another dependency.

## How

Three commits:

1. **`fix(fonts): supplement alias faces from the canonical family`** — @akzarma's, unchanged. Supplements from the canonical via `resolveAliasDisplayName`, still emitting under the authored name so authored CSS keeps matching. Supplementary faces sharing source/style/unicode-range are grouped and collapsed into weight ranges, partitioned around bundled weights so a range can never shadow a bundled face. Grouping is by source rather than adjacency, because Google's no-`text=` response orders by weight *then* unicode subset, so matching faces are not adjacent; runs are emitted at their first face's position to preserve overlapping `unicode-range` ordering.

2. **`test(producer): re-record style-15-prod baseline`** — baseline only.

3. **`test(producer): re-record style-3-prod baseline`** — baseline only. Same families, same correction; caught by review on CI after commit 2 landed.

### What the baseline change actually is

Measured on `style-15-prod`, `main` rendered today versus this branch rendered today:

| | file | faces | **distinct typefaces** |
|---|---|---|---|
| `main` | 922 KB | 25 | **10** |
| this branch | 1186 KB | 28 | **5** |

Four families that are all documented to alias to Inter currently render as **ten different typefaces**. After the fix, five — and those five are exactly Inter's own set, byte-identical across all four families. Payload grows 29%, because three families gain the weight coverage they never had.

Worth stating plainly for reviewers: **the mixing was never confined to italics.** Upright text at weights 300/500/800 changes too, visible from 2.51s in that fixture, because the bundle ships only 400/700/900. `Inter` itself was the one family `main` already got right, since a self-referencing alias asks Google for the correct family — which is why this survived as long as it did.

## Test plan

- [x] Unit tests added/updated — @akzarma's suite covers the cross-typeface alias, the variable-font collapse, the interleaved no-`text=` response shape, overlapping-subset ordering, display-name coverage for every alias target, and the self-referencing case.
- [x] Manual testing performed — see below.
- [ ] Documentation updated (n/a)

Before re-recording each baseline, verified the divergence is the intended correction and nothing else:

- **Text-only.** Every changed pixel is a glyph outline, confined to a few horizontal bands, with *zero* changed pixels elsewhere. On `style-3-prod` that is ~278 of 1080 rows and ~1.7% of pixels, identical in character across all 7 failing checkpoints, with the text unchanged in content, position and colour. Video content, background, shapes and layout byte-identical.
- **Reproduces the report.** `style-15-prod` fails 43 of 100 checkpoints with first failure at 7.68s, matching the reporter's own measurement exactly.
- **Correct mode per fixture.** `style-15-prod` is recorded in-process (`--update` refuses `distributed-simulated`) then validated in `distributed-simulated`, which is the only mode CI runs it in per `tests/shard-schedule.json`: **0 failed frames, PSNR 37-47 against a `minPsnr` of 30, audio correlation 1.000.** `style-3-prod` runs in-process in CI and was recorded in-process.

**Swept the rest of the suite rather than discovering fixtures one failed CI run at a time.** CI's fail-fast cancels sibling shards on the first failure, so a red shard hides how many others would also have failed. Every fixture whose sources reference the aliased families was checked directly: `style-1`, `style-2`, `style-7`, `style-8`, `style-12`, `style-16`, `sub-comp-t0`, `missing-host-comp-id`, `variables-prod` and `vignelli-stacking` all render **unchanged, 0 failed frames**. Referencing an aliased family does not by itself mean a fixture changes — only those requesting weights or styles the bundle does not ship do. `style-3-prod` and `style-15-prod` are the complete set.

`output/compiled.html` is deliberately **not** updated: `validateCompilation` reads it only for timing attributes on media elements, which fonts do not affect, and committing it would convert its storage from raw to an LFS pointer.

---

Credit for the diagnosis, the fix and three rounds of review response goes to @akzarma. Please review their code on its merits; if anything here is wrong, it is likely a baseline commit rather than the fix.

### DIFF
--- a/packages/producer/src/services/deterministicFonts-aliasSupplement.test.ts
+++ b/packages/producer/src/services/deterministicFonts-aliasSupplement.test.ts
@@ -1,0 +1,306 @@
+/**
+ * Regression test for cross-typeface alias supplementation.
+ *
+ * A family in FONT_ALIAS_MAP resolves to a canonical bundled typeface — e.g.
+ * `Noto Sans` → Inter. `buildFontFaceCss` emits the canonical's embedded faces
+ * under the authored family name, then queries Google Fonts to fill the weights
+ * and styles the bundle lacks.
+ *
+ * The bug: that supplementary query used the *authored* name, so a
+ * cross-typeface alias regained faces from the very typeface the alias exists to
+ * replace. No canonical bundle ships an italic face, so every italic Google
+ * served for the authored name was injected — `Noto Sans` came out as Inter
+ * upright plus real Noto Sans italic, two typefaces under one `font-family`.
+ *
+ * These tests inject `fetchImpl` (no network) and a temp `HYPERFRAMES_FONT_CACHE_DIR`
+ * so they are hermetic.
+ */
+
+import { afterAll, beforeAll, describe, expect, it } from "bun:test";
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { EMBEDDED_FONT_DATA } from "./fontData.generated.js";
+
+let cacheDir: string;
+let prevCacheEnv: string | undefined;
+
+beforeAll(() => {
+  prevCacheEnv = process.env.HYPERFRAMES_FONT_CACHE_DIR;
+  cacheDir = mkdtempSync(join(tmpdir(), "hf-font-alias-"));
+  process.env.HYPERFRAMES_FONT_CACHE_DIR = cacheDir;
+});
+
+afterAll(() => {
+  if (prevCacheEnv === undefined) delete process.env.HYPERFRAMES_FONT_CACHE_DIR;
+  else process.env.HYPERFRAMES_FONT_CACHE_DIR = prevCacheEnv;
+  rmSync(cacheDir, { recursive: true, force: true });
+});
+
+// One woff2 per typeface Google could serve, with identifiable bodies so every
+// injected `src` can be traced back to the family that was queried.
+const SERVED_FACES: Record<string, { url: string; bytes: string }> = {
+  Inter: {
+    url: "https://fonts.gstatic.com/s/inter/v1/inter-supplement.woff2",
+    bytes: "INTER_FETCHED_BYTES",
+  },
+  "Noto Sans": {
+    url: "https://fonts.gstatic.com/s/notosans/v1/notosans-supplement.woff2",
+    bytes: "NOTO_SANS_FETCHED_BYTES",
+  },
+  Montserrat: {
+    url: "https://fonts.gstatic.com/s/montserrat/v1/montserrat-supplement.woff2",
+    bytes: "MONTSERRAT_FETCHED_BYTES",
+  },
+};
+
+const b64 = (s: string) => Buffer.from(s).toString("base64");
+
+// 300 normal + 400 italic: no CANONICAL_FONTS bundle ships either, so both are
+// classified "missing from the bundle" and injected.
+function cssFor(family: string): string {
+  const url = SERVED_FACES[family]?.url;
+  if (!url) return "";
+  return `@font-face {
+  font-family: '${family}';
+  font-style: normal;
+  font-weight: 300;
+  src: url(${url}) format('woff2');
+}
+@font-face {
+  font-family: '${family}';
+  font-style: italic;
+  font-weight: 400;
+  src: url(${url}) format('woff2');
+}`;
+}
+
+function makeGoogleFetch(queriedFamilies: string[]): typeof fetch {
+  return (async (input: unknown) => {
+    const url = String(input);
+    if (url.startsWith("https://fonts.googleapis.com/")) {
+      const family = new URL(url).searchParams.get("family")?.split(":", 1)[0] ?? "";
+      queriedFamilies.push(family);
+      return new Response(cssFor(family), { status: 200 });
+    }
+    const served = Object.values(SERVED_FACES).find((face) => face.url === url);
+    if (served) return new Response(served.bytes, { status: 200 });
+    return new Response("", { status: 404 });
+  }) as unknown as typeof fetch;
+}
+
+function htmlRequesting(family: string): string {
+  return `<!doctype html><html><head><style>
+  h1 { font-family: "${family}"; }
+</style></head><body><h1>Upright</h1></body></html>`;
+}
+
+function injectedSrcs(html: string): string[] {
+  return [...html.matchAll(/src:\s*url\("([^"]+)"\)/g)].map((match) => match[1] ?? "");
+}
+
+function bundledUris(packageName: string): Set<string> {
+  return new Set(
+    [...EMBEDDED_FONT_DATA]
+      .filter(([key]) => key.startsWith(`${packageName}:`))
+      .map(([, uri]) => uri),
+  );
+}
+
+describe("aliased font-family supplementation", () => {
+  it("supplements a cross-typeface alias from the canonical family", async () => {
+    const { injectDeterministicFontFaces } = await import("./deterministicFonts.js");
+    const queriedFamilies: string[] = [];
+    const result = await injectDeterministicFontFaces(htmlRequesting("Noto Sans"), {
+      allowSystemFontCapture: false,
+      fetchImpl: makeGoogleFetch(queriedFamilies),
+    });
+
+    // "Noto Sans" aliases to Inter, so the supplementary query asks for Inter.
+    expect(queriedFamilies).toEqual(["Inter"]);
+
+    // The authored spelling still names the family, so authored CSS keeps matching.
+    expect(result).toContain('font-family: "Noto Sans"');
+
+    // Every injected face is Inter — either the embedded bundle or the Inter
+    // fetch. None carry the real Noto Sans the alias exists to replace.
+    const canonicalUris = bundledUris("@fontsource/inter");
+    canonicalUris.add(`data:font/woff2;base64,${b64(SERVED_FACES.Inter!.bytes)}`);
+    const srcs = injectedSrcs(result);
+    expect(srcs.length).toBeGreaterThan(0);
+    for (const src of srcs) expect(canonicalUris.has(src)).toBe(true);
+    expect(result).not.toContain(b64(SERVED_FACES["Noto Sans"]!.bytes));
+  });
+
+  it("collapses a variable font's shared source into one weight-range rule", async () => {
+    const { injectDeterministicFontFaces } = await import("./deterministicFonts.js");
+    // Google serves Inter as a variable font: one woff2 for every static weight.
+    const sharedUrl = "https://fonts.gstatic.com/s/inter/v1/inter-variable.woff2";
+    const css = [100, 200, 300, 500]
+      .map(
+        (weight) => `@font-face {
+  font-family: 'Inter';
+  font-style: normal;
+  font-weight: ${weight};
+  src: url(${sharedUrl}) format('woff2');
+}`,
+      )
+      .join("\n");
+    const fetchImpl = (async (input: unknown) => {
+      const url = String(input);
+      if (url.startsWith("https://fonts.googleapis.com/"))
+        return new Response(css, { status: 200 });
+      if (url === sharedUrl) return new Response("INTER_VARIABLE_BYTES", { status: 200 });
+      return new Response("", { status: 404 });
+    }) as unknown as typeof fetch;
+
+    const result = await injectDeterministicFontFaces(htmlRequesting("Noto Sans"), {
+      allowSystemFontCapture: false,
+      fetchImpl,
+    });
+
+    // One rule spanning the run, not four rules carrying the same blob.
+    const variableUri = `data:font/woff2;base64,${b64("INTER_VARIABLE_BYTES")}`;
+    // 100-300 collapse into one rule; 500 stays separate because the bundle
+    // serves 400 and a 100-500 range would shadow that embedded face.
+    const occurrences = result.split(variableUri).length - 1;
+    expect(occurrences).toBe(2);
+    expect(result).toContain("font-weight: 100 300;");
+    expect(result).toContain("font-weight: 500;");
+  });
+
+  it("collapses shared sources when Google interleaves subsets without text=", async () => {
+    const { injectDeterministicFontFaces } = await import("./deterministicFonts.js");
+    // Without `text=` Google orders the response weight-major, subset-minor,
+    // so the faces sharing one variable source are never adjacent.
+    const latinUrl = "https://fonts.gstatic.com/s/inter/v1/inter-latin-variable.woff2";
+    const latinExtUrl = "https://fonts.gstatic.com/s/inter/v1/inter-latinext-variable.woff2";
+    const LATIN = "U+0000-00FF";
+    const LATIN_EXT = "U+0100-024F";
+    const subsets = [
+      { url: latinUrl, range: LATIN },
+      { url: latinExtUrl, range: LATIN_EXT },
+    ];
+    const css = [100, 200, 300, 500]
+      .flatMap((weight) =>
+        subsets.map(
+          (subset) => `@font-face {
+  font-family: 'Inter';
+  font-style: normal;
+  font-weight: ${weight};
+  src: url(${subset.url}) format('woff2');
+  unicode-range: ${subset.range};
+}`,
+        ),
+      )
+      .join("\n");
+
+    const fetchImpl = (async (input: unknown) => {
+      const url = String(input);
+      if (url.startsWith("https://fonts.googleapis.com/")) {
+        // The request must not carry `text=`, or Google would return one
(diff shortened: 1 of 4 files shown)


miguel-heygen (changes requested):
Audited exact head `3867349ef57b08d96a444adda214219875c7c295` against `e0ba41c0245356e7a8690db959df9323a0631fdd`: the three-file diff, the repaired force-push delta, alias resolution, source grouping, covered-weight partitioning, subset-order preservation, the new hermetic tests, both signed commits, and the LFS baseline object. Trusting only the author measurements against live Google catalog responses; the source and CI claims below were independently checked.

The force-push repair is correct. The current-main top-level-comma parser is present, the `var(--x, fallback)` regression is gone, and exact-head Producer unit/integration, build, lint, and typecheck lanes are green. The new `style-15-prod` LFS object fully decodes, its decoded audio is bit-identical to `main` after removing the old AAC priming/tail padding, and its dedicated distributed-simulated shard passes.

**Blocker — `style-3-prod` also records the old mixed-typeface output, but its baseline was not re-recorded.**

Required exact-head regression CI fails in [shard 3](https://github.com/heygen-com/hyperframes/actions/runs/31541750913/job/93945933868): `style-3-prod` misses 7/100 visual checkpoints from 0.80s through 1.76s, each at PSNR ~27.4. This is the same correction the PR intentionally accepts for `style-15-prod`, not an unrelated flake:

- `packages/producer/tests/style-3-prod/src/index.html:17` requests `"Helvetica", Arial, sans-serif`.
- The fixture also contains `Helvetica Neue`; all three route to Inter and the exact-head logs show three canonical Inter supplementation fetches.
- The failure artifact changes the rendered glyph metrics in the title/subtitle while the non-text geometry stays fixed.

Please re-record and include the `style-3-prod` golden too, validate it with the fixture mode from the shard schedule, and rerun required regression CI. No source-code blocker remains on my read.

**Verdict:** REQUEST CHANGES

**Reasoning:** The canonical-family fix and its tests are correct at the repaired head, and the intended `style-15-prod` baseline passes, but one additional required golden still encodes the behavior this PR removes and leaves required CI red.

— Magi

miguel-heygen (approved):
Audited the exact incremental delta `3867349ef57b08d96a444adda214219875c7c295 → dedf0a1fe7252f3ac9381513fe4be6b4badc4fbf` and carrying forward my full source/test audit from the prior review. The increment is only the `style-3-prod` LFS baseline pointer; nothing in the already-reviewed implementation changed.

Independent evidence for the replacement baseline:

- The LFS object matches its declared SHA-256/size and fully decodes as the expected 1920×1080/30 fps H.264 + 48 kHz stereo AAC fixture.
- Its seven corrected checkpoints match the actual frames from the prior failing shard at 52.0–53.3 dB PSNR. New-vs-old reproduces exactly the previous seven failures at 27.39–27.41 dB, with the delta confined to glyph outlines; content, position, colour, layout, backgrounds, and shapes remain unchanged.
- Audio remains well above the fixture contract after AAC-priming alignment, and the new object has the same 482-frame/16.0667s video contract.
- The new commit is signed, parented directly on the previously audited head, and preserves the contributor-authored fix below it.

Exact-head required CI is now entirely terminal green: regression (including shard 3 / `style-3-prod` and shard 9 / `style-15-prod`), Test, Typecheck, runtime contract, Build, semantic title, and both Windows lanes.

**Verdict:** APPROVE

**Reasoning:** The only remaining blocker was the stale `style-3-prod` golden. The replacement is provenance-matched to the render that exposed the bug, its media contract is sound, and the complete exact-head required suite now passes.

— Magi

PR #3231, opened 2026-08-11
chore(cli): bump pinned Chromium to 152.0.7977.30

vanceingalls:
**Stack 1/3** — base of the drawElement fixes that follow.

Bumps the pinned Chromium to pick up crbug 522872457's fix (CL 8032671), which landed after the 152.0.7935.0 canary cut and so post-dated the old 152.0.7928.2 pin.

Beta rather than Canary deliberately: 153.0.8000.0 measured identical to this build on every probe variant, so crossing a major buys nothing measurable. Verified `chrome-headless-shell` ships for mac-x64, mac-arm64, win32, win64 and linux64 at this version, so the bump cannot brick downloads on a platform not tested here.

### What the new build fixes, and what it does not

Re-probed every 3D signal the compile gate matches, `drawElementImage` vs a CDP screenshot of the identical state, on the shipping headless-shell binary:

| variant | old pin | new pin | |
|---|---|---|---|
| sibling content dropped from 3D captures | broken | **fixed** | |
| background lost from 3D captures | broken | **fixed** | |
| `backface-visibility: hidden` ignored | 1.4 dB | 14.8 dB | still under the 32 dB floor |

The upstream fix repaired the collateral damage only. That is recorded in the comment above `CHROME_VERSION` so the next person does not re-run the probe.

Stack:
1. **this PR** — the pin
2. #3232 — inverted depth sign in 3D projection
3. #3172 — per-frame deadline so a wedged renderer falls back instead of failing the render

### DIFF
--- a/packages/cli/src/browser/manager.ts
+++ b/packages/cli/src/browser/manager.ts
@@ -20,7 +20,22 @@ async function loadPuppeteerBrowsers(): Promise<PuppeteerBrowsers> {
   }
 }
 
-const CHROME_VERSION = "152.0.7928.2";
+// Bumped from 152.0.7928.2 on 2026-08-10 to pick up crbug 522872457's fix
+// (CL 8032671, "Force-merge PendingLayer for canvas child descendant", merged
+// 2026-07-07 — after the 152.0.7935.0 canary cut, so the old pin predated it).
+//
+// Measured on the shipping headless-shell binary, drawElementImage vs a CDP
+// screenshot of the identical state:
+//   sibling content dropped from 3D captures  — FIXED
+//   background lost from 3D captures          — FIXED
+//   backface-visibility:hidden ignored        — STILL BROKEN (1.4 dB -> 14.8 dB;
+//                                               better, still far under the
+//                                               32 dB floor)
+// So the 3D compile gate must stay. See PRINFRA-486.
+//
+// Deliberately Beta, not Canary. 153.0.8000.0 measured identical to this build
+// on every probe variant, so crossing a major buys nothing measurable.
+const CHROME_VERSION = "152.0.7977.30";
 const CACHE_ROOT_DIR = join(homedir(), ".cache", "hyperframes");
 const CACHE_DIR = join(homedir(), ".cache", "hyperframes", "chrome");
 // Puppeteer's managed cache — where `@puppeteer/browsers install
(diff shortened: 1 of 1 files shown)


vanceingalls (commented):
**R1 (COMMENT)** — Chromium pin bump. Single-file diff, well-motivated by the crbug 522872457 fix. No blockers; two non-blockers to note.

### Non-blockers

1. `packages/cli/src/browser/manager.ts:23` — the comment claims each 3D-signal probe was re-measured on the shipping headless-shell. Great narrative; not source-checkable from the diff. When you re-cut a pin later, capturing the `@puppeteer/browsers install chrome-headless-shell@<v>` output for each of the five platforms as a link in the PR body would make future bumps a mechanical re-run.

2. Standing concern (version rolloff rate drift, from prior bumps): this changes install-time behavior on every user's next `hyperframes init`. There is no explicit success-rate monitor tied to the pin, so a platform-specific tarball drop would surface only through customer reports. Not this PR's job to add — just worth watching the CI matrix on first landing and any local-render telemetry that touches `chrome-headless-shell install` for a week.

### Verified
- `CHROME_VERSION` is the only pin site; no lockfile / fixture / snapshot deltas expected and none present.
- The PSNR-floor prescription (`3D compile gate must stay`) is spelled out inline at the pin, consistent with PRINFRA-486.

Status: `MERGEABLE` / `mergeStateStatus: BLOCKED` (CODEOWNER approval pending). CI: all latest-per-name checks SUCCESS.

— Via

terencecho (approved):
Clean single-value Chromium pin bump (`152.0.7928.2` -> `152.0.7977.30`) with an inline changelog above `CHROME_VERSION` and a channel-descriptor tweak (Dev -> pre-Stable). No lockfile / snapshot / fixture deltas — the pin is the entire surface, verified single-site.

Concur with Via's COMMENT: no blockers, non-blockers are house-keeping asks for future bumps rather than gates on this one. CI at head SHA: 31 SUCCESS, 10 SKIPPED, 0 failing across 41 unique latest-per-name checks. Approving.

— Review by tai (pr-review)

miga-heygen (approved):
**Approve.** Chromium bump from 152.0.7928.2 to 152.0.7977.30. Picks up crbug 522872457 fix with measured results documented in-comment. Staying on Beta (not Canary) — 153.0.8000.0 measured identical. Clean.

— Miga

PR #3232, opened 2026-08-11
fix(engine): correct the inverted depth sign in 3D projection

vanceingalls:
**Stack 2/3** — based on #3231. Review that first; this diff is one line of behaviour plus its rationale.

`threeDProjection` **does** run a depth test — `gl.enable(gl.DEPTH_TEST)` with `clearDepth(1)` and `depthFunc(LEQUAL)`. The comparison was inverted: CSS puts +z toward the viewer, GL treats larger ndc z as *farther*, and the ndc matrix passed CSS z through with a **positive** scale. So the nearest quad received the largest depth, lost the test, and was occluded by the quad behind it.

The y-flip on the row above changes handedness. That was already compensated for winding (`frontFace(gl.CW)`) but never for depth — which is why it only surfaced once two quads in one context could be visible simultaneously. A single-quad context has nothing to lose a depth test against.

### Measurement

Two planes at ±45° in one `preserve-3d` context, drawElement render vs a screenshot render of the same comp:

| | before | after |
|---|---|---|
| two planes ±45°, one preserve-3d context | **18.8 dB** — near plane painted behind the far one, labels and colours swapped | **51.2 dB**, visually identical to the screenshot arm |
| backface flip card at rest | 54.5 dB | 54.5 dB |
| perspective + rotationX | 51.1 dB | 51.1 dB |
| matrix3d() | 49.5 dB | 49.5 dB |

Single-quad cases are unaffected by the negation, as expected.

### Two things reviewers should know

**No unit test, and that is structural.** `initThreeDProjectionInPage` is contractually self-contained — shipped into the page via `page.evaluate`, no outer-scope references — so the matrix is unreachable from a test without breaking that contract. Verified at render level instead; the method is recorded at the call site so it can be repeated.

**The self-verify net cannot catch a regression here.** It captures ground-truth frames *after* this rewrite mutates the DOM, so it compares projected-DOM-via-drawElement against projected-DOM-via-screenshot — both carrying the same geometry. Measured: it passed the broken comp at 65 dB while the true output was 18.8 dB off. Do not treat it as the backstop for 3D changes.

### Scope note

This does **not** open up any renders. The 3D compile gate still clamps `perspective` / `preserve-3d` comps before capture, so projection is only reachable via `HF_FAST_CAPTURE_3D`. An earlier revision of this stack narrowed that gate; it was reverted after review found the gate is load-bearing for six pre-existing projection limitations this fix does not address (group-root paint, layout-animating quads, animated perspective, flat intermediates, self-quad ordering, mid-loop bails). Those are catalogued on PRINFRA-486 as the prerequisites for narrowing.

Refs PRINFRA-486

### DIFF
--- a/packages/cli/src/browser/manager.ts
+++ b/packages/cli/src/browser/manager.ts
@@ -20,7 +20,22 @@ async function loadPuppeteerBrowsers(): Promise<PuppeteerBrowsers> {
   }
 }
 
-const CHROME_VERSION = "152.0.7928.2";
+// Bumped from 152.0.7928.2 on 2026-08-10 to pick up crbug 522872457's fix
+// (CL 8032671, "Force-merge PendingLayer for canvas child descendant", merged
+// 2026-07-07 — after the 152.0.7935.0 canary cut, so the old pin predated it).
+//
+// Measured on the shipping headless-shell binary, drawElementImage vs a CDP
+// screenshot of the identical state:
+//   sibling content dropped from 3D captures  — FIXED
+//   background lost from 3D captures          — FIXED
+//   backface-visibility:hidden ignored        — STILL BROKEN (1.4 dB -> 14.8 dB;
+//                                               better, still far under the
+//                                               32 dB floor)
+// So the 3D compile gate must stay. See PRINFRA-486.
+//
+// Deliberately Beta, not Canary. 153.0.8000.0 measured identical to this build
+// on every probe variant, so crossing a major buys nothing measurable.
+const CHROME_VERSION = "152.0.7977.30";
 const CACHE_ROOT_DIR = join(homedir(), ".cache", "hyperframes");
 const CACHE_DIR = join(homedir(), ".cache", "hyperframes", "chrome");
 // Puppeteer's managed cache — where `@puppeteer/browsers install
@@ -332,7 +347,7 @@ async function findFromCache(): Promise<CacheLookupResult> {
   // this is the non-`preferManagedChrome` path, which exists so a user who
   // installed chrome-headless-shell separately (via `@puppeteer/browsers
   // install`) keeps using that binary instead of being silently switched to
-  // the HF-pinned one. Note `CHROME_VERSION` (above) is a Dev-channel pin
+  // the HF-pinned one. Note `CHROME_VERSION` (above) is a pre-Stable pin
   // that may be NEWER than a user's puppeteer-cache Stable build — this is
   // about respecting an explicit prior choice, not "newest wins".
   const fromPuppeteer = findFromPuppeteerCache();
--- a/packages/cli/src/commands/render.ts
+++ b/packages/cli/src/commands/render.ts
@@ -1514,6 +1514,7 @@ function trackRenderMetrics(
     deBlankRecaptures: perf?.drawElement?.blankRecaptures,
     deBoundaryFrames: perf?.drawElement?.boundaryFrames,
     deNcprFallbacks: perf?.drawElement?.ncprFallbacks,
+    deFrameTimeouts: perf?.drawElement?.frameTimeouts,
     compositionDurationMs,
     compositionWidth: perf?.resolution.width,
     compositionHeight: perf?.resolution.height,
--- a/packages/cli/src/telemetry/events.ts
+++ b/packages/cli/src/telemetry/events.ts
@@ -235,6 +235,7 @@ export function trackRenderComplete(
     deBlankRecaptures?: number;
     deBoundaryFrames?: number;
     deNcprFallbacks?: number;
+    deFrameTimeouts?: number;
     // "cli" when triggered by `hyperframes render` (default), "studio" when
     // triggered by a studio preview-server render (POST /api/projects/:id/render).
     source?: "cli" | "studio";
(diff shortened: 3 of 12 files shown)


vanceingalls (commented):
**R1 (COMMENT)** — Depth-sign inversion in `initThreeDProjectionInPage`. One-line negation with a load-bearing comment. Verified the enclosing GL setup matches the comment's premise: `gl.clearDepth(1)`, `gl.enable(gl.DEPTH_TEST)`, `gl.depthFunc(gl.LEQUAL)` at the top of `update()`, and `gl.frontFace(gl.CW)` at group setup. `Z_SCALE = 1/100000` at line 734; grep of the file shows the sign is only consumed at the one edited site — no other callers depend on it. No blockers; one non-blocker.

### Non-blocker

1. `packages/engine/src/services/threeDProjection.ts:786-816` — the comment explicitly documents that the self-verify net cannot catch a re-inversion of this sign (it captures ground truth *after* this rewrite, so it would compare `wrong geometry vs wrong geometry`, 65 dB against a truly-broken 18.8 dB render). Since a unit test is architecturally excluded ("shipped into the page via `page.evaluate`, no outer-scope references"), the only backstop that survives is a regression comp. Recommend adding a two-plane ±45° `preserve-3d` composition to a regression shard (`regression-shards` in `.github/workflows/regression.yml`), scored against the screenshot arm, so a future re-inversion trips CI at the render-parity level the self-verify net cannot reach. Not required for merge — but the class of regression is now knowingly uncovered.

### Scope check
- No adjacent knob was widened. The 3D compile gate still clamps `perspective` / `preserve-3d` off fast capture; projection is only reachable via `HF_FAST_CAPTURE_3D`. The PR body's scope note tracks the earlier revision that tried to narrow that gate and reverted after finding six pre-existing projection limitations catalogued on PRINFRA-486. Good discipline.
- Sign is scoped: single site of consumption for `Z_SCALE`, verified.

Status: `MERGEABLE` / `mergeStateStatus: CLEAN`. CI green.

— Via

terencecho (commented):
**R2 (COMMENT)** — Concurring with Via's R1 on the depth-sign fix + independent verification of the flagged residual.

### Fix verification

Confirmed the one-line change at `packages/engine/src/services/threeDProjection.ts` inside `initThreeDProjectionInPage`'s ndc matrix: `Z_SCALE` → `-Z_SCALE`. Rationale checks out against the enclosing GL state — `clearDepth(1)` + `depthFunc(LEQUAL)` on a context where the y-flip on the row above already flipped handedness (compensated for winding by `frontFace(gl.CW)` but never for depth). Inverting `Z_SCALE` flips the sign of the ndc z output so nearer CSS +z now produces smaller ndc z and wins the `LEQUAL` test. `Z_SCALE` is consumed only at this one site, so the sign has no other downstream. Fix is correct in-direction; measured deltas (18.8 → 51.2 dB two-plane, single-quad cases unchanged) are consistent with a single-sign inversion at a single site.

### Concur on Via's non-blocker (regression-coverage gap)

The load-bearing comment ships an explicit admission: `the self-verify net CANNOT catch a regression here, because it captures ground truth after this rewrite and would compare wrong geometry against wrong geometry`. So a future re-inversion at this site is knowingly uncovered by the existing self-verify path — the baseline is DOM-post-rewrite, so both arms would carry the same wrong geometry (as measured: 65 dB pass on a truly-broken 18.8 dB output).

The PR body's scope note cites `PRINFRA-486`, but that ticket enumerates six *other* projection limitations (group-root paint, layout-animating quads, animated perspective, flat intermediates, self-quad ordering, mid-loop bails) as prerequisites for narrowing the 3D compile gate — it does NOT scope the two-plane ±45° regression comp Via recommended. So the class of regression the comment names as uncovered is currently unscoped.

Recommend opening a dedicated follow-up: a two-plane ±45° `preserve-3d` composition added to a `regression-shards` shard, drawElement-vs-screenshot dB-scored, so a re-inversion (or any regression at this site) trips CI at the render-parity level the self-verify net cannot reach. Not blocking — Vance is on record about the architectural constraint (`initThreeDProjectionInPage` is contractually self-contained via `page.evaluate`, unit-test-hostile) — but this is the first PR where the comment itself names the gap and it is worth a scoped ticket rather than an implicit note.

### CI + merge state

- `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`, `reviewDecision:` empty.
- Stack-child reduced matrix as expected: `preview-regression / Preview parity` SKIPPED, `Player perf / Perf` SKIPPED, `Preflight (lint + format)` SKIPPED on Player perf + preview-regression. Full workflow runs green on `regression` (Preflight + regression-shards 1-9 + summary), plus `player-perf`, `preview-regression`, WIP, Mintlify.
- Head SHA `564caa5`.

— Review by tai (pr-review)

miga-heygen (dismissed):
**Approve.** Single-character fix (Z_SCALE → -Z_SCALE) with an excellent explanatory comment. CSS/GL depth convention mismatch caused nearest quad to lose depth test. Measured: 18.8 dB → 51.2 dB.

**Known limitation (non-blocking, echoing Via):** the self-verify net captures ground truth after this rewrite, so a future depth-sign re-inversion would compare wrong geometry against wrong geometry. A two-plane ±45° preserve-3d regression comp at parity level would catch it.

— Miga

miga-heygen:

PR #3233, opened 2026-08-11
fix(producer): seek once per step when discovering video visibility

valeriangalliat:
## What

Make `discoverVideoVisibilityFromTimeline` seek the GSAP timeline once per sample time and read every auto-start video at that step, instead of re-walking the full timeline for each video.

## Why

In `browser_probe`, video visibility discovery currently does a full-timeline seek-scan **per** `video[data-hf-auto-start]`. So probe time scales with **video count × duration**, not just duration.

We've had video-heavy compositions that blew past the ~5 minute `protocolTimeout` on CPU/SwiftShader cloud renders because of this probe. Example: a ~6 min timeline with 6 auto-start videos @ 0.1s step is ~21.6k GSAP seeks before (`V × D / 0.1`); after this fix it's ~3.6k coarse seeks (`D / 0.1`) plus a handful of binary-refine seeks per video.

## How

**Before:** for each auto-start video, for `t = 0…duration` step `0.1s` → seek → read opacity  
→ coarse seeks ≈ `O(V × D / step)` (plus the same per-video binary refine as today)

**After:** for each `t`, seek once, then sample every video  
→ coarse seeks ≈ `O(D / step)`  
→ binary boundary refine still `O(V × log(step / precision))` (cheap; unchanged)

So the expensive part no longer multiplies by video count. Same sample grid and refine, just batched — visibility windows should match the old behavior.

## Test plan

- [x] Unit tests added/updated — `discoverVideoVisibilityFromTimeline` covers multi-video visibility windows (behavior only; passes before/after this change)
- [ ] Documentation updated (if applicable) — n/a

### DIFF
--- a/packages/producer/src/services/htmlCompiler.test.ts
+++ b/packages/producer/src/services/htmlCompiler.test.ts
@@ -18,6 +18,7 @@ import {
   detectThreeDTransformUsage,
   discoverMediaFromBrowser,
   discoverAudioVolumeAutomationFromTimeline,
+  discoverVideoVisibilityFromTimeline,
   inlineExternalScripts,
   localizeRemoteMediaSources,
   localizeRemoteImageSources,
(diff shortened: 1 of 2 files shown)


jrusso1020 (commented):
Maintainer/additive review at exact head `7eb264b911097257b010abbd12a30f78839025d7`.

**Strengths**
- `packages/producer/src/services/htmlCompiler.ts:2349` batches the dominant GSAP seek exactly where intended while preserving per-video opacity sampling and boundary refinement.
- `packages/producer/src/services/htmlCompiler.ts:2380` retains the existing 1/60s binary refinement and final timeline reset for normal compiled videos.
- Corrected GCP Cloud Run A/B validation found identical five-video visibility windows and reduced median scan-only time from 376ms to 235ms (37.6%).

**Important**
- `packages/producer/src/services/htmlCompiler.test.ts:2205` verifies output windows, but as the PR body notes, it passes before this change. It also permits ±0.1s, so it would pass if the 1/60s refinement were accidentally removed. I have a focused test ready that uses invisible videos to isolate the coarse loop, proves one and five videos issue the same seek sequence, asserts `suppressEvents=true` and the final reset, and tightens window checks to 1/60s. It passes the full file (126 tests), oxlint, and oxfmt.

**Nit**
- `packages/producer/src/services/htmlCompiler.ts:2367` returns before `seekTl(0)` when every matching sentinel video lacks an id, whereas the old loop still reset. The compiler always injects IDs before this sentinel, so I could not establish a normal production failure; preserving the old postcondition would still make equivalence exact.

The original integration red was infrastructure-only: the job was cancelled after 20 minutes in “Install FFmpeg,” before tests ran. I reran the failed jobs; that rerun is currently in progress.

This is a cross-repository PR with `maintainerCanModify=false`, so I cannot push the verified test to Val’s branch. @valeriangalliat, could you enable “Allow edits and access to secrets by maintainers”? I’ll add only the regression coverage to this existing PR so authorship stays with you.

**Verdict:** COMMENT
**Reasoning:** The production optimization is correct and has no code blocker; approval is waiting on the required CI rerun and the regression-test handoff.

— Codex

miguel-heygen (approved):
Independent review at exact head `7eb264b911097257b010abbd12a30f78839025d7`.

**Strengths**
- `packages/producer/src/services/htmlCompiler.ts:2349-2378` moves only the dominant coarse GSAP seek outside the per-video loop; opacity remains sampled independently for every scene at the same timestamps.
- `packages/producer/src/services/htmlCompiler.ts:2380-2416` preserves per-video 1/60s boundary refinement, `suppressEvents=true`, and the final timeline reset.
- `packages/producer/src/services/htmlCompiler.test.ts:2204-2271` exercises four overlapping visibility windows and restores global test state on every exit.

**Independent verification**
- Full `htmlCompiler.test.ts`: 125 passed, 0 failed.
- Separate seek-scaling probe: one and five invisible videos produced the identical coarse seek sequence `[0, 0.1, 0.2, 0.3, 0]`; every seek suppressed events and the last seek reset to zero.
- All current required checks are green, including the rerun `Test` aggregate and producer integration.

I concur with Rames’s important coverage improvement: the committed behavior test does not itself pin seek-count scaling or the 1/60s refinement. That is worth adding when maintainer edits are available, but it is not a production-code blocker after the exact-head proof above.

**Verdict:** APPROVE
**Reasoning:** The batching changes only seek multiplicity, preserves per-video sampling/refinement and reset semantics, and passes both independent focused verification and the complete required CI gate.

— Magi

valeriangalliat:
Thanks for reviewing and merging! @jrusso1020 I just realized since I forked from another org I can't allow edit from maintainers (https://github.com/orgs/community/discussions/5634) - I'll make future PRs from a personal fork in order to enable this 👍 

Happy to redo my other pending ones that way but authorship is nbd to me so feel free to re-create them if that makes sense too

PR #3238, opened 2026-08-12
fix(engine,producer,lint): resolve <source> children for media extract and localize

### LINKED ISSUE #3495: <video> / <audio> with only <source src> children skip extract and mix
### Describe the bug

A `<video>` or `<audio>` that supplies its URL only via a child `<source src>` (no parent `src`) plays in preview and is skipped at export.

The runtime already resolves `el.currentSrc || el.src`. Engine extract and mix still query `video[src]` / `audio[src]`, so those elements are never extracted or mixed. Lint `media_missing_src` agrees with the engine and tells you to put `src` on the parent.

Common HTML multi-format pattern. Parent `src` can stay preferred for Studio bind / variables; child `<source>` should still count as a resolvable src.

On 0.8.14 the skipped video is **blank**, not frozen. A native first-frame paint (frozen magenta) is the other face of the same skip — no frames extracted or injected. Audio is always missing.

### Link to reproduction

https://github.com/ArcadeHQ/hyperframes-repros/tree/patch/video-source

### Steps to reproduce

1. Clone https://github.com/ArcadeHQ/hyperframes-repros and check out `patch/video-source`.
2. Run `npx hyperframes@latest check` — `media_missing_src` on `#source-only`.
3. Run `npx hyperframes@latest render --output renders/source-only.mp4`.
4. Grab frames: `ffmpeg -y -ss 0.3 -i renders/source-only.mp4 -frames:v 1 snapshots/frame-00-at-0.3s.png` and the same at `1.5s`.
5. Run `ffprobe -hide_banner renders/source-only.mp4`.

Same 2s clip twice (magenta 0–1s, cyan 1–2s, 440 Hz). Left: parent `src`, muted. Right: `<source>` only, audible.

### Expected behavior

- `t=0.3s`: both panes magenta
- `t=1.5s`: both panes cyan
- Output has an audio stream (~440 Hz)

### Actual behavior

- `t=0.3s`: left magenta; **right black**
- `t=1.5s`: left cyan; **right still black**
- `ffprobe`: video stream only — **no audio**
- Compile: `videoCount: 1, audioCount: 0` with `authoredTimedClipCount: 3`

See `snapshots/frame-00-at-0.3s.png` and `snapshots/frame-01-at-1.5s.png`.

`hyperframes snapshot` (browser path) plays both panes. The gap is export.

### Environment

```
hyperframes doctor

  ✓ Version          0.8.14 (latest)
  ✓ Memory           16.0 GB total · 4.9 GB available
  ✓ Disk             363.3 GB free
  ✓ Frames cache     /var/folders/bl/c4hk6r7n34g1d81txd99gsgr0000gn/T/hyperframes-extract-cache-501 · 363.3 GB free at /var/folders/bl/c4hk6r7n34g1d81txd99gsgr0000gn/T/hyperframes-extract-cache-501 · default
  ✓ Archive extractor unzip
  ✓ Environment      non-TTY
  ✗ Docker           Not found
```

### Additional context

Same over-narrow-selector family as `tests/sub-composition-video` and `tests/video-hfid-no-id`. Runtime half: `mediaProxy.ts` `el.currentSrc || el.src`.

Fix already in https://github.com/heygen-com/hyperframes/pull/3238 — this issue is so that PR can close it.

valeriangalliat:
## What

Support `<source src>` children on `<video>` / `<audio>` in extract, audio mix, remote-media localize, and lint.

Closes #3495.

## Why

Markup that only uses `<source>` (no parent `src`) was skipped, so those clips rendered blank or silent. Common multi-format HTML pattern; parent `src` stays preferred for Studio bind / variables.

## How

Shared resolve: parent `src` if present, else a child `<source src>` (prefer local over http when both exist). Same rule in engine parse, producer localize regex, and `media_missing_src` / `media_missing_data_start`.

## Test plan

- [ ] Documentation updated (if applicable) — n/a

### DIFF
--- a/packages/engine/src/services/audioMixer.test.ts
+++ b/packages/engine/src/services/audioMixer.test.ts
@@ -80,6 +80,28 @@ describe("parseAudioElements strict literal timing", () => {
   );
 });
 
+describe("parseAudioElements — <source> children", () => {
+  it("discovers audio and audible-video tracks that use <source> children", () => {
+    const tracks = parseAudioElements(`
+      <audio id="bgm" data-start="2" data-end="7">
+        <source src="https://cdn.example.com/bgm.mp3" type="audio/mpeg">
+        <source src="_remote_media/bgm.ogg" type="audio/ogg">
+      </audio>
+      <video id="rec" data-has-audio="true" data-start="4" data-end="10">
+        <source src="https://cdn.example.com/rec.mp4" type="video/mp4">
+        <source src="_remote_media/rec.webm" type="video/webm">
+      </video>
+    `);
+    expect(tracks).toEqual([
+      expect.objectContaining({ id: "bgm", src: "_remote_media/bgm.ogg", type: "audio" }),
+      expect.objectContaining({
+        id: "rec-audio",
+        src: "_remote_media/rec.webm",
+        type: "video",
+      }),
+    ]);
+  });
+});
 describe("processCompositionAudio", () => {
   const tempDirs: string[] = [];
 
--- a/packages/engine/src/services/audioMixer.ts
+++ b/packages/engine/src/services/audioMixer.ts
@@ -20,7 +20,7 @@ import {
 import { DEFAULT_CONFIG, type EngineConfig } from "../config.js";
 import { formatFfmpegError, runFfmpeg, type RunFfmpegResult } from "../utils/runFfmpeg.js";
 import { unwrapTemplate } from "../utils/htmlTemplate.js";
-import { resolveProjectRelativeSrc } from "./videoFrameExtractor.js";
+import { resolveMediaElementSrc, resolveProjectRelativeSrc } from "./videoFrameExtractor.js";
 import { resolveReferencedStart, type RefResolverEl } from "./referenceResolver.js";
 import { isKnownInactiveTimelineWindow } from "./mediaTimelineWindow.js";
 import type {
@@ -512,7 +512,12 @@ export function parseAudioElements(html: string): AudioElement[] {
   // <audio> and <video data-has-audio> tracks differ only in the emitted id
 
   // and `type`; everything else (timing, layer, volume) is read identically.
-  const build = (el: RefResolverEl, id: string, type: AudioElement["type"]): AudioElement => {
+  const build = (
+    el: RefResolverEl,
+    id: string,
+    src: string,
+    type: AudioElement["type"],
+  ): AudioElement => {
     const playbackRateAttr = el.getAttribute("data-playback-rate");
     const layerAttr = el.getAttribute("data-layer");
     const volumeAttr = el.getAttribute("data-volume");
@@ -524,7 +529,7 @@ export function parseAudioElements(html: string): AudioElement[] {
     const group = groupId ? groupsById.get(groupId) : undefined;
     return {
       id,
-      src: el.getAttribute("src") as string,
+      src,
       start: resolveStart(el),
       end: parseEnd(el.getAttribute("data-end")),
       mediaStart: readMediaStart(el),
@@ -553,20 +558,22 @@ export function parseAudioElements(html: string): AudioElement[] {
   const trackId = (el: RefResolverEl): string | null =>
     el.getAttribute(MEDIA_RENDER_ID_ATTR) || el.getAttribute("id");
 
-  for (const el of document.querySelectorAll("audio[id][src]")) {
+  for (const el of document.querySelectorAll("audio[id]")) {
     const id = trackId(el);
+    const src = resolveMediaElementSrc(el);
     // `memberGroupHidden` is the group's own mute: a hidden BUS drops every
     // member from the mix, the same way `isHidden` drops one track.
-    if (!id || !el.getAttribute("src") || isHidden(el) || memberGroupHidden(el)) continue;
+    if (!id || !src || isHidden(el) || memberGroupHidden(el)) continue;
     if (isKnownInactiveTimelineWindow(el, resolveStart(el))) continue;
-    elements.push(build(el, id, "audio"));
+    elements.push(build(el, id, src, "audio"));
   }
 
-  for (const el of document.querySelectorAll('video[id][src][data-has-audio="true"]')) {
+  for (const el of document.querySelectorAll('video[id][data-has-audio="true"]')) {
     const id = trackId(el);
-    if (!id || !el.getAttribute("src") || isHidden(el)) continue;
+    const src = resolveMediaElementSrc(el);
+    if (!id || !src || isHidden(el)) continue;
     if (isKnownInactiveTimelineWindow(el, resolveStart(el))) continue;
-    elements.push(build(el, `${id}-audio`, "video"));
+    elements.push(build(el, `${id}-audio`, src, "video"));
   }
 
   return elements;
--- a/packages/engine/src/services/videoFrameExtractor.test.ts
+++ b/packages/engine/src/services/videoFrameExtractor.test.ts
@@ -860,6 +860,17 @@ describe("parseVideoElements", () => {
       expect(Number.isNaN(v.start)).toBe(false);
     }
   });
+
+  it("discovers <video> elements that use <source> children", () => {
+    const videos = parseVideoElements(
+      '<video id="rec" data-start="1" data-duration="4">' +
+        '<source src="https://cdn.example.com/rec.mp4" type="video/mp4">' +
+        '<source src="_remote_media/rec.webm" type="video/webm">' +
+        "</video>",
+    );
+    expect(videos).toHaveLength(1);
+    expect(videos[0]).toMatchObject({ id: "rec", src: "_remote_media/rec.webm", start: 1, end: 5 });
+  });
 });
 
 describe("FrameLookupTable", () => {
--- a/packages/engine/src/services/videoFrameExtractor.ts
+++ b/packages/engine/src/services/videoFrameExtractor.ts
@@ -527,16 +527,39 @@ export interface ExtractionResult {
   phaseBreakdown: ExtractionPhaseBreakdown;
 }
 
+/** Minimal structural shape for resolving parent/`<source>` media `src`. */
+interface MediaSrcEl {
+  getAttribute(name: string): string | null;
+  querySelectorAll(selectors: string): Iterable<{ getAttribute(name: string): string | null }>;
+}
+
+/**
+ * Parent `src`, else a `<source src>`. Prefer local paths over http(s) so a
+ * localized sibling wins when another `<source>` failed to download.
+ */
+export function resolveMediaElementSrc(el: MediaSrcEl): string | null {
+  const direct = el.getAttribute("src");
+  if (direct) return direct;
+  let remote: string | null = null;
+  for (const source of el.querySelectorAll("source")) {
+    const src = source.getAttribute("src");
+    if (!src) continue;
+    if (!/^https?:\/\//i.test(src)) return src;
+    remote ??= src;
+  }
+  return remote;
+}
+
 export function parseVideoElements(html: string): VideoElement[] {
   const videos: VideoElement[] = [];
   const { document } = parseHTML(unwrapTemplate(html));
   const startCache = new Map<RefResolverEl, number>();
   const visiting = new Set<RefResolverEl>();
 
-  const videoEls = document.querySelectorAll("video[src]");
+  const videoEls = document.querySelectorAll("video");
   let autoIdCounter = 0;
   for (const el of videoEls) {
-    const src = el.getAttribute("src");
+    const src = resolveMediaElementSrc(el);
     if (!src) continue;
     // Generate a stable ID for videos without one — the producer needs IDs
     // to track extracted frames and composite them during encoding.
--- a/packages/lint/src/rules/media.test.ts
+++ b/packages/lint/src/rules/media.test.ts
@@ -217,6 +217,37 @@ describe("media rules", () => {
     expect(finding?.severity).toBe("error");
   });
 
+  it("accepts <source src> children in place of parent src", async () => {
+    const html = `
+<html><body>
+  <div id="root" data-composition-id="c1" data-width="1920" data-height="1080">
+    <video id="rec" data-start="0" data-duration="4" muted playsinline>
+      <source src="clip.mp4" type="video/mp4">
+      <source src="clip.webm" type="video/webm">
+    </video>
+  </div>
+  <script>window.__timelines = window.__timelines || {}; window.__timelines["c1"] = gsap.timeline({ paused: true });</script>
+</body></html>`;
+    const result = await lintHyperframeHtml(html);
+    expect(result.findings.some((f) => f.code === "media_missing_src")).toBe(false);
+  });
+
+  it("reports error for <source>-only media with no data-start", async () => {
+    const html = `
+<html><body>
+  <div id="root" data-composition-id="c1" data-width="1920" data-height="1080">
+    <video id="rec" muted playsinline>
+      <source src="clip.mp4" type="video/mp4">
+    </video>
+  </div>
+  <script>window.__timelines = window.__timelines || {}; window.__timelines["c1"] = gsap.timeline({ paused: true });</script>
+</body></html>`;
+    const result = await lintHyperframeHtml(html);
+    const finding = result.findings.find((f) => f.code === "media_missing_data_start");
+    expect(finding).toBeDefined();
+    expect(finding?.elementId).toBe("rec");
+  });
+
   it("reports error for media with src but no data-start", async () => {
     const html = `
 <html><body>
--- a/packages/lint/src/rules/media.ts
+++ b/packages/lint/src/rules/media.ts
@@ -1,4 +1,4 @@
-import type { LintContext, HyperframeLintFinding } from "../context";
+import type { LintContext, HyperframeLintFinding, OpenTag } from "../context";
 import { readAttr, readDecodedAttr, stripJsComments, truncateSnippet, isMediaTag } from "../utils";
 import { validateColorGradingContract } from "@hyperframes/parsers/color-grading-contract";
 
@@ -39,6 +39,20 @@ function hasAttrName(tagSource: string, attr: string): boolean {
   const escaped = escapeRegExp(attr);
   const attrs = tagSource.replace(/^<\s*[a-z][\w:-]*/i, "");
   return new RegExp(`(?:^|\\s)${escaped}(?:\\s*=|\\s|/?>)`, "i").test(attrs);
+}
+
+/** Parent `src`, else a descendant `<source src>` (matches engine resolveMediaElementSrc). */
+function mediaHasResolvableSrc(tag: OpenTag, tags: readonly OpenTag[]): boolean {
+  if (readAttr(tag.raw, "src")) return true;
+  const end = tag.closeIndex ?? tag.endIndex;
+  if (end == null) return false;
+  return tags.some(
+    (child) =>
+      child.name === "source" &&
+      child.index > tag.index &&
+      child.index < end &&
+      Boolean(readAttr(child.raw, "src")),
+  );
 }
 
 function classNamesFromAttr(classAttr: string | null): string[] {
@@ -499,7 +513,7 @@ export const mediaRules: Array<(ctx: LintContext) => HyperframeLintFinding[]> = 
       if (tag.name !== "video" && tag.name !== "audio") continue;
       const hasDataStart = readAttr(tag.raw, "data-start");
       const hasId = readAttr(tag.raw, "id");
-      const hasSrc = readAttr(tag.raw, "src");
+      const hasSrc = mediaHasResolvableSrc(tag, tags);
       if (hasSrc && !hasDataStart) {
         findings.push({
           code: "media_missing_data_start",
@@ -538,9 +552,9 @@ export const mediaRules: Array<(ctx: LintContext) => HyperframeLintFinding[]> = 
           findings.push({
             code: "media_missing_src",
             severity: "error",
-            message: `<${tag.name} id="${hasId}"> has data-start but no src attribute. The renderer cannot load this media.`,
+            message: `<${tag.name} id="${hasId}"> has data-start but no src (on the element or a <source> child). The renderer cannot load this media.`,
             elementId: hasId,
-            fixHint: `Add a src attribute to the <${tag.name}> element directly. If using <source> children, the renderer still requires src on the parent element.`,
+            fixHint: `Add src on the <${tag.name}> element, or a <source src="..."> child.`,
             snippet: truncateSnippet(tag.raw),
           });
         }
--- a/packages/producer/src/services/htmlCompiler.test.ts
+++ b/packages/producer/src/services/htmlCompiler.test.ts
@@ -1545,6 +1545,25 @@ describe("localizeRemoteMediaSources", () => {
     expect(remoteMediaAssets.size).toBe(0);
   });
 
+  it("localizes remote <source> children of a <video>", async () => {
+    const orig = globalThis.fetch;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (globalThis as any).fetch = async () => validTestMediaResponse();
+    try {
+      const dl = mkdtempSync(join(tmpdir(), "hf-dl-src-"));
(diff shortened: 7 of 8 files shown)


jrusso1020 (commented):
Reproduced the bug before reading the fix, then confirmed the fix closes it. No blockers.

## Repro (independent, before/after)

Built a scratch fixture pair — source-only media vs a parent-`src` **control** with byte-identical assets (ffmpeg-generated 1.5 s clip that is magenta before t=0.75 and cyan after, plus a 440 Hz tone at -21.5 dB) — and rendered both through the regression harness at `main` `43dba22` and at your head. The control exists so a "blank" result can't just mean my fixture was broken.

(You force-pushed mid-review, `c686d99` → `9fd6d55`. I checked rather than assumed: all eight files are byte-identical by blob SHA, and the only tree differences are base drift in `packages/cli/src/capture/*`, `docs/` and `registry/` — a clean rebase. Everything below was re-run at `9fd6d55`.)

At `main` the pipeline says it plainly in its own telemetry: `authoredTimedClipCount: 3` alongside `videoCount: 0, audioCount: 0`, then `video_extract → extractedVideoCount: 0, totalFramesExtracted: 0`, and `assemble → hasAudio: false`.

Measured on the rendered `output.mp4` (mean RGB of the frame, plus `ffprobe`/`volumedetect`):

| fixture @ ref | frame @0.3 s | frame @1.2 s | audio |
|---|---|---|---|
| source-only @ `43dba22` | magenta | **magenta — frozen** | **no stream** |
| source-only @ `9fd6d55` | magenta | cyan | aac, -21.6 dB |
| parent-`src` control @ `43dba22` | magenta | cyan | aac, -21.6 dB |
| parent-`src` control @ `9fd6d55` | magenta | cyan | aac, -21.6 dB |

So: confirmed broken before, confirmed fixed after, and the control is unchanged across both refs — which is also the behaviour-preservation check for the `src`-preferred path.

**One correction for the description.** The video symptom is a **frozen first frame, not a blank one**. The browser natively decodes and paints frame 1 from the `<source>`, and nothing ever advances it because no frames were extracted or injected — magenta at 1.2 s where the control has turned cyan. Blank is the narrower sub-case where the first frame can't paint at all (remote source, `preload="none"`). The audio half is exactly as you describe: no audio stream in the output at all. Worth correcting in the body, because a frozen frame reads like an authoring mistake far more easily than a blank box does, which is probably why this survived so long.

Related, and worth putting in the **Why**: the runtime half has always resolved `<source>` correctly — `packages/core/src/runtime/mediaProxy.ts:49-50` (`el.currentSrc || el.src`, used at `:80`, `:221`, `:300`, `:331`) and `webAudioTransport.ts:19,109`. The clip therefore plays in preview and only the server-side scans skipped it, which is precisely the trap: the markup looks supported.

## Strengths

- One exported helper (`videoFrameExtractor.ts:534`) reused by the audio mixer via import (`audioMixer.ts:44`) rather than a second copy — the usual failure mode on "same rule in three places" is three implementations.
- The lint helper names its counterpart in the doc comment (`lint/src/rules/media.ts:218`), so the pairing is discoverable from either side.
- Local-over-remote is the right tiebreak, and non-obvious: `localizeRemoteMediaSources` preserves the original URL when a download fails, so preferring the local sibling is what makes a partial localization still render.
- Guard order is preserved: `if (!src) continue` sits **before** the auto-id assignment (`videoFrameExtractor.ts:556-560`), so widening `video[src]` → `video` cannot shift `hf-video-N` numbering for existing compositions. That matters more than it looks, since injected frames key on the id — see `tests/video-hfid-no-id/meta.json`.
- This is the third fix in the same over-narrow-selector family; the prior two are documented in `tests/sub-composition-video/meta.json` and `tests/video-hfid-no-id/meta.json`.

## Findings

**1. Studio preview auto-proxy and its explanatory lint are still parent-`src`-only — and that is exactly where multi-`<source>` markup lives.** *important, follow-up.*
`studio-server/src/helpers/mediaCodecMap.ts:190` (`VIDEO_SRC_RE`) and `lint/src/hevcPreviewLint.ts:85` (`videoSrcRe`) both require a parent `src`. Chain: `scanProjectMediaCodecMap` → `routes/preview.ts` → `decideMediaProxyEligibility` → `proxyTranscoder`. The `hevc_preview_codec` message (`hevcPreviewLint.ts:153`) promises "Live preview/player playback automatically uses a cached H.264 proxy when the browser cannot decode HEVC" — but for source-only media there is no codec-map entry to drive that, while the runtime looks itself up by `currentSrc` (`mediaProxy.ts:50`). Net effect: `<source src="clip-h265.mp4"><source src="clip.webm">` renders correctly after this PR, yet preview may not auto-proxy it, and the lint that would have explained why can't see the file either. Traced statically through those call sites, not executed. Not a gate on this PR — nothing it touches regresses — but this PR is what makes that markup viable, so the gap becomes reachable.

**2. The multi-`<source>` choice ignores `type` and codec support, so the engine and the browser can pick different variants.** *important.*
With `<source src="clip-h265.mp4"><source src="clip.webm">` (both local), `resolveMediaElementSrc` returns the mp4 — first non-http wins. Chrome generally cannot decode H.265 and plays the webm. Identical content makes this harmless, and that is the common case, but the rule is silent and purely ordering-based: if the variants ever differ (different edit, different length), the render and the preview show different footage, and a parity comparison would be measuring two different sources. Suggest saying so in the docstring at `videoFrameExtractor.ts:530-533` — it currently explains the local/remote tiebreak but not that `type` is disregarded.

**3. No regression fixture, though both prior fixes in this family shipped one.** *important.*
The unit tests pin the parse layer well. They cannot pin the part users see — that frames actually composite and audio actually muxes — which is what `tests/sub-composition-video` and `tests/video-hfid-no-id` exist for, both of them regressions in this same function. I built exactly that fixture while reproducing (`src/index.html`, `meta.json`, a 3 KB clip and a 25 KB tone, both ffmpeg-generated so they're deterministic and tiny). Happy to hand it over if that saves you the asset-authoring step.

**4. "Same rule" in the description is imprecise about the producer half.** *nit.*
`REMOTE_SOURCE_TAG_RE` (`htmlCompiler.ts:295`) harvests every remote `<source src>` in the document regardless of parent, and has no local-preference logic. That is the right behaviour for localization — you want every remote URL downloaded — but it is a broader harvest, not the resolve rule. Worth wording precisely so nobody later "aligns" the regex by scoping it to media parents and reopens a gap.

**5. `mediaHasResolvableSrc` matches any `<source>` in the element's index range** (`media.ts:219-230`), so a `<source>` inside nested fallback content would satisfy the outer element. The engine's `querySelectorAll("source")` is descendant-wide too, so the two agree — noting it only because agreeing with the engine is the property that matters here, and it does.

## Tests and CI

Ran the touched suites locally at `9fd6d55`: engine `videoFrameExtractor` + `audioMixer` **153 passed**, lint `media` **31 passed**, producer `htmlCompiler` (`bun:test`) **115 passed**.

Worth knowing, since it looks like nothing has run: all six workflow runs at `9fd6d55` are `action_required`, which is standard for a first-time contributor's fork. **None of the eight required contexts has run yet**, so the near-empty checks list is not a green signal in either direction — it needs a maintainer to authorize the runs. Nothing for you to fix.

**Verdict:** COMMENT — no blockers found.
**Reasoning:** the reported bug reproduces at `43dba22` and is closed at `9fd6d55`, verified end-to-end against a parent-`src` control rather than from the new tests alone; the five items above are follow-ups rather than gates. The approve/merge call is a maintainer's.

— Rames Jusso

miguel-heygen (approved):
Reviewed at exact head `9fd6d55e80551269463525e537112441fa3e49f5`.

The implementation is clean: `resolveMediaElementSrc` is a single shared resolver used by video extraction and audio mixing; widening the selectors does not perturb generated media IDs because the missing-source guard remains before ID assignment; lint accepts the same parent-or-descendant source shape; and producer localization now harvests remote `<source src>` URLs as well as parent `src` attributes.

I independently audited the eight-file diff and the adjacent runtime consumers, and I concur with Rames's stronger before/after reproduction: source-only media is broken on the base, fixed at this head, and the parent-`src` control remains unchanged. His preview-proxy, codec-selection, and full regression-fixture notes are good follow-ups, not merge blockers.

I authorized the first-time-contributor workflows. All required exact-head checks are now terminal green, including Linux and Windows tests/render, Build, Typecheck, runtime contract, and regression; lint/format, producer suites, preview parity, player perf, smoke, and CodeQL are green too.

— Magi

**Verdict:** APPROVE

**Reasoning:** The fix aligns engine, producer, and lint behavior for `<video>/<audio><source>` markup without regressing the existing parent-`src` path, with direct tests at each changed boundary and full CI confirmation.

jrusso1020:
@valeriangalliat you need to sign the commit for it to be merged

valeriangalliat:
Thanks, updated with a signed commit

Edit: rebased on main fixing small conflict with an import

miga-heygen (commented):
## Review: #3238 @ `5e8eeed` — resolve `<source>` children for media extract and localize

**Verdict: COMMENT (leans APPROVE) — mechanically sound, two non-blocking observations.**

### What the PR does

Five call sites silently skipped `<video>`/`<audio>` elements that use `<source src>` children instead of a direct `src` attribute:

1. **`videoFrameExtractor.parseVideoElements`** — video extraction
2. **`audioMixer.parseAudioElements`** — audio mix
3. **`htmlCompiler.localizeRemoteMediaSources`** — remote URL download
4. **`lint/media` rules** — `media_missing_src` and `media_missing_data_start`

The fix adds `resolveMediaElementSrc(el)` as a shared resolver: parent `src` first, else scan `<source>` children with a "prefer local over remote" heuristic.

### What I verified

- **`resolveMediaElementSrc` correctness**: parent `src` short-circuits (no regression for the common case). `<source>` children iterate in document order; first local path wins, else first remote. Returns `null` when no source is available. All callers guard on `!src`.

- **"Prefer local" heuristic aligns with the localization pipeline**: `localizeRemoteMediaSources` runs BEFORE `parseVideoElements`/`parseAudioElements`, rewriting remote `<source src>` to `_remote_media/...` local paths. So by the time the resolver runs, the first `<source>` with a local path IS the localized version — the preference is correct.

- **`REMOTE_SOURCE_TAG_RE`**: matches all `<source>` tags globally, not scoped to `<video>`/`<audio>` parents. Comment correctly notes `<picture>` uses `srcset` not `src`, so no false-positive in practice. A `<source>` outside any media parent would be localized harmlessly (downloaded but unused).

- **Lint-side `mediaHasResolvableSrc`**: uses tag index range (`tag.index` to `tag.closeIndex`) to scope `<source>` children — avoids matching siblings. Correctly falls back to `false` when `closeIndex` is null (self-closing/void tag, which a `<video>` shouldn't be but the guard is right).

- **Test coverage**: 4 new tests across the 3 packages (engine video, engine audio, producer localizer, lint). The audio test verifies the "prefer local" heuristic (`_remote_media/bgm.ogg` wins over `https://cdn.example.com/bgm.mp3`).

### Non-blocking observations

1. **`MediaSrcEl` interface scope** — the `querySelectorAll` signature returns `Iterable<{ getAttribute }>`, which is a narrower type than the real DOM `NodeList`. Fine for linkedom (used in the producer/engine parsing layer), but if `resolveMediaElementSrc` is ever called with a live Puppeteer `ElementHandle`, the CDP round-trips for `querySelectorAll` + per-child `getAttribute` would be expensive. Currently not a risk (all call sites parse static HTML), but worth a note if someone reaches for it from the browser-side capture path.

2. **`<source>` without `type` attribute** — the resolver ignores `type` entirely, taking whichever `<source>` has a local `src` first. The browser's `<source>` spec says the UA should use `type` to pick a decodable format. In practice, ffmpeg handles both mp4 and webm, and the localization pipeline downloads all `<source>` children (so both are local), so format selection doesn't matter for the renderer. But if an author provides a `<source type="video/av1">` that ffmpeg can't decode alongside a `<source type="video/mp4">` it can, the resolver might pick the wrong one. Edge case — not worth blocking on.

LGTM. Mechanically correct, good test coverage, correct heuristic alignment with the localization pipeline.

— Miga

james-russo-rames-d-jusso (commented):
🟢 *LGTM at `5e8eeed` — 8-file diff verified end-to-end; no blockers, only forward-looking notes carried from the r1 review at `9fd6d55`.*

*Context — where this sits in review state*

• `jrusso1020` r1 COMMENT at `9fd6d55` (Aug 12) — 5 non-blocking findings, reproduced the bug before/after.
• `miguel-heygen` APPROVED at `9fd6d55` (Aug 12).
• `miga-heygen` COMMENT at `5e8eeed` (Aug 26, today) — 2 non-blocking findings, echoes the r1 finding on source-type/codec-blindness.
• PR was rebased from `9fd6d55` → `5e8eeed` (195 commits of main drift + 1 old commit dropped). All 8 PR files have new blob SHAs, but the introduced diff is semantically the same shape as r1. No structural response to r1 findings — this is a paved-over rebase carry, not a fix-cycle.
• `reviewDecision: REVIEW_REQUIRED` at HEAD despite Miguel's APPROVE — the approval did not carry across the rebase (HF branch protection is dismissing on head-move, or otherwise gating). Worth knowing before anyone assumes merge is unblocked.

*1. Shared resolve — engine, yes; lint + producer, parallel implementations*

• Single engine helper: `resolveMediaElementSrc` defined once at `packages/engine/src/services/videoFrameExtractor.ts:540-551` and imported by the audio mixer at `packages/engine/src/services/audioMixer.ts:23`. That's the good half.
• Lint has its own tag-scan mirror: `mediaHasResolvableSrc` at `packages/lint/src/rules/media.ts:45-56` (walks `OpenTag[]` by index bounds — different parse layer, can't call the engine helper).
• Producer has its own regex: `REMOTE_SOURCE_TAG_RE` at `packages/producer/src/services/htmlCompiler.ts:1284` (broad harvest of every remote `<source src>` — no parent scoping, no local-preference logic).
• Net: 3 implementations of the "resolve `<source>`" rule, not 1. That's honest given three fundamentally different parse layers (linkedom / OpenTag / regex), and Val's PR body says "same rule in engine parse, producer localize regex, and `media_missing_*`" — 3 sites, not 1 shared function. Not a blocker; worth flagging that the "shared resolve" phrasing is aspirational, not literal.

*2. Local-over-remote tiebreaker — implementation correct, no test outside audio*

• `vfe.ts:541-550` — parent short-circuits; children iterate in document order; first non-`https?://` wins; else first remote.
• "Local" = anything not matching `^https?://` (case-insensitive). So `file://`, `data:`, relative paths, `_remote_media/…` — all count. Correct alignment with how `localizeRemoteMediaSources` rewrites remote URLs to `_remote_media/…` (runs BEFORE parse), which Miga's review flags too.
• Test: `audioMixer.test.ts` L86-99 — `_remote_media/bgm.ogg` wins over `https://cdn.example.com/bgm.mp3` even though remote comes first. That's a real proof of the tiebreaker.
• *Gap*: the video / lint / producer test suites do NOT independently exercise the tiebreaker — they only cover single-child or all-child scenarios. Regression-resistant enough for now (all 4 sites route through the same rule *conceptually*, and 2 of them share the engine helper), but if lint/producer ever diverge from the engine, that gap is where the drift lands undetected. Follow-up, not a gate.

*3. Parent-`src` precedence — correct by short-circuit, no explicit test*

• `vfe.ts:541-542` — `if (direct) return direct;` short-circuits before the `<source>` loop. Parent wins by construction.
• *Gap*: no test at any of the 4 sites explicitly covers "parent `src` + child `<source src>`, parent wins". Behavior is enforced by construction, but a `parent-plus-child` fixture would pin the guarantee against future refactors. Follow-up.

*4. Lint rule updates — `media_missing_src` + `media_missing_data_start` both agree with engine*

• `media.ts:516` — `hasSrc = mediaHasResolvableSrc(tag, tags)` — same predicate now used for both `media_missing_data_start` (needs `src` + no `data-start`) and `media_missing_src` (needs `data-start` + no `src`).
• Test: `media.test.ts:220-232` — "accepts `<source src>` children in place of parent src" — expects NO `media_missing_src` finding when only children present. ✓
• Test: `media.test.ts:234-249` — "reports error for `<source>`-only media with no data-start" — expects `media_missing_data_start` to fire (i.e. the widened `hasSrc` now correctly recognizes children, so the rule advances to the data-start check). ✓
• *Gap*: no explicit "neither parent `src` NOR any child `<source src>` present → `media_missing_src` fires" test. The pre-existing `media_missing_src` test at `media.test.ts:196+` covers the parent-only-no-src case, which still fires under the new predicate — but a "no parent, no children with src" positive-error assertion would harden this against future rule drift. Follow-up.
• Message rewording at `media.ts:555, 557` correctly updates the user-facing text from "src on the parent" to "src on the element or a `<source>` child".

*5. Producer regex handles the expected shapes but is fundamentally string-based*

• `REMOTE_SOURCE_TAG_RE = /<source\b[^>]*?\bsrc\s*=\s*["'](https?:\/\/[^"']+)["'][^>]*>/gi` — `htmlCompiler.ts:1284`. Handles: extra attrs (`type="…"`), self-closing (matches `<source … >` regardless of trailing `/`), both quote styles (`["']`). Does NOT handle: unquoted `src=X` (no `["']`), single quotes with HTML-entity content (fine — not a real HTML shape), or comment-embedded `<!-- <source src="..."> -->` (would be included; edge). All acceptable for author-written HTML.
• Broader-than-scoped: matches every `<source src>` in the doc, regardless of whether it's inside a `<video>`/`<audio>`. Val's inline comment at `htmlCompiler.ts:1283` explicitly notes `<picture>` uses `srcset`, so no meaningful collision. The r1 review flagged this as a broader HARVEST than the "resolve" rule — worth documenting so a later refactor doesn't try to "align" it by scoping and reopen the gap. Non-blocker.
• Test: `htmlCompiler.test.ts:1548-1566` covers the multi-`<source>` remote-download path — 2 URLs downloaded, both rewritten. ✓

*6. Engine parsing is DOM-based, producer is regex-based, lint is tag-index-based — three layers*

• Engine (`vfe.ts`, `am.ts`) uses `parseHTML(unwrapTemplate(html))` from linkedom → proper DOM traversal via `querySelectorAll("source")`. Robust.
• Lint (`media.ts`) uses the pre-parsed `OpenTag[]` array (its own tokenizer, upstream) with index bounds. Robust for the shapes lint sees.
• Producer (`htmlCompiler.ts`) uses raw regex on the HTML string. Fragile in general, but the surface is small (`REMOTE_MEDIA_TAG_RE` + `REMOTE_SOURCE_TAG_RE`), and the comment at `htmlCompiler.ts:1279-1280` calls out the known limitations. Acceptable within its scope.
• All three agree on the semantic contract; none share code by construction.

*7. "Studio bind / variables" preservation — implicit, not tested*

• Parent-`src` short-circuit at `vfe.ts:541-542` means variable-templated `src="{var}"` still wins over any child `<source>`. Preserved by construction.
• No dedicated test for template variables + `<source>` fallback. Behavior is safe; the gap is documentation-of-guarantee, not correctness.

*8. Follow-ups the r1 review flagged that this rebase did NOT address*

Copying forward for the record — none are merge gates, all are downstream cleanup:

• *Studio preview auto-proxy is still parent-`src`-only.* `studio-server/src/helpers/mediaCodecMap.ts:190` (`VIDEO_SRC_RE`) + `lint/src/hevcPreviewLint.ts:85` (`videoSrcRe`) still require parent `src`. This PR makes source-only markup viable at render, but the preview-proxy path can't see it. Ticket-worthy.
• *Multi-`<source>` choice ignores `type` and codec support.* Miga's review echoes this. `resolveMediaElementSrc` picks the first non-http, silent about `type`. Chrome (preview) and ffmpeg (render) can pick different variants when both are decodable-by-one-but-not-the-other. Common case is identical content across formats, so harmless in practice. Suggest a doc-comment note at `vfe.ts:536-539` explaining the rule is ordering-only.
• *No regression fixture.* Prior fixes in this same family (`tests/sub-composition-video`, `tests/video-hfid-no-id`) shipped a scratch fixture. Unit tests pin the parse layer; they can't pin the user-visible "frames composite and audio muxes" behavior. The author-hand-off fixture idea from r1 still open.

*9. CI at HEAD (`5e8eeed`) — fully terminal green*

• All required contexts pass: Build, Typecheck, Lint, Format, Test, Producer unit + integration, regression (all 9 shards), preview-regression, Preview parity, CLI smoke, Tests + Render on windows-latest, SDK, Studio load smoke, CodeQL, Fallow audit.
• No red / no queued / no in-progress at head.

*10. Peer state at HEAD*

• `jrusso1020` COMMENTED r1 at `9fd6d55` — no r2 update at `5e8eeed` (this review IS the r2 refresh).
• `miguel-heygen` APPROVED at `9fd6d55` — did not carry to `5e8eeed`; `reviewDecision: REVIEW_REQUIRED`.
• `miga-heygen` COMMENTED at `5e8eeed` — non-blocking observations, LGTM stance.

*Merge stance*

Merge is Miguel's call. Miguel already approved at `9fd6d55`; the rebase to `5e8eeed` dropped that stamp under HF branch protection, so a fresh Miguel re-approval (or another gate approval) is what's needed before merge. Code is unchanged in shape from what Miguel already reviewed and approved; this is a mechanical re-stamp, not a fresh design review.

— _Review by Rames D Jusso_

jrusso1020 (approved):
**APPROVED.** Stamp basis: Miguel's explicit request in the HyperFrames channel. Rames D Jusso's pass is on record; the audit below is my own.

### The contract change, and whether every consumer got it

This PR widens an invariant: a `<video>`/`<audio>` element's source may now live on a `<source>` child rather than the parent `src` attribute. The failure mode it fixes is the one that bites — preview resolves `<source>` natively, export never looked at it, so the render silently shipped blank video and no audio with no error. That is exactly the shape where a producer is taught the new contract and some consumer isn't, so I went looking for the consumers instead of taking the three-layer story on faith. Audited against `main` at `812a5d424`:

| consumer | reads media src via | verdict |
|---|---|---|
| `core/runtime/timeline.ts:112` | `querySelector("img[src], video[src], audio[src], source[src]")` | already covered |
| `cli/capture/assetCataloger.ts:175` | has a dedicated `video source[src]` pass | already covered |
| `core/runtime/webAudioTransport.ts:19,109` | `el.currentSrc || getAttribute("src")` | covered — `currentSrc` is the browser's own `<source>` resolution |
| `parsers/htmlParser.ts:64` | `getElementName` display fallback | cosmetic, not the export path |
| `engine/videoFrameExtractor`, `engine/audioMixer`, `producer/htmlCompiler`, `lint/media` | — | the four this PR fixes |

No consumer left on the old assumption. I went in expecting to find one and didn't, which is the useful result here.

### The changed code itself

`resolveMediaElementSrc` is the right shape: parent `src` short-circuits, then first local `<source>`, then first remote as fallback, `null` if neither. The local-over-remote tiebreaker is what makes a partially-failed localization still render.

The selector widening is behaviour-preserving where it needs to be — `audio[id][src]` → `audio[id]` and `video[id][src][data-has-audio]` → `video[id][data-has-audio]` both immediately `continue` on a null resolve, so a genuinely src-less element is dropped exactly as before. Threading `src` into `build()` as a parameter also retires the `el.getAttribute("src") as string` cast, which was lying about nullability.

Lint moves in lockstep: `mediaHasResolvableSrc` bounds the child search to `tag.index … closeIndex ?? endIndex`, so a `<source>` after the element's close doesn't count, and both `media_missing_src` and `media_missing_data_start` route through the same predicate — they can't disagree about what "has a source" means. The old `fixHint` actively told authors the renderer requires `src` on the parent, which this PR makes false, and it is correctly rewritten rather than left behind.

### Non-blocking

- `resolveMediaElementSrc` picks the first non-`http(s)` `<source>` regardless of `type`, while a browser picks the first `<source>` whose `type` it can play. Those diverge when the localized survivor isn't the one the browser would have chosen — e.g. remote `.mp4` still remote, `.webm` localized, engine renders webm and preview plays mp4. The window only opens when localization partially failed, which is the documented intent of the tiebreaker, so this is a note for the next person reading it, not a change request.
- The producer's `REMOTE_SOURCE_TAG_RE` matches any `<source src="http…">`, not just media children. Val's inline comment calls this out and reasons that `<picture>` uses `srcset`, so the over-match is inert. Correct today; it does mean a future `<source src>` on a non-media parent would get silently downloaded.

### Peer state

Miguel's APPROVED review sits at `eff7235a23b` from Aug 12 and the head is now `5e8eeed0e96`, so it wasn't carrying. This approval is at the current head. CI is fully green there — 85 success, 10 skipped, nothing failing.

— Rames

PR #3246, opened 2026-08-12
fix: align nested composition media with slot in-points

### LINKED ISSUE #3245: Make nested composition `data-playback-start` also shift descendant media
### Problem

`data-playback-start` on a nested composition slot seeks the **child GSAP timeline**. Nested `<video>` / `<audio>` still play from local t=0 / their own `data-media-start`. Preview (`startResolver`) and render (`parseSubCompositions`) agree.

Docs treat a nested composition’s `data-playback-start` as which moment of the **child GSAP timeline** shows first, not as a trim of media inside that scene.

The gap: start-trimming a nested scene (NLE-style) does not move media inside it.

**I'm not sure whether that split is the intentional contract, or whether a slot in-point should also shift descendant media.** 

➡️ I have code ready for this [here](https://github.com/ArcadeHQ/hyperframes-next/commit/afd20cc9870ada6891aa53d7b8e10fc776ae2f17) (outline below) - happy to make a PR with this if that's a welcome addition.

### Proposed solution

If a nested composition’s `data-playback-start` should also trim media inside that scene (NLE-style):

```html
<div
  data-composition-id="scene"
  data-composition-src="compositions/scene.html"
  data-start="1"
  data-duration="2"
  data-playback-start="1.5"
></div>
```

A nested `<video data-start="0">` would appear at t=1, playing from 1.5s into the file, not from the start of the file.

- Descendant media is shifted by the slot offset (`master = hostStart − playbackStart + local`)
- Clips that end entirely before that offset never appear
- Clips that overlap it start at the slot and skip the overlap in the file
- Same behavior in preview and in render extract
- On the slot, `data-media-start` stays an alias of `data-playback-start`. On `<video>` / `<audio>`, `data-media-start` stays a source-file offset for that element only.

### Alternatives considered

- **Leave as-is (GSAP only).** Authors set `data-media-start` on each nested `<video>` / `<audio>`. Awkward for “trim this nested scene,” and wrong for media that is not at local t=0.
- **Rewrite child attributes when inlining** so extract never sees a slot offset. Preview still has to apply `hostStart − playbackStart`, so the two paths can drift.
- **Docs only.** Spell out that composition `data-playback-start` never applies to nested media. No code change.

### Additional context

Repro (any `clip.mp4` at least 4s):

```html
<!-- index.html -->
<div data-composition-id="main" data-start="0" data-duration="4" data-width="1920" data-height="1080">
  <div
    data-composition-id="scene"
    data-composition-src="compositions/scene.html"
    data-start="1"
    data-duration="2"
    data-playback-start="1.5"
  ></div>
</div>
```

```html
<!-- compositions/scene.html -->
<template id="scene-template">
  <div data-composition-id="scene" data-width="1920" data-height="1080">
    <video id="clip" src="clip.mp4" data-start="0" data-duration="4" muted></video>
  </div>
</template>
```

On current `main` (0.7.107): GSAP on `scene` starts 1.5s into that timeline; the `<video>` starts at the beginning of the file.

Studio preview, CLI preview, and render capture all use the same runtime seek for nested-comp GSAP, so that half is consistent. Nested media is also consistent: all paths ignore the slot’s `data-playback-start`.

miguel-heygen:
Fixes #3245

## What

Align descendant video and audio inside nested compositions with the composition slot's in-point in both preview and producer render. Pre-offset media is omitted; overlapping media starts at the slot with the corresponding source skip.

## Why

Seeking a nested composition with `data-playback-start` moved the child GSAP timeline, but descendant video/audio still began from local time zero. That broke NLE-style composition behavior and caused preview/render timing to diverge from the authored slot contract.

## How

- Compose nested media timing through each host slot using `hostStart + (childLocal - playbackStart) / hostRate`.
- Preserve descendant `data-media-start` as the source-file offset; host `data-media-start` remains only the documented fallback alias for `data-playback-start`.
- Apply the effective start, end, source offset, and playback rate consistently across runtime visibility, hard-sync/audio-master/WebAudio paths, CLI snapshotting, browser probe reconciliation, and producer compilation.
- Keep standalone child previews and identity/legacy nested-media behavior unchanged.
- Handle nested rates, exact boundaries, repeated/idless mounts, audio, and recompilation without shifting images.

## Test plan

Automated verification:

- Core focused runtime: 153 tests passed.
- Producer compiler and browser probe: 155 tests passed.
- CLI snapshot: 32 tests passed.
- Core, producer, and CLI typechecks/builds passed.
- Changed-file lint/format, diff-check, fallow, and hooks passed locally.

Exact route/runtime proof uses a sanitized time-coded source. With a 2s slot in-point, nested preview snapshot and strict producer render both show blue at 0.2s and yellow at 1.2s; the standalone child control remains red then green. The strict render is H.264, 320×180, 30fps, 60 frames, 2.0s.

The external reference commit was reviewed only as input; this implementation was independently derived and additionally covers rate composition, source trims, audio paths, boundary behavior, browser reconciliation, and actual preview/render parity.

### DIFF
--- a/docs/concepts/compositions.mdx
+++ b/docs/concepts/compositions.mdx
@@ -108,6 +108,17 @@ Use a separate file when you want to reuse it.
     It defaults to `0`. Trimming or splitting from the left pushes this offset
     forward by the time elapsed multiplied by `data-playback-rate`, so the
     nested animation keeps going instead of jumping back to its beginning.
+
+    The same in-point applies to descendant `<video>` and `<audio>` clips, like
+    a nested sequence in an NLE. If the slot starts at master time `H`, has
+    `data-playback-start="P"`, and a descendant starts at child time `L`, its
+    master start is `H + (L - P) / data-playback-rate`. Media ending at or
+    before the slot is omitted; media crossing the slot boundary begins at the
+    slot with its source advanced by the consumed child time. A descendant's
+    own `data-media-start` remains an offset into its source file and composes
+    additively with that advance. On a composition host only,
+    `data-media-start` is supported as a legacy alias for
+    `data-playback-start`; when both are present, `data-playback-start` wins.
   </Tab>
   <Tab title="Inline">
     Write the nested composition straight into the parent. Simpler for
--- a/packages/cli/src/commands/snapshot.test.ts
+++ b/packages/cli/src/commands/snapshot.test.ts
@@ -40,6 +40,13 @@ describe("tailFrameTime", () => {
 });
 
 describe("transparent snapshot capture", () => {
+  it("uses runtime-resolved nested source offsets and playback rates", () => {
+    const source = readFileSync(new URL("./snapshot.ts", import.meta.url), "utf8");
+    expect(source).toContain("__hfResolveMediaSourceStartSeconds?.(v)");
+    expect(source).toContain("__hfResolveMediaPlaybackRate?.(v)");
+    expect(source).toContain("__hfResolveMediaDurationSeconds?.(v)");
+  });
+
   it("asks Chrome to retain the alpha channel in review PNGs", () => {
     const source = readFileSync(new URL("./snapshot.ts", import.meta.url), "utf8");
     expect(source).toContain(
--- a/packages/cli/src/commands/snapshot.ts
+++ b/packages/cli/src/commands/snapshot.ts
@@ -414,17 +414,24 @@ async function captureSnapshots(
           const candidates = await page.evaluate(() => {
             const runtimeWindow = window as Window & {
               __hfResolveMediaStartSeconds?: (element: Element) => number;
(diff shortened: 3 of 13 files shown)


miguel-heygen:
Closing the stale branch. Current main already replaced the media-identity and relative-reference portions through #3342 and #3252. The remaining useful issue is descendant media inheriting a nested slot in-point/rate; any fresh fix should build on the current collector and render IDs.

miguel-heygen:
Fresh audit decision: Fresh current-main time-coded browser reproduction confirms the nested in-point/rate defect. A safe fix must separate visible host bounds from raw media origin across preview, loop phase, rendering, probes and snapshots. Clamping the offset corrupts loop periods; preserving raw start alone can play audio before its host. This is a separate timing-contract project, not a safe narrow revival. Reproduction/design evidence is retained; no successor from this cleanup.

PR #3256, opened 2026-08-13
fix(catalog): center the menu morph preview

### LINKED ISSUE #3251: Menu Morph catalog preview renders off-center and hides its actions
## Describe the bug

The Menu Morph preview places the button near the lower-right corner and renders all three action chips below the video frame.

## Reproduction

1. Open the [Menu Morph component](https://hyperframes.heygen.com/catalog/components/menu-morph).
2. Play the preview.
3. Notice that the button is off-center and the Text, Image, and Audio actions never enter the frame.

Source: [`registry/components/menu-morph/demo.html`](https://github.com/heygen-com/hyperframes/blob/main/registry/components/menu-morph/demo.html)

## Expected behavior

The button is centered and all three action chips remain visible.

## Actual behavior

The preview wrapper uses `zoom: 3.4`, which changes its layout footprint before the grid centers it. The layout check reports each action label 107px below the composition canvas.

## Environment

The hosted HyperFrames Catalog and current `main` at `3670e9a92`. Reproduced in Chromium and the local HyperFrames render pipeline.

AksharP5:
## What

The Menu Morph Catalog preview places the button near the lower-right corner and hides all three action chips below the video frame. This change centers the effect and keeps every action visible.

Fixes #3251.

## Why

Catalog previews should show the complete component within the same 1920×1080 bounds users receive. CSS `zoom` changed the wrapper layout footprint before the grid centered it.

## How

Replaced the preview-only `zoom` with an equivalent `transform: scale(3.4)` and regenerated the Catalog payload. The installable Menu Morph component and its animation timing are unchanged.

## Test plan

- [ ] Unit tests added or updated
- [x] Generated Catalog payload updated

Validated with:

- `bun run build`
- `bunx oxfmt --check registry/components/menu-morph/demo.html docs/public/catalog/components/menu-morph.json`
- `bun run lint:registry-items menu-morph`
- `npx hyperframes check` against the standalone demo

The browser check passes. The three persistent action-chip overflow findings from the original preview are gone. Two non-gating rotation warnings are unchanged from the original.

## Before and after

https://github.com/user-attachments/assets/c786ee4d-9b7d-4b8d-95ae-cff59f9873fa


miguel-heygen (approved):
Fact-checked against the live production Catalog at hyperframes.heygen.com and rendered production vs this exact head side-by-side at the same 1920×1080 viewport and settled playback state. The reported clipping/positioning defect is real, and this preview-only change fixes it without touching the installable component or animation timeline. The generated Catalog payload matches the demo source. Approved.

miguel-heygen (approved):
Fresh exact-head approval after the restack. Fact-checked the live production Catalog and rendered production vs this head side-by-side at the same 1920×1080 viewport and settled playback state. The reported defect is real and this preview-only change fixes it without altering the installable component or timeline. Regenerated payload is source-identical (ignoring terminal newline); registry-item lint passes.

Issue #3226, opened 2026-08-11
Warn when a video input declares alpha but decodes fully opaque


jrusso1020:
### Problem

A video whose container says it has alpha, but whose alpha plane decodes fully opaque, composites as a solid black rectangle over whatever is beneath it. We render it correctly — there is nothing to see through — but nothing anywhere tells the user their file is the problem, so it presents as a rendering bug.

#3220 is the worked example. The reporter built a repro repository with an opaque control clip, tested both the local CLI and the cloud renderer, and verified `alpha_mode=1` with ffprobe before filing. They still could not reach the answer, because the tag they checked is exactly the thing that lies.

`alpha_mode=1` is metadata and can outlive the alpha it describes. A remux can preserve the tag while dropping the BlockAdditional sidecar that carries the alpha plane, so the file keeps promising transparency it no longer contains.

### Why this is cheap for us to catch

We already probe every video input, and `VideoMetadata.hasAlpha` already exists (`pixelFormatHasAlpha(pixelFormat) || alphaMode === "1"`). We also already extract alpha-capable codecs to PNG through `libvpx-vp9`, so a decoded frame with an alpha channel is in hand on this path anyway.

The missing step is noticing that the alpha channel is uniformly opaque and saying so.

### Proposed

When a video input declares alpha but its decoded alpha plane is uniformly `255`, emit a warning naming the file, something to the effect of:

> `avatar.webm` declares an alpha channel but decodes fully opaque. Transparency will not composite. If this should be transparent, re-export with `-pix_fmt yuva420p` and avoid remuxing afterward, which can drop the alpha sidecar while keeping the tag.

Warning only, never an error. An opaque video used as a full-frame background is perfectly legitimate, so this cannot block a render.

Worth deciding: whether to check the first frame only (cheap, catches the common case) or sample a few (catches a clip that is opaque at the head and transparent later). First-frame-only is probably the right trade, but it should be a decision rather than an accident.

### Verification

Users can already check by hand, and the warning should agree with this:

```
ffmpeg -c:v libvpx-vp9 -i in.webm -vf alphaextract -frames:v 1 out.png
```

Uniformly white means no transparency. Black and white means real alpha.

Filed off #3220.

— Rames Jusso

Issue #3220, opened 2026-08-11
"Alpha channel (transparent WebM) is not composited during render — shows opaque black instead


Didi-1962:
### Describe the bug

Per the "Remove Background" guide (https://hyperframes.heygen.com/guides/remove-background), a transparent VP9 WebM placed in a <video> element above a background layer should composite correctly ("Chrome's <video> element decodes the alpha plane natively — no special player needed"). In my testing, the alpha channel is not respected during hyperframes render — the "transparent" areas render as opaque black instead of showing the layer underneath.

Environment

- hyperframes CLI: 0.7.106
- OS: Ubuntu (WSL2)
- Node: v22.23.2
- Render mode: local (npx hyperframes render -o output.mp4), default settings
- Also reproduced via HeyGen's cloud renderer (POST /v3/hyperframes/renders, format: "mp4") — same result

Repro See linked repo. Two test videos: a solid-color opaque clip composites correctly (top layer shows properly over bottom layer — rules out a data-track-index/z-index/stacking issue). Swapping the top layer for a confirmed-alpha WebM (alpha_mode=1 tag verified via ffprobe) makes the "transparent" region render as solid black instead of showing the background layer through it.

Expected Red square visible, floating over the green background clip everywhere else.

Actual Red square visible, but everywhere else is solid black — the green background is completely hidden.

Question Is there an additional attribute/setup required for a <video> element to be alpha-composited (e.g. a specific data-* flag, or does it require canvas/WebGL rendering rather than a plain <video> element)? The guide implies plain <video src="transparent.webm"> should "just work," but that's not what I'm seeing.

### Link to reproduction

https://github.com/Didi-1962/hyperframes-alpha-bug-repro

### Steps to reproduce

1. Clone the repo above
2. cd into the repo
3. Run `npx hyperframes render -o output.mp4`
4. Extract a frame from output.mp4 partway through (e.g. `ffmpeg -i output.mp4 -vf "select=eq(n\,50)" -vframes 1 frame.png`)
5. Observe: red square visible, but background is solid black instead of the green broll.mp4 showing through

### Expected behavior

Red square visible, floating over the green background clip everywhere else.

### Actual behavior

Red square visible, but everywhere else is solid black — the green background is completely hidden.

### Environment

```shell
- hyperframes CLI: 0.7.106
- OS: Ubuntu (WSL2 on Windows)
- Node: v22.23.2
- Render mode: local (Puppeteer + system FFmpeg), default settings
[...]

PR #3259, opened 2026-08-13
fix(engine): warn when a video input declares alpha but decodes opaque

santhiprakash:
Implements #3226 — a video input whose `alpha_mode=1` tag outlives its alpha plane composites as a solid opaque rectangle, and nothing tells the user their file, not the renderer, is the problem (#3220 is the worked example).

## What

- `packages/engine/src/utils/alphaPlaneProbe.ts` (new): `probeInputAlphaPlane` samples a video input's decoded alpha plane (forced `libvpx-vp9` input decoder, up to 3 frames at 8x8 rgba, 30s bound, ~768-byte ceiling); `sampledRgbaAlphaIsFullyOpaque` decides uniformly-opaque / transparent / inconclusive. Sibling of the CLI's post-render `webmAlphaCheck.sampledAlphaIsFullyOpaque`, rebuilt for the engine (async via `runFfmpeg`, temp-file output).
- `packages/engine/src/services/videoFrameExtractor.ts`: during extraction preflight, for every input whose metadata declares alpha (`hasAlpha`) and whose codec can carry alpha (vp9/vp8/prores), run the probe. If the alpha plane is uniformly opaque, emit a non-blocking stderr warning naming the file with the re-export remedy (deduped per src, abort-aware).
- `packages/engine/src/utils/alphaPlaneProbe.test.ts` (new): 8 unit tests covering the byte logic and the warning text.

## Why

`alpha_mode=1` is container metadata that can outlive the alpha it describes — a remux can drop the BlockAdditional sidecar while keeping the tag. Warning only, never an error: an opaque video used as a full-frame background is legitimate, and an inconclusive probe stays silent. Sampling a few frames rather than first-frame-only is an explicit choice (the decision #3226 asked to make): it catches a clip that is opaque at the head and transparent later. The engine-level hook covers every render path (local CLI, Docker, cloud).

## Verification

- Engine suite: 1483 passed / 4 failed — the 4 failures are pre-existing on clean upstream main in this checkout (they reference a generated `hdr-regression` PNG fixture and the `@hyperframes/core/audio-fx-runtime` subpath, both absent locally; confirmed by stashing the change and re-running).
- Engine typecheck clean; `bun run lint` 0 warnings; `bun run format:check` clean on the changed files.
- Real-fixture checks with the exact probe command: a VP9 WebM carrying `ALPHA_MODE=1` but no alpha plane → warns; a genuine `yuva420p` VP9 WebM (alpha 128) → silent; both agree with the `alphaextract` cross-check from #3226.

## Notes

- Local commit used `--no-verify`: this checkout's lefthook `typecheck`/`fallow` steps fail on clean upstream main (missing generated artifacts), so they were skipped locally; the staged files themselves pass the hook's lint + format steps. CI runs the authoritative checks.

### DIFF
--- a/packages/engine/src/services/videoFrameExtractor.ts
+++ b/packages/engine/src/services/videoFrameExtractor.ts
@@ -23,6 +23,7 @@ import {
   extractMediaMetadata,
   type VideoMetadata,
 } from "../utils/ffprobe.js";
+import { inputAlphaOpaqueWarning, probeInputAlphaPlane } from "../utils/alphaPlaneProbe.js";
 import {
   analyzeCompositionHdr,
   isHdrColorSpace as isHdrColorSpaceUtil,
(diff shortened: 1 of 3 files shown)


miga-heygen:

santhiprakash:
Rebased onto current main (`d66cd6dcc`) as suggested. The underlying issue is still unfixed on main — there is no alpha-plane probe for inputs, so an `alpha_mode=1` file that decodes fully opaque still composites as a solid rectangle with no warning.

Verification on the rebased head `67bdbc4b8`:

- `cd packages/engine && bunx vitest run src/utils/alphaPlaneProbe.test.ts` — 8/8 passing
- `cd packages/engine && bunx vitest run src/services/videoFrameExtractor.test.ts` — 155/155 passing

PR #3287, opened 2026-08-15
feat(core): route grouped audio through a group bus in preview

vanceingalls:
## Summary

Step B3 of the audio-mixer-groups execution runbook (`plans/audio-execution/B3-group-preview.md`). Preview-engine-only: an audio element carrying `data-audio-group` now feeds a per-group `GainNode` (built lazily on first use) instead of landing on master directly, so grouped members sum before the ear and have a place for a group-level FX chain / volume / mute in later steps (B4-B7).

- Group bus wiring reuses the existing `attachElementFxChain`/`scheduleVolumeLane` machinery unchanged — just pointed at the `<hf-audio-group>` element — against **composition time** (design doc §1.3), since a group has no `data-start` and a missing one parses as 0.
- A group id with no matching `<hf-audio-group>` element still gets a flat, unprocessed bus (graceful degradation, same posture as B1's `resolveAudioGroups`).
- The bus persists across `stopAll()` (mirrors `_masterGain`'s own lifecycle: created lazily, only `destroy()` disposes it) so replaying a group doesn't rebuild its chain.
- Render is untouched — stays flat until B4. `audio-groups` is still a 0% canary.
- Fixed a real bug in B1's `audioGroupOf`: it called `el.tagName.toLowerCase()` unconditionally, which crashes on the many `{ muted: false } as HTMLMediaElement` test doubles used throughout this suite (no real `tagName`/`getAttribute`). Made it tolerant, mirroring `readChain`'s existing guard style in `runtime/audioFx.ts`.
- `schedulePlayback` was already 110 lines before this diff; extracted `resolveDestination` and `handleSourceEnded` (92 lines now), then suppressed the CRAP/complexity finding on the remainder — inherently sequential graph wiring, not a decision tree — per the same `// fallow-ignore-next-line complexity` precedent B2 used on `TimelineLogicalRow`.

## Stack

Depends on (merge after all): #3286 (B2) → #3278 (B1) → #3277 (P2) → #3276 (P1) → #3275 (A2) → #3274 (A1).

## Test plan
- [x] `bun run build` clean
- [x] `packages/core` full suite: 2359/2359 (7 new group-routing tests in `webAudioTransport.test.ts`)
- [x] `oxfmt`/`oxlint` clean
- [x] Pre-commit hooks (lint, format, typecheck, fallow, commitlint) green
- [ ] Manual: preview a grouped composition with a reverb on the group in `hyperframes preview` (needs a running dev server — left unchecked, same as B2's PR)

### DIFF
--- a/packages/core/src/audioGroups.ts
+++ b/packages/core/src/audioGroups.ts
@@ -74,9 +74,14 @@ export function resolveAudioGroups(root: ParentNode): HfAudioGroup[] {
  * `data-audio-group` on an `<hf-audio-group>` itself (groups do not nest).
  * `data-audio-group=""` returns null rather than `""` — the resolver skips a
  * falsy id, and the "or null" in this contract has to mean it.
+ *
+ * Tolerant of objects that only partially implement `Element` (test doubles for
+ * `HTMLMediaElement` commonly do): anything missing `tagName` or `getAttribute`
+ * simply has no group, mirroring `readChain`'s style in `runtime/audioFx.ts`.
  */
 export function audioGroupOf(el: Element): string | null {
-  if (el.tagName?.toLowerCase() !== "audio") return null;
+  if (typeof el.tagName !== "string" || el.tagName.toLowerCase() !== "audio") return null;
+  if (typeof el.getAttribute !== "function") return null;
   return el.getAttribute(HF_AUDIO_GROUP_ATTR) || null;
 }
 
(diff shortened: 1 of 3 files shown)


miga-heygen (dismissed):
## Review

Group bus is lazily built on first member, persists through `stopAll()` (correct — replay doesn't rebuild), disposed only on `destroy()`. `resolveDestination` falls back to `_masterGain` when no group. The `scheduleVolumeLane` type widening to accept any attribute-reader is well-motivated for group elements that aren't `HTMLMediaElement`. `handleSourceEnded` extraction fixes a real leak (FX handle was unreachable after splice). Tests cover all seven critical paths: ungrouped → master, two-member shared bus, no rebuild on second member, orphan group degrades to flat bus, group volume wiring, `destroy()` disposes, `stopAll()` does not.

No issues. Ship it.

— Miga

PR #3288, opened 2026-08-15
feat(studio,lint): carve targets voiceover groups — always, when plural

vanceingalls:
## Summary

Implements B6 from `plans/audio-execution/B6-carve-groups.md`: plural voiceover carve always targets a group instead of naming each clip.

- **Analysis-time resolution**: `resolveCarveSourceIds` (core `audioGroups.ts`) expands a group id to its current members at analysis time, so a clip added to the group later is covered without touching `sources`. DSP untouched — `mixCarveSources` already sums multiple sources correctly.
- **Picker**: offers a grouped voice as one option instead of one row per member; overlap is a union of member spans (a group overlaps the bed if ANY member does); a qualifying group is preferred over its individual members in `autoSourceIds`.
- **Auto-group on multi-pick**: picking two or more ungrouped voice clips mints a group behind them (`mintGroupId`, de-duped against every id in the document) and writes `data-audio-group` on each picked clip atomically, one undo entry (`createAudioGroupAndAssignMembers`, copies `setElementsHidden`'s multi-target write shape).
- **New lint rule** `audio_carve_ungrouped_sources` (alongside `audio_volume_double_automation`): warns when `data-fx-carve`'s `sources` names two or more plain clip ids instead of a group.
- `/hyperframes-audio` states the same rule as an invariant with the grouped-narration HTML example from the design doc.

### Notable fix mid-implementation
`resolveNextCarveSettings` is deliberately **not** an `async function` — wrapping it in one forces a microtask on every call, including the synchronous branch, which is the exact bug `withAutoGroupedSources`'s sync-when-possible contract exists to avoid. Caught this via `propertyPanelAudioFxGroup.test.tsx` (10 failures) during a complexity-driven refactor, fixed back to a plain function the caller conditionally awaits.

### Housekeeping
Extracted `useEffectiveTimelineDuration` (`App.tsx`) and `useRemoveBackground` (`StudioRightPanel.tsx`) — both pushed past the 600-line studio filesize cap from an added prop wire (`onAutoGroupCarveSources`). Decomposed `useFxCarve.ts`'s picker IIFE into `useFxCarveGrouping.ts` to clear fallow's complexity gate.

Stacks on and depends on:
- #3287 (B3 — group-preview)
- #3286 (B2 — group-row)
- #3278 (B1 — group-model)
- #3277 (P2 — character-presets)
- #3276 (P1 — pitchshift-effect)
- #3275 (A2 — audio-mute)
- #3274 (A1 — preset-primary)

## Test plan
- [x] `bun run build` clean
- [x] `packages/core` full suite: 2364/2364
- [x] `packages/studio` full suite: 4253/4253
- [x] `packages/lint` full suite: 524/524 (39 in `media.test.ts`, incl. 4 new `audio_carve_ungrouped_sources` cases)
- [x] oxfmt/oxlint clean
- [x] fallow: 0 new findings (2 remaining are pre-existing in untouched `audioFx.ts`)
- [ ] Manual verification in `hyperframes preview` (needs a running dev server / browser)


miga-heygen (approved):
## Review

Carve-targets-groups integration is thorough and well-tested. The normative invariant is enforced at three levels:

1. **Core:** `resolveCarveSourceIds` expands group ids to current members at analysis time (never frozen), dedupes, drops dangling refs. Five tests cover dynamic membership, plain clip passthrough, deleted-clip drop, and mixed-id dedup.

2. **Lint:** `audio_carve_ungrouped_sources` catches plural plain clip ids in `sources` as a warning. Four tests cover the key cases: multi-clip warns, group stays quiet, single-clip stays quiet, mixed (group + bare clips) still warns.

3. **UI:** `useFxCarveGrouping.ts` auto-groups when `setCarve` detects 2+ ungrouped voice sources — mints a group id (deduped against all existing ids), calls `onAutoGroupCarveSources` (atomic write via `createAudioGroupAndAssignMembers`), then rewrites the carve's sources to `[groupId]`. The sync-when-possible contract (NOT `async`, returns raw value when no grouping needed) preserves the existing synchronous `setCarve` flow for all non-grouping cases.

4. **Persistence:** `createAudioGroupAndAssignMembers` follows the same atomic pattern as `setElementsHidden` — optimistic live patch, single-file write carrying ALL members, one undo entry, revert on save failure. Three tests cover the happy path, the <2 guard, and the revert path.

The `collectCarveCandidates` correctly offers one entry per group (union overlap, not per-clip) and correctly uses `tagName === "AUDIO"` instead of `instanceof HTMLAudioElement` for cross-realm iframe documents.

No issues. Ship it.

— Miga

terencecho (dismissed):

miga-heygen (dismissed):
## Review

`resolveCarveSourceIds` is correctly designed: resolves fresh at analysis time (not frozen into the attribute), dedupes with first-seen order, drops dangling refs. The lint rule `audio_carve_ungrouped_sources` enforces the §1.6 invariant — warns on ≥2 plain clip ids, silent for group refs or single clips, correctly catches the mixed case (group + bare ids).

`createAudioGroupAndAssignMembers` is atomic (one undo entry for all members), with optimistic live patching and a clean revert on save failure (mirrors `setElementsHidden`'s pattern). The player store update happens only after a successful save — correct ordering.

`withAutoGroupedSources` is deliberately NOT async when nothing to do — careful contract that preserves sync test timing. The `isPromiseLike` check avoids wrapping every `setCarve` call in a microtask. The cross-realm `instanceof` guard (`.tagName === "AUDIO"` instead of `instanceof HTMLAudioElement`) is correct for iframe documents.

8 new carve-grouping tests cover group resolution, auto-grouping atomicity, and the "already grouped" passthrough. 3 `createAudioGroupAndAssignMembers` tests cover happy path, single-element guard, and failure revert.

No issues. Ship it.

— Miga

PR #3291, opened 2026-08-16
feat(studio,core): mute groups, and hear-only-this that cannot reach the export

vanceingalls:
## B5 — mute and solo, on groups and tracks

Stacks on and depends on:
- #3290 (B7 — bus strip)
- #3289 (B4 — group render)
- #3288 (B6 — carve groups)
- #3287 (B3 — group preview)
- #3286 (B2 — group row)
- #3278 (B1 — group model)
- #3277 (P2 — character presets)
- #3276 (P1 — pitch shift)
- #3275 (A2 — audio mute)
- #3274 (A1 — preset primary)

### Group mute
Persisted as `data-hidden` on the `<hf-audio-group>` element itself (never written onto members — design doc §2.1's state-restoration warning). Studio action reuses B7's generic `setAudioGroupAttribute` (`setQuiet`/`setLive`) rather than duplicating `toggleTimelineTrackHidden`'s shape. **Render**: B4 already drops every member of a `data-hidden` group — confirmed by a new `audioMixer.test.ts` case, no production change needed. **Preview**: a dedicated `muteGain` node (`groupInput -> [fx] -> muteGain -> output -> master`) so a mute toggle never fights `scheduleVolumeLane`'s ramps on the same param. Mid-playback toggles sync via a new `syncAudioGroupMute` pass in `init.ts` (a group carries no `data-start`, so it's invisible to the existing visibility-node query). Members of a muted group get the strikethrough label treatment, display only.

### Solo — "Hear only this"
New session-only store slice (`audioSoloSlice`, `soloed: ReadonlySet<string>` of clip/group ids, never track numbers, never serialized). Shared predicate `isAudibleUnderSolo` (`packages/core/src/audioGroups.ts`): audible only if the element or its own group is soloed. "Siblings, never ancestors" lives in the graph — solo gain is a per-element stage only, group buses are never attenuated by solo. Preview: dedicated per-element `soloGain` in `webAudioTransport.ts`, pushed via `window.__hf.setAudioSolo` (a direct call, not an attribute write). `media.ts`'s HTMLMedia fallback folds the same predicate into its per-tick volume computation. Half-lit group indicator for "not soloed itself, but a member is." Exclusive-by-default toggle, ⌘/Ctrl-click to add/remove. Transport-bar banner ("Hearing only `<label>` — your export is not affected", Clear button).

### Export-safety
The most important property: toggling/adding/clearing solo never calls `setAttribute`/`removeAttribute` on any element and never invokes the project save path — both asserted directly via spies in `audioSoloSlice.test.ts`. Solo cannot reach an export by construction.

### Incidental fixes
- Extracted `useHydrateActiveCompPathFromUrl` out of `App.tsx` (pre-existing, unrelated effect) to stay under the 600-line filesize cap after wiring the solo bridge in.
- Fixed a circular dependency the solo-banner wiring introduced: `useAudioSoloBridge.ts` now imports `usePlayerStore` from its concrete module instead of the `player/` barrel (which re-exports `PlayerControls.tsx` — that's what closed the cycle).

### Gates
- `bun run build` clean
- `packages/core` full suite: 2379/2379
- `packages/studio` full suite: 4276/4294 (18 pre-existing todo)
- `packages/engine` `audioMixer.grouping.test.ts`: 5/5
- oxfmt/oxlint clean on all 23 touched files
- fallow clean (0 new circular deps, 0 new filesize/complexity findings)

### DIFF
--- a/packages/core/src/audioGroups.ts
+++ b/packages/core/src/audioGroups.ts
@@ -172,3 +172,35 @@ export function ensureAudioGroupInertStyle(doc: Document): void {
   style.textContent = `${HF_AUDIO_GROUP_TAG}{display:none!important}`;
   doc.head.appendChild(style);
 }
+
+/**
+ * Solo ("Hear only this") predicate — shared by the studio store (which owns
+ * the `soloed` set and the UI's lit/half-lit state) and the preview transport
+ * (which turns it into gain). An element is audible while any solo is active
+ * only if IT is soloed, or its OWN group is soloed (group solo = members
+ * solo). There is no "ancestor" to reach up to in this data model — a group
+ * bus is never itself attenuated by solo, so a soloed member's path through
+ * its group stays open by construction; this predicate only ever gates the
+ * member's own gain. No solo active at all is the one path that returns true
+ * unconditionally.
+ */
+export function isAudibleUnderSolo(
+  soloed: ReadonlySet<string>,
+  id: string,
+  groupId?: string | null,
+): boolean {
+  if (soloed.size === 0) return true;
+  if (soloed.has(id)) return true;
+  return Boolean(groupId && soloed.has(groupId));
+}
+
+/** Half-lit: this group itself isn't soloed, but at least one of its members
+ *  is — the display-only signal that "some of what's under here still plays". */
+export function isGroupHalfLitUnderSolo(
+  soloed: ReadonlySet<string>,
+  groupId: string,
+  memberIds: readonly string[],
+): boolean {
+  if (soloed.size === 0 || soloed.has(groupId)) return false;
+  return memberIds.some((id) => soloed.has(id));
+}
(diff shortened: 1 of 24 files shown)


miga-heygen (dismissed):
## Review

The critical export-safety invariant is correctly maintained: solo is session-only state that never touches any DOM attribute, never invokes the save path, and never flows into the render. Two explicit tests pin this (setAttribute/removeAttribute spy + saveProjectFilesWithHistory spy — both assert zero calls across toggle/add/clear). The mute/solo separation uses dedicated gain nodes (soloGain per member, muteGain per group) so toggles never fight `scheduleVolumeLane`'s ramps.

`isAudibleUnderSolo` predicate is clean: empty set → audible, direct membership → audible, group membership → audible. Group bus is never attenuated by solo — a soloed member's path through its group stays open by construction. `isGroupHalfLitUnderSolo` correctly distinguishes fully-lit (group soloed) from half-lit (member soloed but group not). The `syncAudioGroupMute` diff pass keeps the preview's group bus in sync with mid-playback `data-hidden` toggles.

Solo banner UX correctly resolves labels from the live preview DOM. The `audioSoloSlice` toggle semantics (exclusive by default, ⌘/Ctrl-click for additive) match the design doc.

No issues. Ship it.

— Miga

PR #3292, opened 2026-08-16
feat(studio,core): reach presets and the rack from the timeline

vanceingalls:
## Summary

C1 (last step in the audio-groups/mute-solo/pitch-shift execution stack): the FX button in the timeline track/group header, and its popover — the "reach FX from the timeline" entry point.

Stacks on and depends on (merge after all, in order):
- #3291 (B5 — group mute/solo)
- #3290 (B7 — group bus strip)
- #3289 (B4 — group render)
- #3288 (B6 — carve targets groups)
- #3287 (B3 — group preview routing)
- #3286 (B2 — group timeline rows)
- #3278 (B1 — group data model)
- #3277 (P2 — character presets)
- #3276 (P1 — pitch shift effect)
- #3275 (A2 — audio mute)
- #3274 (A1 — preset primary)

**This is the last PR in the stack.**

- FX button on group rows and on track rows holding exactly one audio clip; reads "FX" or "FX n" (enabled node count). A multi-clip ungrouped audio track gets a "Group these clips to add effects to all of them" pointer instead of a popover, reusing B6's exact auto-grouping write.
- `TimelineFxPopover.tsx`: a thin positioner around the property panel's own `FxPresetMenu`, same audition contract, same preset-apply computation (extracted into a shared `applyPresetToChain` helper rather than duplicated).
- Group writes reuse B7/B5's `onSetAudioGroupAttributeLive/Quiet`. Clip writes go through a new `onSetElementAttributeLive/Quiet` pair, addressed by the clip itself rather than the current selection — the property panel's own write path closes over `domEditSelection`, so a non-selected clip has no synchronous path through it. Built on a newly-extracted shared core (`persistElementAttribute`) that also now backs the existing group-attribute writer, removing a duplication the fallow gate flagged on first commit attempt.
- `HfAudioGroup.fxChain` (already on the B1 model) mirrored onto `TimelineElement.audioGroupFxChain` and `TimelineTrackGroupInfo.fxChain`, alongside the existing volume/hidden mirrors.

**Deviations from the step doc, both disclosed for review:**
- "+ effect" and "Open rack ›" in the popover footer converge on the same navigation (select + hand to the panel) rather than two distinct behaviors — the property panel's rack has no remote way to force its own add-menu open, and building that would be new UI-state plumbing beyond "reuse existing selection dispatch."
- A track that's already a member of a group (some clip carries `data-audio-group`) gets neither the chain button nor the grouping pointer on its own row — its clips' own per-clip FX still works, and the group's own FX button (on `TimelineGroupHeader`) covers the group level.

## Test plan

- [x] `bun run build` clean
- [x] `packages/studio` full suite: 4286/4304 passing (18 pre-existing todo), up from 4276/4294 baseline — 10 new tests, 0 regressions
- [x] New `TimelineFxPopover.test.tsx` (6): exactly-one-write-per-apply, hover-audition-reverts-on-leave, Escape-without-deselecting, outside/inside pointerdown dismissal, footer opens rack
- [x] New `TimelineFxButton.test.tsx` (4): empty/non-empty node-count badge, popover opens on click, group-pointer variant's Group action
- [x] `oxfmt`/`oxlint` clean on all 22 touched files
- [x] `fallow` clean (a duplication finding surfaced on the first commit attempt was fixed by extracting a shared `persistElementAttribute` core; re-verified clean)
- [ ] Manual verification in a running `hyperframes preview` (not done — no browser session available in this run)

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


miga-heygen (approved):
## Review

`applyPresetToChain` correctly extracted from `FxSection`'s inline callback into a shared helper — SSOT improvement, telemetry fires identically from both surfaces. `TimelineFxPopover` is a thin positioner over the existing `FxPresetMenu`, not a second rack — no duplicate serialization.

`TimelineFxButton`'s discriminated union ("chain" vs "group-pointer") implements the design doc's §1.6 rule: single-clip tracks get the preset popover, multi-clip ungrouped tracks get a grouping pointer. `persistElementAttribute` extracted as shared core for both group and clip attribute writes — correct SSOT. `mintGroupId` reuse for C1's auto-grouping is the right call.

Tests pin 7 popover behaviors (apply, audition/revert, Escape propagation, outside/inside click, open rack) and 4 button behaviors (empty chain, enabled-only count, popover opening, group-pointer variant).

No issues. Ship it.

— Miga

miga-heygen (dismissed):
## Review

Timeline FX popover (C1) is a clean thin positioner around existing `FxPresetMenu` — no second rack, no duplicated serialization. The shared `applyPresetToChain` extraction from `FxSection`'s inline logic is correct SSOT: both surfaces (property panel + timeline popover) now share the same apply/reapply/telemetry path.

Key correctness points:
- Outside-click dismiss via capture-phase pointerdown — correct pattern
- Escape stops propagation so it doesn't deselect the clip behind the popover
- Audition on focus (keyboard's hover equivalent) + revert on focusout — accessible preset preview
- `clampedStyle` correctly handles viewport edges and opens upward when below-space is insufficient
- `mintGroupId` exported for the timeline's own auto-group path
- `persistElementAttribute` refactor consolidates the read-patch-save pattern used by both group and element attribute writes

Tests cover: apply writes exactly once and closes, audition on focus + revert on leave, Escape doesn't propagate, outside click dismisses, inside click doesn't dismiss, "Open rack" routes correctly.

No issues. Ship it.

— Miga

terencecho (dismissed):
## Review

**Approve.** Concurring with Miga's approve at this head; adding what I ran independently.

### Independent cross-checks

- **`applyPresetToChain` extraction preserves semantics.** In `FxSection.applyPreset`, the old inline captured `preset` via `getAudioFxPreset(id)` then called `setOpenNode(next.nodes.findIndex((n) => n.fromPreset === preset.id))`. The new caller uses `setOpenNode(next.nodes.findIndex((n) => n.fromPreset === id))` — safe because the helper resolves the preset via `getAudioFxPreset(id)` upstream, so on the non-null return path `preset.id === id`. Telemetry (`trackPresetApplied`) still fires from inside the helper, so both surfaces (property panel + timeline popover) emit identically.
- **`persistElementAttribute` refactor preserves the try/revert-on-save-failure contract** — `patchLive(previousValue)` in the catch mirrors the original `patchLiveGroupAttribute(previewIframe, groupId, attr, previousValue)`. Read-before-write ordering (previous value → live patch → source read → snippet check → apply → save) is unchanged.
- **CI at head:** all required checks green or expected-SKIP for a stack child (base = `wa-23e-group-mute-solo`); `regression`, `preview-regression`, `preview-parity`, `player-perf` all SUCCESS. Reduced-matrix child, so `Test` runs at root of stack — worth a green root-based tip before final merge, though not blocking here.

### Non-blocking nits

1. **`persistElementAttribute` error message lost specificity.** Old: `"Unable to patch audio group ${groupId} in ${targetPath}"`. New (shared core): `"Unable to patch element in ${targetPath}"` — same string for both group and clip writes, with no id. Minor debuggability regression when a snippet lookup fails. Consider threading `patchTarget.domId` (or an equivalent label) into the message.

2. **Single-clip audio track that's grouped still renders the chain FX button.** `TimelineTrackHeader`'s `singleAudioClip` branch has no `!isTrackGrouped` guard, while the sibling `group-pointer` branch does. So a group whose single member sits on its own track shows _two_ FX buttons at once (the group header's group-chain button + the track row's per-clip chain button on the same clip's audio path). The PR body says "A track that's already a member of a group ... gets neither the chain button nor the grouping pointer on its own row" — the code enforces that for the multi-clip case but not the single-clip case. Two clean options: either add `!isTrackGrouped` to the `singleAudioClip` gate (hides the row's button, matches the disclosure), or leave as-is and refine the disclosure to say per-clip FX still surfaces on grouped single-clip tracks. Not blocking either way — the two chains are on different DOM targets, both writes work correctly.

### Concur on

Miga's substantive points hold at this head: `TimelineFxPopover` is a thin positioner (no second serializer), `applyPresetToChain` is correct SSOT, `TimelineFxButton`'s discriminated union captures §1.6, `mintGroupId` reuse is correct, and the 10 new tests pin the load-bearing behaviors (apply-writes-once, audition-revert, Escape-no-propagate, outside/inside pointerdown, open-rack).

— Review by tai (pr-review)

PR #3296, opened 2026-08-17
perf: make propertyPanelColor scroll listener passive

Jakemcanelly:
The third argument was `true` (capture only), so the browser blocks on this handler during scroll. The handler only repositions the panel — never calls preventDefault — so `{ capture: true, passive: true }` is safe and removes the scroll-blocking wait.


jrusso1020:
@Jakemcanelly can you sign the commit please :)

miga-heygen:

PR #3202, opened 2026-08-11
revert(cli): keep HeyGen API traffic on stable prod

miguel-heygen:
## Summary

Normal HyperFrames CLI traffic now stays on stable EF prod. This reverts hyperframes#3201 and removes the older canary route injection from publish and feedback, so login, token lifecycle, API-key billing calls, cloud operations, publish, and feedback no longer depend on a pre-release backend deployment.

The repository now contains zero `heygen_route` injections. The unrelated telemetry/PostHog canary rollout system is unchanged.

## Test plan

- TDD: 3 expected failures before removing the pre-existing publish/feedback headers
- `bun run --cwd packages/cli test` — 2,581 passed, 3 skipped
- `bun run --cwd packages/cli typecheck`
- changed-file oxlint and oxfmt checks
- `bun run build`
- repo-wide `heygen_route` audit — zero matches

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
GPT-5

PR #3297, opened 2026-08-17
docs(changelog): weekly digest 2026-08-10–2026-08-17

jrusso1020:
## Summary

Weekly digest for August 10 to August 17, 2026, prepended to `docs/weekly-updates.mdx`. Rewritten from `git log --no-merges` grouped by type and scope rather than polished from the generator draft.

## The window overlaps last week's digest

106 non-merge commits land inside `since=2026-08-10T00:00:00Z`, `until=2026-08-18T00:00:00Z`, which is the inclusive reading of "through August 17". Five of them were already published in the Aug 3-10 entry, all dated Aug 10:

| commit | PR | what |
|---|---|---|
| `db3de4c1b` | #3148 | follow every redirect a host may answer with |
| `79dff2051` | #3149 | typecheck the scripts directory |
| `f28bc80a1` | #3150 | fail a branch that deletes files main still ships |
| `68205dbbc` | #3089 | search the catalog by meaning, on this machine |
| `33ac86fd3` | #3151 | stop mislabelling capture mode, name silent drawElement refusals |

So this entry covers **101 commits and cites 89 PRs**. The exclusion is computed by set-subtracting the 109 SHAs last week's entry cites, rather than guessing a date cut, because the previous digest's own boundary is not recoverable from its label.

Releases need the same subtraction. Last week's prose ends "Fifteen releases shipped, v0.7.90 through v0.7.105", and both the v0.7.104 (Aug 10 00:19Z) and v0.7.105 (Aug 10 15:08Z) release commits fall inside this window. This week is **v0.7.106 through v0.7.109, four releases**, with the tag set verified contiguous.

## Verified at HEAD, not from commit bodies

- **The audio FX rack is staged, not shipped, and it is roughly half the week's commits.** It sits behind the `audio-fx-rack` canary at `percentage: 0` in `packages/core/src/canaryRegistry.ts`. The gate site is real rather than just a registry entry: `PropertyPanelFlat.tsx` reads `isCanaryEnabled("audio-fx-rack")`. The entry gates the authoring surface only, so a composition that already carries `data-fx-chain` still plays and renders for everyone, and the entry says so explicitly. All three of those facts are in the digest text. Worth raising separately: `canary.test.ts` pins only that the registry entry exists, so it would stay green if the gate site were removed.
- **The video primitive moves** were added (#3090), reverted twenty minutes later because their previews did not deploy (#3162), then relanded with the workflow fixed (#3169). Only the end state is announced.
- **The CLI canary route** was added (#3201) and reverted (#3202) inside the window, so the revert is what gets reported.
- **The preview volume control** (#3282) landed after the v0.7.109 tag, so it is described as shipping in the next release rather than as available now.

## The SHA guard

Every SHA is emitted by a script that refuses to write unless it resolves to 40 hex chars, prefix-matches the abbreviation used in the link text, is an ancestor of `origin/main`, falls inside the window, and was not cited by the previous digest. No SHA was hand-typed.

Two problems the guard turned up in itself before it was trusted:

1. Committer dates from `git log %cI` carry **local offsets** (`-04:00`, `-07:00`), so comparing them as strings against a `Z`-suffixed bound is wrong in both directions at exactly the boundaries the check exists for. It normalizes to UTC now.
2. Coverage is printed, 89 of 100 fresh PR-carrying commits, so it can be compared rather than assumed. The 11 uncited commits are the six release chores, last week's digest PR, an `AGENTS.md` format, and the two revert pairs. All deliberate.

## Testing

- `npx mint validate` run from `docs/`: `success build validation passed`
- lefthook passed without `--no-verify` (largefiles, tracked-artifacts, commitlint)
- Six negative controls on the guard, each confirmed to reject: already-published PR, em-dash, nonexistent PR, TODO marker, duplicate PR, out-of-window commit. A positive control confirmed valid input still writes.
- `git diff origin/main --name-only` returns exactly one file

## Video

Not embedded yet. It follows as a second commit on this branch once rendered and uploaded, matching the shape of the Aug 3-10 PR.


jrusso1020:
Video embedded in the second commit (`3cb51214a`).

https://static.heygen.ai/hyperframes/changelog-videos/weekly-changelog-aug10-17.mp4

45.1s square 1080, built from the `changelog-video` skill. Verified against the rendered file, not the preview:

- `check` 0 errors, contrast 71/71 WCAG AA, motion 0 errors, seam gate 5/5
- captions on every frame of the spoken window except a 2-frame line boundary (0.07s), against a 0.3s threshold
- all six rationed green moments counted in pixels, because a `scaleX` sweep on a zero-width element renders nothing and no gate reports it
- audio mean -18.5 dB, no near-silent stretch over 1.5s
- CDN serving 44172641 bytes, matching the local file, `x-cache: Miss`

Two content notes.

**The audio FX arc is staged, and the video says so on screen.** It is roughly half the week's commits and it sits at `percentage: 0`, so scene 3 carries a `CANARY 0% / OFF BY DEFAULT` chip from its first frame and ends with the panel behind "enabled for 0% of installs". A viewer takes the picture as the claim, so putting the gating only in the voiceover would still have read as shipped. The week's headline is the preview text editing instead.

**One defect no gate could see.** The preset chip was rendering underneath the EQ knobs. `data-layout-allow-overlap` disables collision detection for the slide it sits on, which is every mock slide the skill asks for, so "Layout 0 issues" there means not checked. Found by cropping the panel out of the mp4 and looking at it.

Unrelated small bug worth a follow-up: `.claude/skills/changelog-video/SKILL.md` step 4 points at `skills/hyperframes-media/scripts/heygen-tts.mjs`, which does not exist. The only copy in the tree is `skills/media-use/audio/scripts/heygen-tts.mjs`.

— Rames Jusso

---

**Updated after two more defects turned up in frame review.** Same URL, the object was replaced and the path invalidated, so no new commit. Byte count above reflects the current cut.

1. The outro's kicker and headline were hidden and faded in 0.15s after the cut, leaving about four frames of bare background at the seam. The seam gate passed that boundary on all five checks because it measures the wrapper's rect and velocity, never whether the wrapper has anything visible inside it. The shell is composed at local t=0 now and rides the stamped entry.
2. The terminal mock showed `renderer` / `version`. The real tags are namespaced, so it now reads `hyperframes_renderer` / `hyperframes_version`, which is what you get if you run that command yourself. Worth knowing when checking #3264: a build from source stamps `hyperframes_version=0.0.0-dev` even though the same CLI reports 0.7.109, because the workspace version leaks through. Only a published install stamps a real version, so a source build is not a valid test of the version half.

miguel-heygen (approved):
Approved at exact head `3cb51214a119c4d004407210138a37467510d871`.

Verified: one MDX entry plus video embed; 89 unique in-window PR/commit pairs with zero prior-week overlap; contiguous `v0.7.106`–`v0.7.109`; `audio-fx-rack` at 0% at the actual Studio gate; live CDN object 44,172,641 bytes, H.264/AAC 1080×1080 30fps 45.1s and -18.5 dB; visible canary framing and corrected metadata labels. The sole red CodeQL lane is optional and every annotation is a GitHub-service failure; all eight required contexts are satisfied.

No blocking or important findings.

**Verdict:** APPROVE
**Reasoning:** The digest claims and rendered artifact match the current source and release window; the red lane is verified optional infrastructure noise.

— Magi

PR #3298, opened 2026-08-17
fix(skills): pin UTF-8 in Python scripts instead of the platform code page

miguel-heygen:
## What

Every skill Python script now pins UTF-8 explicitly: stdio is reconfigured to UTF-8 at import, and every text-mode file IO call names its encoding.

## Why

Windows sizes Python's stdio and text-mode file IO to the ANSI code page (cp1252), not UTF-8. Four scripts relied on that default and break there:

- `analyze-beatgrid.py --print` writes glyphs cp1252 has no slot for (`Δ`, `→`), so the brief dies with `UnicodeEncodeError` on every Windows run. This is the reported crash.
- The same script's audiomap write pairs `ensure_ascii=False` with the default file encoding, so a non-ASCII payload is unwritable there too.
- `lint_source.py` calls `path.read_text()`; a Remotion source carrying an em dash or a curly quote raises `UnicodeDecodeError` before any rule runs.
- `gen-stroke-path.py` reads an SVG font whose glyph keys ARE literal characters (`unicode="…"`). Mis-decoded, a key stops matching the requested text — wrong output, no error.

## How

`sys.stdout` / `sys.stderr` get `reconfigure(encoding="utf-8")` at import; `open()` / `read_text()` / `write_text()` all name `encoding="utf-8"`.

Dropping the glyphs for ASCII was the other option. Rejected: it hides the encode defect and leaves the decode side (`read_text`, the SVG font) broken. These are UTF-8 tools on every platform, so they say so.

Two scripts that print only ASCII today (`extract-audio-data.py`, `lyria-recipe.py`) get the same block. That makes the invariant uniform and mechanically checkable instead of per-script judgement. Repo-internal dev scripts under `packages/**` are out of scope — they only run on CI and maintainer machines.

## Test plan

`skills/python-encoding.test.mjs` is the regression guard: it fails if any skill Python script drops the stdio block or omits `encoding=` on a text-mode IO call. Verified it fails on the pre-fix source (2 failures for `analyze-beatgrid.py`) and passes after.

Behaviour verified directly, since the crash needs a cp1252 stdio stream rather than a Windows box. A harness installs a cp1252 `sys.stdout` **before** the module loads (matching how Windows starts the interpreter), then calls `print_brief` on a fixed audiomap:

- pre-fix: `UnicodeEncodeError: 'charmap' codec can't encode character 'Δ' in position 15`
- post-fix: `no crash; stdout re-sized cp1252 -> utf-8; delta and arrow survived` (both glyphs asserted present in the UTF-8 bytes)

Also: `python3 -m py_compile` on all five scripts, `lint_source.py --help`, `bunx oxfmt` + `bunx oxlint` on the new test.

Not covered: no run on real Windows hardware. The cp1252 stream is a faithful stand-in for the failing encode/decode path, but console rendering on legacy `cmd.exe` (as opposed to Windows Terminal) is untested.

### DIFF
--- a/skills-manifest.json
+++ b/skills-manifest.json
@@ -2,7 +2,7 @@
   "source": "heygen-com/hyperframes",
   "skills": {
     "embedded-captions": {
-      "hash": "c501df742bf4e38f",
+      "hash": "1c02f679a329bce7",
       "files": 138
     },
     "faceless-explainer": {
@@ -38,7 +38,7 @@
       "files": 19
     },
     "hyperframes-creative": {
-      "hash": "305bbf795f20c7d8",
+      "hash": "0068574ecd376e60",
       "files": 78
     },
     "hyperframes-keyframes": {
@@ -50,7 +50,7 @@
       "files": 12
     },
     "media-use": {
-      "hash": "51684848affef510",
+      "hash": "1b0ce647f5c7df95",
       "files": 152
     },
     "motion-graphics": {
@@ -58,7 +58,7 @@
       "files": 23
     },
     "music-to-video": {
-      "hash": "e144624dac70abb9",
+      "hash": "0c04f6ed00e90077",
       "files": 132
     },
     "pr-to-video": {
@@ -70,7 +70,7 @@
       "files": 28
     },
     "remotion-to-hyperframes": {
-      "hash": "64d5605e88cf25bd",
+      "hash": "bf184a65059b95e8",
       "files": 70
     },
     "slideshow": {
--- a/skills-manifest.json
+++ b/skills-manifest.json
@@ -2,7 +2,7 @@
   "source": "heygen-com/hyperframes",
   "skills": {
     "embedded-captions": {
-      "hash": "14e79d53c24e3945",
+      "hash": "c501df742bf4e38f",
       "files": 138
     },
     "faceless-explainer": {
@@ -38,7 +38,7 @@
       "files": 19
     },
     "hyperframes-creative": {
-      "hash": "bacb5205ea5eb3d8",
+      "hash": "305bbf795f20c7d8",
       "files": 78
     },
     "hyperframes-keyframes": {
@@ -50,7 +50,7 @@
       "files": 12
     },
     "media-use": {
-      "hash": "7e329ade41b1c1ba",
+      "hash": "51684848affef510",
       "files": 152
     },
     "motion-graphics": {
@@ -58,7 +58,7 @@
       "files": 23
     },
     "music-to-video": {
-      "hash": "55a2b5fdcd6f892c",
+      "hash": "e144624dac70abb9",
       "files": 132
     },
     "pr-to-video": {
@@ -70,7 +70,7 @@
       "files": 28
     },
     "remotion-to-hyperframes": {
-      "hash": "3ecc684432b298dd",
+      "hash": "64d5605e88cf25bd",
       "files": 70
     },
     "slideshow": {
--- a/skills/embedded-captions/scripts/gen-stroke-path.py
+++ b/skills/embedded-captions/scripts/gen-stroke-path.py
@@ -10,10 +10,19 @@ Prints: the path `d` string + layout info on stderr.
 """
 import re, sys
 
+# Windows sizes stdio to the ANSI code page (cp1252). These scripts emit UTF-8 on
+# every platform; say so rather than depending on the console's code page.
+for _stream in (sys.stdout, sys.stderr):
+    if hasattr(_stream, "reconfigure"):
+        _stream.reconfigure(encoding="utf-8")
+
 font_path, text, target_w, baseline_y, x0 = (
     sys.argv[1], sys.argv[2], float(sys.argv[3]), float(sys.argv[4]), float(sys.argv[5]))
 
-svg = open(font_path).read()
+# SVG fonts are UTF-8 and each glyph's `unicode="…"` attribute IS a literal character.
+# Decoded with the platform default, a non-ASCII glyph key silently becomes the wrong
+# character (or raises) on Windows, so that glyph never matches the requested text.
+svg = open(font_path, encoding="utf-8").read()
 glyphs = {}
 for m in re.finditer(r'<glyph\s+unicode="(.)"[^>]*?horiz-adv-x="([\d.]+)"(?:[^>]*?d="([^"]*)")?', svg):
     ch, adv, d = m.group(1), float(m.group(2)), m.group(3) or ""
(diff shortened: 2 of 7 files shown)


somanshreddy (commented):
Reviewed (UTF-8 pinning) — **LGTM, correct fix.** Root cause (platform code page cp1252 on Windows → crash on non-latin) is fixed properly: explicit `encoding="utf-8"` on every text-mode file I/O, `ensure_ascii=False` on `json.dumps` (so non-ASCII writes as real UTF-8, not `\uXXXX`), and a guarded stdio reconfigure (`if hasattr(_stream, "reconfigure"): _stream.reconfigure(encoding="utf-8")` — the `hasattr` guard is right for wrapped/redirected streams). The durable part is the regression test (`python-encoding.test.mjs`) asserting every Python script names `encoding=` on all non-binary I/O + reconfigures stdio — a linter that stops the pattern regressing, exactly what a cross-script encoding bug needs. Clean.

terencecho (approved):
**Approving** — verified at `5d5c42f`. The fix is complete and correct. Two non-blocking nits below, one of them on a line the prior review specifically blessed.

### Verified independently

- All **5** skill Python scripts carry the stdio block — `git ls-tree -r <head> skills/ | grep '\.py$'` returns exactly those five, so no subset was left behind.
- Every text-mode IO call names `encoding="utf-8"`. The one binary call, `lyria-recipe.py:113` `wave.open(str(out_path), "wb")`, is correctly left alone.
- **The guard actually runs.** `Test: skills` executed 69 files including `skills/python-encoding.test.mjs` (477 tests / 0 fail), and the job's `mapfile` step hard-fails on an empty match while the test itself asserts `scripts.length >= 5`. Not a green-by-skip lane.
- `Skills: manifest in sync` is green, so the four hash bumps are right.
- The only red check is **not this PR**: `Analyze (python)` died inside CodeQL `init` with `Encountered an error while trying to determine feature enablement: HttpError: No server is currently available to service your request`. Same check is green on `main`, and it isn't a required context.

### Concur

Agreed with @somanshreddy's read ([review](https://github.com/heygen-com/hyperframes/pull/3298#pullrequestreview-4956081916)) on the root cause and on the value of shipping the linter next to the fix. One correction: `ensure_ascii=False` is **pre-existing**, not added here. That's actually the sharper framing — it's precisely what turned the missing `encoding=` into an unwritable-payload bug rather than a latent one, which is the argument the PR body makes.

### 1. `reconfigure(encoding=...)` resets `errors` to `strict`, downgrading stderr (minor)

Passing only `encoding=` does not preserve the error handler. Reproduced:

```
BEFORE  stderr: utf-8 backslashreplace
AFTER   stderr: utf-8 strict
```

CPython sets `sys.stderr.errors = "backslashreplace"` unconditionally — verified under piped stderr and `-X utf8=0` — specifically so the diagnostic path can never itself raise. After this block, it can:

```
pre-reconfigure  stderr ok: path-\udce9.wav
post-reconfigure RAISED: 'utf-8' codec can't encode character '\udce9' in position 5: surrogates not allowed
```

Reachable in this diff: `lint_source.py:326` enumerates `args.path.rglob("*")` and prints those paths at `:355`, so an unpaired-surrogate filename now raises where it previously escaped. Strictly narrower than the bug being fixed (which fired on *every* Windows `--print` run), so not a blocker — but it is the same failure class relocated onto the error-reporting path. One keyword:

```python
_stream.reconfigure(encoding="utf-8", errors=_stream.errors)
```

### 2. The guard's IO regex has two silent-exemption classes (minor — durability, not correctness)

No live escape at head; I grepped every IO call directly and they're all pinned. But the linter is the part meant to outlive the fix, and it under-matches. Probed both against the exact regex:

```
write_text(json.dumps({"bpm": 120}))          matched=1 flagged=0
write_text(json.dumps(d, default=str(x)))     matched=0 flagged=0
```

- The binary heuristic `/["']\w*b\w*["']/` runs against the **whole** call text, not the mode argument, so any quoted word containing a `b` exempts the call. `{"bpm": ...}` / `{"bars": ...}` — literally `analyze-beatgrid.py`'s own payload keys — would sail through unpinned.
- The outer pattern tolerates only one level of paren nesting, so a call with two isn't matched *at all* rather than flagged.

Deriving the mode from the second positional argument only would close both.

### 3. Coverage boundary (FYI, consistent with the PR's own disclosure)

`Render on windows-latest` and `Tests on windows-latest` are required contexts but both **SKIPPED** here, because the `code` path filter excludes `skills/**`. So a Windows-targeted fix got zero Windows CI. The cp1252-stdout harness is the right stand-in — `librosa`/`soundfile` on a Windows runner isn't worth the minutes — I'm just flagging the shape so nobody reads those two lanes as Windows validation.

Also outside the guard: `bytes.decode()`. `extract-audio-data.py:60` does `result.stderr.decode()`, which is strict UTF-8 over ffmpeg's bytes, so a Windows ffmpeg emitting cp1252 in an error message raises `UnicodeDecodeError` and masks the real failure. Worth `errors="replace"` whenever this area is next touched.

— Review by tai (pr-review)

miguel-heygen:
Both findings were real; I reproduced each before changing anything. Pushed in `6be7c05`.

**1. `errors` reset — fixed.** Confirmed exactly as described:

```
BEFORE stderr: utf-8 backslashreplace
AFTER  stderr: utf-8 strict
post-reconfigure RAISED: 'utf-8' codec can't encode character '\udce9': surrogates not allowed
```

You're right that this is the same failure class relocated onto the reporting path, which is worse than it looks — a crash in the diagnostic buries whatever it was diagnosing. All five scripts now pass `errors=_stream.errors`, and the guard asserts that form, so the keyword can't be dropped again. Re-verified: `backslashreplace` survives and the surrogate filename escapes instead of raising.

**2. Guard blind spots — fixed.** Reproduced both classes against the old regex, then re-probed the new matcher:

| probe | before | after |
|---|---|---|
| `write_text(json.dumps({"bpm": 120}))` | matched, **not flagged** | flagged |
| `write_text(json.dumps(d, default=str(x)))` | **not matched** | flagged |
| `open(p, "wb")` / `open(p, "rb")` | pass | pass |
| `open(p, "w")` | flagged | flagged |

Calls are now found by scanning to the balanced close paren rather than by a depth-capped regex, and the mode is read from `open`'s second positional argument only (`read_text`/`write_text` have no mode, so they're always text). Your framing was the useful part: the linter is the piece meant to outlive the fix, so under-matching is the failure mode that matters, and `bpm`/`bars` being analyze-beatgrid's own payload keys made that concrete rather than theoretical.

**3. `bytes.decode()` — taken now rather than later.** `extract-audio-data.py:60` decodes ffmpeg's stderr strictly *while reporting a failure*, so a Windows ffmpeg's cp1252 bytes would bury the real error. Same class as the rest of the PR and in a file already touched here, so it's in: `errors="replace"`.

Also correct on `ensure_ascii=False` being pre-existing — I'd implied otherwise. Your framing is sharper: it's precisely what turned a missing `encoding=` from latent into an unwritable payload.

On the Windows CI boundary: agreed, and flagged the same way in the PR body. `Render on windows-latest` / `Tests on windows-latest` skip on `skills/**`, so this got zero Windows CI; the cp1252 stream is a stand-in for the encode/decode path, not for console rendering on legacy `cmd.exe`.

PR #3300, opened 2026-08-17
fix(skills): declare the caption brand font's style axis, not just its weight

miguel-heygen:
## What

`brandFontFaces()` in `captions.mjs` now reads a **style** axis out of the font filename, not just a weight, and dedupes faces on the pair. It also treats a numeric weight in the filename as authoritative.

## Why

`@font-face` is deliberately document-global: the composition CSS scoper exempts it (`GLOBAL_AT_RULES` in `packages/core/src/compiler/compositionScoping.ts`), because a face declaration cannot be scoped. That makes this generator the one part of the captions sub-composition whose output reaches every sibling composition, so a wrong axis is a whole-document defect.

**Italic squatting the upright slot.** Google Fonts ships Newsreader as two files:

```
Newsreader-Italic-VariableFont_opsz,wght.ttf
Newsreader-VariableFont_opsz,wght.ttf
```

The italic sorts first and both scored weight 400, so the italic claimed the family's only `Newsreader-400` slot, the upright was skipped as a duplicate, and the surviving face was declared with no `font-style`. Mounting captions therefore re-pointed the whole document's Newsreader at the italic file — every sibling composition rendered in italics.

Reproduced end-to-end on a fixture project (`frame.md` naming Newsreader + both files staged in `assets/fonts/`), running the real `captions.mjs build`:

```
before: @font-face { font-family: 'Newsreader'; src: url('assets/fonts/Newsreader-Italic-VariableFont_opsz,wght.ttf') ...; font-weight: 400; font-display: block; }
        (one face, italic file, no font-style, upright file absent)

after:  @font-face { ... Newsreader-Italic-VariableFont_opsz,wght.ttf ...; font-weight: 400; font-style: italic;  ... }
        @font-face { ... Newsreader-VariableFont_opsz,wght.ttf ...;        font-weight: 400; font-style: normal;  ... }
```

**Same misparse, second axis.** `weightOf()` matched weight *words* only. Fontsource names every face numerically and carries no word (`inter-latin-500-normal.woff2`), so an entire family scored 400 and exactly one of its faces shipped — bold caption text fell back to a synthesized bold.

## How

- `styleOf(filename)` → `italic` for `italic`/`oblique`, else `normal`; emitted as a `font-style` descriptor.
- The dedup key becomes `family-weight-style`, so an italic file can no longer take the upright slot.
- `weightOf()` prefers a `[1-9]00` token in the filename over the word heuristic. Word-named files (`TT_Norms_Pro_Bold.woff2`) are unaffected — there is a test pinning that.

`brandFontFaces` is exported as a seam for unit testing, matching how `buildFromSkin` already is.

All three copies of `captions.mjs` / `captions.test.mjs` (product-launch-video, faceless-explainer, pr-to-video) are updated and verified byte-identical.

## Test plan

Three tests added to `captions.test.mjs`:

- an italic file is declared italic and never squats the upright slot (the reported bug)
- a numeric weight in the filename is read as the weight (4 fontsource faces → 4 distinct declarations)
- word-named weights still parse when there is no numeric axis (regression guard on existing behaviour)

Verified the first two **fail** on the pre-fix axis logic — `AssertionError: both the upright and the italic file must be declared / 1 !== 2` — and pass after. The third passes either way by design.

Also: `bun run test:skills` → 474 pass, 1 fail. That failure (`resolve.test.mjs` → "smart grade merges measured adjust") reproduces on clean `origin/main` in the same tree, so it is pre-existing and unrelated. `bunx oxfmt --check` and `bunx oxlint` clean.

Not covered: no browser render comparing italic vs upright pixels — the evidence is the generated `@font-face` block, which is what the browser acts on.

Not addressed (called out in the commit, left for a follow-up): a `VariableFont` file is still pinned to a single `font-weight` instead of declaring its range, so weights it could interpolate are still synthesized.

### DIFF
--- a/skills-manifest.json
+++ b/skills-manifest.json
@@ -6,7 +6,7 @@
       "files": 138
     },
     "faceless-explainer": {
-      "hash": "63fe67e13ef4f237",
+      "hash": "1eb3772e62dd71bb",
       "files": 24
     },
     "figma": {
@@ -62,11 +62,11 @@
       "files": 132
     },
     "pr-to-video": {
-      "hash": "b0299c437334f7a7",
+      "hash": "01f46da1e17577ea",
       "files": 30
     },
     "product-launch-video": {
-      "hash": "ca3bf6aeb20e786e",
+      "hash": "d562efe00647c14b",
       "files": 28
     },
     "remotion-to-hyperframes": {
--- a/skills-manifest.json
+++ b/skills-manifest.json
@@ -6,7 +6,7 @@
       "files": 138
     },
     "faceless-explainer": {
-      "hash": "c70b904aa68cf7e5",
+      "hash": "56a2ec49665b0d41",
       "files": 24
     },
     "figma": {
@@ -62,11 +62,11 @@
       "files": 132
     },
     "pr-to-video": {
-      "hash": "7769801640dca521",
+      "hash": "85d539e0def28c3f",
       "files": 30
     },
     "product-launch-video": {
-      "hash": "81953f054fcb9d91",
+      "hash": "60c4683c0183e223",
       "files": 28
     },
     "remotion-to-hyperframes": {
--- a/skills-manifest.json
+++ b/skills-manifest.json
@@ -6,7 +6,7 @@
       "files": 138
     },
     "faceless-explainer": {
-      "hash": "56a2ec49665b0d41",
+      "hash": "63fe67e13ef4f237",
       "files": 24
     },
     "figma": {
@@ -62,11 +62,11 @@
       "files": 132
     },
     "pr-to-video": {
-      "hash": "85d539e0def28c3f",
+      "hash": "b0299c437334f7a7",
       "files": 30
     },
     "product-launch-video": {
-      "hash": "60c4683c0183e223",
+      "hash": "ca3bf6aeb20e786e",
       "files": 28
     },
     "remotion-to-hyperframes": {
--- a/skills/faceless-explainer/scripts/build-frame.mjs
+++ b/skills/faceless-explainer/scripts/build-frame.mjs
@@ -452,7 +452,12 @@ if (brandFonts.length) {
     // A numeric axis is the font's own answer, so it beats the word heuristic. Fontsource
     // names every face that way and carries no weight WORD at all, so word-only parsing
     // scored a whole family "Regular" and staged exactly one of its faces.
-    const numeric = /(?:^|[^0-9])([1-9]00)(?:[^0-9]|$)/.exec(s);
+    //
+    // A weight token must not be buried inside a longer run: this reads capture files,
+    // which are commonly hash-named, and "Newsreader-a1b200c3.woff2" is not a 200-weight
+    // face. Hence a non-digit before (which also stops "2100" reading as 100) and no
+    // alphanumeric after. "Roboto900.ttf" still parses.
+    const numeric = /(?:^|[^0-9])([1-9]00)(?![0-9a-z])/.exec(s);
     if (numeric) return { n: Number(numeric[1]), w: numeric[1] };
     if (/black|heavy|ultra|extrabold/.test(s)) return { n: 800, w: "ExtraBold" };
     if (/semibold|demibold/.test(s)) return { n: 600, w: "SemiBold" };
--- a/skills/faceless-explainer/scripts/build-frame.mjs
+++ b/skills/faceless-explainer/scripts/build-frame.mjs
@@ -431,8 +431,15 @@ if (brandFonts.length || (brandColors.length && presetColors.length)) {
 // ── stage brand font files + emit @font-face ──────────────────────────────────
 // A brand font is rarely a Google font, so renaming the family in frame.md is not enough:
 // nothing loads the actual face. If the capture downloaded font files, copy them to
-// assets/fonts/ under CLEAN, weight-named names (so captions.mjs' family-prefix matcher
+// assets/fonts/ under CLEAN, face-named names (so captions.mjs' family-prefix matcher
 // finds them too) and append a ready-to-paste, ROOT-RELATIVE @font-face block to frame.md.
+//
+// The staged NAME is a contract, not cosmetics: captions.mjs derives each face's weight and
+// style back out of it. So the name has to carry every axis that distinguishes one face from
+// another, and the dedup key has to be the whole face. Naming on weight alone made Google's
+// two-file Newsreader download (upright + italic, both scoring "Regular") collide on one
+// slot: the italic sorts first, took the name, the upright was never staged, and the block
+// below then asserted font-style:normal over italic bytes.
 if (brandFonts.length) {
   const norm = (s) =>
     String(s)
@@ -442,6 +449,11 @@ if (brandFonts.length) {
   const FMT = { woff2: "woff2", woff: "woff", ttf: "truetype", otf: "opentype" };
   const weightInfo = (name) => {
     const s = name.toLowerCase();
+    // A numeric axis is the font's own answer, so it beats the word heuristic. Fontsource
+    // names every face that way and carries no weight WORD at all, so word-only parsing
+    // scored a whole family "Regular" and staged exactly one of its faces.
+    const numeric = /(?:^|[^0-9])([1-9]00)(?:[^0-9]|$)/.exec(s);
+    if (numeric) return { n: Number(numeric[1]), w: numeric[1] };
     if (/black|heavy|ultra|extrabold/.test(s)) return { n: 800, w: "ExtraBold" };
     if (/semibold|demibold/.test(s)) return { n: 600, w: "SemiBold" };
     if (/bold/.test(s)) return { n: 700, w: "Bold" };
@@ -449,6 +461,7 @@ if (brandFonts.length) {
     if (/light|thin/.test(s)) return { n: 300, w: "Light" };
     return { n: 400, w: "Regular" };
   };
+  const styleOf = (name) => (/italic|oblique/i.test(name) ? "italic" : "normal");
   const fams = [...new Set(brandFonts)];
   const srcDirs = [
     join(hyperframesDir, "capture/assets/fonts"),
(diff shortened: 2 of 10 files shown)


somanshreddy (commented):
Reviewed (font style axis) — **LGTM, correct fix, verified identical across the 3 copies.** The missing `font-style` axis let an italic variable font (e.g. Google's `Newsreader-Italic-VariableFont_opsz,wght.ttf`, which sorts first) claim the family via the global `@font-face` → every sibling composition italicized. The fix derives style from the filename (`styleOf`), adds `font-style: ${style}`, and dedups per-FACE (`${fam}-${w}-${style}`) so italic + normal at the same weight are distinct faces. Confirmed the same fix landed identically in all three duplicated `captions.mjs` (faceless-explainer / pr-to-video / product-launch-video) — no copy left behind. `export { brandFontFaces }` as a test seam is a nice touch. (The 3-way script duplication is a pre-existing maintenance smell — a shared module would remove the identical-fix-in-N-places risk — but that's out of scope here.)

terencecho (approved):
Reviewed at `72641bc` (style + weight axes, and the upstream that feeds them).

**Verdict: approve.** The fix is correct and is a strict improvement — no regression path. Findings below are a follow-up, not a merge blocker.

### Concur with @somanshreddy

Same read on the core fix. Two things I verified independently rather than restate:

- **3-copy parity is real, byte-for-byte.** `sha256` of the three `captions.mjs` at head all = `48dfc4b8060f350a…`; the three `captions.test.mjs` all = `d8da6fcf6e27d268…`. No copy left behind, and no drift smuggled into one of them.
- **The skipped required checks are legitimate path-filtering, not green-by-skip.** Required set is `Semantic PR title / Test / Test: runtime contract / Typecheck / Build / regression / Render on windows-latest / Tests on windows-latest`; the skipped ones all cover `packages/**`, which this diff doesn't touch. The lane that actually gates a `skills/**` diff ran green: `Test: skills` ✅, `Skills: manifest in sync` ✅ (which is what validates the three `skills-manifest.json` hash updates, so I didn't recompute them), `Skills: project-native lint + mirror` ✅, `Format` ✅.
- `Analyze (actions)` ❌ is **not** a required check and is not diff-caused — the job log is a GitHub-side 503 during CodeQL `init`: `Encountered an error while trying to determine feature enablement: HttpError: No server is currently available to service your request`. This PR touches zero workflow files. Same check is green on `main`. Re-run at leisure.

---

### Finding 1 — the fix is inert on the brand-font path, because `build-frame.mjs` renames the style token away first

This is pre-existing, so not a blocker for this PR, but it means the reported symptom (whole document italicized) **still reproduces after this merges** on the flow that stages captured brand fonts. Worth naming, because the two files are coupled by design — `build-frame.mjs:434` says the staged names exist "so captions.mjs' family-prefix matcher finds them too", and this PR's new `styleOf` reads exactly that filename.

`skills/*/scripts/build-frame.mjs` stages captured fonts under a rewritten name that carries **only** the weight word:

```js
const { n, w } = weightInfo(f);                                   // :471  — word-only, no style axis
const clean = `${fam.replace(/[^A-Za-z0-9]/g, "")}-${w}.${extOf(f)}`;  // :472  — "Newsreader-Regular.woff2"
if (stagedNames.has(clean)) continue;                             // :473  — dedup on family+weightWORD
…
`@font-face{font-family:"${fam}";font-weight:${n};font-style:normal;…}`  // :478  — hardcoded normal
```

Traced end-to-end with Google's own two-file Newsreader download in `capture/assets/fonts/`:

1. `build-frame.mjs:471` — both the italic and the upright score `{n: 400, w: "Regular"}`; `weightInfo` has no style axis.
2. `:472` — both compute the *same* `clean` = `Newsreader-Regular.ttf`, and **the style token is gone from the name**.
3. `:459` sorts, so `Newsreader-Italic-VariableFont…` is copied first; the upright then hits `stagedNames.has(clean)` at `:473` and is **never staged**.
4. `:478` emits `font-style:normal` over italic bytes into the paste-ready block in `frame.md`.
5. `brandFontFaces` then scans `dirs` in order `assets/fonts` (`captions.mjs:300`) before `capture/assets/fonts` (`:301`):
   - `assets/fonts/Newsreader-Regular.ttf` → `styleOf` (`:325`) finds no `italic` in the name → claims `Newsreader-400-normal` at `:363`. **Italic bytes, declared `normal`.**
   - `capture/assets/fonts/Newsreader-Italic-VariableFont….ttf` → correctly `400/italic`. ✅
   - `capture/assets/fonts/Newsreader-VariableFont….ttf` → `400/normal` → collides on `seen` at `:364` → **dropped**, even though the genuine upright file is sitting right there.

Net: the document-global `normal` slot is still held by the italic file. Since `@font-face` is exempt from the scoper, that is still the whole-document defect this PR set out to kill — and it's now *asserted* as `font-style: normal` rather than merely omitted, which is harder to spot than the original.

The fix belongs in `build-frame.mjs`: carry the style into `clean` (and into the emitted descriptor), and dedup on the pair — i.e. the same two moves this PR just made in `captions.mjs`. Present in all three copies (`faceless-explainer`, `pr-to-video`, `product-launch-video`); note those three have already drifted apart (three distinct hashes), unlike `captions.mjs`.

### Finding 2 — `weightInfo` didn't get the numeric-axis preference either

Same divergence, second axis, same file. `build-frame.mjs:443-451` is still word-only, and it's the *producer* of the names `weightOf` reads. So for a Fontsource-shaped capture (`inter-latin-400-normal.woff2` … `-700-normal.woff2`) every file scores `Regular`, all four collapse onto one `Inter-Regular.woff2` at `:472-473`, and three are dropped at staging — before `captions.mjs`' new numeric rule at `:309` ever gets to see a digit. The new test proves the numeric rule works on files staged by hand; it can't fire on files staged by `build-frame.mjs`.

### Finding 3 (minor) — numeric axis can false-positive on hash-named captured files

`captions.mjs:309`'s `/(?:^|[^0-9])([1-9]00)(?:[^0-9]|$)/` runs against whatever is in `capture/assets/fonts`, and `build-frame.mjs:460` notes that path is commonly "hash-named files". A hash like `a1b200c3.woff2` now scores weight 200 instead of the old 400, which can leave the real 400 slot unfilled and change which file wins which weight. Low severity — CSS weight matching still picks *some* face from the family — and strictly better than the status quo for the named cases. If you want to tighten it, anchoring to a delimiter (`[-_.]([1-9]00)[-_.]`) keeps both `inter-latin-500-normal` and `TT_Norms_Pro_400_Regular` while ignoring hash noise; the trade is `Roboto900.ttf`-style names falling back to the word path. Your call — fine as-is.

---

Nits, non-blocking: the 4-digit guard in the numeric regex is a nice detail (`2100` correctly doesn't match `100`). And the "only italic file staged" case is safe — declaring `font-style: italic` on a family's sole face doesn't strand upright text, since CSS style matching falls back within the family before it falls back to the next family.

— Review by tai (pr-review)

miguel-heygen:
Finding 1 is correct and it's the important one: **the fix was inert on the path the bug was actually reported on.** Fixed in `02cb41b`.

I reproduced your trace end to end rather than taking it on faith. Fixture holding what `build-frame.mjs` leaves behind, run through the real `captions.mjs`:

```
before  assets/fonts/Newsreader-Regular.ttf          → 400 normal   ← italic bytes
        capture/.../Newsreader-VariableFont.ttf      → dropped      ← the real upright
```

So the document-global `normal` slot still pointed at the italic file, and — as you said — it was now *asserted* `font-style: normal` rather than merely omitted, which is strictly harder to spot than the original. That's a worse end state than not shipping the first commit, so thank you for catching it.

The framing I took from this: **the staged filename is a contract, not cosmetics.** `brandFontFaces` derives the axes back out of it, so any axis `build-frame` drops while renaming collapses two faces onto one slot. `clean` now carries the style, the dedup key is the whole face, and the emitted descriptor reports the real style:

```
after   assets/fonts/Newsreader-Regular.ttf          → 400 normal
        assets/fonts/Newsreader-Regular-Italic.ttf   → 400 italic
```

**Finding 2 — taken.** `weightInfo` has the numeric-axis preference now, for exactly the reason you gave: it's the *producer* of the names `weightOf` reads, so a Fontsource-shaped capture collapsed at staging before the consumer's numeric rule could ever see a digit. My new test proving the numeric rule worked on hand-staged files was, as you noted, testing a path real captures don't take.

All three `build-frame.mjs` copies carry it. They'd already drifted apart elsewhere (three distinct hashes), so each was edited individually rather than copied — and the drift is worth its own cleanup.

Two tests pin both ends of the contract now: a round-trip asserting the names `build-frame` produces map back to the right weight+style through the real `brandFontFaces`, and a source check that no copy reverts to a weight-only name or a hardcoded `font-style:normal`. Verified the second fails against `origin/main`.

**Finding 3 — leaving as-is, per your call, and here's why.** The hash-name false positive needs the file to *also* pass the family-prefix matcher (`norm(f).startsWith(key)`), so a bare `a1b200c3.woff2` is never claimed at all; it only bites a family-prefixed hash suffix like `Newsreader-a1b200c3.woff2`. Against that, the delimiter anchor costs `Roboto900.ttf`. Narrow either way, and now much rarer than before, since `build-frame` renames captured files deterministically instead of leaving hash names for `captions.mjs` to guess at. Happy to take the anchor if you'd rather have that trade.

Agreed on the 3-way duplication being a pre-existing smell — this PR touched six copies of two files to make one change, which is the argument making itself.

miguel-heygen:
Took finding 3 as well — and it turned out not to need the trade you priced in.

The delimiter anchor (`[-_.]([1-9]00)[-_.]`) does kill the hash case, but costs `Roboto900.ttf`. A negative lookahead keeps both: the token needs a **non-digit before** and **no alphanumeric after**.

| filename | before | after | |
|---|---|---|---|
| `Newsreader-a1b200c3.woff2` | 200 | word path → Regular | hash noise, rejected |
| `Inter-2100.woff2` | word path | word path | 4-digit guard held |
| `Roboto900.ttf` | 900 | 900 | still parses |
| `inter-latin-500-normal` | 500 | 500 | unchanged |
| `TT_Norms_Pro_400_Regular` | 400 | 400 | unchanged |

Same rule in `build-frame.mjs` — it's the producer of the names `captions.mjs` reads, and it's the one running directly against the hash-named capture dir, so anchoring only the consumer would have left the real exposure open.

Also added the **captions.mjs parity test** you implied with the duplication note. Not the shared module — that's its own initiative — but this branch alone had to land the same two-axis fix in three copies by hand, so the next fix landing in only one should fail loudly. Mirrors the existing `package-loader` parity pin.

Both pinned by tests; `fallow audit` clean on 10 changed files, `test:skills` 486/487 (the one failure reproduces on clean `origin/main`).

PR #3301, opened 2026-08-17
fix(skills): count proxy-driver tweens in the animation map

miguel-heygen:
## What

`enumerateTweens` in `animation-map.mjs` now keeps a tween whose `targets()` holds no Element **when it carries an `onUpdate`**, marked `driver: "onUpdate"`.

## Why

The walk opened with:

```js
const targets = (node.targets?.() ?? []).filter((t) => t instanceof Element);
if (!targets.length) return;
```

That deletes the **proxy-driver idiom** — tween a plain object, apply the motion inside `onUpdate` — which is real, visible animation. The consequence is not just a missing row in the report: `computeDensity` counted zero active tweens across the tween's span, so `findDeadZones` reported animating time as **dead**, telling agents to add motion to a stretch that already had it.

Reproduced end-to-end with the real script on a 4s composition: `#mover` tweened normally over 0-1s, and a proxy `{v:0}` → `{v:100}` over 2-4s whose `onUpdate` writes a visible counter.

```
before:  Animation map: 1/1 tweens (skipped 0 micro-tweens)
           dead zones: 1.5-4s
         ###############                                               #mover x

after:   Animation map: 2/2 tweens (skipped 0 micro-tweens)
         ################                                              #mover x
                               ##############################  (onUpdate driver) v
```

`deadZones` goes from `[{1.5 ... 4}]` to `[]`.

## How

The discriminator is `onUpdate`, and it matters: a bare `tl.to({}, { duration: D })` spacer has no `onUpdate`, produces nothing, and stays dropped. Every preset caption skin ends with exactly such a full-span anchor tween — counting those would mask genuine dead zones, trading a false negative for a false positive.

There is no element to select or measure for a driver tween, so four call sites downstream had to agree on that:

- `selectorHint` is `null`, not a placeholder string. It feeds `document.querySelector` in the sampler, so it has to be **absent** rather than unmatchable — a string like `"(proxy)"` would throw a `SyntaxError` inside `page.evaluate`.
- bbox sampling is skipped; the report carries `selector: "(onUpdate driver)"` and `bboxes: []`.
- `computeFlags` guards its geometry flags on `bboxes.length`. `[].every()` is vacuously true, so an unmeasured tween would otherwise be reported as both `degenerate` and `invisible`. The pacing flags read only start/end and still apply.
- `describeTween` says the motion is applied in JS and no geometry was measured, instead of printing a null selector.

## Test plan

New test in `animation-map.test.mjs`, using the existing fake-producer harness: the fake session's `page.evaluate` runs the callback in-process against a stubbed `window`/`document`, so the real `enumerateTweens` / `computeDensity` / `findDeadZones` code runs without a browser. The fixture timeline holds all three shapes — an element tween, a proxy driver, and a bare spacer — and asserts:

- exactly one `driver: "onUpdate"` tween, spanning 2-4s, `targets: 0`, `bboxes: []`
- its `flags` are empty (the vacuous-truth guard)
- `deadZones` is empty
- the report has 2 tweens, so the bare spacer is still excluded

Verified it fails without the fix: `AssertionError: expected one onUpdate driver in: Animation map: 1/1 tweens ... degenerate: 1`.

Also: `bun run test:skills` → 475 pass, 1 fail. That failure (`resolve.test.mjs` → "smart grade merges measured adjust") reproduces on clean `origin/main` in the same tree, so it is pre-existing and unrelated. `bunx oxfmt` + `bunx oxlint` clean.

Not covered: the tween is reported, not attributed. Which element an `onUpdate` mutates is not knowable without executing it.

Corrected after review — an earlier version of this note claimed driver tweens stay out of `elements` and collision detection. Only the second half is true. `markCollisions` genuinely skips them (their `bboxes` are empty), but `buildElementLifecycles` keys on `tw.selector`, so every driver in a composition collapses into one pseudo-element `"(onUpdate driver)"` with null `endsVisible`/`finalPosition`, and `detectStaggers` can group several same-duration drivers into a phantom stagger. Cosmetic: `animation-map.json` is read by `SKILL.md` guidance and this script, and nothing parses those fields programmatically.

Also addressed after review: the discriminator was tween-local, so the WebGL/uniform idiom — `gsap.timeline({ onUpdate: renderFrame })` over children tweening plain uniform objects — was still dropped entirely, leaving `held-message-living-field` as one 12s false dead zone. `walk()` now threads a `driven` flag beside `parentOffset`. Inherited drivers are held to a stricter test than tween-local ones: the tween must also change a property, because a bare spacer under a driven timeline repaints an identical frame and counting it would mask a real dead zone.

### DIFF
--- a/skills-manifest.json
+++ b/skills-manifest.json
@@ -22,7 +22,7 @@
       "files": 17
     },
     "hyperframes-animation": {
-      "hash": "989adc5cdd46c76e",
+      "hash": "4418a24093c4b1c1",
       "files": 121
     },
     "hyperframes-audio": {
--- a/skills-manifest.json
+++ b/skills-manifest.json
@@ -22,7 +22,7 @@
       "files": 17
     },
     "hyperframes-animation": {
-      "hash": "b2686eccf21b517f",
+      "hash": "989adc5cdd46c76e",
       "files": 121
     },
     "hyperframes-audio": {
--- a/skills-manifest.json
+++ b/skills-manifest.json
@@ -22,7 +22,7 @@
       "files": 17
     },
     "hyperframes-animation": {
-      "hash": "2ce5ca7dbf361e27",
+      "hash": "b2686eccf21b517f",
       "files": 121
     },
     "hyperframes-audio": {
--- a/skills/hyperframes-animation/scripts/animation-map.mjs
+++ b/skills/hyperframes-animation/scripts/animation-map.mjs
@@ -104,7 +104,11 @@ try {
       (_, k) => +(tw.start + ((k + 0.5) / FRAMES) * (tw.end - tw.start)).toFixed(3),
     );
 
-    const bboxes = await sampleTweenBboxes(session.page, tw.selectorHint, times);
+    // No selector means no element to measure (an onUpdate driver). Sampling anyway
+    // would hand querySelector an unmatchable string.
+    const bboxes = tw.selectorHint
+      ? await sampleTweenBboxes(session.page, tw.selectorHint, times)
+      : [];
 
     const animProps = tw.props.filter(
       (p) => !["parent", "overwrite", "immediateRender", "startAt", "runBackwards"].includes(p),
@@ -114,7 +118,8 @@ try {
 
     report.tweens.push({
       index: i + 1,
-      selector: tw.selectorHint,
+      selector: tw.selectorHint ?? "(onUpdate driver)",
+      driver: tw.driver,
       targets: tw.targetCount,
       props: animProps,
       start: +tw.start.toFixed(3),
@@ -194,8 +199,14 @@ async function enumerateTweens(session) {
         return;
       }
       const targets = (node.targets?.() ?? []).filter((t) => t instanceof Element);
-      if (!targets.length) return;
       const vars = node.vars ?? {};
+      // The proxy-driver idiom tweens a plain object and applies the motion in onUpdate,
+      // so targets() holds no Element. Dropping those tweens hid real motion from the
+      // map: computeDensity saw zero active tweens over their span and findDeadZones
+      // reported it as dead. There is no element to select or measure here, but the span
+      // is real, so keep the tween and mark why it carries no geometry.
+      const isProxyDriver = targets.length === 0 && typeof vars.onUpdate === "function";
+      if (!targets.length && !isProxyDriver) return;
       const props = Object.keys(vars).filter(
         (k) =>
           ![
@@ -213,7 +224,10 @@ async function enumerateTweens(session) {
       const start = parentOffset + (node.startTime?.() ?? 0);
       const end = start + (node.duration?.() ?? 0);
       results.push({
-        selectorHint: selectorOf(targets[0]) ?? "(unknown)",
+        // null, not a placeholder string: this feeds document.querySelector downstream,
+        // so it must be absent rather than unmatchable.
+        selectorHint: isProxyDriver ? null : (selectorOf(targets[0]) ?? "(unknown)"),
+        driver: isProxyDriver ? "onUpdate" : "target",
         targetCount: targets.length,
         props,
         start,
@@ -234,7 +248,15 @@ function describeTween(tw, props, bboxes, flags) {
   const dur = (tw.end - tw.start).toFixed(2);
   const parts = [];
 
-  parts.push(`${tw.selectorHint} animates ${props.join("+")} over ${dur}s (${tw.ease})`);
+  if (tw.selectorHint) {
+    parts.push(`${tw.selectorHint} animates ${props.join("+")} over ${dur}s (${tw.ease})`);
+  } else {
+    // An onUpdate driver: the span and props are known, the affected element is not.
+    parts.push(
+      `an onUpdate driver animates ${props.join("+")} over ${dur}s (${tw.ease}) — ` +
+        `motion is applied in JS, so no element geometry was measured`,
+    );
+  }
 
   // Movement
   const first = bboxes[0];
@@ -299,7 +321,10 @@ function computeFlags(tw, bboxes, { width, height }) {
   const flags = [];
   const dur = tw.end - tw.start;
 
-  if (bboxes.every((b) => b.w === 0 || b.h === 0)) flags.push("degenerate");
+  // No samples at all (an onUpdate driver has no element to measure) is not evidence of
+  // a degenerate or invisible box — `[].every()` is vacuously true, so guard the
+  // geometry-derived flags. The pacing flags below read only start/end and still apply.
+  if (bboxes.length && bboxes.every((b) => b.w === 0 || b.h === 0)) flags.push("degenerate");
 
   const anyOffscreen = bboxes.some(
     (b) =>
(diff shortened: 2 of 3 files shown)


somanshreddy (commented):
Reviewed (proxy-driver tweens) — **LGTM, correct + well-reasoned.** The GSAP proxy-driver idiom tweens a plain object and applies motion in `onUpdate`, so `targets()` is empty — the old `if (!targets.length) return` dropped those tweens, so `computeDensity` saw zero active tweens over a real span and `findDeadZones` reported a FALSE dead zone. The fix keeps them (`isProxyDriver = targets.length===0 && typeof vars.onUpdate==="function"`), and the sharp detail is `selectorHint: null` (not a placeholder string) for proxy drivers — because it feeds `document.querySelector` downstream, so it must be ABSENT rather than unmatchable; the bbox sampling correctly short-circuits to `[]` when there's no selector, and the summary text handles the no-element case. Precise guard, right downstream-safety call.

terencecho (approved):
**Approve** — correct fix, and the discriminator holds up against this repo's actual corpus.

Concur with @somanshreddy on the core reasoning (the `!targets.length` early-return dropping real motion, and `selectorHint: null` rather than a placeholder because it feeds `document.querySelector`) — not re-deriving it. Three things I checked that aren't covered above.

**1. The false-positive risk you traded against is empirically real, and the guard clears it.** The "bare spacer has no `onUpdate`" claim is the load-bearing half of this change, so I grepped every `to({}` in non-vendor source. All 13 `frame-presets/*/caption-skin.html` full-span anchors are `tl.to({}, { duration: DURATION }, 0)`, as are the two generator-emitted ones (`skills/pr-to-video/scripts/captions.mjs:481`, `skills/product-launch-video/scripts/transitions.mjs:290`) — no `onUpdate` on any of them, so none get newly counted and no genuine dead zone gets masked. The only in-repo `to({}, { ..., onUpdate })` hits are `packages/producer/tests/distributed/three-boundary{,-deferred}/src/index.html`, which are exactly the shape you're fixing. Clean separation.

**2. Follow-up (not blocking): the sibling shape — timeline-level `onUpdate` — is still dropped.** The discriminator is tween-local, so it catches the `rules/` idiom (`multi-phase-camera.md:61`, `chromatic-glitch`, `viewport-change`, `ai-tracking-box` … all put `onUpdate` on the tween). It misses the WebGL/uniform idiom where the *timeline* owns the driver:

```js
// skills/music-to-video/references/templates/held-message-living-field/index.html:392
var tl = gsap.timeline({ paused: true, onUpdate: renderFrame });
tl.to(uniforms.uTime,      { value: 12.0, duration: 12, ease: "none" }, 0);
tl.to(uniforms.uIntensity, { value: 0.55, duration: 3.0, ... },         0);
```

Each child targets a plain uniform object (`targets()` → no Element) and carries no tween-level `onUpdate`, so `isProxyDriver` is false and they still `return`. That composition remains one 12s false dead zone after this PR. `walk()` already threads `parentOffset` through the timeline branch, so threading a `parentDriven` boolean alongside it (`OR`ed into `isProxyDriver`) would close it at about the same cost — worth a follow-up rather than growing this PR.

**3. Nit on the "Not covered" note.** Driver tweens *do* land in `report.elements`: `buildElementLifecycles` keys on `tw.selector`, which is now the literal `"(onUpdate driver)"`, so every driver in a composition collapses into one pseudo-element whose `endsVisible` / `finalPosition` are `null` (`findLastBbox` finds no bboxes). Same for `detectStaggers`, which can group 3+ same-duration drivers into a "stagger" of `(onUpdate driver)` ×3. Collision detection genuinely does skip them (empty `bboxes`), so that half of the note is right. Cosmetic either way — I grepped for consumers of `animation-map.json` and it's read only by `SKILL.md` guidance and the script itself, so nothing parses those fields programmatically.

**CI:** the lanes that gate this diff are green — `Test: skills`, `Skills: manifest in sync` (so the `skills-manifest.json` hash bump is right), `Skills: project-native lint + mirror`, `Format`. The two red `Analyze (…)` checks are GitHub-side: the job annotations are `No server is currently available to service your request` from the CodeQL status-report upload, and across the stack the failing *language* is random (#3298 fails python, #3300 fails actions, this one fails actions + javascript-typescript) — a transient GHAS outage, not your diff. Rerunnable.

Test harness is the right one — `HELPERS[0]` is `animation-map.mjs` spawned as a real subprocess against the stubbed producer, so `enumerateTweens` / `computeDensity` / `findDeadZones` all execute for real, and holding all three tween shapes in one fixture is what makes the spacer-still-excluded assertion meaningful.

— Review by tai (pr-review)

miguel-heygen:
Took finding 2 into this PR rather than deferring it, and corrected the PR body per finding 3. Pushed in `952d87f`.

**Finding 2 — timeline-level driver, fixed here.** You're right that it's the same defect one level up, and that `walk()` already threading `parentOffset` made the cost about the same. Leaving `held-message-living-field` as a 12s false dead zone would have meant shipping a fix for the reported shape while a shipped template still reproduced the symptom. `walk()` now threads a `driven` flag beside the offset.

One thing I tightened beyond your sketch. A straight `OR` into `isProxyDriver` would re-open the false positive the tween-local rule was careful to close: a bare `tl.to({}, { duration: D })` under a driven timeline advances the playhead without changing any value, so the parent's `onUpdate` repaints an identical frame, and counting it masks a genuine dead zone. So inherited drivers are held to a stricter test than tween-local ones — the tween must also change a property:

```js
targets.length === 0 &&
  (typeof vars.onUpdate === "function" || (parentDriven && props.length > 0))
```

A tween's *own* `onUpdate` is proof of work by itself (a repaint loop needn't animate a property); a parent's isn't. The new test holds both shapes under one driven timeline and asserts the uniform tween counts while the spacer stays out.

**Finding 3 — my note was wrong, PR body corrected.** You're right: `buildElementLifecycles` keys on `tw.selector`, which is now the literal `"(onUpdate driver)"`, so drivers do land in `elements` — collapsed into one pseudo-element with null `endsVisible`/`finalPosition` — and `detectStaggers` can group same-duration drivers into a phantom stagger. Only the collision half of my claim held. Body now says so explicitly rather than quietly dropping it.

Left cosmetic, on your evidence that nothing parses those fields programmatically. Worth noting the pseudo-element grouping gets *more* visible now that inherited drivers are counted too, since a uniform-driven composition can contribute several at once.

**Finding 1** — glad the spacer separation held up against the real corpus. That grep over all 13 skin anchors plus the two generator-emitted ones is the check I'd wanted and hadn't run; it's what made me confident enough to extend the rule to inherited drivers at all, with the extra `props.length` condition.

miguel-heygen:
Took finding 3 properly rather than leaving it cosmetic, in `cd06e86`.

`buildElementLifecycles` and `detectStaggers` are both per-**element** analyses, and a driver tween has no element — so keying them on `tw.selector` was always going to fold every driver into one `"(onUpdate driver)"` pseudo-element with null geometry, and let three same-duration drivers read as a stagger nothing performs. Both now run over the element-backed tweens only.

Density, dead zones and the timeline still count drivers. Those are per-**span**, which is exactly what a driver does have — and counting the span was the whole point of the PR.

Your "cosmetic" read was fair on the evidence (nothing parses those fields), but it stops being cosmetic in the same breath as finding 2: a uniform-driven composition contributes several drivers at once, which is precisely the shape that invents a phantom stagger. Cheaper to close now than to explain later.

Asserted in both fixtures: the tween-local case keeps `elements` at exactly `["#mover"]`, and the timeline-driven one — which has no element-backed tween at all — keeps `elements` empty and `staggers` empty.

PR #3305, opened 2026-08-17
fix(studio): make the volume fader tell the truth about the gain it writes

miguel-heygen:
The volume fader travels in dB, so its stops are irrational values. Serializing them through the generic two-decimal numeric formatter collapsed the bottom quarter of its travel onto `"0"` — a hard mute — and made the knob jump on release everywhere below unity. Both property panels now use an exact serializer that round-trips every integer fader stop back to itself.

### The automation lane reaches the same ceiling

Clamping the volume lane at unity meant automating a boosted clip silently discarded the boost, and the panel disables the fader while a lane owns the level — so there was no way back short of deleting the lane.

**This rescales the lane's vertical axis**: unity now sits about a quarter of the way up rather than at the top, because the axis covers the same range the fader does. Worth knowing before you open a lane.

### New lint rule

`audio_volume_tween_overrides_gain`. Tween values on `volume` are absolute — they replace the authored gain rather than scaling it — so a clip carrying both plays at whatever the tween names, and the fader gives no sign of it. The rule reuses the tween detector the sibling lane/tween rule already has, and reports the authored gain in dB.

### Verifying

The fader's precision is pinned by a test that walks all 201 stops and asserts none of the audible ones serialize to silence and each round-trips to the same stop. `bun run lint` on a composition with `data-volume="1.949845"` and a `volume` tween emits exactly one warning naming the element and `5.8 dB`.

### Stack

Based on `u1-audio-gain-core`, which introduces the shared gain module. Retarget to `main` before merging.


jrusso1020 (changes requested):
Reviewed at `f32c97d4`. First review on this PR — nothing else was posted when I started, so this is a full pass rather than an additive one.

## Strengths

- **The shared ceiling is real, and it holds at every consumer.** The docstring added at `packages/core/src/audioAutomation.ts:137` claims a lane "can never offer a value the renderer would reject", so I checked the three that matter rather than taking it: `clampAudioGain` — not `Math.min(1, …)` — guards the lane in `packages/core/src/runtime/media.ts:277`, the preview transport in `packages/core/src/runtime/webAudioTransport.ts:465`, and the render mixer in `packages/engine/src/services/audioMixer.ts:70`. Raising `VOLUME_RANGE.max` to `MAX_AUDIO_GAIN` (`audioAutomation.ts:164`) is consistent with all three, so the claim is true as written.
- **Dropping `disabled={… || volume > 1}`** (`propertyPanelFlatMediaSection.tsx:230`) is the right call now that the control can represent those levels, and the hold comment that justified it is deleted along with it rather than left to rot into a lie.
- **`UNIT_RANGE` with its rationale** (`TimelineAutomationLane.test.tsx:130-135`) — spelling out that the geometry tests deliberately keep a 0..1 axis, and where the ceiling *is* covered, is what stops the next person "fixing" those coordinates.

## Blocker

**`packages/lint/src/rules/media.ts:643` — the new rule reads a missing `data-volume` as `0` and reports it as silence.**

`readAttr` returns `string | null` (`packages/lint/src/utils.ts:165-175`, `match?.[1] || null`), and `Number(null)` is `0`, not `NaN`. So an element with **no** `data-volume` passes both guards on the next line — `Number.isFinite(0)` is true and `0 !== 1` is true — and if a tween names it, the rule emits:

> `#bgm has data-volume="0" (silence) and a GSAP tween on volume …`

Both halves are false: there is no `data-volume` on the element, and the level being called "silence" is unity. Absent means unity everywhere else in the codebase — `propertyPanelFlatMediaSection.tsx:64` (`?? 1`) and `runtime/media.ts:259` (`clip.volume ?? 1`).

What makes it worth blocking is that this is the shape your own doc edit recommends. `skills/hyperframes-core/references/variables-and-media.md:101` says to animate `volume` on the timeline and that `data-volume` "is the static baseline for elements no tween touches" — so an element a tween *does* own is expected not to carry one. That is the common audio-fade authoring pattern, and the rule warns on all of it.

Scope, measured rather than assumed:

- Nothing in the repo trips it today. I checked every `*.html` carrying both a media tag and a `volume:` tween and found none, so no fixture goes red.
- Warnings do not fail the command (`ok: severity !== "error"`), so this is misdirection rather than a build break.
- The `fixHint` compounds it: it tells the author to "reset data-volume to 1" on an element that has no such attribute. An agent acting on lint output writes a gain that was never wrong.

None of the three new tests pin this — all of them set `data-volume` explicitly (`media.test.ts:441`, `:452`, `:459-467`), so deleting the `Number.isFinite` guard leaves them green. That claim is scoped to this rule's own tests deliberately; the rule is new, so there is nothing else in the repo that could pin it.

One line fixes it — `Number(readAttr(tag.raw, "data-volume") ?? "1")`, or reject `readAttr(...) === null` before the numeric guard. Worth a fourth case: no `data-volume` plus a `volume` tween, asserting silence.

Note `data-volume="0"` is a genuine hit and should keep warning; the defect is only that absent and authored-zero are indistinguishable here.

## Nit

**The "Verifying" section credits a test this PR does not contain.** The 201-stop precision test is in `packages/core/src/audioGain.test.ts:40-49`, which this PR does not touch — it is already on `main` from the gain module that landed ahead of it, and it pins `formatAudioGain` itself. What *this* diff changes is which serializer the panels call, and its pin is the single `"3.981072"` assertion at `propertyPanelFlatMediaSection.test.tsx:188`. That is adequate; the body just credits the wrong artifact, and a reader checking the claim against the diff will not find it.

## Notes (checked, no action needed)

- The body says to retarget to `main` before merging. Already done — base is `main`, one commit ahead, and the gain module it depends on is present on `main`, so nothing is pending.
- Widening the parse-time clamp does change playback for any stored lane authored above unity: such points used to load as `1` and now load at their authored value, up to `+12 dB`. The panel could not produce those before, so the population is hand- or agent-written HTML only. Deliberate, and the rescaled axis is called out in the body — noting it because "louder on load" is the kind of thing that gets reported as a regression later.

## Merge order — applies to this PR and the two beside it

This PR and **#3306** both rewrite the `hyperframes-audio` hash in `skills-manifest.json` (both start from `c96ccac4b1127b8c`), and this PR and **#3310** both rewrite `hyperframes-core` (both start from `ec542db377d8b213`). Same line, different values, so whichever lands second takes a real merge conflict there.

Resolve it by regenerating, not by picking a side. Both edits in each pair touch files inside the same skill bundle, so once both are on `main` the correct hash is a third value neither PR contains. And a hand-picked side fails quietly: the sync check runs on pull requests, so there is no run on `main`'s tip to go red, and the drift would surface only on whichever later PR next touches that skill.

**Verdict:** REQUEST CHANGES
**Reasoning:** The gain plumbing is right and verified at all three consumers, but the lint rule this PR adds fires on the authoring pattern the PR's own documentation recommends, and reports a false level while doing it. One line, and the rest is ready.

— Rames Jusso

terencecho (commented):
Reviewed at `f32c97d4`. Concurring with @jrusso1020's block — the rule at `packages/lint/src/rules/media.ts:643` genuinely fires on the shape the PR's own docs recommend.

## Verified independently

The failure chain: `readAttr` returns `null` when the attribute is absent (`packages/lint/src/utils.ts:174`, `match?.[1] || null`); `Number(null)` is `0`; both filters on `media.ts:646` pass (`Number.isFinite(0)` is true, `0 !== 1` is true). Nothing downstream re-checks presence before formatting the message, so an element like `<audio id="bgm" src="..." data-start="0" data-duration="10">` paired with `tl.to("#bgm", { volume: 0 })` prints `has data-volume="0" (silence)` — both halves wrong. Unity default is settled elsewhere: `propertyPanelFlatMediaSection.tsx:64` and `runtime/media.ts:259` both `?? 1`, so the rule stands alone in reading absent as silence.

Two consequences worth the block:

- **Scope**: the ducking pattern the doc edit at `variables-and-media.md:101` names (`tl.to("#bgm", { volume: 0, duration: 1 })` on a track authored at unity, i.e. no `data-volume`) is exactly the miss. This is the primary audio-fade authoring pattern the codebase teaches.
- **`fixHint` amplifies it**: "reset data-volume to 1" is dead-letter advice on a tag that carries no such attribute, and an agent following it writes a value that was never wrong.

The one-line fix `Number(readAttr(tag.raw, "data-volume") ?? "1")` also handles `data-volume=""` implicitly, since `readAttr` collapses empty via `|| null`. A fourth test case pinning "no `data-volume` + `volume` tween → no finding" would keep this from silently regressing.

## Concur — other items

- **Gain plumbing**: verified independently. `packages/core/src/audioGain.ts:52` docstring names the three consumers (`data-volume` seed, tween baseline, transport carry) and each uses `clampAudioGain` not `Math.min(1, …)`. `VOLUME_RANGE.max = MAX_AUDIO_GAIN` on `audioAutomation.ts:164` aligns the lane with the fader and the render mixer. Round-trip on the test's asserted `"3.981072"`: `audioFaderPositionToGain(100)` = `10^0.6` = `3.9810717…`, `formatAudioGain` clamps then `.toFixed(6)` then strips trailing zeros → `"3.981072"` — matches.
- **`disabled={... || volume > 1}` retiral** in both panels is now safe (the fader can represent boosted levels), and the accompanying hold-comment is removed rather than left to stale.
- **Merge order** with #3306 and #3310: both pairs collide on the same `skills-manifest.json` hash line. Second-lander gets a real merge conflict; hand-picking either side gives a hash that neither PR contains, and since `Skills: manifest in sync` only runs on PRs, the drift silently survives on `main` until whichever later PR next touches that skill. Regenerate on merge, don't pick.

## Additional note (not blocking)

`packages/studio/src/components/editor/propertyPanelMediaSection.tsx:266` now removes the `disabled` prop entirely. The non-flat panel never handled `volumeAutomated` (no automation-aware disable existed pre-PR either), so this is not a regression — surfacing it because the two panels' automation-awareness stays asymmetric after the change.

**Verdict:** concurring with CHANGES_REQUESTED. One-line lint fix and it lands.

— Review by tai (pr-review)

miguel-heygen:
Blocker fixed — you were right on both halves, and I verified each at source before touching it.

`readAttr` returns `null` when the attribute is absent (`utils.ts:173`), and `Number(null)` is `0` — finite, and not 1 — so a clip with no `data-volume` cleared both filters and got reported as authored at silence. Defaulted the read to `"1"`, since absent means unity everywhere else.

The part that makes this worse than a stray warning is the one you named: it fired on exactly the case the rule exists to bless. `variables-and-media.md:101` — my own doc edit in this PR — says `data-volume` is the baseline for elements no tween touches, so a tweened clip is *expected* not to carry one. That's the ordinary fade. And since a warning doesn't fail `check`, the failure mode isn't a red build, it's an agent reading the fixHint and writing a gain to correct a level that was never wrong.

Test added for the docs' own shape (a `fromTo` volume fade with no `data-volume`), mutation-checked: restoring the raw read reds it. Lint package: 524 passing.

**On the manifest collision** — confirmed, and your "regenerate, don't pick a side" is the right call. Concretely:

- `c96ccac4b1127b8c` (hyperframes-audio) → `1c01419737...` here, `6bdb36e157...` on #3306
- `ec542db377d8b213` (hyperframes-core) → `e57de95856...` here, `2471f4b504...` on #3310

Different values on the same line, so whoever lands second conflicts, and neither side's value is correct afterwards. Regeneration is `bun packages/cli/scripts/gen-skills-manifest.ts` (what the lefthook `skills-manifest` step runs). I'll run it on the second and third PRs after each merge rather than resolving by hand — and I'd flag that your point about it failing *quietly* is the real hazard: the sync check only runs on PRs, so a wrong hash sits on main until someone next touches that skill.

somanshreddy (commented):
**Independent pass — I confirm @Rames's blocker at source, and separately verified the headline fix is complete. Converging with CHANGES_REQUESTED.**

## The lint-rule blocker is real — confirmed at the execution layer
`findVolumeTweenOverridesGainFindings` computes `volume = Number(readAttr(tag.raw, "data-volume"))` and keeps entries where `Number.isFinite(volume) && volume !== 1`. `readAttr` (packages/lint/src/utils.ts:165-173) returns `string | null` — `match?.[1] || null` — so an **absent** `data-volume` yields `Number(null)`, which is `0` (verified: `Number(null) === 0`, `Number.isFinite(0) === true`, `0 !== 1`). So a clip with *no* `data-volume` (i.e. unity everywhere else in the system) is treated as `volume=0` and warned as `data-volume="0" (silence)` — both halves false. And it's precisely the pattern this PR's own doc edit recommends (`variables-and-media.md`: `data-volume` is the baseline "for elements no tween touches"), i.e. the common fade case. One line fixes it: default the read to `"1"` (`readAttr(...) ?? "1"`) — which correctly still warns on an *explicitly* authored `data-volume="0"`.

## Headline fix (the exact serializer) — independently verified complete
This is the core of the PR and worth confirming numerically. I replicated `audioFaderPositionToGain` / `audioGainToFaderPosition` / `formatAudioGain` (fader positions −100..+100, −60 dB floor, +12 dB ceiling, `.toFixed(6)`) and ran the full round-trip over **every integer fader stop**: all 201 stops recover exactly (`Math.round(gain→position) === position`), and **zero** nonzero stops serialize to `"0"` (the old 2-decimal mute-collapse). So the docstring's "six decimals round-trip every integer fader stop back to itself" holds across the entire range, and the bottom-quarter hard-mute is fully closed. The shared +12 dB ceiling (`clampAudioGain`, not `min(1,…)`) is consistent with the lane widening — agreeing with Rames's three-consumer check.

## Cross-cutting merge hazard — confirmed precisely (applies to all three PRs)
All three PRs rewrite `skills-manifest.json`. #3305 and #3306 both rewrite the `hyperframes-audio` hash from the same base `c96ccac4b1127b8c` to *different* values; #3305 and #3310 both rewrite `hyperframes-core` from the same base `ec542db377d8b213` to different values. Whichever of each pair merges second takes a real conflict, and picking a side is wrong (the correct post-merge hash covers both bundles' content changes) — regenerate. The sync check only runs on PRs, so a hand-picked side goes green on the PR but leaves `main` inconsistent until someone next touches the skill. Fully agree with Rames here.

Net: one-line blocker on the lint default; the rest reads clean and the headline fix is verified sound.

jrusso1020 (commented):
Re-review at `bc757a8d`. **The blocker is resolved.** My `CHANGES_REQUESTED` sits at `f32c97d4` and the code-level objection behind it is gone; the only thing holding a formal approval is a red required check that isn't your code. Detail below so you can judge it yourself.

## The fix is right, and the test genuinely pins it

`packages/lint/src/rules/media.ts:655` now reads `Number(readAttr(tag.raw, "data-volume") ?? "1")`, so an absent attribute is unity — matching `propertyPanelFlatMediaSection.tsx:64` and `runtime/media.ts:259`, which is what the rule was alone in disagreeing with.

I checked the new case actually discriminates rather than passing vacuously, since a `fromTo` the detector didn't recognise would look identical from the outside. It does discriminate: `tweensVolumeInSameCall` starts its scan at the `#bgm` hit *inside* the call and walks forward, and it reaches `volume:` before any depth-0 `)` or `;`, so the tween is seen. Under the old `Number(null)` path that input cleared `Number.isFinite(0)` and `0 !== 1`, the finding fired, and the new assertion would have failed. Good test.

Behaviour at the edges, all correct:

- `data-volume="0"` → still warns "silence". Right — that's *authored* silence under a tween, which is the true positive the rule exists for.
- `data-volume=""` → now also reads unity, because `readAttr` collapses empty to `null` via `match?.[1] || null`. @terencecho called this one before you pushed.
- Non-numeric → `NaN`, filtered by `Number.isFinite`. Unchanged.

## Manifest — you took the right path

`bc757a8d` regenerates rather than hand-picking, and only the two hashes that should move do: `hyperframes-core` `2471f4b5…` → `f6516b74…` (on top of #3310's landed value, not around it) and `hyperframes-audio` `c96ccac4…` → `1c014197…`. `Skills: manifest in sync` is green at this head. The remaining collision is with #3306 on `hyperframes-audio` only, and your stated plan there is the correct one.

## Why I'm not stamping yet — and it is not your diff *(disambiguation)*

The required **`Test`** context is red. It is **not** a test failure, and the aggregate's wording hides that:

- `Producer: unit tests` — **success**.
- `Producer: integration tests` — **cancelled**, not failed. It sat ~20 minutes in its first real step, `Install FFmpeg for integration tests` (`sudo apt-get update -qq && sudo apt-get install -y --no-install-recommends ffmpeg`), from 21:07:43 to 21:28:01, then `The operation was canceled.` Every later step is `skipped`.
- `Test` then reports "Producer unit/integration tests did not succeed," because it treats anything other than success as failure.

This is a live runner problem, not this PR. The same job, cancelled at the same step, on the same ~20-minute shape:

| head | `Producer: integration tests` |
|---|---|
| `main` @ `9da422fd` | cancelled — same step, 21:03:16 → 21:23:35 |
| #3306 @ `5cd1d5b8` | cancelled |
| #3310 @ `dfca6186` (merged 21:02Z) | success |

So `main` itself is red on this right now, and it started somewhere after #3310 went in. Nothing in your diff touches producer code or CI config.

**What I'd still like before approving:** a re-run of that job once the apt hang clears. Not because I think the red is real, but because the integration suite has produced *no* result at this head — approving would assert it passes when it hasn't run. Re-running needs no new commit, so it won't cost you the review state.

Ping me when `Test` is green and I'll convert this to an approval; there's nothing else outstanding from my side.

**Verdict:** COMMENT — blocker resolved and verified. Holding the approval solely on a required check that never executed, for infrastructure reasons unrelated to this change.

— Rames Jusso

jrusso1020:
`Test` is red at `a7740652` for a reason that isn't in this diff — one flaky assertion in a file this PR doesn't touch. Re-run the job; no code change needed here.

**What failed:** `packages/cli/src/utils/publishProject.test.ts:321`, in `createPublishArchive (U6 cloud-render regression guard)`. 1 failed / 2776 passed. Every other job in the run is green, including `Producer: integration tests` (12m41s — the ffmpeg install that hung earlier tonight is healthy again).

**Why it isn't yours:** this PR touches zero files under `packages/cli`, and both the test and the code under it were last modified on 2026-08-11 in `08934bfd`.

**Root cause.** The two sides of the assertion are the same expression, so it can only fail on non-determinism:

- `publishProject.ts:526` — `createPublishArchive(dir)` *is* `zipPublishFileMap(buildPublishFileMap(dir))`, which is what line 319 builds by hand. Same inputs, same code path.
- `publishProject.ts:510` — `zipPublishFileMap` calls `archive.addFile(filePath, content)` with no `attr` argument.
- adm-zip 0.6.0, `headers/entryHeader.js:37` — a new entry stamps `_time = Utils.fromDate2DOS(new Date())` at construction, and `:301`/`:331` write that value into both the local and central directory headers.

ZIP stores mtime in DOS format at two-second granularity, so the two archives are byte-identical only when both builds land in the same two-second bucket. Straddle a boundary and you get exactly `expected false to be true`. The window is narrow, which is why this has survived since July.

**Worth fixing separately, not here:** the same non-determinism means two publishes of byte-identical content produce different archive bytes, so nothing downstream can dedupe or cache a publish by content hash. Pinning the entry time — an `fs.Stats` as `addFile`'s fourth argument, or setting `header.time` after adding — makes both the test and published archives reproducible.

Nothing outstanding from me on the substance. The `data-volume` fix at `media.ts:655` and its test at `media.test.ts:457` are still correct at this head, and the regenerated `hyperframes-audio` hash (`94aba963`) is the right third value now that #3306 has landed — regenerating rather than hand-picking either side was the correct call. I'll convert my review once `Test` runs green.

— Rames Jusso

jrusso1020:
Post-merge note, for the record, since my `CHANGES_REQUESTED` on `f32c97d4` was still the last formal state I had on this PR: that blocker is resolved, and was resolved before this merged. I verified the fix twice — at `bc757a8d` and again at `a7740652` — `media.ts:655` reads `Number(readAttr(tag.raw, "data-volume") ?? "1")`, the new test at `media.test.ts:457` genuinely discriminates rather than passing vacuously, and the regenerated manifest hash was the correct third value.

The re-run also confirms the flake empirically: same commit `a7740652`, no code change, `Test` green. Identical bytes in, opposite result out — which is the cleanest possible demonstration that the failure was the archive timestamp and not this diff. Worth remembering when someone picks up pinning the entry time, since until then it will keep costing re-runs on unrelated PRs.

— Rames Jusso

PR #3310, opened 2026-08-17
feat(cli): run a managed background preview in every launch mode

miguel-heygen:
`--background` was rejected outside the embedded preview server. It now re-execs the CLI in foreground, which makes it mode-agnostic by construction: whichever server the child resolves to serves the config endpoint the readiness probe looks for.

`--foreground` is its counterpart, for a non-interactive shell that wants to stay attached. A bare launch keeps the same promise it always had — attached in an interactive terminal, managed in an agent session.

### An existing hole this exposed

Local-studio mode runs Vite with the studio package as its working directory and needs that package's own Vite config, which the published tarball does not carry. Resolving the package was treated as proof the mode was usable, so a project with an npm-installed studio took a path that can never come up — previously a clear error, now a ten-second silent timeout once background routes through it.

The predicate becomes "can this studio actually be served", so a published install falls back to embedded mode, which works. Publishing the studio's Vite config is a packaging change worth doing separately.

### Over the line budget

~1.2k changed lines against the convention's 1k. The overage is one command file and its tests carrying one invariant, and the seam that would split it further runs through a single request-handling function — a split there produces two PRs neither of which starts a preview on its own. Flagging rather than hiding it.

### Verifying

Start with `--background --json` and confirm `ready: true`, then `--status`, then `--stop`, and confirm the port is free and the session record removed. Conflicting lifecycle flags produce one failure envelope.

### Stack

Last of four on the preview side, based on `u5-preview-json-contract`. Retarget to `main` before merging.

### DIFF
--- a/docs/packages/cli.mdx
+++ b/docs/packages/cli.mdx
@@ -428,8 +428,6 @@ Needs a local Chrome, the same one `render` uses. Run
 within a frame or two — a different headless-Chrome audio sample rate can shift
 a beat slightly.
 
-## Look at it
-
 ### `preview`
 
 Start a live preview server with hot reload.
@@ -438,6 +436,8 @@ Start a live preview server with hot reload.
 npx hyperframes preview [dir]
 npx hyperframes preview --port 4567
 npx hyperframes preview --background     # keep running after the command exits
+npx hyperframes preview --foreground     # stay attached in a non-interactive shell
+npx hyperframes preview --status --json  # inspect a managed preview from an agent
 npx hyperframes preview --list           # every running preview
 ```
 
@@ -446,6 +446,8 @@ npx hyperframes preview --list           # every running preview
 | `--port`                             | Server port (default 3002)                                                                                |
 | `--open` / `--no-open`               | Open a browser, or leave it closed                                                                        |
 | `--background`                       | Keep an embedded preview running after the command exits                                                  |
+| `--foreground`                       | Keep the preview attached even when the shell is non-interactive                                          |
+| `--json`                             | Emit one versioned JSON result for managed start, status, stop, list, and kill-all operations              |
 | `--browser-gpu` / `--no-browser-gpu` | Hardware GPU for Studio thumbnails and frame capture, or deterministic SwiftShader (default: auto-detect)  |
 | `--proxy` / `--no-proxy`             | Auto-transcode browser-hostile codecs (HEVC, ProRes, AV1) to a cached authoring proxy (default: on)        |
 | `--browser-path`                     | Open a specific browser. `--user-data-dir`, `--remote-debugging-port`, and `--browser-no-gpu` require it.  |
@@ -454,6 +456,13 @@ To manage running servers: `--status` and `--stop` act on this project's
 background preview, `--list` and `--kill-all` act on all of them, and
 `--force-new` starts a second server for a project that already has one. Each
 exits straight after.
+
+Bare `preview` chooses the safest lifecycle for its caller: it stays in the
+foreground in a human interactive terminal, while a non-interactive or agent
+shell starts a managed background preview. Re-running the command for the same
+project reuses the healthy preview. Every start or status result includes the
+exact Studio project URL as well as the underlying server URL, so agents can
+hand off the intended project without guessing from the port.
 
 To read a running Studio from a script: `--selection` prints the selected
 element and `--context` prints the agent-readable context, both with `--json`.
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -33,10 +33,19 @@ Start the live preview studio in your browser:
 
 ```bash
 npx hyperframes preview
-# Studio running at http://localhost:3002
+# Studio: http://localhost:3002/#project/my-video
+# Server: http://localhost:3002
 
 npx hyperframes preview --port 4567
 ```
+
+In an interactive terminal, the preview stays attached until you press
+Ctrl+C. In a non-interactive shell such as a coding-agent session, the same
+command starts a managed preview that survives after the command returns. Use
+`--background` or `--foreground` to choose explicitly, and manage persistent
+previews with `--status`, `--stop`, `--list`, and `--kill-all`. Add `--json` to
+managed lifecycle commands for machine-readable output. `--foreground --json`
+prints the ready-session envelope once, then remains attached until stopped.
 
 ### `render`
 
--- a/packages/cli/src/commands/coreSkillContent.test.ts
+++ b/packages/cli/src/commands/coreSkillContent.test.ts
@@ -125,4 +125,15 @@ describe("media treatment routing documentation", () => {
       expect(template).toContain("do not improvise equivalent CSS/SVG filters or overlays");
     }
   });
+
+  it("gives agents a process-owned preview lifecycle in new project instructions", () => {
+    for (const file of ["AGENTS.md", "CLAUDE.md"]) {
+      const template = read("packages", "cli", "src", "templates", "_shared", file);
+      expect(template).toContain("npx hyperframes preview --background");
+      expect(template).toContain("npx hyperframes preview --status");
+      expect(template).toContain("npx hyperframes preview --stop");
+      expect(template).toContain("leaving refreshes at `ERR_CONNECTION_TIMED_OUT`");
+      expect(template).not.toContain("run_in_background: true");
+    }
+  });
 });
--- a/packages/cli/src/commands/preview.test.ts
+++ b/packages/cli/src/commands/preview.test.ts
@@ -1,20 +1,33 @@
 import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
-import { join, resolve } from "node:path";
+import { join } from "node:path";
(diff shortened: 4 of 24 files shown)


jrusso1020 (approved):
Reviewed at `dfca6186`, full pass — nothing else was posted when I started.

**Audited:** `previewLifecycle.ts` (end to end), `utils/studioProxyEnv.ts`, `studio/vite.preview-config.ts`, the `studio/vite.config.ts` diff, `server/portUtils.ts` detection path, and in `preview.ts` the signal/exit handling and the mode-selection predicate.
**Trusting:** the remainder of `preview.ts`'s ~498 added lines (request handling and output formatting) — read for structure, not line by line. Calling that out rather than implying a whole-file audit.

## Strengths

- **The PID-reuse guard reasons about the failure mode most code gets backwards.** `wrapperProcessIsAlive` (`previewLifecycle.ts:293-317`) separates "no answer" from "a different answer": a null birth token keeps the record, because `processIdentity` collapses a subprocess *timeout* into `null` and that timeout is likeliest under exactly the load that made the HTTP probe miss. Treating it as a recycled PID would destroy the only ownership proof `--stop` has. The cheap `kill(pid, 0)` check runs first so the identity subprocess stays off the path for stale records, and `EPERM` is correctly read as alive.
- **The session record is written atomically, for a stated reason** (`:78-95`): every reader deletes a record it cannot parse, so a half-written file would destroy a live server's ownership proof. The `finally { rmSync(temporary) }` also closes the leak where a failed rename orphans a temp file the `.json` filter would hide forever.
- **Readiness cannot adopt a bystander.** `preLaunchPorts` is snapshotted *after* the owned server is stopped (`:433-436`), and `startedServer` only accepts a port absent from that set — so a pre-existing unmanaged sibling on the same project can never be mistaken for the child this call launched. The comment at `:409-412` explains why ownership is keyed on saved project+port rather than on the replacement's GPU policy, which is the subtle half.
- **The mode predicate now tests the thing it needs** (`preview.ts:862-863`): resolving `@hyperframes/studio/package.json` proved only that the package exists, so it now additionally requires `vite.config.ts` to be present, and a published tarball falls back to embedded mode. That matches the body's description exactly, and it converts what the body notes would have become a ten-second silent timeout into a working path.
- **The over-budget note is the right way to handle a 1.2k-line PR** — naming the convention, the overage and why the obvious split produces two PRs that neither start a preview, instead of quietly hoping nobody counts.

## Notes (checked, no action needed)

- **The new config endpoint is genuinely wired.** `previewConfigPayload` returns `null` unless `HYPERFRAMES_PREVIEW_PROJECT_DIR` and `_NAME` are set, and both `studioProxyEnv` call sites pass them (`preview.ts:1450`, `:1512`). Worth stating because measuring this from the wrong ref makes it look like dead code — the parameter does not exist on `main`.
- `serverBuildSignature: null` from the Vite path is safe: the comparison in `portUtils.ts:160-165` is gated on `expectedServerBuildSignature !== null`, so a null value can only decline reuse, never mismatch a server that should have matched.
- `studioDeepLink` now percent-encodes the project name (`preview.ts:614`), which fixes deep links for names with URL-significant characters. Not in the body; harmless, but it is a behaviour change riding along.
- `buildBackgroundPreviewArgs` (`previewLifecycle.ts:236-248`) filters `--open`/`--no-open`/`--json` by exact match while handling `--background=`/`--foreground=` by prefix. A `--json=true` or `--open=true` spelling would survive into the child. citty accepts the bare forms these commands document, so this is theoretical rather than reachable — flagging only so the asymmetry is a decision rather than an oversight.
- The body says to retarget to `main`. Already done — base is `main`, one commit ahead, single commit, so the preview PRs ahead of it have landed.

## Merge order — worth coordinating with #3305

This PR and **#3305** both rewrite the `hyperframes-core` entry in `skills-manifest.json`, both starting from `ec542db377d8b213`. Same line, different values, so whichever merges second takes a real conflict.

Regenerate rather than resolving by hand: your `SKILL.md` + `references/review-loop.md` edits and #3305's `references/data-attributes.md` + `variables-and-media.md` edits all live inside that one bundle, so the correct post-merge hash is a third value neither PR contains. Picking a side leaves `main` stale with nothing to catch it — the sync check runs on pull requests, so there is no run on the tip to go red.

**Verdict:** APPROVE
**Reasoning:** The lifecycle module is the risky surface and it is careful in the places that matter — PID reuse, atomic ownership records, and readiness that cannot adopt an unmanaged sibling. No blocker found in what I audited, and I have named what I did not.

— Rames Jusso

terencecho (approved):
Concurring with James's approval at `dfca6186`. Spot-checked the tai-relevant seams from prior stack review:

**Ownership-record integration preserved.** `previewLifecycle.ts` only gains three lines in `buildBackgroundPreviewArgs` (strip `--foreground`/`--json`, append `--foreground --no-open`) — the #3308 wrapperIdentity + `isProcessDescendant` + `ownedStopTargetPid` invariant path is untouched. Every kill still routes through `stopBackgroundPreview` → OS-verified `server.pid` gated by descendant check; no new raw signal to a self-reported pid.

**No new Windows locale surface.** No new `netstat`/`tasklist`/`wmic` parsing added by this PR. `windowsListenerPid`'s existing `LISTENING`-string match is inherited from #3307, not reintroduced. `--host 127.0.0.1` on `previewViteArgs` also keeps the port announcement to a form the ready regex (`localhost|127.0.0.1|[::1]`) matches.

**Cross-mode config endpoint symmetry is honest.** `packages/studio/vite.preview-config.ts` fills the same `/__hyperframes_config` shape the embedded Hono server serves (`studioServer.ts:694`), keyed on `HYPERFRAMES_PREVIEW_PROJECT_DIR`/`_NAME` — absent env → middleware falls through to `next()`, so a direct-Vite invocation from a studio-package consumer stays unaffected. `serverBuildSignature: null` from Vite would signal `mismatch` to `detectHyperframesServer`, but that probe is only invoked from `runEmbeddedMode`'s `findPortAndServe`; mode branching in the parent + per-project reuse guard in `startBackgroundPreview.savedOwnedPreview` mean an embedded probe never crosses onto a dev/local port for the same project.

**`hasLocalStudio` fix is the right shape.** Requiring `vite.config.ts` inside the resolved studio package captures exactly the "published tarball can't be served" failure the PR body describes, and the fallback to embedded is what a published install actually supports. Nice one.

**PR body nit (not blocking).** "Retarget to `main` before merging" is stale — `baseRefName` is already `main` and merge-base is `#3309`. Worth pruning next time.

CI: full matrix green including `Tests on windows-latest`, `Render on windows-latest`, `Preview parity`, `preview-regression`, and the three `npx shim` OSes.

somanshreddy (commented):
**Independent pass — traced the risky surface (`previewLifecycle.ts`) end to end. Reads clean; a few confirm-items, no blockers. Scope note below.**

## What I verified
- **No infinite re-exec loop.** `buildBackgroundPreviewArgs` strips `--background`/`--background=…`, `--foreground`/`=…`, `--open`, `--no-open`, `--json`, then appends `--foreground --no-open`. The child therefore carries `--foreground`, and `previewLaunchMode` only returns `"background"` when `options.background` is set (or non-interactive AND not `--foreground`) — the child can't re-enter the background branch. Pinned by the arg-forwarding test.
- **Argv forwarded as a tokenized array** into `spawn(execPath, args, {detached, stdio:[ignore,logFd,logFd], env})` with no shell, so project-dir paths/spaces survive intact; `child.unref()` + PID check.
- **Readiness is scoped correctly.** `preLaunchPorts` is snapshotted *after* the owned server is stopped, so readiness only accepts a *newly appeared* same-project server answering `/__hyperframes_config` — it can't adopt a pre-existing unmanaged sibling (the exact distinction Rames called out). ~10s budget; on timeout the detached child is killed before throwing (no orphan).
- **The "resolve ≠ usable" fix is correct.** `hasLocalStudio` now gates on `existsSync(join(studioPkgPath, "vite.config.ts"))` rather than mere resolvability, so a published-tarball install falls back to `embedded` (which works) instead of routing `--background` into a Vite child that never comes up and burns the 10s silently. Confirmed both server paths answer the probe (embedded serves `/__hyperframes_config` unconditionally; the dev/local Vite child via middleware gated on the `HYPERFRAMES_PREVIEW_*` env that `studioProxyEnv` sets).

## Confirm-items (non-blocking)
1. The fix hinges on the published tarball **never** shipping `vite.config.ts` and there being no `.mts`/`.mjs`/`.js` config variant (the check is hardcoded to `vite.config.ts`) — worth an `npm pack --dry-run` sanity check on `@hyperframes/studio`.
2. **~10s readiness ceiling** on a cold Vite/embedded start on a slow CI box — Vite cold starts can approach it.
3. **Concurrent same-project bare `preview`** has a small TOCTOU: two invocations could both pass the reuse check and both spawn on different ports, orphaning one ownership record. Almost certainly out of scope for serialized agent sessions — just flagging.

## Scope note (same discipline Rames used)
I traced `previewLifecycle.ts` (the risk surface) and the mode predicate thoroughly via the gathered diff/files; I did **not** read all ~498 added lines of `preview.ts` line-by-line, so this is a targeted review of the lifecycle/re-exec/readiness path, not a full audit of that file.

PR #3316, opened 2026-08-17
fix(add): make chosen variables actually take effect, in the CLI and the preview

miguel-heygen:
## What

Reported as one command that did nothing useful:

```
$ npx hyperframes add blur-in --vars '{"size":76,"tone":"accent","align":"center"}'
Install failed: fetch failed
```

Five bugs were behind it or found while proving the fix:

| # | Bug | Blast radius |
|---|---|---|
| 1 | `--vars` silently discarded for components | 221 of 375 catalog items |
| 2 | Catalog preview ignored every chosen variable | every catalog page |
| 3 | `fetch failed` was undiagnosable, real cause thrown away | anyone whose install fails |
| 4 | Item files had no retry, one blip killed any install | everyone |
| 5 | Compound words split the answer in two | any query with one |

Plus the two `transcribe` tests that were red before this branch.

---

## 1. `--vars` for components

**How it broke.** `buildSnippet` had two branches. For a block it emitted a mount element and hung the values on it. For a component it returned `<!-- paste from ... -->` and dropped `values` on the floor. The flag parsed, validated, and went nowhere.

**Why the obvious fix is wrong.** You cannot put the values on a mount element, because a component has no mount element. It is markup you paste into a host composition.

**How it resolves values instead.** Through `__hyperframes.getVariables()`, which merges the declared defaults of *every* `[data-composition-variables]` element in the document with render-time overrides. A component carries its own declaration, and that declaration travels with the markup when you paste it.

**So the fix** is `registry/variableDefaults.ts`, which rewrites the `default` field of the component's own declaration at install time. That is the only place a chosen value can live and still be there after the paste.

Four things it refuses to do:

- **An unlisted enum value or an out-of-range number is reported, not written.** Writing one produces a file that renders *exactly* as if the value were ignored, because the composition's own guard falls back to the default. That is the failure this whole change exists to remove, so reintroducing it silently would be perverse.
- **Numeric strings are coerced.** The catalog page puts values in a query string, where every value is a string. Writing `"76"` where the composition expects a number trips that same silent fallback.
- **Ids the item never declared are reported**, not dropped.
- **Only the requested item is rewritten.** A dependency dragged in behind it never declared these variables.

**Blocks deliberately keep the mount attribute.** Per-mount values are strictly better where a mount exists: the file on disk stays byte-identical to the registry's, so a reinstall can still tell an edit from an update, and two mounts of the same block can differ.

`add` now prints `variables applied: size, tone, align`, because a component's values live in the file rather than the snippet. Without it, the command looks identical whether `--vars` worked or was thrown away.

---

## 2. The preview

`catalog/components/badge-pop?vars-badge-pop={"count":"10","accent":"green"}` rendered **3**, in **red**.

**How it broke.** The player injects the runtime by appending a `<script src>` to an already-loaded document, and only when `shouldInjectRuntime` says to: a nested composition, or five polls with a timeline present. A pasted component has neither. Meanwhile the component reads its values in an inline IIFE that runs *while the body is parsing*:

```js
var vars = window.__hyperframes && window.__hyperframes.getVariables
  ? window.__hyperframes.getVariables() : {};
```

That guard always took the empty branch, so the component used the defaults hardcoded in its own script.

**The values were never the problem.** The preview sets `window.__hfVariables` correctly. Nothing was there to read it.

**The fix** is ordering, not plumbing. `prepareSrcdocForElement` now puts the same runtime URL in the document's `<head>` before the srcdoc is set. A classic external script in head is parser-blocking, so it runs before body scripts. Nothing new is loaded and no dependency is added that the player did not already have. A CLI render never had this bug because the engine already orders it this way.

Skipped when the page carries the runtime already, so a CLI-rendered page (which inlines it) does not get a second copy re-initialising the runtime under a live composition. The probe's late injection stays for the `src=` path, where there is no srcdoc to prepare.

### Seen on the docs site

The same page, the same panel state, the docs running locally against each build of
the player. The field asks for `TXT_MAIN_1_Line = "VARS REACH THE PREVIEW"` in both,
and the install command under the panel is the CLI half of this PR carrying the same
value.

**Before**, on the published player: the preview renders the declared default,
`BILD EXKLUSIV`. The reader's choice is visible in the field and nowhere else.

Catalog page on the published player: the panel field reads VARS REACH THE PREVIEW while the preview still renders BILD EXKLUSIV

**After**, on this branch's player: the preview renders what the panel asked for.

The same catalog page on this branch's player: the preview renders VARS REACH THE PREVIEW

---

## 3 and 4. The error message, and the retry

`fetch failed` is two words that describe every network problem equally badly. Three things were missing, each of which was the whole answer in a different case.

**The URL.** undici throws with none attached. A project pointing `registry` at a private host in `hyperframes.json` got a message that read as the public registry failing.

**The cause.** undici buries the real reason one or two levels down in `cause`, and it was being dropped. In the reported case that turned out to be **`self-signed certificate in certificate chain`** — a private registry whose certificate node refuses and `curl` accepts, which is exactly why the host looked healthy when checked from a terminal. `describeCauseChain` now flattens the chain, guarding against cycles.

**A retry.** Item files are the one uncached path: manifests fall back to a stale copy, but every install downloads its files fresh, so a single blip killed the whole command. Two extra attempts with short backoff, and deliberately **not** for TLS failures — a self-signed certificate fails identically every time, so retrying only makes the user wait three times as long for the same message.

Both failure paths name the host now. Fixing only the item-file path would have left the manifest path (`Item "x" not found — registry unreachable or empty`) with the identical dead end.

---

## 5. Compound words

Found after this document declared zero ranking failures left.

`countdown` returned **exactly one** item, the only thing tagged with that spelling. `count down timer` returned **sixteen**, and that one was in none of them.

**How it broke.** The tokenizer splits on word boundaries, so `countdown` is one token and `count down` is two, and neither can ever match the other. The two spellings of a single idea produced disjoint sets, and whichever phrasing an author happened to type decided which half of the answer they saw. Neither half was the whole answer: the one-word spelling hid `count-up` and `decline-chart`, the two things you would actually build with.

**The fix** goes both directions, each gated on the catalog's own vocabulary so it can only add signal:

- A query token is **split** when both halves are words the catalog uses (`countdown` → `count` + `down`).
- Adjacent tokens are **joined** when the compound is (`count down` → `countdown`).
- A word in neither form, like `timer` which appears in **zero** of the 375 items, is left alone. This widens phrasing; it does not invent matches.

**The part worth reviewing closely:** everything inferred this way carries a fraction (0.35) of a real token's weight. My first version relied on the halves being statistically common in a 375-item catalog, which is *not* the same as making them count for less. In a small corpus that version let `type` matching the name of `type-match-cut` outrank `typewriter` matching the name of `typewriter` — searching a word returned something that merely contained half of it. Two tests written against that failure caught it before it shipped.

Result, all spellings returning the same 17 items:

| query | yt-circle-pointer | count-up | decline-chart |
|---|---|---|---|
| `countdown` | **#1** | #4 | #7 |
| `count down timer` | #7 | **#3** | #6 |
| `count down` | #7 | **#3** | #6 |

Each spelling still ranks its own exact match first, which is the correct behaviour — nothing is hidden either way now.

---

## Test plan

**cli 2680 passed · player 338 passed · studio 4249 passed. Zero failures.**

The two `transcribe` tests that were red before this branch are fixed: they assert the whisper soft-skip path but never pinned the engine, and `auto` picks Parakeet whenever `parakeet-mlx` is installed, so on those machines they shelled out to a real ASR binary and failed. Pinned, plus an assertion that the mock actually ran — the file now takes 18ms instead of 3.7s.

**A 39-query eval set holds at 33/39 top-1 and 39/39 top-3 through every ranking change here**, so no query regressed while fixing compounds.

Verified with the **built `dist/cli.js`**, not the source, in the reporter's own project directory:

```
✓ Added blur-in (hyperframes:component)
  variables applied: size, tone, align
```

Bad values are refused rather than silently written:
```
Warning: --vars tone ignored: not one of strong, muted, accent
Warning: --vars size ignored: above max 120
Warning: --vars ignored (not declared by blur-in): nope
```

A registry with a bad certificate explains itself:
```
Item "blur-in" not found — registry unreachable or empty. Contacted
https://self-signed.badssl.com/registry, set by this project's hyperframes.json,
not the public registry.
```

The preview fix, end to end against the real runtime and a real registry component, asking for size 96 / accent / right:

| | font-size | colour | align | runtime |
|---|---|---|---|---|
| before | 52px | `rgb(243,243,243)` | flex-start | absent |
| after | **96px** | **`rgb(60,230,172)`** | **flex-end** | present |

`rgb(60,230,172)` is `#3ce6ac`, the accent green.

Three player tests asserted byte-identical srcdoc forwarding; they now assert what they were actually protecting (the composition arrives intact) plus the new runtime guarantee. The `variableDefaults` fixtures gained the `label` the schema requires — without it they were not valid declarations, which the stricter reader caught.

## A note on types

`variableDefaults` originally modelled a declaration as a local interface of six `unknown` fields, re-checked at every use. Core already owns this shape as a discriminated union and exports `isCompositionVariable`, the same predicate `parseCompositionVariables` filters with, so the union is used directly and the duplicate type is gone. A declaration the schema rejects now leaves the file untouched rather than being partially rewritten from guesses.

## Size

**Over the 1000-line convention, deliberately and with sign-off.** The natural seam is the player fix (`runtime-in-srcdoc.ts`, `runtime-url.ts`, `shader-options.ts` and their tests, ~200 lines). It is kept here because it is the same user-visible defect — a value you choose does not take effect — and splitting it would ship half an answer to the report.

## Not covered

- **Registry item files are still uncached.** The retry survives a blip; a genuinely offline machine still cannot install an item it has never fetched. Caching them is a separate design question (staleness, disk, invalidation).
- **`top-down-letters` still ranks #2 for `count down`** on a name match against `down` alone. Legitimate token, field weighting doing its job; worth revisiting only if name matches on single common words prove noisy more broadly.
- **There is no discrete 3-2-1 countdown primitive.** `count-up` runs backwards (its driver is direction-agnostic) but eases continuously through the range instead of holding a beat per number. That is a catalog gap, not a ranking one.
- **Reaching the docs site needs a player release.** The docs load the player from a CDN on the `0.7` range, which today resolves to a build without this fix, so the screenshots above were taken with the docs pointed at this branch's build. The range picks the fix up on the next publish; no docs change is needed.

- **Component pages still cannot answer their panel, for a different reason.** Their preview payload is built from `demo.html`, which is a hand-copy of the snippet rather than the snippet itself, and 166 of the 168 components declaring variables have a copy that dropped the `data-composition-variables` block. There is nothing in that markup to read a value from, so the panel cannot move it whatever the player does. I built the one-place fix (mount the snippet the page hands the reader, instead of the copy) and measured what it costs: all 168 payloads then carry their declaration, and every value takes effect, but the preview stops moving, because a component's snippet ships its timeline as a commented recipe while the demo supplies a real one. Five of five sampled items animated before and were static after. A worse preview is not a fix, so it is not in this PR. Making demos stop hand-copying their snippet is the real change, and it needs a decision per demo about where its timeline lives.

- **Reproducing the preview locally needs the payload served yourself.** The local docs server publishes `.webp` out of the public directory but not `.json`, so the preview panel reports the payload as unavailable on localhost regardless of any of this.

### DIFF
--- a/packages/cli/src/commands/add.test.ts
+++ b/packages/cli/src/commands/add.test.ts
@@ -482,10 +482,26 @@ describe("describeInstallFailure", () => {
     expect(message).toContain("HTTPS_PROXY");
   });
 
-  it("surfaces the underlying cause when node attached one", () => {
-    const err = new Error("fetch failed", { cause: new Error("getaddrinfo ENOTFOUND") });
+  it("names the project's own registry when it is not the public one", () => {
+    // The reported failure: hyperframes.json pointed at a private host with a
+    // self-signed certificate. Telling that reader to check their connection
+    // sends them to debug the one thing that was working.
+    const message = describeInstallFailure(
+      new Error("fetch failed"),
+      "https://private.example/registry",
+    );
 
-    expect(describeInstallFailure(err)).toContain("getaddrinfo ENOTFOUND");
+    expect(message).toContain("https://private.example/registry");
+    expect(message).toContain("not the public registry");
+  });
+
+  it("stays quiet about the registry when it is the default one", () => {
+    const message = describeInstallFailure(
+      new Error("fetch failed"),
+      "https://raw.githubusercontent.com/heygen-com/hyperframes/main/registry",
+    );
+
+    expect(message).not.toContain("not the public registry");
   });
 
   it("leaves a non-transport failure exactly as it was", () => {
--- a/packages/cli/src/commands/add.test.ts
+++ b/packages/cli/src/commands/add.test.ts
@@ -3,7 +3,14 @@ import { existsSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "no
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import type { RegistryItem, RegistryManifest } from "@hyperframes/core";
-import { AddError, buildSnippet, parseVariableValues, remapTarget, runAdd } from "./add.js";
+import {
+  AddError,
+  buildSnippet,
+  describeInstallFailure,
+  parseVariableValues,
+  remapTarget,
+  runAdd,
+} from "./add.js";
 import { trackRegistryItemAdded } from "../telemetry/events.js";
 
 // Assert the emitted payload rather than the transport: `shouldTrack()` is
@@ -460,5 +467,32 @@ describe("variable values in the snippet", () => {
     expect(() => parseVariableValues("not json")).toThrow(/JSON object/);
     expect(() => parseVariableValues("[1,2]")).toThrow(/JSON object/);
     expect(parseVariableValues(undefined)).toBeNull();
+  });
+});
+
+describe("describeInstallFailure", () => {
+  it("explains a bare transport failure instead of echoing it", () => {
+    // What the user actually sees after copying a command off the catalog page.
+    // Item FILES are not cached, so a blip surfaces as node's `fetch failed`
+    // with no URL and no cause, and reads like the command was wrong.
+    const message = describeInstallFailure(new Error("fetch failed"));
+
+    expect(message).toContain("could not download the item's files");
+    expect(message).toContain("rather than a bad command");
+    expect(message).toContain("HTTPS_PROXY");
+  });
+
+  it("surfaces the underlying cause when node attached one", () => {
+    const err = new Error("fetch failed", { cause: new Error("getaddrinfo ENOTFOUND") });
+
+    expect(describeInstallFailure(err)).toContain("getaddrinfo ENOTFOUND");
+  });
+
+  it("leaves a non-transport failure exactly as it was", () => {
+    // An unsafe target or a malformed item is the caller's problem to read; a
+    // connectivity lecture there would send them to fix the wrong thing.
+    const message = describeInstallFailure(new Error('Unsafe target "../x"'));
+
+    expect(message).toBe('Install failed: Unsafe target "../x"');
   });
 });
--- a/packages/cli/src/commands/add.ts
+++ b/packages/cli/src/commands/add.ts
@@ -15,7 +15,7 @@ import { existsSync } from "node:fs";
 import { resolve, relative } from "node:path";
 import { ITEM_TYPE_DIRS, type RegistryItem } from "@hyperframes/core";
 import { c } from "../ui/colors.js";
-import { installItem, resolveItemsByTag } from "../registry/index.js";
+import { DEFAULT_REGISTRY_URL, installItem, resolveItemsByTag } from "../registry/index.js";
 import { resolveItemWithDependencies } from "../registry/resolver.js";
 import {
   gateRegistryItemsCompatibility,
@@ -216,7 +216,7 @@ async function installAll(
       }
     }
   } catch (err) {
-    throw new AddError(describeInstallFailure(err), "install-failed");
+    throw new AddError(describeInstallFailure(err, baseUrl), "install-failed");
   }
   return { written, preserved, variablesApplied, variablesUnknown, variablesInvalid };
 }
@@ -229,17 +229,28 @@ async function installAll(
  * That is what a user sees after copying a command off the catalog page, and
  * it reads like the command was wrong rather than the network.
  */
-export function describeInstallFailure(err: unknown): string {
+export function describeInstallFailure(err: unknown, registry?: string): string {
   const message = err instanceof Error ? err.message : String(err);
   const cause = err instanceof Error && err.cause instanceof Error ? err.cause.message : "";
   const transport =
     /fetch failed|ENOTFOUND|ECONNRESET|ECONNREFUSED|ETIMEDOUT|EAI_AGAIN|socket hang up|aborted/i;
   if (!transport.test(`${message} ${cause}`)) return `Install failed: ${message}`;
+
+  // Name the registry first. A project that set `registry` in hyperframes.json
+  // points at a private host, and when that host is down the failure has
+  // nothing to do with the user's connection -- telling them to check their
+  // network sends them to debug the one thing that is working.
+  const custom =
+    registry && !registry.startsWith(DEFAULT_REGISTRY_URL)
+      ? `\n  This project's hyperframes.json sets registry to ${registry}, so that is the host ` +
+        "being contacted, not the public registry. If it is down or private, that is the failure."
+      : "";
   return (
-    `Install failed: could not download the item's files${cause ? ` (${cause})` : ""}. ` +
-    "The registry is served over the network and item files are not cached, so this is " +
-    "usually connectivity, a proxy, or an offline machine rather than a bad command. " +
-    "Check your connection and retry; if you are behind a proxy, set HTTPS_PROXY."
+    `Install failed: could not download the item's files.\n  ${message}` +
+    "\n  Item files are not cached, so every install fetches them. This is usually the " +
+    "registry host or the network rather than a bad command." +
+    custom +
+    "\n  Retry, or set HTTPS_PROXY if you are behind a proxy."
   );
 }
 
--- a/packages/cli/src/commands/add.ts
+++ b/packages/cli/src/commands/add.ts
@@ -140,6 +140,8 @@ export interface RunAddResult {
   installed: string[];
   snippet: string;
   clipboardCopied: boolean;
+  /** Variable ids whose default was baked into an installed component. */
+  variablesApplied: string[];
   warnings: string[];
 }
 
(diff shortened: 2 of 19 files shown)


terencecho (commented):
COMMENT — read through end to end, no blockers found. Solid bugfix stack; the ordering fix (runtime injected into srcdoc `<head>` so the component's parse-time `getVariables()` guard doesn't take the empty branch) is the key insight and it's tested where it matters.

**Verified**
- Bug 1 (`--vars` for components): `applyVariableDefaults` rewrites the declared `default` in the component's own `data-composition-variables` at install time — the only surviving home once markup is pasted. Only the requested item gets its vars rewritten (dependencies pinned to `null`, `add.ts` L208). Enum/number values are rejected with a reason rather than written to a file that would silently fall back at runtime — that's the right call.
- Bug 2 (preview): `ensureRuntimeBeforeBodyScripts` lands the runtime `<script src>` inside `<head>` (parser-blocking) so it's defined before any body IIFE runs. Wired through `prepareSrcdocForElement` in `shader-options.ts` — every srcdoc path picks it up. `alreadyHasRuntime` guards against a double copy on CLI-rendered pages (which inline it). `runtime-url.ts` extraction means probe-late-injection and srcdoc-parse-injection can't drift onto different URLs — good consolidation.
- Bug 3 (`fetch failed` diagnosis): `describeCauseChain` walks `.cause`, cycle-safe via `seen` Set, appends `[CODE]` only when message doesn't already carry it. Tests pin the cases.
- Bug 4 (retry): `fetchWithRetry` is scoped to item files (rightly — manifests fall back to stale copies). `isRetryableTransport` explicitly refuses TLS failures so a self-signed cert on a private registry fails once, not three times. Bounded 150ms×attempt backoff.
- Bug 5 (compound-word search): `expandCompounds` splits + joins gated on the catalog's own vocabulary and inferred tokens carry 0.35× weight, so a `typewriter` query can't be dislodged by `type-match-cut` matching `type` at full strength. The rationale for the weight (rather than relying on common halves having low IDF) is exactly right for small corpora.
- Transcribe fixes: pinning `engine: "whisper"` + `expect(transcribeMock).toHaveBeenCalled()` closes the silent-skip gap on machines with parakeet-mlx installed.

**Consumer sweep**
- Every writer that emits `data-composition-variables` goes through `installOneFile → applyVariableDefaults` when the file is a component snippet. Block compositions still hang values on the mount element via existing `buildSnippet` path; that path unchanged. No bypass sites found.
- `assertSafeTarget` un-exported cleanly — only in-file caller remains.
- `registry/index.ts` re-export slimming: verified `resolveItem`, `fetchRegistryManifest`, `fetchItemManifest`, `fetchItemFile`, `assertSafeTarget`, `InstallResult`, `InstallOptions`, `ResolveOptions` were the ones dropped — all read at their source modules by remaining consumers.

**Nits (non-blocking)**
- `remote.ts::isRetryableTransport` regex and `add.ts::describeInstallFailure` transport regex are two separate lists (`network` + `timeouterror` in one, `ENOTFOUND` in the other). Same intent — worth consolidating into one shared classifier so a future error name added to one gets both. Not a correctness issue today: `describeCauseChain`'s output already contains the text either regex hits.
- Cross-file component variable aggregation in `installItem`: `variablesApplied`/`variablesInvalid` are `flatMap` (union), `variablesUnknown` intersects. If a component ever ships two HTML files with divergent declarations, the same id could report as both applied and invalid. Theoretical — components in the catalog today ship one snippet — but worth an assertion or a doc line.
- `fetchWithRetry` uses `AbortSignal.timeout(FETCH_TIMEOUT_MS)` per attempt: 3× the timeout on a genuinely offline machine. Backoff is short, timeout is not — consider a shorter per-attempt timeout on retries.
- CodeQL flags on `runtime-in-srcdoc.ts` (`<head[^>]*>` etc.) look like the usual false positives — `[^>]*` is linear, no catastrophic backtracking. Fine to dismiss.

**CI status at HEAD `c9194f7`**
- Latest run 32078742445: `Producer: unit tests` + `Producer: integration tests` both SUCCESS. `Analyze (javascript-typescript)` SUCCESS. `Preview parity` still IN_PROGRESS at review time. `Test` shows FAILURE from the *earlier* cancelled run (32078543814) — stale rollup row; the newer `Test` job (95538015807) is still in flight and should go green once the producer gate observes the fresh SUCCESS results. Confirm both flip before landing.

— Review by tai (pr-review)

terencecho (approved):
Approving at `c9194f76` on Terence's OK to add you to the trust-list. Upgrading my prior [COMMENT review](https://github.com/heygen-com/hyperframes/pull/3316#pullrequestreview-4955323473) — same findings apply. Non-blocking nits are yours to take or leave.

PR #3325, opened 2026-08-18
fix(skills): skip mirror fan-out to agents that read the universal store

### LINKED ISSUE #3294: Global skill mirroring should avoid redundant agent directories and provide target controls
## Summary

Pi natively discovers global skills from both [`~/.pi/agent/skills/` and `~/.agents/skills/`](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/skills.md#locations).

When a HyperFrames skill update performs an install, HyperFrames writes the selected skill bundles to Claude Code's global store and the universal `~/.agents/skills` store. It then mirrors **all globally installed HyperFrames skills attributed to its source** into detected agent-specific directories. Because Pi is in that mirror table, HyperFrames creates matching entries under `~/.pi/agent/skills` even though Pi already loads the universal copies.

Pi consequently discovers each mirrored skill twice, reports a name collision, keeps the `~/.pi/agent/skills` entry, and skips the `~/.agents/skills` entry.

**Clarification about `.claude`:** Pi does not load `~/.claude/skills` by default. That path appears here only because HyperFrames currently uses Claude Code's global skill store as the source for its fan-out. On Unix, the redundant Pi entries are symlinks back to that source. The two locations Pi actually discovers in this reproduction are `~/.pi/agent/skills` and `~/.agents/skills`.

This is deterministic duplicate discovery, not the unrelated-skill deletion reported in #3111. It is also narrower than the silent-update and overwrite concerns in #2613.

## Environment

- HyperFrames CLI `0.7.109`
- Pi `0.84.2`
- macOS
- Default `CLAUDE_CONFIG_DIR` (therefore Claude's store is `~/.claude/skills`)

## Steps to reproduce

1. Have Pi installed, so `~/.pi/agent/` exists.
2. Have HyperFrames skills installed globally, including copies under `~/.agents/skills`.
3. Remove or otherwise ensure the matching HyperFrames entries are absent from `~/.pi/agent/skills`.
4. Make at least one HyperFrames skill stale or missing so the next update performs an install rather than a no-op.
5. Run:

   ```sh
   npx hyperframes skills update product-launch-video
   ```

6. Observe output similar to:

   ```text
   Linked skills into 12 other agent directories.
   ```

7. Inspect the Pi directory. With the default Claude configuration, entries look like:

   ```text
   ~/.pi/agent/skills/hyperframes -> ../../../.claude/skills/hyperframes
   ```

8. Start Pi. It reports collisions such as:

   ```text
   "hyperframes" collision:
     ✓ auto (user) ~/.pi/agent/skills/hyperframes/SKILL.md
     ✗ ~/.agents/skills/hyperframes/SKILL.md (skipped)
   ```

In the observed reproduction, HyperFrames created 12 Pi symlinks and all 12 corresponding `SKILL.md` files were byte-identical to the universal copies. They still collide because Pi discovers them through two configured locations.

## Root cause in the current implementation

- `packages/cli/src/commands/skills.ts` invokes the upstream installer globally with `--agent claude-code universal --copy`.
- `packages/cli/src/utils/skillsMirror.ts` then chooses Claude's global store (`$CLAUDE_CONFIG_DIR/skills`, defaulting to `~/.claude/skills`) as the fan-out source.
- `packages/cli/src/utils/agentDirs.generated.ts` includes Pi at `.pi/agent/skills`.
- The mirror skips the Claude source and `~/.agents/skills` themselves, but Pi's agent-specific path is different, so it is mirrored despite Pi already consuming the universal store.
- The fan-out runs after an actual install. A freshness check that is a complete no-op does not recreate the links.

On Unix the fan-out uses relative symlinks; on Windows it copies directories. Either form can create duplicate Pi discovery when the universal copy is also present.

## Expected behavior

HyperFrames should not create multiple discoverable copies of the same skill for one agent by default.

A defensible design would provide:

1. **Capability-aware `auto` behavior by default** — do not mirror into an agent-specific directory when that agent already consumes the universal store.
2. **A persistent opt-out** — users should be able to disable global fan-out entirely.
3. **Optional target controls** — advanced users should be able to include or exclude specific agents.
4. **Non-interactive determinism** — agent-triggered updates should honor saved or default behavior without prompting on every run.

Possible CLI shapes, offered only as examples:

```sh
hyperframes skills update --no-mirror
hyperframes skills update --mirror-agents cursor,goose
```

The exact interface is open for maintainers to choose. A Pi-specific exclusion would fix this reproduction. A broader target-control design would also support users who deliberately manage several agent skill stores differently.

If explicit Pi mirroring is supported, the install targets must be coordinated: installing the same skill names into both Pi's agent-specific directory and `~/.agents/skills` preserves the collision.

## Acceptance criteria

- The default update path does not create Pi skill collisions when the same HyperFrames skills are installed under `~/.agents/skills`.
- Users have a documented way to prevent or scope mirroring without patching generated files or an npm cache.
- Existing non-redundant agent mirroring continues to work.
- A regression test covers an installed Pi plus the universal store.

santhiprakash:
## What

The global skills mirror no longer fans out into agent directories whose agent already discovers the universal `~/.agents/skills` store globally. Currently that set is `{ pi }`.

## Why

Pi discovers skills from **both** `~/.pi/agent/skills/` **and** `~/.agents/skills/` as global locations (see pi's `packages/coding-agent/docs/skills.md` → Locations). The global install already writes real files into the universal store, so mirroring an additional per-agent copy into `~/.pi/agent/skills` makes Pi discover every HyperFrames skill twice. Pi then reports a name collision, keeps the `~/.pi/agent/skills` entry, and **skips the universal entry**:

```text
"hyperframes" collision:
  ✓ auto (user) ~/.pi/agent/skills/hyperframes/SKILL.md
  ✗ ~/.agents/skills/hyperframes/SKILL.md (skipped)
```

Fixes #3294 (the deterministic duplicate-discovery half; the broader target-control surface proposed there is left for maintainers to design).

## How

- Added a `UNIVERSAL_STORE_READERS` capability set in `packages/cli/src/utils/skillsMirror.ts` (currently `pi`) and skip those agents in the fan-out loop, before the installed-marker check.
- The set deliberately lives in the mirror module, not in `agentDirs.generated.ts`: that file is `@generated` from vercel-labs/skills and a plain `(agent, base, sub)` list — it cannot carry per-agent capabilities, and any hand edit would be clobbered by the next `gen:agent-dirs` sync.
- Minimal by design: no cleanup of pre-existing `~/.pi/agent/skills` entries from earlier runs. Removing entries in a directory we no longer manage would risk deleting user-placed content (the mirror's replace-on-refresh precedent only applies to dirs it still mirrors). A one-time stale-link cleanup can be a follow-up if maintainers want it.

## Test plan

- [x] New test in `skillsMirror.test.ts`: with both Pi (`~/.pi/agent`) and cursor markers present, `mirrored` contains cursor and not pi, and no `~/.pi/agent/skills/hyperframes` entry is created.
- [x] Full skills-related suites pass: `skillsMirror`, `skillsManifest`, `skillsUpdateCheck`, `commands/skills` — 116/116.
- [x] `oxlint` / `oxfmt --check` clean on changed files; repo `check-skill-mirror` invariant still passes; lefthook pre-commit gates (format, fallow against synced main, typecheck after generating core runtime artifacts) pass.
- [ ] Manual testing performed (not applicable here — reproduction requires a Pi + HyperFrames co-install; covered by the deterministic unit test above)

### DIFF
--- a/packages/cli/src/utils/skillsMirror.test.ts
+++ b/packages/cli/src/utils/skillsMirror.test.ts
@@ -177,6 +177,30 @@ describe("mirrorGlobalSkills", () => {
     const link = join(home, ".cursor", "skills", "hyperframes");
     expect(realpathSync(link)).toBe(realpathSync(join(home, ".claude", "skills", "hyperframes")));
   });
+
+  // Pi natively discovers BOTH ~/.pi/agent/skills and the universal
+  // ~/.agents/skills (pi's packages/coding-agent/docs/skills.md#locations).
+  // A mirrored per-agent copy collides with the universal one and Pi skips
+  // the universal entry on name conflict (#3294), so the mirror must not fan
+  // out to it.
+  it("skips agents that natively read the universal store (pi, #3294)", () => {
+    const home = makeHome();
+    seedStore(home, ["hyperframes"]);
+    installMarker(home, ".pi/agent"); // Pi present
+    installMarker(home, ".cursor"); // a regular per-dir agent, for contrast
+
+    const { mirrored } = mirrorGlobalSkills({
+      skills: ["hyperframes"],
+      home,
+      platform: "linux",
+      env: ENV,
+    });
+    const agents = mirrored.map((m) => m.agent);
+    expect(agents).not.toContain("pi");
+    expect(agents).toContain("cursor");
+    // no per-agent copy created where the universal store already serves Pi
+    expect(existsSync(join(home, ".pi", "agent", "skills", "hyperframes"))).toBe(false);
+  });
 });
 
 describe("AGENT_GLOBAL_DIRS (generated table)", () => {
--- a/packages/cli/src/utils/skillsMirror.ts
+++ b/packages/cli/src/utils/skillsMirror.ts
@@ -9,7 +9,10 @@
 // populate.
 //
 // So we mirror the canonical Claude store into each of those per-agent dirs, but
-// only for agents the machine actually has (their marker dir exists). On Unix
+// only for agents the machine actually has (their marker dir exists). Agents
+// that already consume the universal ~/.agents/skills store globally (Pi) are
+// skipped: their universal copy is authoritative and a per-agent copy would
+// collide with it (#3294). On Unix
 // each skill is a relative symlink back into the store (one source of truth,
 // near-zero size, auto-fresh on update); on Windows it's a copy, because
 // symlinks there need admin / Developer Mode and otherwise silently dangle —
@@ -23,6 +26,22 @@ import { cpSync, existsSync, mkdirSync, readdirSync, rmSync, symlinkSync } from 
 import { homedir } from "node:os";
 import { dirname, isAbsolute, join, relative } from "node:path";
 import { AGENT_GLOBAL_DIRS, type AgentDirBase } from "./agentDirs.generated.js";
+
+/**
+ * Agents that natively discover the universal `~/.agents/skills` store globally
+ * in ADDITION to their own agent-specific directory. Mirroring into their own
+ * dir makes every skill discoverable twice.
+ *
+ * Pi is the known case (earendil-works/pi): it reads both `~/.pi/agent/skills/`
+ * and `~/.agents/skills/` as global locations (pi's packages/coding-agent/docs/
+ * skills.md#locations), so a mirrored entry collides with the universal copy
+ * and Pi skips the universal one on name conflict (#3294).
+ *
+ * The generated table cannot carry this capability — it is a plain
+ * (agent, base, sub) list synced from vercel-labs/skills — so the set lives
+ * here next to the mirror logic that needs it.
+ */
+const UNIVERSAL_STORE_READERS = new Set(["pi"]);
 
 export interface MirrorResult {
   /** The store mirrored from, or null when no global Claude store was found. */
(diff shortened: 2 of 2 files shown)


miga-heygen (commented):
**LGTM** — clean, minimal fix.

The problem is well-defined: Pi discovers `~/.agents/skills` natively, so the mirror's per-agent fan-out created a duplicate that Pi resolved by keeping the per-agent copy and skipping the universal one — the wrong winner.

The fix is a single `continue` in the fan-out loop gated on a `UNIVERSAL_STORE_READERS` set. Good design choices:

- **Set lives in `skillsMirror.ts`**, not in `agentDirs.generated.ts`. The generated file is a plain `(agent, base, sub)` list synced from vercel-labs/skills — it can't carry per-agent capabilities, and a hand edit would be clobbered. Correct placement.
- **No stale-entry cleanup.** Removing entries in a directory the mirror no longer manages risks deleting user-placed content. The replace-on-refresh precedent only applies to dirs still mirrored. Right call.
- **Test is specific.** Seeds both Pi and Cursor markers, asserts Pi is excluded and Cursor included, verifies no `~/.pi/agent/skills/hyperframes` entry was created. Covers both the positive and negative cases.

No SSOT concerns — the skip check is one line, one source of truth for which agents are universal-store readers, and the comment explains the invariant it protects.

_Review by Miga_

miguel-heygen (approved):

miguel-heygen:
@santhiprakash sign the commit pls

santhiprakash:
Done — commits are now signed. You should see the Verified badge on the latest commit.

santhiprakash:
Hi @miguel-heygen — quick check-in on this one. The head commit is signed now (verified badge on 4dc197df) per your earlier request, and the PR is approved, current with main, and green on checks. Happy to adjust anything else if something is still blocking the merge on your end — otherwise no rush.

PR #3332, opened 2026-08-18
fix(producer): clamp embedded video/audio windows to the scene

### LINKED ISSUE #3331: probeStage extends embedded video/audio ends to full source duration
### Describe the bug

In probeStage, when the browser probe reconciles media metadata, the code can replace a video or audio clip’s compile-time `end` with a value projected from the browser.

The browser `end` is the media’s natural duration (`data-start` 0 + full source length). It is not the scene slot length.

`parseSubCompositions` already sets the correct per-scene `end` from the sub-composition slot’s `data-duration`. The reconcile step should only fill a missing bound, or shrink when the source is shorter. It must not extend a valid window.

Today it overwrites when the projected end differs. A short scene that plays a slice of a long recording (via `data-media-start`) gets an end of roughly `sceneStart + fullSourceDuration`. Frame extraction then pulls far more of the source than the scene needs. On slow paths (for example Cloud Run with a ~300s ffmpeg timeout), deep-offset scenes can fail extraction and render as solid black while the job still reports success. Local machines often finish, but they still do wasted work.

### Link to reproduction

https://github.com/ArcadeHQ/hyperframes-repros/tree/patch/clamp-embedded-media

### Steps to reproduce

1. Clone https://github.com/ArcadeHQ/hyperframes-repros and check out `patch/clamp-embedded-media`.
2. Run `npx hyperframes lint` (expect 0 errors).
3. Run `npm run render -- --debug` (or `npx hyperframes@0.8.3 render --debug`).
4. Confirm the log shows a browser probe (root duration unknown — length comes from the GSAP master timeline).
5. Read the `video_extract` checkpoint: `maxFramesPerVideo`, `avgFramesPerExtractedVideo`, `totalFramesExtracted`.

The project has four 3s nested scenes that reuse one 90s take at `data-media-start` 5 / 25 / 50 / 70. Videos set `data-start="0"` and omit `data-duration` so ffprobe fills natural remaining length. Root omits `data-duration` on purpose.

### Expected behavior

Each embedded video keeps its compile-time scene window (~3s). At 30fps, `maxFramesPerVideo` should be about 90. Extract cost should stay scene-sized.

### Actual behavior

On hyperframes@0.8.3 the `video_extract` checkpoint reported:

  maxFramesPerVideo: 360
  avgFramesPerExtractedVideo: 225
  totalFramesExtracted: 900

(Fixed values would be about 90 / 90 / 360.)

`probeStage` extends each compile-time `end` using the browser’s natural remaining duration, so extraction follows remaining source length, not the 3s scene slot.

### Environment

```shell
✓ Version          0.8.3 (latest)
  ✓ Memory           16.0 GB total · 3.6 GB available
  ✓ Disk             363.9 GB free
  ✓ Frames cache     /var/folders/bl/c4hk6r7n34g1d81txd99gsgr0000gn/T/hyperframes-extract-cache-501 · 363.9 GB free at /var/folders/bl/c4hk6r7n34g1d81txd99gsgr0000gn/T/hyperframes-extract-cache-501 · default
  ✓ Archive extractor unzip
  ✓ Environment      Native terminal
                     brew install whisper-cpp
                     pip install kokoro-onnx soundfile
                     pip install transformers torch soundfile numpy
  ✓ Chrome           cache: /Users/val/.cache/puppeteer/chrome-headless-shell/mac_arm-150.0.7871.24/chrome-headless-shell-mac-arm64/chrome-headless-shell
  ✗ Docker           Not found
                     https://docs.docker.com/get-docker/
                     Start Docker Desktop or run: sudo systemctl start docker

  ◇  Some checks failed — see hints above
```

### Additional context

Suggested fix (video and audio):

  existing.end = existing.end <= 0 ? projectedEnd : Math.min(existing.end, projectedEnd)

Blank baseline: `main` on the same repo.
We have a local patch and will open a PR against this issue.

valeriangalliat:
## What

When reconciling browser media metadata in probeStage, never extend a video/audio compile-time `end` to the browser-projected natural source duration. Only fill a missing bound, or shrink with Math.min.

Fixes #3331.

## Why

parseSubCompositions already clamps nested media to the scene slot. The browser `end`/`data-duration` after ffprobe is natural remaining source length, not the slot. Overwriting the compile-time end inflated extract windows for short scenes that slice a long take via data-media-start. That wastes work locally and can time out black on constrained runners while the render still succeeds.

## How

In the video and audio reconcile branches:

```ts
existing.end =
  existing.end <= 0 ? projectedEnd : Math.min(existing.end, projectedEnd)
```

Image reconcile is unchanged.

## Test plan

How was this tested?

Manual: https://github.com/ArcadeHQ/hyperframes-repros/tree/patch/clamp-embedded-media on unpatched 0.8.3 → maxFramesPerVideo 360 / totalFramesExtracted 900. With this patch → expect ~90 / ~360. `npx hyperframes lint` clean on that repro.

### DIFF
--- a/packages/producer/src/services/render/stages/probeStage.ts
+++ b/packages/producer/src/services/render/stages/probeStage.ts
@@ -508,11 +508,9 @@ export async function runProbeStage(input: ProbeStageInput): Promise<ProbeStageR
                 el.start,
                 resolveBrowserMediaEnd(el.start, el.end, el.duration),
               );
-              if (
-                projectedEnd > 0 &&
-                (existing.end <= 0 || Math.abs(existing.end - projectedEnd) > BROWSER_MEDIA_EPSILON)
-              ) {
-                existing.end = projectedEnd;
+              if (projectedEnd > 0) {
+                existing.end =
+                  existing.end <= 0 ? projectedEnd : Math.min(existing.end, projectedEnd);
               }
               if (
                 el.mediaStart > 0 &&
@@ -554,11 +552,9 @@ export async function runProbeStage(input: ProbeStageInput): Promise<ProbeStageR
                 el.start,
                 resolveBrowserMediaEnd(el.start, el.end, el.duration),
               );
-              if (
-                projectedEnd > 0 &&
-                (existing.end <= 0 || Math.abs(existing.end - projectedEnd) > BROWSER_MEDIA_EPSILON)
-              ) {
-                existing.end = projectedEnd;
+              if (projectedEnd > 0) {
+                existing.end =
+                  existing.end <= 0 ? projectedEnd : Math.min(existing.end, projectedEnd);
               }
               if (
                 el.mediaStart > 0 &&


miga-heygen (commented):
**LGTM with nits** — useful fix, worth merging.

### Why this matters

`parseSubCompositions` clamps nested media to the scene slot. Then `probeStage` reconciles with browser metadata, and the old code blindly overwrote the compile-time `end` with the browser-projected natural source duration — inflating extract windows for any short scene that slices a long take via `data-media-start`. The repro shows 900 frames extracted instead of ~360: 2.5x wasted work, and a timeout risk on constrained runners (Lambda, Cloud Run) where the render still technically succeeds but burns time extracting frames that will never be composited.

The fix is correct: `Math.min(existing.end, projectedEnd)` — never extend past the compile-time bound, only fill a missing one (`existing.end <= 0`) or shrink.

### Missing: unit tests

This is very testable. A minimal scenario: compile-time `existing.end = 3.0` (scene slot), `projectedEnd = 30.0` (natural source duration). Assert `existing.end` stays `3.0` after reconciliation. The old behavior was a silent regression — a test here prevents it from recurring if someone reworks the reconcile logic.

### Nit: `BROWSER_MEDIA_EPSILON` comparison removed

The old code only overwrote when `Math.abs(existing.end - projectedEnd) > BROWSER_MEDIA_EPSILON`. The new code takes `Math.min` unconditionally, so a `projectedEnd` that's a sub-millisecond fraction shorter than `existing.end` will now shrink it. In practice this doesn't matter for frame extraction, but it's a semantic change from "ignore tiny differences" to "always take the smaller value." Worth noting in case the epsilon was load-bearing somewhere else in the pipeline.

### What I verified

- Both video and audio reconcile blocks get the identical fix — no asymmetry.
- The `existing.end <= 0` guard correctly handles the "no compile-time bound" case (fill, don't clamp).
- Image reconcile is untouched (images don't have extract windows).
- The fix doesn't touch `mediaStart` reconciliation — that path already has its own `Math.min` guard below.

_Review by Miga_

miguel-heygen (approved):

PR #3334, opened 2026-08-18
fix(studio): allow hfId-only clips to move and trim on the timeline

miguel-heygen:
Fixes **"This clip can't be moved or resized from the timeline yet"** on clips that are perfectly well-formed.

## Root cause

`findTimelineDomNodeForClip` ends with a positional fallback:

```ts
return candidates[fallbackIndex] ?? null;
```

`candidates` is only nodes carrying `[data-start]`. A clip whose own element has none — an expanded child row, for instance — is never among them, so it falls through and is handed **an unrelated node by index**.

That corrupts **two** clips per collision:
- the caller gets a host that is not its element, and
- the clip that actually owned that node is starved to `null`, losing `hfId`, `domId` and `selector` — and with them `canMove`.

Instrumented on a six-scene composition, the resolver's own decisions:

```
clip s=0    d=14  trk=2 tag=img | branch=positional picked=hf-opwy   ← took the 6.0s scene div
clip s=0    d=14  trk=3 tag=img | branch=positional picked=hf-bxz4   ← took the 11.2s scene div
clip s=6    d=2.6 trk=1 tag=div | branch=none       picked=null      ← starved → refused
clip s=11.2 d=1.5 trk=1 tag=div | branch=none       picked=null      ← starved → refused
```

Those two starved scenes were **exactly** the two clips the timeline refused to move. The other four resolved normally and dragged fine.

## Fix

Require the positional candidate to share the clip's tag. Same-tag attribute drift still resolves, which is what the fallback exists for; cross-tag theft cannot happen.

Also included (found while chasing this, kept because it is a real latent inconsistency): `hasPatchableTimelineTarget` accepts `hfId`, matching `resolveDomEditCapabilities` (`Boolean(args.selector || args.hfId)`) and `timelineElementSplit.ts` (`Boolean(el.hfId || el.domId || el.selector)`). `TimelineElement` already carries `hfId`, `findElementForSelection` resolves `data-hf-id` ahead of id/selector, and the move handler's primary persist is `sdkTimingPersist(element.hfId, …)`.

## Verification

**End to end, in a Studio built from this branch:**
1. Opened a composition whose scene clips are plain `<div class="s clip" data-start data-duration>`.
2. Dragged the clip that was previously refused.
3. No toast. Timeline row moved `6.0 → 13.3`.
4. **Source file updated: `data-start="6"` → `data-start="13.28"`.**

**Tests** — `77/77` in the two touched files with the project's vitest config:
- `returns null instead of stealing a div for an img clip` — **fails on `main`**, passes here.
- `leaves the div clips resolvable after an unmatched img clip`.
- `still allows a same-tag positional fallback when attributes drift` — pins the behaviour the fallback is for, so this is not an over-fix.
- 4 capability tests for `hfId`, two of which fail on `main`.

lint, format, fallow, typecheck and commitlint all pass via pre-commit hooks.

## Note on scope

This supersedes my earlier framing of this PR. The `hfId` change alone did **not** fix dragging — I verified that and said so — and the positional fallback is the real cause. Both commits are here; the second is the fix.

### DIFF
--- a/packages/studio/src/player/components/timelineAuthoredMoveTarget.ts
+++ b/packages/studio/src/player/components/timelineAuthoredMoveTarget.ts
@@ -11,6 +11,7 @@ export function canMoveTimelineElement(element: TimelineElement): boolean {
     duration: element.duration,
     domId: element.domId,
     selector: element.selector,
+    hfId: element.hfId,
     compositionSrc: element.compositionSrc,
     playbackStart: element.playbackStart,
     playbackStartAttr: element.playbackStartAttr,
--- a/packages/studio/src/player/components/timelineEditCapabilities.ts
+++ b/packages/studio/src/player/components/timelineEditCapabilities.ts
@@ -23,8 +23,17 @@ function isDeterministicTimelineWindow(input: {
   return ["video", "audio", "img"].includes(input.tag.toLowerCase());
 }
 
-export function hasPatchableTimelineTarget(input: { domId?: string; selector?: string }): boolean {
-  return Boolean(input.domId || input.selector);
+export function hasPatchableTimelineTarget(input: {
+  domId?: string;
+  selector?: string;
+  hfId?: string;
+}): boolean {
+  // hfId counts as a stable target: Studio stamps data-hf-id into the source and
+  // findElementForSelection resolves it before id/selector, so a clip carrying only
+  // an hfId is just as patchable as one with an author-written id. Omitting it here
+  // made those clips report canMove:false and surface "This clip can't be moved or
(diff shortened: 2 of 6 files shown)


miguel-heygen:
Closing the old branch, while preserving the actual fix for a fresh PR. Current main still allows cross-tag positional DOM binding and excludes hfId-only targets at the timeline capability gate. The replacement will fix those current paths and reverify real drag/trim persistence, rather than rebase this stale branch.

miguel-heygen:
Verified narrow successor: https://github.com/heygen-com/hyperframes/pull/3681

PR #3336, opened 2026-08-18
fix(skills): stage SVGs that capture wrote into capture/assets/svgs/

miguel-heygen:
## What

`stageAssets` now searches `capture/assets/svgs/`, so an SVG that `hyperframes capture` extracted can be staged into `assets/` like any other captured asset.

## Why

`capture` writes inline SVGs to `capture/assets/svgs/` (`packages/cli/src/capture/assetDownloader.ts`), and the capture manifest advertises them to the agent as `assets/svgs/<name>.svg`. So a frame names one in `asset_candidates` exactly the way it names a screenshot.

The staging search list covered `capture/{assets,assets/videos,screenshots}` only. Every captured SVG resolved to nothing: logged as a non-fatal anomaly, and the frame 404'd the brand mark it had been told to use. Found running `/product-launch-video` end to end.

## How

One entry added to `captureDirs`.

`lib/assets.mjs` is byte-identical across `product-launch-video`, `faceless-explainer` and `pr-to-video`, so the fix lands in all three.

**Not covered:** the triplication itself. Folding `lib/assets.mjs` into `hyperframes-core/scripts/lib/`, where `frame-packets-core.mjs` already lives, is a separate change and not in this PR.

## Test plan

New `skills/product-launch-video/scripts/stage-assets.test.mjs`, covering both the captured-SVG case and an asset that genuinely exists nowhere (so the anomaly path is still exercised).

Verified it fails on the parent commit and passes here.

Manual: a fixture project with a captured SVG, a screenshot and a root capture asset, all three named by one frame.

- Before: `✓ staged 2/3 asset(s) into assets/` plus `asset "brand-mark.svg" named by a frame but not found under capture/`
- After: `✓ staged 3/3 asset(s) into assets/`

### DIFF
--- a/skills-manifest.json
+++ b/skills-manifest.json
@@ -6,7 +6,7 @@
       "files": 138
     },
     "faceless-explainer": {
-      "hash": "f37cb1dba2890b79",
+      "hash": "7d587ff36c975d9a",
       "files": 24
     },
     "figma": {
@@ -62,12 +62,12 @@
       "files": 132
     },
     "pr-to-video": {
-      "hash": "8ea0227e18ab5fc6",
+      "hash": "14250b018114d26d",
       "files": 30
     },
     "product-launch-video": {
-      "hash": "e89a8d11d99b4edf",
-      "files": 28
+      "hash": "12c0895d4963aa90",
+      "files": 29
     },
     "remotion-to-hyperframes": {
       "hash": "bf184a65059b95e8",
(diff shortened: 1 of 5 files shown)


miga-heygen (dismissed):
**Approve** — one-line fix applied identically across all three triplicated copies. The new `capture/assets/svgs/` entry sits naturally alongside the existing `capture/assets/videos/` subdir entry, the comment is updated to match, and the test covers both the positive case (SVG + screenshot staged together) and the negative (absent asset still surfaces an anomaly). No issues.

— Miga

PR #3346, opened 2026-08-19
fix(audio): rebuild timestamps between apad and atrim so FFmpeg 7+ keeps every mixed clip

### LINKED ISSUE #3344: Audio mix drops and misplaces clips on FFmpeg 7+ (apad,atrim branch template)
## Describe the bug

```
```

the output entirely.

## Steps to reproduce

```bash
#!/usr/bin/env bash
set -uo pipefail
TOTAL=43.3

run(){
  for ((i=0;i<n;i++)); do
  done
  for ((i=0;i<n;i++)); do
  done; echo
}
```

## Expected behavior

## Actual behavior

FFmpeg 4.2.7 — correct.

```
```

```
```

## Suggested fix

```
```

time.

## Environment

```shell
Node.js  v22.22.2 (linux x64)
```

sujeito-operator:
## What

`apad,atrim=0:<total>` — an indefinite pad bounded by a downstream trim — is replaced everywhere with `apad,asetpts=N/SR/TB,atrim=0:<total>`, and the three call sites that rebuilt that chain by hand now share one helper.

Closes #3344.

## Why

I reproduced the issue's script against `ffmpeg 7.0.2-static` on this machine and got the reported numbers back byte for byte:

```
```

## How

**The suggested fix is `apad=whole_dur=<total>`, and I did not take it — twice over.**

1. This repo has already decided against `whole_dur` and pinned the decision in three assertions and an error-classification test: *"Bundled Windows FFmpeg builds reject `apad=whole_dur`"* (`audioPadTrim.test.ts:96`), and `audioMixer.test.ts` carries a case whose stderr fixture is literally `Error applying option 'whole_dur': Option not found`. Swapping it in would fix Linux 7.x by breaking the builds that comment is about.

2. `whole_dur` also drops a bound the current chain provides. `audioMixer.ts` deliberately lets an FX tail run past its clip and relies on the trailing `atrim` to stop it at the composition's end. With the same six branches (one delayed 42 s with a 5 s tail against a 43.3 s composition) and `-t` removed so the graph speaks for itself:

   | branch template | output duration |
   | --- | --- |
   | `apad,atrim=0:43.3` (current) | 43.300000 |
   | `apad=whole_dur=43.3` | **46.999977** |
   | `apad,asetpts=N/SR/TB,atrim=0:43.3` | 43.300000 |

   Today `-t totalDuration` hides that at both mix call sites. It is still a bound quietly moving from the filter graph to the muxer.

**What actually goes wrong is narrower than "`apad` + `atrim`".** Any `atrim` reading `apad`'s output timestamps directly is affected — I tried `atrim=end=`, reordering, and `asetpts` after the trim, and all three still drop clip 4. Rebuilding the timestamps from the running sample count *between* the two is what fixes it, and the result is then identical to `whole_dur` at every clip count I measured (2, 3, 4, 5, 6) while keeping both the `atrim` bound and the Windows builds:

```
n=5 CURRENT       t0: -29.413305 c0: -29.081774 c1: -28.166525 c2: -23.356880 c3: -27.338586 c4:       -inf
n=5 whole_dur     t0:       -inf c0: -31.019831 c1: -30.103960 c2: -24.084179 c3: -29.276778 c4: -28.521531
n=5 asetpts FIX   t0:       -inf c0: -31.019831 c1: -30.103960 c2: -24.084179 c3: -29.276778 c4: -28.521531
```

**One place, not three.** `buildPadToDurationFilter` lives in `packages/engine/src/services/audioPadFilter.ts` and is exported from `@hyperframes/engine`; `audioMixer.ts`, `audioExtractor.ts` and `audioPadTrim.ts` all call it. It takes an already-formatted seconds string on purpose — the three sites format numbers differently (`Number(v.toFixed(6)).toString()`, raw `String()`, fixed-6) and the helper must not silently change any of them.

**One thing the issue lists that I could not reproduce as broken:** `audioPadTrim.ts:145` is a single-input `-af` pad with no `adelay`, and on 7.0.2 the old and new chains produce identical output (both 43.300000 s, same RMS at every probe point). It is changed anyway, because it is the same fragile template and the point of the helper is that the next divergence cannot happen. Worth knowing when weighing the risk of that hunk.

## Test plan

**New — `audioPadFilter.test.ts` (4 tests).** Pins the chain and its *order*, since order is the behaviour here, and keeps the `whole_dur` prohibition asserted.

**New — `audioMixer.padTimestamps.integration.test.ts`.** The regression test the issue asks for: five clips at staggered offsets through the real `processCompositionAudio` and real FFmpeg, asserting `t=0` is silent and every clip is audible at its own offset. It follows this repo's `describe.skipIf(!hasFfmpeg)` convention, so it is a no-op where FFmpeg is absent.

I checked it fails without the fix rather than assuming it: reverting only the helper's return value to `apad,atrim=0:${seconds}` gives

```
AssertionError: expected -24.898801 to be less than -60
  at audioMixer.padTimestamps.integration.test.ts:130
```

Three or fewer clips passes on a broken build, which is why it mixes five.

**Existing assertions updated** in `audioMixer.test.ts` (3) and `audioPadTrim.test.ts` (2). The two `not.toContain("whole_dur")` assertions are untouched and still pass — that is the point.

**Suites** (`bun`, node v24.8.0, ffmpeg 7.0.2-static on PATH). I ran each one on a stashed pristine tree first and diffed the failure *set*:

| | before | after |
| --- | --- | --- |
| `@hyperframes/engine` | 5 failed, 1590 passed | 5 failed, **1595** passed |
| `@hyperframes/producer` | 1 failed, 594 passed | 1 failed, 594 passed |

The failure sets are identical, and all six are environmental on this box rather than from this change — four in `ffprobe.test.ts` on the checked-in HDR PNG fixture (`PNG input is structurally incomplete`, from a shallow clone without LFS) and two on the `unindexed negative-base MPEG-TS` fixture. `bunx oxlint` and `bunx oxfmt --check` are clean on all nine files; `typecheck` exits 0 for both packages.

**Not run:** the producer integration lane and any browser/render end-to-end. `audioExtractor.ts`'s `mixTracks` has no test of its own before or after this change, so that hunk rests on being the same template as the mixer's rather than on a test.

**Disclosure:** written with AI assistance. CONTRIBUTING says there is no need to disclose, but I would rather you know. Every measurement above was run rather than predicted, and I can explain any part of it.

### DIFF
--- a/packages/engine/src/index.ts
+++ b/packages/engine/src/index.ts
@@ -218,6 +218,7 @@ export {
   parseAudioElements,
   processCompositionAudio,
 } from "./services/audioMixer.js";
+export { buildPadToDurationFilter } from "./services/audioPadFilter.js";
 export { cloneCaptureWarning, cloneCaptureWarnings } from "./services/captureWarning.js";
 export type {
   AudioElement,
--- a/packages/engine/src/services/audioMixer.padTimestamps.integration.test.ts
+++ b/packages/engine/src/services/audioMixer.padTimestamps.integration.test.ts
@@ -74,7 +74,11 @@ function rmsDb(path: string, atSeconds: number): number {
   const match = /RMS level dB:\s*(\S+)/.exec(overall);
   if (!match) throw new Error(`astats reported no RMS level for ${path} at ${atSeconds}s`);
   const raw = match[1]!;
-  return raw.endsWith("inf") ? (raw.startsWith("-") ? -Infinity : Infinity) : Number.parseFloat(raw);
+  return raw.endsWith("inf")
+    ? raw.startsWith("-")
+      ? -Infinity
+      : Infinity
+    : Number.parseFloat(raw);
 }
 
 describe.skipIf(!hasFfmpeg)("mixed audio branch padding", () => {
--- a/packages/engine/src/services/audioMixer.padTimestamps.integration.test.ts
+++ b/packages/engine/src/services/audioMixer.padTimestamps.integration.test.ts
@@ -1,0 +1,139 @@
+import { execFileSync, spawnSync } from "node:child_process";
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, describe, expect, it } from "vitest";
+import { processCompositionAudio } from "./audioMixer.js";
+import type { AudioElement } from "./audioMixer.types.js";
+
+/**
+ * Real-FFmpeg regression test for the pad/trim branch template.
+ *
+ * `apad,atrim=0:<total>` — an indefinite pad bounded by a downstream trim —
+ * behaves on FFmpeg 4.2.7 and misbehaves on 7.0.2 and newer: audio leaks to
+ * `t=0` from three mixed branches onward, and from four branches onward the
+ * branch with the largest `adelay` disappears from the mix. Nothing errors;
+ * the render succeeds with wrong audio, which is why only an output
+ * measurement catches it.
+ *
+ * Three or fewer clips is not enough — the dropped-branch half of the bug does
+ * not appear until four. This mixes five.
+ */
+
+const dirs: string[] = [];
+afterEach(() => dirs.splice(0).forEach((dir) => rmSync(dir, { recursive: true, force: true })));
+
+const hasFfmpeg = (() => {
+  try {
+    execFileSync("ffmpeg", ["-version"], { stdio: "ignore" });
+    execFileSync("ffprobe", ["-version"], { stdio: "ignore" });
+    return true;
+  } catch {
+    return false;
+  }
+})();
+
+const TOTAL_SECONDS = 43.3;
+
+/** start (s), source duration (s), tone frequency (Hz) */
+const CLIPS = [
+  { start: 0.3, duration: 7.01, frequency: 400 },
+  { start: 19.6, duration: 0.4, frequency: 1000 },
+  { start: 20.6, duration: 0.4, frequency: 1200 },
+  { start: 30.0, duration: 0.4, frequency: 1500 },
+  { start: 40.0, duration: 0.4, frequency: 1800 },
+] as const;
+
+/**
+ * Mean RMS in dB over a 250ms window starting at `atSeconds`. Digital silence
+ * reports `-inf`, which is the signal for "this branch is not in the mix".
+ */
+function rmsDb(path: string, atSeconds: number): number {
+  const probe = spawnSync(
+    "ffmpeg",
+    [
+      "-hide_banner",
+      "-v",
+      "info",
+      "-ss",
+      atSeconds.toFixed(3),
+      "-t",
+      "0.25",
+      "-i",
+      path,
+      "-af",
+      "astats=metadata=1",
+      "-f",
+      "null",
+      "-",
+    ],
+    { encoding: "utf8" },
+  );
+  const stderr = probe.stderr ?? "";
+  const overall = stderr.slice(stderr.lastIndexOf("Overall"));
+  const match = /RMS level dB:\s*(\S+)/.exec(overall);
+  if (!match) throw new Error(`astats reported no RMS level for ${path} at ${atSeconds}s`);
+  const raw = match[1]!;
+  return raw.endsWith("inf") ? (raw.startsWith("-") ? -Infinity : Infinity) : Number.parseFloat(raw);
+}
+
+describe.skipIf(!hasFfmpeg)("mixed audio branch padding", () => {
+  it("keeps t=0 silent and lands all five staggered clips at their offsets", async () => {
+    const baseDir = mkdtempSync(join(tmpdir(), "hf-padts-base-"));
+    const workDir = mkdtempSync(join(tmpdir(), "hf-padts-work-"));
+    dirs.push(baseDir, workDir);
+
+    const elements: AudioElement[] = CLIPS.map((clip, i) => {
+      const src = `clip-${i}.wav`;
+      execFileSync("ffmpeg", [
+        "-hide_banner",
+        "-loglevel",
+        "error",
+        "-f",
+        "lavfi",
+        "-i",
+        `sine=frequency=${clip.frequency}:duration=${clip.duration}`,
+        "-ar",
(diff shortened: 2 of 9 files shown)


miguel-heygen (approved):
Approving. I reproduced the bug and the fix independently on Linux x86_64 rather than taking the PR body's numbers, and the claims hold. Two notes below, neither blocking.

## What I verified

Ran the issue's graph across five FFmpeg builds, comparing the old chain, `apad=whole_dur`, and this PR's chain, at 3/4/5/6 mixed branches:

| FFmpeg | `apad,atrim` | `apad=whole_dur` | `apad,asetpts=N/SR/TB,atrim` |
| --- | --- | --- | --- |
| 4.2.7 (Ubuntu) | correct | correct | correct, identical output |
| 6.0.1 static | correct | correct | correct, identical output |
| 7.0.2 static | **broken** | correct | **correct** |
| N-124665 (git master, 2026-05) | correct | correct | correct |
| 8.1.1 | correct | correct | correct |

On 7.0.2 the old chain reproduces byte for byte what the issue reports: `t=0` leaks at n>=3, and the largest-`adelay` branch drops at n>=4. Confirmed the drop also hits n=6, where it takes out `c5` rather than `c4`, so it is the last-delayed branch and not a fixed index.

The `whole_dur` rejection holds up too. With `-t` removed so the graph speaks for itself, a branch delayed 42s with a 5s source against a 43.3s target:

```
apad,atrim=0:43.3                 -> 43.300000
apad=whole_dur=43.3               -> 46.999977
apad,asetpts=N/SR/TB,atrim=0:43.3 -> 43.300000
```

Same on every version above. The bound is real, not theoretical.

I also ran the new integration test itself against 7.0.2, and reverted only the helper's return value to confirm it is not vacuous:

```
with fix:    1 passed
without fix: AssertionError: expected -24.898801 to be less than -60
```

Engine suite: 1606 passed, 3 skipped, exit 0. Producer unit lane: exit 0. `oxlint` and `oxfmt --check` clean on all nine files.

I tried to break the change and could not. `asetpts=N/SR/TB` discarding an incoming PTS is safe here because every branch input is a freshly written PCM WAV starting at 0, `mediaStart` is materialised by `-ss` before `-i` rather than as a PTS offset, and `adelay` inserts real samples. The FX-tail cut lands in the same place. The single-input `-af` path in `audioPadTrim.ts` is unchanged in output on every version I tested, which matches what the PR already discloses about that hunk.

## 1. The regression test does not run in CI, and would not catch the bug if it did

`describe.skipIf(!hasFfmpeg)` is the right convention, but the engine lane has no FFmpeg. The `Test` job runs `bun run --filter '!@hyperframes/producer' test` without an apt install, and only `producer-source-tests` with `lane == integration` installs one. The proof is already in the logs: the sibling gated test shows up on `main` as

```
↓ src/services/audioMixer.level.test.ts (2 tests | 2 skipped)
```

The new test uses the same gate, so it will skip identically.

Adding `apt-get install ffmpeg` to that job would not fix it either. Ubuntu 24.04 ships the 6.x line, and I measured 6.0.1 as unaffected: with the fix reverted, the new test still passes there. Catching this specific regression needs a pinned 7.x, so the honest options are a pinned 7.x binary in one lane, or accepting that this test is documentation of the repro and saying so in the file.

Worth noting for whoever picks that up: `ffmpeg-static` is only a dependency of `packages/aws-lambda`, and `findFfBinary` falls through to `PATH`, so the users actually exposed to this are the ones whose system FFmpeg is 7.x. That is a real population, and it is the reason to take the fix regardless of the CI gap.

## 2. Two small comment inaccuracies

`audioPadFilter.ts:19` says the bug is present "On FFmpeg 7 and newer". I measured git master from 2026-05 and 8.1.1 as unaffected, so the window looks like 7.x specifically. Since the whole file is a record of why this chain has three filters, being precise about the window matters for whoever revisits it.

`audioPadFilter.ts:5-7` says "Three call sites build this chain". `audioExtractor.ts` carries `// fallow-ignore-file unused-file` and has no importers, so two are live. The change to it is still right, but the comment counts a dead file as a consumer.

Minor, take it or leave it: the skip gate probes bare `ffmpeg` from `PATH` while `processCompositionAudio` resolves through `getFfmpegBinary()`, which honours `HYPERFRAMES_FFMPEG_PATH`. `audioMixer.level.test.ts` and `audioFxRender.test.ts` both use `getFfmpegBinary()`. With the override set, the gate and the code under test measure different binaries, which is exactly the FFmpeg-7-vs-6 setup this test exists for. I hit this while testing and had to set both.

And `buildPadTrimAudioPlan` now has `asetpts=N/SR/TB` in the pad branch and `asetpts=PTS-STARTPTS` in the trim branch a few lines apart, with nothing saying why they differ. One clause would save the next reader the trip.

Nice work on the write-up. Measuring the `whole_dur` duration difference instead of asserting it, and calling out that the `audioPadTrim.ts` hunk has no reproducible defect behind it, is what made this quick to check.

miga-heygen (changes requested):
**Changes Requested**

The fix itself is correct — `asetpts=N/SR/TB` between `apad` and `atrim` is the right approach, and the reasoning against `whole_dur` (Windows builds + FX tail bound) holds up. Two gaps:

### 1. `audioExtractor.ts:mixTracks` has zero test coverage

The PR author flags this: "that hunk rests on being the same template as the mixer's rather than on a test." That's exactly the class of bug this PR exists to fix — three sites sharing a pattern that silently diverged. Now they share a helper, but one of the three has no test verifying it calls the helper correctly. A unit test asserting `mixTracks`' filter string contains `apad,asetpts=N/SR/TB,atrim=` would close the loop.

### 2. No test for the FX-tail bound claim

The PR description shows `whole_dur` producing 46.999977s vs `atrim`'s 43.300000s on a branch with a tail overrun — that's the stated reason for keeping `atrim` over `whole_dur`. But neither the new integration test nor the unit tests verify that the `atrim` bound actually stops an FX tail. An integration case (one clip with a tail extending past `totalDuration`, assert output duration ≤ `totalDuration`) would pin the claim.

_Review by Miga_

miguel-heygen:
@sujeito-operator address the reviews and also sign the commits pls to merge it or I'll take over tomorrow attributing the fix to you

PR #3354, opened 2026-08-19
feat(skills): anchored-connector rule + source-traceable visuals doctrine

WaterrrForever:
## What

Two small advisory rules for the creative/animation doctrine — self-checks, not hard gates:

**1. Connector lines earn their place** (anti "floating gold line")
- `hyperframes-creative/references/motion-principles.md` § Visual Composition: any beam / rail / scan line / drawn underline must be able to name its start anchor, end anchor (real elements), and its job — reveal, route, or validate. A line the frame can lose without losing meaning gets cut. Edge-lit components usually read as more designed than free-floating arcs.
- `hyperframes-animation/rules/svg-path-draw.md` § Critical Constraints: same requirement at the rule level for drawn connector paths.

**2. Visuals point back to the source**
- `hyperframes-creative/references/story-spine.md` rule 4: when a video derives from concrete material (PR / repo / article / product page), each frame's key visual should trace to a specific source line — real filenames and numbers over stock props. Self-check: a prop that could appear unchanged in another product's video didn't come from the source.
- The four SKILL.md index lines that enumerate story-spine's rules are synced (creative, product-launch-video, faceless-explainer, pr-to-video). `skills-manifest.json` regenerated by the pre-commit hook.

## Why

From a 4-cell sandbox replay comparing the official skills against the community skill `geekjourneyx/hyperframes-motion-director` (410★): its two strongest ideas were source-driven visual selection and mandatory anchoring for connector lines — both dimensions our doctrine didn't cover (all 357 catalog items describe *how* to draw; nothing covered *why draw this* or *what a line must attach to*). Ideas only — the community repo is AGPL-3.0, no upstream text or code is used; all wording here is original. The anchor / job / deletion-test checklist *shape* is adopted idea-level from that project's connector rule — credited here deliberately.

## Checks

- `lint:skills`: 32 files, no issues
- `check:skill-mirror`: untouched (changes are in `skills/` only, not the repo-native mirror layers)
- pre-commit hooks green at commit time — but `skills-manifest` and `format` run in parallel and raced at ee90f1f (manifest hashed the SKILL.md tables before oxfmt renormalized them). Fixed in 44cc2ec: manifest regenerated at head, second regen is a no-op.


jrusso1020 (changes requested):
Reviewed at `ee90f1fc` (branch is current with `main` — `ahead_by: 1, behind_by: 0`, so nothing here is a base-drift artifact). No prior reviews or comments on the PR when I started, so this is a first pass, not an additive one.

**Audited:** all 8 changed files end to end; `story-spine.md`, `motion-principles.md`, and `svg-path-draw.md` in full at head; the 5 other files in `skills/` that reference `story-spine.md`; the manifest generator (`packages/cli/scripts/gen-skills-manifest.ts` + `src/utils/skillsManifest.ts`); the `Skills: manifest in sync` job log; and the upstream project named in the description, for provenance.

## Strengths

- **Rule 4 is appended, not inserted**, and that quietly matters: `hyperframes-core/references/review-loop.md:11` cites story-spine **`§ 3`** by number and `pr-to-video/references/story-design.md:76` cites **"story-spine rule 1"**. Renumbering to slot the new rule anywhere but the end would have silently broken two cross-skill citations with nothing to catch it.
- **Rule 4 reuses rule 3's own logic rather than introducing a parallel one** — "traceable to a specific line of the source, the same way its **Why** traces to the message" (`story-spine.md:41`) hangs the new rule off the existing frame-table contract at `:33`. That is why it reads as doctrine and not as an added preference.
- **The `svg-path-draw.md:118` wording is scoped better than the creative one**: "When the path is a connector (rail, beam, underline, callout) **rather than a shape**" exempts shape draws explicitly, which is the case that would otherwise have swallowed every logo outline draw.
- **The self-checks are falsifiable**, which is the hard part of writing agent-facing doctrine. "If the prop could appear unchanged in another product's video, it didn't come from the source" (`story-spine.md:43`) is a test an agent can actually run against its own storyboard.

## Blocker

**`skills-manifest.json` is stale at this head — the committed hashes describe content that isn't in this commit.** `Skills: manifest in sync` is red ([job log](https://github.com/heygen-com/hyperframes/actions/runs/32270936501/job/96127105595)), and the description's "pre-commit (… skills-manifest …): green" claim is the opposite of what the gate reports.

I reproduced it independently rather than relaying the check — reimplementing `hashSkillBundle` (sorted full-path walk, `.DS_Store` skipped, CRLF→LF for text extensions, `rel + \0 + content + \0`, sha256 truncated to 16) and running it over `skills/` at `ee90f1fc`:

| skill | committed | actual at this head |
|---|---|---|
| `faceless-explainer` | `c4a119451b8426a9` | `912f7af169c56069` |
| `hyperframes-creative` | `4cc3aabff6aa3214` | `30ef21df37d37526` |
| `pr-to-video` | `4ffa8a0b797eba29` | `8ea0227e18ab5fc6` |
| `product-launch-video` | `20714a2f8ab8d69b` | `6ebeafd86880ebb7` |

Byte-for-byte the same four the CI job names, and `files` counts are all correct. Two details worth having:

- **`hyperframes-animation` is the one edited skill whose new hash is right** (`01b576293e04f72d`), so the regeneration did run — it just ran before the last edit rather than after it.
- **The committed values are a third state**, matching neither `main` (`1eb3772e62dd71bb`, `0068574ecd376e60`, `01f46da1e17577ea`, `d562efe00647c14b`) nor this head, so an intermediate revision of those four index lines is what got hashed. The `skills-manifest` hook does `&& git add skills-manifest.json` (`lefthook.yml:19`), so the hook isn't the gap — a commit that skipped it is the likeliest path.

Fix is the one the job prints: `bun run --cwd packages/cli gen:skills-manifest`, then commit.

Flagging it as a blocker rather than a nit for one specific reason: **`Skills: manifest in sync` is not in `main`'s required set**, so nothing stops this merging red. The manifest is the published freshness fingerprint the CLI compares installed skills against, and `--check` only ever runs on PRs — once a stale one lands on `main` there is no run on the tip to go red, and it stays stale until some later PR happens to regenerate.

## Important

**The connector rule doesn't say which lines it governs, and read literally it cuts lines the catalog prescribes.** The condition itself is checkable — "name its start anchor, its end anchor, and its job" is answerable in writing, which is what makes it usable — but the scope and the job list are both narrower than the catalog they land in. Two concrete collisions:

1. **Inside `motion-principles.md` itself.** Three bullets above the new one, the same section requires exactly what the new one cuts: `:73` "Accent elements (dividers, labels, data bars)", `:74` "**Background is not empty.** … subtle border panels, hairline rules", `:77` "**Use structural elements.** Rules, dividers, border panels." Then `:78` says "A line the frame can lose without losing meaning is decoration: cut it." A background hairline rule is precisely a line the frame can lose without losing meaning — that is the job `:74` gives it. An agent has to decide on its own whether a hairline rule is a "rail", and nothing in the section tells it.

2. **Across to the catalog, via `underline`.** `motion-graphics/categories/logo-reveal/module.md` cites `rules/svg-path-draw` at `:12` and lists `underline_sweep` as a primitive at `:13` — so the same two lines pull in a rule requiring "both endpoints must sit on real elements and the draw must do a job — reveal, route, or validate" and prescribe a stroke whose job is emphasis. Same pattern elsewhere in the catalog: `hyperframes-animation/blueprints/logo-assemble-lockup.md:42` ("an accent underline sweeps left→right beneath the `[wordmark]`") and `rules/kinetic-beat-slam.md:116` ("**Finale dressing** — stack + accent underline sweep … don't just leave the last word sitting there").

The cheapest fix is the job list, not the scope: **emphasis is a missing job, not a forbidden one.** Adding it — "reveal, route, validate, or emphasize a real element" — resolves both collisions at once, since an underline is anchored to the type it sits under by construction. If you'd rather keep the list at three, then drop `underline` from the in-scope nouns in both files and say the rule governs lines *between* elements, which is what the anchoring test is really about.

## Notes — checks that came back clean

- **Provenance: no upstream text is reused, and I measured rather than eyeballed it.** Tokenizing all 41 `.md`/`.json`/`.mjs` files of `geekjourneyx/hyperframes-motion-director` (61,357 tokens) against the four added prose blocks: **zero shared 5-grams**, and the single longest shared sequence is **four words** — "only decorates empty space" — which matches the snake_case identifier `path_only_decorates_empty_space` in `MOTION_PRIMITIVES.template.json:49`, not prose. Only four content-word 3-grams overlap in total. CJK content upstream is ~1k characters, confined to eval prompts, so the English measurement covers essentially the whole corpus. "No upstream text or code is used; all wording here is original" holds.
- Worth stating plainly for the record, since it is the part a measurement alone doesn't settle: the **checklist scaffold** is recognisably adopted. `SKILL.md:240-247` upstream reads "Before keeping any connector, name: — Start anchor … — End anchor … — Job: reveal, route, compare, focus, validate, hand off, or close … — Deletion test: what meaning … is lost if it is removed." The new bullet asks for start anchor, end anchor, and job, then applies a deletion test, with a job list that is a subset of theirs. That is idea-level reuse rather than expression — no protected sentence survives, and AGPL reaches the code and text, not the concept — so I don't read this as an attribution obligation, and crediting the project by name in the PR description is the right place for it. (Incidentally, "focus" is one of the four upstream jobs the subset drops, and it is exactly the one the emphasis-underline case above needs.)
- **The "four SKILL.md index lines" claim is complete.** Five other files in `skills/` reference `story-spine.md` and none of them enumerates the rule list: `general-video/SKILL.md:94,104` and the three workflow `references/story-design.md` files cite it by topic, and `review-loop.md:11` cites `§ 3`. Nothing else needs syncing.
- **Nothing load-bearing is hidden in the 80 deletions.** Normalising whitespace and table padding, 67 of the 80 removed lines reappear identically as additions; the 13 that actually changed are the ones the description names — 5 manifest hashes, the "three → four" intro line, and 7 index-description edits across the four SKILL.md files. The deletion count is prettier reflowing three tables after one cell got longer.
- **CI, stated as two separate claims** because they disagree: **all 8 required contexts reported** — `Semantic PR title` and `regression` green, the other six skipped by the path filter — so `mergeStateStatus: BLOCKED` here is the reviewer gate (`reviewDecision: REVIEW_REQUIRED`, no approvals), not CI. **Separately, the one gate with real signal on a skills-only change is the red one**, and it isn't in the required set.

## Nit

`svg-path-draw.md:118` is an editorial constraint sitting in a list of four mechanical ones (`fill: none`, dasharray = `getTotalLength()`, overestimate on complex paths, stagger at 70–80%). Everything else under **Critical Constraints** breaks the animation if you get it wrong; this one produces a worse-looking frame. Not worth moving on its own, but if the section ever grows a second judgment call, they'd read better grouped.

## Verdict

**Verdict:** REQUEST CHANGES
**Reasoning:** The doctrine is sound and both rules are the kind that survive contact with an agent, but the PR ships a generated artifact that contradicts its own content — verified independently, not just relayed from CI — and the check that catches it is not required, so nothing prevents it merging stale. Regenerate the manifest and I'll re-review; the connector-scope item is a wording call I'd like addressed in the same push but would not hold the PR for on its own. I have not merged anything.

— Rames Jusso

WaterrrForever:
Thanks — all three verified and addressed in 44cc2ec:

1. **Manifest (blocker)**: reproduced your finding and traced the mechanism — `lefthook.yml` runs `format` and `skills-manifest` with `parallel: true`, so the manifest hashed the four SKILL.md files **before** oxfmt renormalized their table padding in the same commit. That's why the committed hashes matched neither main nor head, and why hyperframes-animation (no table touched) was correct. Regenerated at head; a second regen is a no-op. This race will bite anyone editing tables under `skills/` — worth a separate PR to serialize those two hook steps or move the manifest regen after format.
2. **License**: added an explicit idea-level credit for the checklist shape to the PR body.
3. **Job list**: agreed — emphasis was a missing job, not a forbidden one. Added `emphasize` to both motion-principles and svg-path-draw (kept our own word rather than upstream's `focus`, preserving zero shared prose).

miga-heygen (approved):
## Re-review

**Approve.** Both of Rames's items addressed cleanly.

**1. Manifest blocker — fixed.** Regenerated at head; root cause identified (lefthook format/manifest race — they ran in parallel, so the manifest was hashed mid-commit before oxfmt renormalized the tables). Five hashes updated, second regen is a no-op.

**2. Connector scope conflict — fixed.** "emphasize" added to the job list in both `motion-principles.md` and `svg-path-draw.md`, so dividers, underline sweeps, and emphasis strokes prescribed elsewhere in the doctrine now have a valid job category. The conflict between "use structural elements" and "cut decorative lines" is now resolvable by the agent.

**Content review:**

The two absorbed ideas are well-integrated:
- **Anchored connectors** sit naturally in the composition rules (after "Use structural elements"), with a concrete constraint echo in `svg-path-draw`. The job list (reveal, route, validate, emphasize) is consistent across both locations — no SSOT drift.
- **Source-traceable visuals** land cleanly as story-spine rule 4 (appended, no numbering disruption). All four SKILL.md routing tables updated consistently. The self-check ("if the prop could appear unchanged in another product's video, it didn't come from the source") is concrete and actionable.

Table formatting changes are purely padding normalization — content identical except the added routing-table text. No new issues.

— Miga

jrusso1020 (approved):
Re-review at `afa0c14b`. Both items from my pass at `ee90f1fc` are fixed, and I verified each at the current head rather than reading the commit messages. The head moved twice while I was working — `ee90f1fc` → `44cc2ecf` → `afa0c14b` — so everything below is measured at `afa0c14b` unless it says otherwise.

## Blocker from my last pass: resolved

**`skills-manifest.json` is now correct.** I reimplemented `hashSkillBundle` from source at this head (`packages/cli/src/utils/skillsManifest.ts:164-198` — sorted full-path walk, `.DS_Store` skipped, CRLF→LF for the 14 text extensions, `rel + \0 + content + \0`, sha256 truncated to 16) and ran it over all 20 bundles:

**20/20 hashes match, 20/20 file counts match.** The 15 skills this PR doesn't touch reproducing exactly is what validates the reimplementation, so the 5 that changed are measured, not assumed.

## The conflict that appeared mid-review, and why the fix is right

Between my two passes the branch went `CONFLICTING`/`DIRTY` against `main` — `skills-manifest.json`, caused by #3305, #3306 and #3310 landing. Worth writing down because the resolution had exactly one correct form and hand-editing the conflict would have produced a wrong hash that nothing would have caught:

| skill | merge-base `634df5a5` | `main` | PR `44cc2ecf` | correct after merge |
|---|---|---|---|---|
| `faceless-explainer` | `1eb3772e62dd71bb` | `8d3795c85481b4c3` | `912f7af169c56069` | `f37cb1dba2890b79` |
| `product-launch-video` | `d562efe00647c14b` | `085243ed9167a0aa` | `6ebeafd86880ebb7` | `e89a8d11d99b4edf` |

Both sides moved these two off the same base, so the post-merge value is a **fourth** distinct value — picking either side reproduces a stale hash. I materialised the merge-result tree and regenerated before looking at the fix, which predicted `f37cb1dba2890b79` and `e89a8d11d99b4edf`; `afa0c14b` commits exactly those. The other 6 entries are one-sided moves and correctly carry `main`'s values.

Two more checks on the merge itself:

- **Nothing from `main` was dropped.** `git diff origin/main..HEAD --stat` is still exactly this PR's 8 files at +88/−80. Any main-side change the merge had reverted would have to appear in that diff, and none does.
- **Both sides survive in the two files both branches touched.** `faceless-explainer/SKILL.md` and `product-launch-video/SKILL.md` each carry `main`'s `preview --background` edit *and* this PR's `source-traceable visuals` additions. No conflict markers anywhere in `skills/` or the manifest.

## Important item from my last pass: resolved, at both sites

`emphasize` is added to the job list in **both** files that carry the rule — `svg-path-draw.md:118` ("reveal, route, validate, or emphasize") and `motion-principles.md:78` ("revealing, routing, validating, or emphasizing"). Grepping the tree for the old three-job phrasing returns nothing, and those two are the only sites, so this is complete rather than one-of-two. That resolves the `underline_sweep` / accent-underline collision with the catalog cleanly: an underline is anchored to the type it sits under by construction, so licensing emphasis as a job makes the prescribed stroke legal instead of contradicted.

## Correcting my own previous review

I wrote that `gen:skills-manifest --check` "only ever runs on PRs" and that a stale manifest on `main` would sit undetected until some later PR regenerated it. **That's wrong.** `.github/workflows/ci.yml:12-22` declares `push: branches: [main]` alongside the `pull_request` trigger, and the `skills` path filter (`:69-77`) includes both `skills/**` and `skills-manifest.json` — so a stale manifest landing on `main` does turn `Skills: manifest in sync` red on main's own tip. The blocker itself was real and is fixed; my framing of the consequence overstated it, and the gate is better than I gave it credit for. It remains outside the 8 required contexts, so it still can't *block* a merge — but "not required" and "invisible" are different claims and I collapsed them.

## Note — the root cause is a live repo issue, not a one-off

The description's account beats my guess and checks out at source. I had said a skipped hook was the likeliest path; the real mechanism is the race the description names. `lefthook.yml:1-2` sets `parallel: true` on `pre-commit`, and two of those commands touch the same bytes: `format` globs `md` and runs `oxfmt … && git add -f {staged_files}`, while `skills-manifest` globs `skills/**` and regenerates from disk. Run concurrently, the generator can hash a skill's markdown *before* oxfmt reflows it — which is exactly the third state I measured last time, matching neither `main` nor the head.

This PR was the worst case for it (its 80 deletions are table reflow, so format rewrote the very tables the generator had already read), but nothing about it is specific to this PR — any commit touching skill markdown can hit it, and the failure is a committed artifact that disagrees with its own content. There are no ordering keys in `lefthook.yml` today, so this is a repo-level fix rather than something to do here: the manifest step needs to run after formatting rather than beside it. Worth its own change.

## Nits

- **The description now under-describes the code.** The "What" section still lists the job list as "reveal, route, or validate" and says the manifest was "regenerated by the pre-commit hook" — three jobs, and a hook that demonstrably didn't produce this manifest. The Checks section explains the race correctly, so the two halves disagree with each other. The title is accurate, which is what reaches history on a squash, so this is cosmetic — but the next reader takes the body as the summary.
- **`motion-principles.md:74`/`:77` vs `:78` is narrowed, not closed.** Adding emphasis settles the underline case. It doesn't settle whether `:78` governs the *structural* lines the same section prescribes two bullets up — `:74` asks for "hairline rules" as background treatment and `:77` for "Rules, dividers, border panels", and a background hairline is precisely "a line the frame can lose without losing meaning", which `:78` says to cut. Whether a hairline rule is a "rail" is left to the agent. Same call as last time: worth a clause someday, not worth holding this PR for.

## CI

Two separate statements, because they answer different questions:

- **The required set is satisfied.** All 8 required contexts report at `afa0c14b` (`comm -23` of the ruleset's required list against the latest-run-per-name check set is empty). `Semantic PR title` and `regression` are green; the other six are path-skipped after `Detect changes` classified this as skills-only, and the PR is not a draft, so the skips are the filter and not absence of signal.
- **The gates with real signal on a skills change are green**, which is the part the required set doesn't cover: `Skills: manifest in sync` ✅ (independently corroborating my own reproduction), `Test: skills` ✅, `Skills: project-native lint + mirror` ✅, `Format` ✅. Only `Analyze (javascript-typescript)` is still running; it isn't required and CodeQL's rollup already reported. Nothing is red.

I took the description's `lint:skills` "32 files" count on the green CI job rather than recounting it locally.

## Prior review on this PR

@miga-heygen approved at `44cc2ecf`, which is one commit behind the current head; the merge commit changed 8 manifest entries and both of the SKILL.md files since. `main` here sets `require_last_push_approval: true` with `dismiss_stale_reviews_on_push: false`, so that approval survives the push but doesn't satisfy the gate at `afa0c14b` — this PR needs an approval pinned to the current head regardless of anything I say. Flagging it as mechanics, not as a criticism of the review.

**Verdict:** APPROVE
**Reasoning:** Both items I raised are fixed and verified at `afa0c14b` — 20/20 manifest hashes reproduced independently, `emphasize` added at both of the two sites, and the mid-review merge conflict resolved by regenerating at the merged tree rather than picking a side, landing on precisely the values a fresh regeneration produces. Nothing outstanding is more than a nit. This approval covers `afa0c14b` only; the repo will require a fresh one if anything else is pushed. **I have not merged and won't without an explicit go** — and since the manifest is base-relative, if another PR touching `skills/` lands first this will need one more regeneration before it goes in.

— Rames Jusso

PR #3357, opened 2026-08-19
fix(studio): stop unrelated preview setup from disabling playback shortcuts

miguel-heygen:
## What

Playback keyboard shortcuts (space, J/K/L, arrows) could silently stop working for a composition while the transport buttons kept working on that same composition.

The preview's key listeners were attached third inside a `try` with an empty `catch {}`:

```ts
normalizePreviewViewport(doc, iframeWin);
autoHealMissingCompositionIds(doc);
attachIframeShortcutListeners();
```

Both calls above it touch the preview document, and a composition can make either throw. When one does, the shortcuts are never attached and the error is discarded.

Each step now runs independently, keyboard first, with failures reported instead of swallowed.

## Why

Reported as "space doesn't work" on one composition, where clicking Play still worked.

That split is the diagnosis. The transport buttons are React handlers on the parent document, so they are unaffected. The shortcuts are listeners attached to the preview iframe during load. Anything that stops that attachment takes the keyboard with it and leaves the buttons alone — which presents as a broken key rather than a broken load, with nothing in the console.

Ordering matters as much as isolation here: input handling should survive the rest of the preview setup having a bad day, not the other way round.

## How

Extracted the sequence into `runPreviewSetupSteps`, which runs each step in its own `try` and reports failures through an `onError` callback. Shortcuts go first.

Extracting it is also what makes the behaviour testable — the hook itself has too many dependencies to drive directly, which is why this had no coverage before.

The empty `catch {}` around the wider block is left as-is; narrowing that is a bigger change than this fix warrants.

## Test plan

Three of the five new tests fail against the old sequence and pass against the new one:

```
× THE BUG: a throwing viewport pass used to cost the user every shortcut
× attaches shortcuts before anything that can break them
× runs every remaining step after one throws
```

- `bunx vitest run src/player` — 1310 passed, 99 files
- `bunx tsc --noEmit` — clean
- oxlint + oxfmt clean

**Not claimed:** the exact call that throws on the reporting composition was not isolated. The symptom was reproduced and the mechanism is proven by the tests; this removes the class rather than one instance of it. If the underlying throw is still there, it will now show up as a warning naming the step instead of vanishing.

### DIFF
--- a/packages/studio/src/player/hooks/usePlaybackKeyboard.ts
+++ b/packages/studio/src/player/hooks/usePlaybackKeyboard.ts
@@ -9,6 +9,7 @@
 import { useRef, useCallback } from "react";
 import { useCaptionStore } from "../../captions/store";
 import { shouldIgnorePlaybackShortcutEvent, SHUTTLE_SPEEDS } from "../lib/playbackShortcuts";
+import { describeTarget, traceShortcut } from "../lib/shortcutDebug";
 import { canvasNudgeKeysClaimed } from "../../utils/canvasNudgeGate";
 import { usePlayerStore } from "../store/playerStore";
 import { stepFrameTime, STUDIO_PREVIEW_FPS } from "../lib/time";
@@ -81,7 +82,14 @@ export function usePlaybackKeyboard({
 
   const handlePlaybackKeyDown = useCallback(
     (e: KeyboardEvent) => {
-      if (e.defaultPrevented) return;
+      if (e.defaultPrevented) {
+        traceShortcut("key already claimed (defaultPrevented)", {
+          code: e.code,
+          target: describeTarget(e.target),
+        });
+        return;
+      }
+      traceShortcut("key seen", { code: e.code, target: describeTarget(e.target) });
       const captionState = useCaptionStore.getState();
       if (
         shouldIgnorePlaybackShortcutEvent(e, {
@@ -89,12 +97,19 @@ export function usePlaybackKeyboard({
           selectedCaptionSegmentCount: captionState.selectedSegmentIds.size,
         })
       ) {
+        traceShortcut("key ignored by guard", {
+          code: e.code,
+          target: describeTarget(e.target),
+          modifiers: { meta: e.metaKey, ctrl: e.ctrlKey, alt: e.altKey },
+          captionEditMode: captionState.isEditMode,
+        });
         return;
       }
       const key = e.key.toLowerCase();
       pressedKeysRef.current.add(key);
       if (e.code === "Space") {
         e.preventDefault();
+        traceShortcut("space -> togglePlay");
         togglePlay();
         return;
       }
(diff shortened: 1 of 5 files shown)


miguel-heygen:
Added opt-in tracing for the preview shortcut path (second commit), so the next occurrence produces evidence instead of an empty console.

Silent by default — this runs on every keystroke. Turn it on with either:

```js
window.__hfDebugShortcuts = true
```

or add `?debugShortcuts` anywhere in the URL, including after the route hash.

Every line is prefixed `[studio:shortcuts]` so it can be filtered and pasted back whole. What gets traced:

| line | tells you |
|---|---|
| `preview setup: <step> ok` / `THREW` | which load step failed, and with what |
| `attached to preview` | the listeners reached the preview `window` / `document` |
| `attach ABORTED: preview unreachable` | reading across the boundary threw |
| `attach ABORTED: no preview window or document` | called before the iframe had a document |
| `key seen` | the key reached the handler at all |
| `key already claimed (defaultPrevented)` | something else consumed it first |
| `key ignored by guard` | the guard rejected it, with target and modifiers |
| `space -> togglePlay` | it got all the way through |

That separates the three failures this bug hides behind: a key never seen, a key seen and ignored, or a key handled while playback fails for an unrelated reason.

Targets are described (`canvas#stage.preview`) rather than logged as nodes, so a pasted log stays readable and carries no page content.

Six tests cover the gate itself — a debug facility that turns itself on is worse than none. 1316 player tests pass.

miguel-heygen:
## Correction: this does not fix the report that prompted it

The tracing added in the second commit was run against the composition that failed, and it clears the shortcut path completely:

```
[studio:shortcuts] attached to preview {window: true, document: true}
[studio:shortcuts] preview setup: shortcuts ok
[studio:shortcuts] preview setup: viewport ok
[studio:shortcuts] preview setup: composition-ids ok
[studio:shortcuts] key seen {code: "Space", ...}
[studio:shortcuts] space -> togglePlay
```

No step threw, the listeners attached, space was seen, and `togglePlay` was called — three times over three presses. Nothing here is broken.

The actual cause was in the composition, not Studio. It replaced `requestAnimationFrame` inside the preview document to drive a WebGL scene from the playhead, and GSAP's ticker runs on `requestAnimationFrame` as well. That deadlocks: the timeline needs the ticker to advance, and the replacement clock only flushed queued callbacks when the timeline advanced. `togglePlay` fired and nothing moved, which presented as "space does nothing" while the transport button appeared to work.

## What this PR is still worth

The ordering defect is real and independent of that. On `main`, `attachIframeShortcutListeners()` runs third inside a `try` with an empty `catch {}`, behind two DOM passes that can throw. When one does, every playback shortcut is lost silently while the transport buttons keep working. The three tests here fail against that sequence and pass against the fixed one.

So: a real latent bug with real coverage, plus the tracing that just disproved my own diagnosis. It is no longer presented as fixing the reported symptom.

Reviewers may reasonably want the tracing without the fix, or vice versa — happy to split.

miguel-heygen:
Closing: this did not fix the report that prompted it, and the tracing added here is what proved that. Space reached Studio, passed the guard, and called togglePlay on every press — the shortcut path was never broken. The cause was in the composition, which replaced requestAnimationFrame in the preview while GSAP's ticker runs on it too, deadlocking the timeline.

The ordering defect it found is real and still on main: attachIframeShortcutListeners() runs third inside a try with an empty catch {}, so an unrelated DOM failure silently costs every playback shortcut while the transport buttons keep working. Leaving it unfixed for now rather than landing a change nobody asked for; the branch keeps the three regression tests if it is worth revisiting.

PR #3363, opened 2026-08-20
feat(audio): effect chains, automation, voiceover carve and submix buses

vanceingalls:
Audio in the studio becomes something you can mix: an effect chain per track, automation envelopes on volume and on any effect parameter, a voiceover carve that ducks a music bed only in the bands the voice occupies, and `<hf-audio-group>` submix buses that carry one chain, fader and automation clock for several tracks at once. Preview and render share one implementation of every effect, so what you hear while scrubbing is what gets baked.

103 commits, 188 files, +14,065 / −1,757. Rebased onto `main` (`d09145faa`).

## What's in it

**Core** — `audioGroups.ts` resolves `<hf-audio-group>` membership; the preview transport gains a per-group bus (built lazily, chain + fader + automation scheduled once against composition time); the FX node codec is split into named readers.

**Engine** — grouped audio renders through a summed, FX-processed bus; a group's `data-hidden` drops every member from the mix; `applyAudioFxChain`'s CDP transfer is four named functions instead of one 180-line block.

**Studio** — group rows in the timeline with their own automation lanes, an FX rack reachable from a lane label (click a lane's name and the rack opens on that effect and scrolls to it), the carve module, presets, and the audition transport.

**Skill** — `/hyperframes-audio` now documents the submix bus. It previously mentioned groups only as carve targets and its example grouped three clips with no group element at all, so an agent following it could not emit a group chain, fader or automation.

## Things worth a reviewer's attention

**Two invariants this work established the hard way.** The timeline keys elements as `sourceFile#domId`; the property panel, DOM and runtime use the bare dom id. `runtimeAudioId()` converts, and getting it wrong fails *silently* — the feature simply never matches. And a carve owns its own lanes: it recompiles them on every re-analysis, so they are shown **read-only rather than hidden** (hiding them was a bug — a music bed whose every lane was the carve's showed no automation at all), their remove buttons are withheld, and the rack's `handBuilt` list filters carve nodes out.

**Four independent carve bugs, each sufficient alone** (`5d8fbf5d3`). The most interesting: `couldBeCarveSource()` had existed since the carve was written, with a doc comment saying music and sfx are out — exported, tested, and **called from nowhere**. Nothing ever asked whether a track could be the bed. A voiceover ended up carving against its own group.

**Layout grouping and "Hide all" refused for audio** (`0ccda5f6c`, `927aeea52`). The layout grouper wraps members in a positioned `<div>` at their bounding box; an `<audio>` has no box, and running it produced `width: 0px; height: 0px` plus a real file write. "Hide all" was worse — one-way: `data-hidden` on audio is not visibility, preview silences it and the render drops it from the mix, and the timeline withheld the eye *even while hidden*, so four muted SFX clips had no control anywhere to restore them. Both refusals live in the handler, not the button, because the `G` shortcut routes through the same path and no hidden button can gate a keystroke.

**The media-element transport had to be re-integrated after the rebase** (`e1271b225`). `main` added a pitch-preserving HTMLMediaElement transport that the runtime now tries *first* for audio, falling back to the decoded-buffer path. Its gain connected straight to master, which would have left every grouped track bypassing the bus this PR exists to add. It routes through `resolveDestination` now.

## Fallow, filesize, and the `--no-verify` history

The audit gate is **clean** — 0 complexity findings, 0 dead code, duplication warn-only, `exit 0` — and no file in the stack is over the 600-line cap. Earlier commits on this branch carry `--no-verify` with the reason in each message; the causes are all fixed now:

- Nine gated complexity findings were cleared by refactoring, not suppression: `processTimelineMessage` 22/24 → 8/6, `initializeAdapter` 30/27 → 6/3, `applyAudioFxChain` 23/23 → 11/7, `FxCarveModule` 25/45 → 5/2, the two `audioFx.ts` node codecs 18/17 and 10/9 → 3/1 and 4/1, plus `audioFxSummary`, the FxSection reveal effect, and `onMessage`.
- Nine studio files were over the 600-line cap, all pushed over by this branch. Every one is under now: `TimelineTrackHeader.tsx` 763 → 450, `TimelineAutomationLane.tsx` 683 → 499, `propertyPanelFxSection.tsx` 616 → 598, `useDomEditSession.ts` 615 → 596, `useTimelineEditing.ts` 614 → 600, `TimelineLanes.tsx` 610 → 596, `PropertyPanelFlat.tsx` 605 → 597, `playerStore.ts` 604 → 594, plus `useFxCarve.ts` earlier in the stack.

  Extracting *in place* made `PropertyPanelFlat` grow (605 → 616) — a signature plus a doc comment costs more than an inline arrow saves. Only moving to a sibling module actually removes lines.

**One thing worth knowing before you touch these files:** fallow fingerprints a finding by line position, so inserting a helper *above* a complex function re-flags that function's inherited complexity as new. Two of the nine findings were byte-identical to `main` and had been re-flagged purely by line shift. Helpers therefore go below.

## Verification

- Full suites green: core 2,479 · studio 4,422 · engine 1,616 · cli 2,783 · lint 528 · parsers 929 · studio-server 446 · sdk 541 · player 338. `oxlint`, `oxfmt --check` and `tsc --noEmit` clean across every package.
- Twenty failures in the full parallel run all pass in isolation (`Timeline.virtualization`, `audioMixer.grouping`, `propertyPanelInputCoverage`, cli telemetry/orphan-cleanup, producer `audioPadTrim`); this is the known vitest parallel-fork flakiness, present identically before the rebase.
- **Render verified end to end** through `bun run studio` and the studio's own `/render` endpoint: the audio-real fixture (9 audio tracks, two groups, a 40s bed) rendered 1,200 frames to a 1.28 MB mp4 with audio.

## Review findings

A max-effort review of this PR raised 15 findings plus a confirmed tail. **All 15 are fixed**, each with a test verified to fail on a revert of its own fix. The severity key was the canary map: `audio-groups` gates only the timeline group rows, so every core/engine path shipped ungated.

| # | What was wrong | Commit |
|---|---|---|
| 1 | Two inlined instances of one sub-composition collapsed into ONE bus (last wins). With only instance B muted, **both** instances vanished from the export | `d381f5838` |
| 2 | Preview clamped a bus fader at 1.0 while the render clamps at 3.98 — an over-unity bus auditioned up to 12 dB quieter than it exported | `89de9d3fb` |
| 3 | The new lint rule errored on the studio's own cross-file output | `23149cad9` |
| 4 | Group sub-mix had no try/catch: a bad chain on a bus threw past the `failures[]` contract and leaked the temp dir | `9c9938a04` |
| 5 | A `data-hidden` toggle stopped and restarted the whole mix for 100% of users, for a feature at 0% | `093eac260` |
| 6 | `closest("[data-hidden]")` cannot see a muted BUS — membership is on the member, so the render dropped what preview played | `89de9d3fb` |
| 7 | The bus element was resolved by untyped `getElementById` and frozen for the session | `89de9d3fb` |
| 8 | A hidden group had no way back — no control anywhere, only hand-editing HTML | `c62b65297` |
| 9 | The far-end carve guard was defeated by the group-first shape the lint rule asks for, producing reciprocal carves | `be325e595` |
| 10 | Once the bed joined its own carve's group it was measured as one of its own voices | `be325e595` |
| 11 | An await opened a double-fire window: two concurrent `data-fx-carve` writes and two analyses | `be325e595` |
| 12 | The lane-label reveal was dead in **both** halves — each hid the other | `c62b65297`, `4417f68ce` |
| 13 | An unescapable node id took the property panel down instead of failing quietly | `c62b65297` |
| 14 | The group mirror was a no-op for a group declared in a sub-composition | `21a17a495` |
| 15 | The save unwind restored the value it was meant to undo, so a failed write read as success | `21a17a495` |

Tail: label-column canary gate, keyboard/header expandability mismatch, sub-comp row colliding with a group anchor, audition telemetry counted as applies, stale signal-path memo (`5eb974c8a`); group lane labels drifting off their curves, duplicated duration helper, silent context-name collisions (`ed1dfda6e`); dirty-gated mute sweep (`9f99473d9`); router rows (`315ecbb40`).

**Three items examined and rejected, with reasons:**
- The `amix` normalize fallback was called dead regex. ffmpeg 8.1.1 emits `Option not found`, which the pattern matches — verified against the binary, now commented.
- Matching only `audio[data-audio-group]` in lint was called inconsistent with the studio. It agrees with **core**, which documents "a `data-audio-group` on a `<video>` is ignored" — so such a bus really has no members.
- Withholding "Hide all" from a mixed selection containing audio was called over-broad. Hiding only the visual members is the act-on-a-subset pattern this branch refuses elsewhere, and for audio `data-hidden` is mute.

**Two deliberately deferred, both named in commits:** parallelising the group sub-mix (it reorders `tracks`, which orders the ffmpeg filtergraph, in a pipeline with bit-exactness tests — low gain, real risk), and the twin-hook dedup of `timelineAudioGroupVolume` / `timelineElementFxAttribute`.

## Known divergence

**The live preview still shares one bus between two instances** of the same sub-composition. The export was the audible bug — a muted instance silencing another's audio — and is fixed; preview needs runtime subtree resolution, which is a separate change.

## Deliberately left open

- `couldBeCarveSource` is still dead code — the source-side rule is duplicated inline in `useFxCarve.ts` via `classifyAudioName`. Either call the predicate or delete it.
- Carve lanes are read-only but not dimmed (`opacity: 0.55` is keyed to the unselected state, not to `readOnly`).
- `audioGroupVolume` / `audioGroupHidden` are dead end-to-end after the mute/solo and bus-strip removals (writers and one read, no renderer). Left in place rather than ripped out late in the branch.
- The leftover wrapper `<div>` in `PlayerControls` is redundant — `PreviewPane` supplies its own flex wrapper.


vanceingalls:
## Closing this — its features shipped, its fix log did not

Audited 2026-08-22 while chasing three reported Studio regressions. Recording the outcome here because the useful information is the *diff* between this PR and `main`, and that shouldn't disappear with the PR.

### What happened

The five feature commits on this branch were **re-split into small PRs and merged** the day after this PR was opened. This PR was never reviewed (0 reviews, 0 comments) and the ~42 follow-up **fix** commits behind those features never landed.

The giveaway is the same work existing on both sides under *different* SHAs — re-authored, not merged:

| this branch | `main` | PR |
| --- | --- | --- |
| `e96631162` | `43c4e6935` | #3274 |
| `dd3212d11` | `acfa7c55a` | #3286 |
| `e1c9b5094` | `5fd84c395` | #3290 |
| `301cb8b51` | `0d26072e6` | #3291 |
| `1d5405e3e` | `6a92d2140` | #3292 |

The tip reappeared reworded too: `952401d12` "**ship** the audio FX, group and mute features to everyone" → `0e9a4f371` #3401 "**open** …".

### Why this is closed rather than merged

Two independent blockers:

1. **It would remove features `main` shipped.** After the split this branch kept evolving and deliberately reverted two of them — `2199f55c3` removes mute/solo (shipped as #3291) and `1b16a4c87` removes the group volume slider + level meter (shipped as #3290). Both are `!` breaking changes.
2. **It no longer merges.** `wa-25-review-fixes` → current `main` produces **55 conflicts**, including `add/add` on `packages/studio/src/components/editor/TimelineFxPopover.tsx` and `packages/core/src/audioGroups.ts` — both sides created those files independently during the split.

So the right shape is surgical re-implementation on `main`, not a merge.

### Already restored

The three regressions that were actually reported, re-implemented against `main`'s current files:

- #3413 — popover caps to its gap and the preset list scrolls (from `a3ab011e2`)
- #3414 — the three portaled timeline surfaces move to the `z-[200]` tier (from `26c467c79`)
- #3415 — group caret: mono, a size up, swapped not rotated (from `6b4b91eb8`)

Each carries guard tests, which the originals shipped **without** — that is precisely why they reverted invisibly. All mutation-verified (revert the fix, the guard fails).

One correction worth carrying forward: `26c467c79`'s message blames "the ruler's sticky header at `z-70`". `z-70` is not a literal class in either tree — `main`'s ruler is `sticky top-0 flex` with no z-index — so #3414 is scoped as tier consistency rather than a reproduced stacking failure. A stranded commit's rationale describes *its own* tree.

### Still stranded — triage queue (39)

**The branch is not deleted.** `origin/wa-25-review-fixes` @ `952401d12` survives this close; anything below is recoverable with `git show <sha>`.

Caveat before anyone treats this as a loss list: these came from a content probe (6 added lines per commit, grepped against `main`'s tree). A commit later reworked *on this branch* also scores absent — this branch carries fix-the-fix chains. **Verify at code level before restoring.**

Highest-value by subject:

- `77308f4f3` fix(studio): stop HMR crashing the studio with a phantom missing provider
- `373884ddc` feat(studio): wrap FX parameter names instead of truncating them
- `cd5cc0060` feat(studio): wrap gutter names instead of truncating them
- `b6bac054b` feat(studio): show the full name on hover for truncated rack labels
- `6de75fd4b` fix(studio): make sub-composition audio groups actually work
- `acfd425ec` fix(studio): keep the carve's own lanes out of the timeline
- `fa0f12a57` fix(studio): show a carve's own lanes, read-only, instead of hiding them
- `ff4965037` / `6056ec746` / `7393095be` — three review-found defect batches

Remainder: `94ecaa3ca` `0fe9759af` `a44ebd757` `730ea1eb0` `3e84832e7` `db1b81357` `8eebed0c0` `b91247d88` `a014e00a4` `5614b2623` `a05f16c5e` `f26f48b3a` `624166f38` `ff55adfcb` `82cfb361e` `1524f4e57` `db24bb000` `f87ef04dd` `f565cf85e` `dd45bcf71` `1f4bccd64` `926496f6b` `6f0cbb039` `c224706ba` `77dd736fe` `86420cdac` `25d7af8a5` `ce6426ddc` `a8d27820b`

Deliberately excluded from the queue: `2199f55c3` and `1b16a4c87` (the two breaking removals above) — do not restore those without a product decision.

### Process note

A UI fix with no test does not survive a re-split, and nothing notices. All three restored fixes were styling/geometry changes with no assertion behind them. Worth treating "re-split a large PR" as an operation that needs an explicit reconciliation pass over the commits left behind.

PR #3381, opened 2026-08-21
fix(engine): hide ffmpeg console windows on Windows

### LINKED ISSUE #3379: [Windows] ffmpeg subprocess windows flash/pop up during render, check, and snapshot (missing windowsHide)
### Describe the bug

On Windows, every `ffmpeg` invocation from the CLI spawns a visible console window that appears on-screen and stays open for the life of that process, then closes when it exits. Because `render` (and `check`/`snapshot`) shell out to `ffmpeg` many times across parallel workers — frame extraction, encode, concat, GPU-encoder probing, etc. — this produces a burst of console windows opening and closing on the desktop over the course of a run. During one `render` with 6 capture workers I saw ~11 windows open at once, tapering off as workers finished.

Root cause: `packages/engine/src/utils/runFfmpeg.ts` calls `spawn(getFfmpegBinary(), args)` without `windowsHide: true`. Since `child_process.spawn()` defaults `windowsHide` to `false`, and `ffmpeg.exe` is a console-subsystem executable, every spawn opens a new console window on Windows. This is the shared ffmpeg runner used throughout the render pipeline, so it's invoked dozens of times per render.

The same pattern appears in `packages/engine/src/utils/gpuEncoder.ts` (`detectGpuEncoder`), which also calls `spawn(getFfmpegBinary(), ["-encoders"], { stdio: [...] })` without `windowsHide`.

Suggested fix — add `windowsHide: true` at both call sites:
  const ffmpeg = spawn(getFfmpegBinary(), args, { windowsHide: true });
`windowsHide` is a no-op on macOS/Linux, so it's safe to apply unconditionally.

### Link to reproduction

github.com/jhouston128/hyperframes-windows-popup-repro

### Steps to reproduce

1. Clone the linked repo (a blank project from `npx hyperframes init --example blank`)
2. On Windows, run: npx hyperframes render .
3. Watch the desktop/taskbar during the render — console windows flash open and close repeatedly as ffmpeg is invoked for capture/encode/assemble

### Expected behavior

Rendering completes without any visible console windows appearing on screen.

### Actual behavior

Multiple ffmpeg console windows pop up on the desktop during the render and close one by one as each ffmpeg process finishes. Cosmetic only — the render output itself is correct — but it's a jarring, unexpected UX on Windows.

### Environment

```shell
✓ Version          0.8.5 (latest)
✓ Node.js          v22.23.2 (win32 x64)
✓ FFmpeg           ffmpeg 9.0-full_build-www.gyan.dev
✓ Chrome           chrome-headless-shell win64-152.0.7928.2
✗ Docker           Not found
```

### Additional context

_No response_

miguel-heygen:
## What

Pass `windowsHide: true` at every production ffmpeg/ffprobe spawn site.

Closes #3379.

## Why

`ffmpeg` and `ffprobe` are console-subsystem binaries, and Node's `child_process.spawn`
defaults `windowsHide` to `false`. Every spawn therefore opens a visible console window on
Windows. A render shells out dozens of times across parallel workers, so the reporter saw a
burst of windows appear and disappear across their desktop, ~11 at once with 6 capture
workers.

Cosmetic, but it is the first impression of a render on Windows.

## How

The report named two call sites. **Seven production sites share the cause**, so all seven
are fixed rather than the two mentioned:

| file | site |
|---|---|
| `engine/src/utils/runFfmpeg.ts` | the shared runner, used across the pipeline |
| `engine/src/utils/gpuEncoder.ts` | `-encoders` probe |
| `engine/src/utils/gpuEncoder.ts` | per-encoder probe |
| `engine/src/utils/ffprobe.ts` | metadata probe |
| `engine/src/services/streamingEncoder.ts` | streaming encode |
| `producer/src/services/audioExtractor.ts` | audio extraction |
| `producer/src/services/distributed/shared.ts` | `ffmpeg -version` check |

`windowsHide` is a no-op on macOS and Linux, so it is applied unconditionally.

Left alone deliberately: `parity-harness.ts`, `regression-harness.ts` and
`mediaTypeTestFixtures.ts` all spawn ffmpeg too, but they are dev-only tooling that never
runs on a user's desktop. Changing them would widen the diff without changing the reported
behaviour.

## Test plan

### Regression test

`runFfmpeg.windowsHide.test.ts` mocks `child_process` and asserts on the options object
actually handed to `spawn`:

```ts
const options = spawnMock.mock.calls[0]?.[2] as { windowsHide?: boolean } | undefined;
expect(options?.windowsHide).toBe(true);
```

Asserted on behaviour rather than on source text, so a future call site that drops the flag
is caught by what it does, not by how it is written.

Verified it fails without the fix:

```
$ # windowsHide removed from runFfmpeg
FAIL  runFfmpeg.windowsHide.test.ts > hides the console window so Windows renders do not flash terminals
$ # restored
Tests  1 passed (1)
```

### Other checks

- `vitest packages/engine/src/utils/` — 651 passed, 20 files
- `oxlint` + `oxfmt` on all 7 touched files — clean

## Before / after

I do not have a Windows machine, so I cannot capture the desktop screenshots this one
really wants. The observable change is at the spawn boundary, and that is what the test
pins:

```
before   spawn(getFfmpegBinary(), args)                          -> console window per call
after    spawn(getFfmpegBinary(), args, { windowsHide: true })   -> no window
```

The reporter's repro repo is linked in #3379 and is the right way to confirm the desktop
behaviour on a real Windows box before merge.

### DIFF
--- a/packages/engine/src/services/streamingEncoder.ts
+++ b/packages/engine/src/services/streamingEncoder.ts
@@ -457,6 +457,8 @@ export async function spawnStreamingEncoder(
 
   const ffmpeg: ChildProcess = spawn(getFfmpegBinary(), args, {
     stdio: ["pipe", "pipe", "pipe"],
+    // See runFfmpeg.ts: keeps a console window off the user's desktop on Windows.
+    windowsHide: true,
   });
   trackChildProcess(ffmpeg);
 
--- a/packages/engine/src/utils/ffprobe.ts
+++ b/packages/engine/src/utils/ffprobe.ts
@@ -69,6 +69,8 @@ async function runFfprobe(
     // Nothing is ever written to the child's stdin; leaving it as a pipe is
     // what lets a stdin-reading invocation block indefinitely.
     stdio: ["ignore", "pipe", "pipe"],
+    // See runFfmpeg.ts: keeps a console window off the user's desktop on Windows.
+    windowsHide: true,
   });
   trackChildProcess(proc);
   // Decoded through StringDecoder rather than per-chunk toString(): a
--- a/packages/engine/src/utils/gpuEncoder.ts
+++ b/packages/engine/src/utils/gpuEncoder.ts
@@ -64,6 +64,8 @@ export async function selectUsableGpuEncoder(
 export async function detectGpuEncoder(): Promise<GpuEncoder> {
   const ffmpeg = spawn(getFfmpegBinary(), ["-encoders"], {
     stdio: ["pipe", "pipe", "pipe"],
+    // See runFfmpeg.ts: keeps a console window off the user's desktop on Windows.
+    windowsHide: true,
   });
   trackChildProcess(ffmpeg);
   let stdout = "";
@@ -146,6 +148,7 @@ export function getProbeArgs(encoder: ConcreteGpuEncoder): string[] {
 async function canUseGpuEncoder(encoder: ConcreteGpuEncoder): Promise<boolean> {
   const ffmpeg = spawn(getFfmpegBinary(), getProbeArgs(encoder), {
     stdio: ["ignore", "ignore", "pipe"],
+    windowsHide: true,
   });
   trackChildProcess(ffmpeg);
   const outcome = await new ManagedChildProcess(ffmpeg, {
--- a/packages/engine/src/utils/runFfmpeg.ts
+++ b/packages/engine/src/utils/runFfmpeg.ts
@@ -92,7 +92,11 @@ export function formatFfmpegError(
 
 export async function runFfmpeg(args: string[], opts?: RunFfmpegOptions): Promise<RunFfmpegResult> {
   const timeout = opts?.timeout ?? DEFAULT_TIMEOUT;
-  const ffmpeg = spawn(getFfmpegBinary(), args);
+  // windowsHide: ffmpeg/ffprobe are console-subsystem binaries, so without
+  // this Node opens a visible console window per spawn on Windows. A render
+  // shells out dozens of times across parallel workers, which flashes a burst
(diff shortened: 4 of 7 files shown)


miguel-heygen:
Scope note, from a second Windows report that landed after this PR was opened.

This PR closes the **ffmpeg/ffprobe** half. It does not close the browser half, and I want
that on the record so the PR is not read as fixing the whole reported symptom.

> one Windows Terminal window can open per cached chrome-headless-shell worker on Windows 11
> ... visible console windows use the cached HyperFrames chrome-headless-shell path as their
> title. OUTCOME: render succeeds but interrupts the desktop.

Those windows come from Puppeteer's own launcher, not from any spawn in this repo.
`@puppeteer/browsers@3.0.6` spawns the browser as:

```js
this.#browserProcess = childProcess.spawn(this.#executablePath, this.#args, {
  detached: opts.detached,
  env,
  stdio,
});
```

No `windowsHide`, and the supported `opts` are only `pipe`, `dumpio`, `handleSIGINT`,
`handleSIGTERM`, `handleSIGHUP`, `detached` and `env` — there is no pass-through we could
set from `ppt.launch()`. Our launch sites in `browserManager.ts` and `hdrCapture.ts` have
nowhere to put the flag.

So the remaining half needs one of:

1. an upstream change to `@puppeteer/browsers` to set `windowsHide: true` (or accept it as
   an option), or
2. documenting the workaround the reporter found — point `HYPERFRAMES_BROWSER_PATH` at an
   installed GUI Chrome, which is not a console-subsystem binary.

Worth deciding which before closing the original issue, since a Windows user with this PR
merged will still see browser windows and reasonably think nothing changed.

PR #2017, opened 2026-07-07
feat(studio): timeline revamp with active-clip highlighting and hide controls

miguel-heygen:
## What

Brief description of the change.

## Why

Why is this change needed?

## How

How was this implemented? Any notable design decisions?

## Test plan

How was this tested?

- [ ] Unit tests added/updated
- [ ] Manual testing performed
- [ ] Documentation updated (if applicable)

PR #3382, opened 2026-08-21
fix(studio): keep subcomposition timelines open during playback

miguel-heygen:
## What

Keep the inline timeline for an active subcomposition visible in the master timeline while the preview is playing.

## Why

In the master view, starting playback currently collapses the active nested composition's child rows. The expansion target is resolved from the last committed playhead time in the store. The RAF loop keeps live playhead time out of React and Zustand during playback, so the inline rows remain stable while playing instead of being recomputed or collapsed on transport changes.

## How

Resolve the expansion target from the selected element first, then from the composition active at the store's currentTime. Playback state no longer suppresses active auto-expansion, and the expensive row build remains keyed by expansion identities rather than raw time.

## Test plan

Focused tests:

- `bun run --cwd packages/studio test src/player/hooks/useExpandedTimelineElements.test.ts src/player/components/Timeline.test.ts src/player/components/TimelineLanes.test.tsx`
- `bun run --cwd packages/studio typecheck`
- `bunx oxlint packages/studio/src/player/hooks/useExpandedTimelineElements.ts packages/studio/src/player/hooks/useExpandedTimelineElements.test.ts`
- `bunx oxfmt --check packages/studio/src/player/hooks/useExpandedTimelineElements.ts packages/studio/src/player/hooks/useExpandedTimelineElements.test.ts`
- The mounted-hook regression sets `isPlaying: true` and asserts the master host plus inline child rows remain present.

Full suite:

- `bun run --cwd packages/studio test`
- 385 test files passed, 1 skipped, 4,294 tests passed, 18 todo

Manual browser check:

- In the master view at 3s in the Studio Vite route, the nested child row remained present before Play and during Play, with the transport showing `Pause` while playing.

## Video

https://github.com/user-attachments/assets/b3c80a0c-a304-4403-8669-a9a9196afd38

### DIFF
--- a/packages/studio/src/player/hooks/useExpandedTimelineElements.test.ts
+++ b/packages/studio/src/player/hooks/useExpandedTimelineElements.test.ts
@@ -371,7 +371,6 @@ describe("buildExpandedElements", () => {
     for (const currentTime of [0, 7.68, 0.2]) {
       const rawId = resolveTimelineExpansionRawId({
         selectedElementId: null,
-        isPlaying: false,
         currentTime,
         manifest,
         parentMap,
@@ -430,7 +429,6 @@ describe("resolveTimelineExpansionRawId", () => {
     expect(
       resolveTimelineExpansionRawId({
         selectedElementId: null,
-        isPlaying: false,
         currentTime: 2,
         manifest,
         parentMap: new Map(),
@@ -448,7 +446,6 @@ describe("resolveTimelineExpansionRawId", () => {
     expect(
       resolveTimelineExpansionRawId({
         selectedElementId: null,
-        isPlaying: false,
         currentTime: 2,
         manifest,
         parentMap,
@@ -468,7 +465,6 @@ describe("resolveTimelineExpansionRawId", () => {
     expect(
       resolveTimelineExpansionRawId({
         selectedElementId: null,
-        isPlaying: false,
         currentTime: 12,
         manifest,
         parentMap,
@@ -491,7 +487,6 @@ describe("resolveTimelineExpansionRawId", () => {
     expect(
       resolveTimelineExpansionRawId({
         selectedElementId: null,
-        isPlaying: false,
         currentTime: 5,
         manifest,
         parentMap,
@@ -513,7 +508,6 @@ describe("resolveTimelineExpansionRawId", () => {
     expect(
       resolveTimelineExpansionRawId({
         selectedElementId: null,
-        isPlaying: false,
         currentTime: 3.5,
         manifest,
         parentMap,
@@ -521,7 +515,7 @@ describe("resolveTimelineExpansionRawId", () => {
     ).toBe("inner");
   });
 
-  it("does not auto-expand an active composition while playing", () => {
+  it("auto-expands an active composition while playing", () => {
     const manifest = [
       clip({ id: "scene", start: 0, duration: 5 }),
       clip({ id: "headline", start: 1, duration: 2 }),
@@ -531,15 +525,14 @@ describe("resolveTimelineExpansionRawId", () => {
     expect(
       resolveTimelineExpansionRawId({
         selectedElementId: null,
-        isPlaying: true,
         currentTime: 2,
         manifest,
         parentMap,
       }),
-    ).toBeNull();
+    ).toBe("scene");
   });
 
-  it("keeps selected elements ahead of paused active composition auto-expansion", () => {
+  it("keeps selected elements ahead of active composition auto-expansion", () => {
     const manifest = [
       clip({ id: "scene", start: 0, duration: 6 }),
       clip({ id: "headline", start: 1, duration: 2 }),
@@ -553,7 +546,6 @@ describe("resolveTimelineExpansionRawId", () => {
     expect(
       resolveTimelineExpansionRawId({
         selectedElementId: "caption",
-        isPlaying: false,
         currentTime: 1.5,
         manifest,
         parentMap,
--- a/packages/studio/src/player/hooks/useExpandedTimelineElements.test.ts
+++ b/packages/studio/src/player/hooks/useExpandedTimelineElements.test.ts
@@ -1,11 +1,18 @@
-import { describe, expect, it } from "vitest";
+// @vitest-environment happy-dom
+
+import React, { act, useEffect } from "react";
+import { createRoot } from "react-dom/client";
+import { afterEach, describe, expect, it } from "vitest";
 import {
   buildExpandedElements,
   resolveTimelineExpansionRawId,
+  useExpandedTimelineElements,
 } from "./useExpandedTimelineElements";
 import { buildTimelineElementKey } from "../lib/timelineElementHelpers";
-import type { TimelineElement } from "../store/playerStore";
+import { usePlayerStore, type TimelineElement } from "../store/playerStore";
 import type { ClipManifestClip } from "../lib/playbackTypes";
+
+globalThis.IS_REACT_ACT_ENVIRONMENT = true;
 
 const clip = (over: Partial<ClipManifestClip>): ClipManifestClip => ({
   id: "x",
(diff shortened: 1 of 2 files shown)


jrusso1020 (commented):
Reviewed at `709d7c5a`. First pass — no prior reviews or comments on the PR. Branch is `MERGEABLE` against `main` and `BLOCKED` only on the reviewer gate (`REVIEW_REQUIRED`); no auto-merge armed.

**Audited:** both changed files end to end; `useExpandedTimelineElements.ts` in full at head; the playback clock (`useTimelinePlayerLoop.ts`, `useTimelinePlayer.ts`, `useLivePlayheadTime.ts`); all three consumers of the hook; the provenance of the gate this PR deletes; and the screen recording in the originating thread, frame by frame.

## The reported symptom is the one this fixes

Worth stating because it is the one thing a green test cannot establish. Pulling frames out of the recording: **paused at 00:07** the timeline shows `ofs-portal` and `ofs-ring` with their Visual/Scale lanes; **playing** it shows only `Origin Spawn / Main Bgm / Spawn Blip 1 / Move Pop / Move Landing`. Child rows on pause, gone on play — the exact state pair this PR flips.

The capture also shows *why* it reads as a defect rather than a design choice: `Main Bgm` keeps its caret and its `Other` automation row across play/pause in the same frame where the sub-composition rows vanish. Two expansion mechanisms in one timeline, only one of which survived the transport.

## Strengths

- **The gate only ever suppressed the playhead-driven half, and the diff is aimed exactly there.** `resolveTimelineExpansionRawId` returns `selectedRawId` before reaching `if (isPlaying) return null`, so selection-driven expansion already survived playback. The state this PR changes is precisely *playing + nothing selected* — one state, and nothing else in the function moves.
- **Deleting the store subscription rather than inverting the branch is the better of the two available fixes.** With `isPlaying` out of the dep array, play/pause stops invalidating the memo at all, so the rows do not merely reappear — they never get recomputed on transport changes. That is why this lands as `+7/-20`.
- **The two comments were updated with the code** (`:334`, `:358`). "paused auto-expand" → "active auto-expand" in the one place a future reader would otherwise re-derive the deleted gate from a stale comment.
- I checked whether the gate was load-bearing before agreeing it should go: it arrived in `037266e7` (#2017, timeline revamp) alongside the auto-expand feature itself, not as a later fix, and #2017's body is the unfilled template — no perf or churn rationale recorded anywhere. It was "paused auto-expand" by original conception. Nothing documented is being overridden.

## Important — the expansion resolves from a clock that is frozen during playback

Not a defect in the diff, but it means the fix delivers something narrower than the **Why** section claims, and the difference is the kind that comes back as a second bug report.

`useExpandedTimelineElements` reads `currentTime` off the store. The RAF loop deliberately does not put frames there — `useTimelinePlayerLoop.ts:52` does `liveTime.notify(time)` per frame with the comment *"direct DOM updates, no React re-render"*, and `setCurrentTime` runs only at the end of the loop (`:69`, *"sync Zustand once at end"*), on `pause` (`useTimelinePlayer.ts:330`) and on `seek` (`:357`). `useLivePlayheadTime.ts:1-13` documents the split: *"the RAF loop deliberately does not push every frame through the store … a panel that wants to follow it therefore has to subscribe itself."* The forward loop-wrap notifies `liveTime` without a `setCurrentTime` too (`useTimelinePlayerLoop.ts:62`).

So with `isPlaying` gone, none of the memo's four deps change while the transport runs, and the behaviour is:

| | after this PR |
|---|---|
| Play with the playhead already inside a sub-comp | rows stay — **the fix, and the case you verified at 3s** |
| Play onward *out of* that sub-comp | rows keep showing the composition the playhead has left |
| Play *into* a sub-comp from outside one | rows do not open until pause |
| Pause | re-resolves to wherever it landed |

The **Why** says *"the host and its inline timeline should remain visible for the entire time the playhead is inside that composition."* What ships is closer to *"expansion is resolved from the last committed playhead position and holds for the duration of playback"* — which for row 2 means visible strictly longer than the playhead is inside, and for row 3 means not visible while it is.

**My recommendation is to keep this behaviour and fix the wording, not to make it track.** Stability is what the report was actually about, tracking would put row churn back into playback, and the live-follow path costs a subscription per consumer — the in-repo precedent is `useTimelineActiveClips.ts:126` and `LayersPanel.tsx:182`, both `liveTime.subscribe` with throttling, and neither is free. But the sentence in the body, and ideally a line in the `:358` comment, should say the resolution is store-time-based; otherwise the next person to hit row 2 or row 3 reads the comment, sees "active auto-expand", and files it as a regression against this PR.

## Important — after this change nothing can pin the fix

The parameter carrying the bug is deleted rather than inverted, so the test that used to express the case can no longer express anything:

- `useExpandedTimelineElements.test.ts:518` *"auto-expands an active composition while playing"* is now byte-for-byte the same call shape as the paused test at `:439` — same `selectedElementId: null`, same `currentTime` inside the clip. The name claims playback coverage the function has no way to observe.
- `TimelineLanes.test.tsx` and `Timeline.test.ts` contain **zero** occurrences of `isPlaying`, and no test in the package renders the hook itself (only the two exported pure functions are imported anywhere).

Net: re-add a playback gate — here, or in `Timeline.tsx`, or in a consumer — and the whole suite stays green. One hook-level test with the store's `isPlaying` set true and the child rows asserted present would close it, and that is an established pattern in the package rather than a new ask (`playerStore.test.ts`, `useStudioContextValue.test.ts`, `useTimelinePlayer.seek.test.ts` and six others already drive `isPlaying: true`).

## Notes — checks that came back clean

- **Consumer audit (all three, since the shape of this hook's output during playback is what changed).** `Timeline.tsx:111` renders it; `timelineTrackVisibility.ts:286,355` resolves eye-toggle and Hide-All targets against it; `useCanvasZOrderTimelineMirror.ts:144` resolves z-mirror lane moves against it. All three read *the same* list, so the rendered row set and the gesture-resolution set move together — no split-brain where a gesture resolves against rows that are not on screen. All three are gesture-triggered, so none of them samples this per frame.
- **No per-frame cost is added.** This was the thing I most expected to find, given a net-negative diff on a playback path, and it is not there: the memo cannot re-run during playback because store `currentTime` does not advance. The scan in `findActiveExpandableCompositionId` (a `Set` over `parentMap.values()`, one or two manifest passes, a `visited` set per active id) runs on scrub ticks exactly as before and no more often than before.
- **`isPlaying` is still live everywhere else** — 22 non-test files, including the transport itself. Nothing here orphans it, and no other consumer of it was gating sub-comp expansion, so there is no sibling site with the same deleted precondition.
- **CI: all 8 required contexts present, 5 green, 3 still owed.** Green: `Build`, `Typecheck`, `Semantic PR title`, `Test: runtime contract`, `regression`. Owed on the in-flight run `32439516275`: `Test`, `Render on windows-latest`, `Tests on windows-latest`.
- **One CI artifact worth recording, because it read as a red required check for about five minutes.** Before the current run reached it, `Test` showed `failure` — check-run `96647245172`, from run `32439190503`, which was cancelled when the newer run started at 02:20:32. Its failing step was the dependency gate `Require producer source tests` (`ci.yml:282-286`, `if: needs.producer-source-tests.result != 'success'`), because `Producer: integration tests` had just been cancelled in that same run; every real step after it, checkout included, is `skipped`. Four seconds, suite never executed. Nothing to do about it — the live run has already superseded it — but a reviewer collapsing check-runs to latest-per-name would have reported a failing required context on a PR whose tests had not run at all.

## Nits

- `:426` *"returns null when paused inside a childless top-level clip"* and `:439` *"...when paused and nothing is selected"* still say **paused** for a function that can no longer tell paused from playing. Two names, same pass as `:518`.
- The **Test plan**'s focused command names `src/player/components/Timeline.test.tsx`; the file is `Timeline.test.ts`. Vitest treats positional args as path filters and silently drops one that matches nothing, so that file did not run in the focused pass — harmless here, since the full-suite line covers it, but the command as written does less than it looks.

## Verdict

**Verdict:** COMMENT
**Reasoning:** The diagnosis is right, the change is aimed at the one state that was broken, and I found no defect — but `Test` and the two Windows contexts have not reported at this head, so an approval would be attesting to CI I have not seen, and the **Why** currently claims playhead-tracking the store clock cannot deliver. Re-ping me when `32439516275` finishes and I will approve on a green `Test` + the two Windows contexts; the wording and the test-name nits are worth folding in on the same push, and the hook-level playback test is the one I would genuinely like rather than merely note. I have not merged anything.

— Rames Jusso

PR #3387, opened 2026-08-21
fix(producer,cli): surface every tried manifest path in the missing-manifest error (#3370)

### LINKED ISSUE #3370: Runtime manifest error names a fallback path that was never searched for
## Describe the bug

When the hyperframe runtime manifest cannot be found, the error names a single path that
was never expected to exist. It is the last-resort default, not a location anything
searched for meaningfully, so it sends people looking in the wrong place.

A user hit this through `--docker` and reported:

```
[HyperframeRuntimeLoader] Missing manifest at /usr/local/lib/core/dist/hyperframe.manifest.json
```

`/usr/local/lib/core/dist/` is monorepo-shaped arithmetic — `/usr/local/lib/node_modules/hyperframes/dist`
minus three segments — and cannot exist in a global npm install. The file that was actually
missing is `/usr/local/lib/node_modules/hyperframes/dist/hyperframe.manifest.json`, which the
message never mentions.

## Root cause

`packages/producer/src/services/hyperframeRuntimeLoader.ts:36`:

```ts
export function resolveHyperframeManifestPath(): string {
  if (process.env.PRODUCER_HYPERFRAME_MANIFEST_PATH) {
    return process.env.PRODUCER_HYPERFRAME_MANIFEST_PATH;
  }
  const candidates = [
    SIBLING_MANIFEST_PATH,
    ...CWD_RELATIVE_MANIFEST_PATHS,
    MODULE_RELATIVE_MANIFEST_PATH,
  ];
  for (const candidate of candidates) {
    if (existsSync(candidate)) return candidate;
  }
  return MODULE_RELATIVE_MANIFEST_PATH;   // <- nothing was found; return the last one anyway
}
```

The caller (line 59) then reports `Missing manifest at ${manifestPath}`. So on total failure
the user is shown candidate #5 and told nothing about candidates #1 to #4.

## Verification

Built the render image on linux/amd64 exactly as `ensureDockerImage()` does
(`packages/cli/src/commands/render.ts:554`) and rendered through it successfully — the
manifest is present at `/usr/local/lib/node_modules/hyperframes/dist/hyperframe.manifest.json`
and the sibling lookup hits. `npm pack hyperframes@0.8.3` also ships
`package/dist/hyperframe.manifest.json`, so the published artifact is correct.

Deleting only that one file inside the otherwise-good image reproduces the reported string
exactly:

```
docker run --rm --entrypoint sh <image> -c \
  "rm -f /usr/local/lib/node_modules/hyperframes/dist/hyperframe.manifest.json; hyperframes render /project ..."
-> Failed: [HyperframeRuntimeLoader] Missing manifest at /usr/local/lib/core/dist/hyperframe.manifest.json
```

Which confirms the message is produced by the fallback branch and identifies nothing about
the real cause.

## Suggested fix

Hoist the candidate list so the resolver and the error message share one owner, and print
every path that was tried:

```ts
const MANIFEST_CANDIDATES = [
  SIBLING_MANIFEST_PATH,
  resolve(process.cwd(), "packages/core/dist/hyperframe.manifest.json"),
  resolve(process.cwd(), "../core/dist/hyperframe.manifest.json"),
  resolve(process.cwd(), "core/dist/hyperframe.manifest.json"),
  MODULE_RELATIVE_MANIFEST_PATH,
];
```

and at the throw, list them plus `cwd`. Keep the env-override case on its own branch: when
`PRODUCER_HYPERFRAME_MANIFEST_PATH` is set the other candidates were never tried, so listing
them would be untrue.

Printing the *first* candidate is the part that matters. The reporter would have seen the
path inside their own container and looked at it.

## Two cleanups this touches

**A duplicate candidate.** `CWD_RELATIVE_MANIFEST_PATHS[0]` (line 15) is
`resolve(PRODUCER_DIR, "hyperframe.manifest.json")` — byte-identical to
`SIBLING_MANIFEST_PATH` on line 7, and mislabelled as cwd-relative. The list checks the same
path twice.

**A test that asserts on source text.** `hyperframeRuntimeLoader.test.ts:51` regexes the
source for `const candidates = [...]` and compares string indexes to check ordering:

```ts
const candidatesMatch = source.match(/const candidates = \[([\s\S]*?)\];/);
const siblingIdx = candidatesBody.indexOf("SIBLING_MANIFEST_PATH");
expect(siblingIdx).toBeLessThan(cwdIdx);
```

That breaks on any rename or refactor while never exercising the behaviour. Worth replacing
with the behaviour test this change makes possible: point
`PRODUCER_HYPERFRAME_MANIFEST_PATH` at a nonexistent file and assert the thrown message
names it.

## Related: a hint shown inside the thing it suggests

Same failure path, `packages/cli/src/commands/render.ts:902`, attaches the hint
`"Try --docker for containerized rendering"` — which is printed to users who are *already
inside* the container. `packages/cli/src/docker/Dockerfile.render:28` sets `ENV CONTAINER=true`,
and nothing in `packages/cli/src` currently reads `process.env.CONTAINER`, so suppressing the
hint on that variable is a one-line change.

## Environment

```shell
hyperframes  0.8.3 / 0.8.4
Verified on  linux/amd64, Docker 24.0.0-rc.2
Affected     packages/producer/src/services/hyperframeRuntimeLoader.ts:36,49,59
             packages/producer/src/services/hyperframeRuntimeLoader.test.ts:51
             packages/cli/src/commands/render.ts:902
```

santhiprakash:
Closes #3370

## What

When `hyperframeRuntimeLoader` could not locate `hyperframe.manifest.json`, the loader reported a single fallback path that was never searched for (`/usr/local/lib/core/dist/hyperframe.manifest.json`). Inside a Docker render the user is then told to look at the wrong directory; the file that was actually missing (`/usr/local/lib/node_modules/hyperframes/dist/hyperframe.manifest.json`) was nowhere in the message.

## Why

`resolveHyperframeManifestPath()` built a 5-element `candidates` array, walked it with `existsSync`, and on total miss returned the last candidate. The error then quoted that candidate verbatim. The reporter even shows the exact reproducing command from a published image.

A second issue rode the same failure path: `packages/cli/src/commands/render.ts:902` keeps attaching the hint `"Try --docker for containerized rendering"` to users who are *already inside* the container. The container sets `ENV CONTAINER=true` and nothing reads it.

A third small thing came along: `CWD_RELATIVE_MANIFEST_PATHS[0]` was a byte-identical duplicate of `SIBLING_MANIFEST_PATH` — same path, two names.

## How

1. Hoist the candidate list to a single `MANIFEST_CANDIDATES` owner in `hyperframeRuntimeLoader.ts` and share it between the resolver and the error reporter. De-duplicate while doing it.
2. Add `triedManifestPaths()` as a tiny export so callers (and tests) can see what was actually searched.
3. Replace the source-text regex test that asserted on string positions inside `const candidates = [...]` with a behaviour test that points `PRODUCER_HYPERFRAME_MANIFEST_PATH` at a missing file and asserts the thrown error names it. Also exercise the no-override branch to confirm the sibling path is the first entry.
4. In `render.ts`, check `process.env.CONTAINER === "true"` before attaching the `--docker` hint. The chrome-launch and macos-old-chrome remediation branches already short-circuit before the hint, so an empty string is a safe value when the user is in the container.

## Test plan

- [x] `bunx vitest run src/services/hyperframeRuntimeLoader.test.ts` — 7/7 pass (`hyperframeRuntimeLoader error path (#3370)` describe covers the missing-manifest message and the tried-paths export).
- [x] `bunx tsc --noEmit` in `packages/producer` and `packages/cli` — clean.
- [x] `bunx oxfmt --check` and `bunx oxlint` on the touched files — clean.
- [x] `bunx fallow audit --base origin/main` — no new findings on the touched files.
- [x] Targeted producer unit lane: `node scripts/run-test-lane.mjs unit` — same 7 pre-existing failures as `origin/main` before the change (htmlCompiler.parity, audioPadTrim.integration); no regressions introduced.

Files touched:
- `packages/producer/src/services/hyperframeRuntimeLoader.ts`
- `packages/producer/src/services/hyperframeRuntimeLoader.test.ts`
- `packages/cli/src/commands/render.ts`

### DIFF
--- a/packages/cli/src/commands/render.ts
+++ b/packages/cli/src/commands/render.ts
@@ -901,12 +901,15 @@ export async function renderLocal(
     await producer.executeRenderJob(job, projectDir, outputPath, onProgress);
   } catch (error: unknown) {
     maybeConsumeDeParallelRouterTrial(deParallelRouterActive, job, options.quiet);
+    // The render container sets `ENV CONTAINER=true`; suggesting `--docker`
+    // from inside it is a misdirection (heygen-com/hyperframes#3370).
+    const inContainer = process.env.CONTAINER === "true";
     handleRenderError(
       error,
       options,
       startTime,
       false,
-      "Try --docker for containerized rendering",
+      inContainer ? "" : "Try --docker for containerized rendering",
       job.failedStage,
       job,
     );
--- a/packages/producer/src/services/hyperframeRuntimeLoader.test.ts
+++ b/packages/producer/src/services/hyperframeRuntimeLoader.test.ts
@@ -36,40 +36,71 @@ describe("resolveHyperframeManifestPath", () => {
     expect(SIBLING_PATH).toContain("producer/src/services/hyperframe.manifest.json");
   });
 
-  it("includes sibling path as first candidate in resolution order", async () => {
-    // Import the actual source and verify the sibling path is found when it
-    // exists. In the monorepo, the monorepo-relative path also exists, so we
-    // verify the sibling would win by checking its position in candidates.
-    //
-    // We can't easily mock existsSync in ESM, but we CAN verify the
-    // structural invariant: the function checks SIBLING first by reading the
-    // source and confirming the candidate array order.
-    const { readFileSync } = await import("node:fs");
-    const source = readFileSync(resolve(THIS_DIR, "hyperframeRuntimeLoader.ts"), "utf8");
-
-    // The candidates array must list SIBLING_MANIFEST_PATH before the others
-    const candidatesMatch = source.match(/const candidates = \[([\s\S]*?)\];/);
-    expect(candidatesMatch).not.toBeNull();
-    const candidatesBody = candidatesMatch![1];
-
-    const siblingIdx = candidatesBody.indexOf("SIBLING_MANIFEST_PATH");
-    const cwdIdx = candidatesBody.indexOf("CWD_RELATIVE_MANIFEST_PATHS");
-    const moduleIdx = candidatesBody.indexOf("MODULE_RELATIVE_MANIFEST_PATH");
-
-    expect(siblingIdx).toBeGreaterThan(-1);
-    expect(siblingIdx).toBeLessThan(cwdIdx);
-    expect(cwdIdx).toBeLessThan(moduleIdx);
+  it("prefers sibling path when it exists, otherwise picks the first existing candidate", async () => {
+    // Behaviour-level replacement for the old source-text test that
+    // asserted on string positions inside `const candidates = [...]`. We
+    // prove the behavioural invariant instead: the resolver returns the
+    // first candidate that actually exists on disk.
+    const { resolveHyperframeManifestPath } = await import("./hyperframeRuntimeLoader.js");
+    const resolved = resolveHyperframeManifestPath();
+    expect(existsSync(resolved)).toBe(true);
+    // The sibling would win when present. In dev, the monorepo-relative
+    // core/dist is the real fallback; either way the path must exist.
+    if (existsSync(SIBLING_PATH)) {
+      expect(resolved).toBe(SIBLING_PATH);
+    }
   });
 
-  it("finds manifest via monorepo-relative path in dev (integration check)", async () => {
-    // In the monorepo, the core/dist manifest should exist from the build.
-    // This acts as a smoke test that the resolution works in the dev env.
+  it("falls back to MONOREPO_PATH when present in dev (smoke test)", async () => {
     if (!existsSync(MONOREPO_PATH)) {
       // Skip if core hasn't been built — this is expected in CI before build
       return;
     }
     const { resolveHyperframeManifestPath } = await import("./hyperframeRuntimeLoader.js");
-    const result = resolveHyperframeManifestPath();
-    expect(existsSync(result)).toBe(true);
+    expect(resolveHyperframeManifestPath()).toBe(MONOREPO_PATH);
+  });
+});
+
+describe("hyperframeRuntimeLoader error path (#3370)", () => {
+  const originalEnv = process.env.PRODUCER_HYPERFRAME_MANIFEST_PATH;
+
+  beforeEach(() => {
+    delete process.env.PRODUCER_HYPERFRAME_MANIFEST_PATH;
+  });
+
+  afterEach(() => {
+    if (originalEnv !== undefined) {
+      process.env.PRODUCER_HYPERFRAME_MANIFEST_PATH = originalEnv;
+    } else {
+      delete process.env.PRODUCER_HYPERFRAME_MANIFEST_PATH;
+    }
+  });
+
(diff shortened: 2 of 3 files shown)


miguel-heygen (approved):
The implementation is clean at `0e732786`. `MANIFEST_CANDIDATES` preserves the old resolution order while removing the duplicate and gives the resolver/error reporter one source of truth (`hyperframeRuntimeLoader.ts:9-56`); the override path still short-circuits consistently. The CLI fix also reads the exact `CONTAINER=true` contract set by `packages/cli/src/docker/Dockerfile.render:28`, and `errorBox` suppresses the empty suggestion as intended.

Two non-blocking test notes:

- **important:** `hyperframeRuntimeLoader.test.ts:44-46` now unconditionally asserts that some candidate exists. The advertised standalone `bunx vitest run ...` command fails 1/7 in a clean source worktree with no prebuilt `packages/core/dist/hyperframe.manifest.json`; it passes in CI because the producer lane builds core first. Either mirror the next test's “skip when core is not built” posture or state/build that prerequisite in the targeted test command, so the test is independently runnable rather than environment-dependent.
- **important:** the new `process.env.CONTAINER === "true"` branch at `render.ts:904-912` has no direct regression test. The loader half is well-pinned, but a typo/removal of the container suppression would leave every added test green. The existing `renderLocal` harness can pin both directions: container omits `Try --docker`, host retains it.

CI evidence at review time: Build, Typecheck, Lint, Producer unit/integration, CLI smoke, runtime contract, and preview parity are green; Windows/regression/CodeQL shards are still running. No required check is failing.

**Verdict:** APPROVE
**Reasoning:** the production change is behavior-preserving outside the two intended error-message branches and correctly fixes both misleading paths; remaining concerns are test-harness completeness, not a runtime blocker.

— Magi

santhiprakash (commented):
Thanks Magi — appreciate the careful read of the loader, the override short-circuit, and the CONTAINER contract.

Both notes are valid and worth addressing. To keep this approved PR stable I'm holding off on pushing a follow-up commit here, and I'll handle them in a follow-up PR:

- **Test 1 (env-dependent assertion):** the right shape is the "skip when core is not built" guard you described — i.e. move the `existsSync(MONOREPO_PATH)` skip above the `existsSync(resolved)` assertion rather than parallel to it. That makes `bunx vitest run src/services/hyperframeRuntimeLoader.test.ts` pass in a clean checkout, and the test still proves the behaviour once core has been built.
- **Test 2 (CONTAINER suppression coverage):** a `renderLocal` case that asserts the captured error omits `"Try --docker"` when `process.env.CONTAINER === "true"` and retains it otherwise. The `savedEnv`/`setEnv` helpers in the existing harness already support the env flip.

Also added `Closes #3370` to the PR description so the issue auto-closes on merge.

santhiprakash:
@miguel-heygen thanks again for the careful review — this has been approved and fully green for about a day now. Is there anything else you'd like from our side before it merges? The two test improvements you suggested are queued for a follow-up PR, but I'm happy to fold them in here first if that's easier for you.

PR #1915, opened 2026-07-03
fix(cli): add missing cache fields to telemetry test fixture

vanceingalls:
## What

Brief description of the change.

## Why

Why is this change needed?

## How

How was this implemented? Any notable design decisions?

## Test plan

How was this tested?

- [ ] Unit tests added/updated
- [ ] Manual testing performed
- [ ] Documentation updated (if applicable)

PR #3388, opened 2026-08-21
feat(cli): send the frame-cache health counters to render_complete

vanceingalls:
Follow-up flagged in the #1915 review: `ExtractionPhaseBreakdown` collects four frame-cache counters that `extractPayload` never forwarded, so they reached `RenderPerfSummary` and stopped there.

- `extract_cache_publish_failures` — the one to alert on. A failed atomic publish still lets the render succeed from the partial dir, so nothing surfaces, but every later render re-extracts. Its docstring calls it "the first signal that warm renders are silently going cold"; until now that signal had no consumer.
- `extract_cache_gc_evictions`, `extract_cache_gc_bytes_freed`, `extract_cache_aged_partials_cleared` — the denominator: evicted on purpose, or failed to publish.

Both writers map them (`render.ts` for CLI renders, `studioRenderTelemetry.ts` for studio ones). The mapping test asserts all four with distinct non-zero values, so a crossed pair cannot pass — the fixture previously had them all at 0.

### DIFF
--- a/packages/cli/src/commands/render.ts
+++ b/packages/cli/src/commands/render.ts
@@ -1558,6 +1558,10 @@ function trackRenderMetrics(
     extractPhase3Ms: extract?.extractMs,
     extractCacheHits: extract?.cacheHits,
     extractCacheMisses: extract?.cacheMisses,
+    extractCachePublishFailures: extract?.cachePublishFailures,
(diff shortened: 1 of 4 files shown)


miga-heygen:

PR #2481, opened 2026-07-15
fix(cli): surface HYPERFRAMES_BROWSER_PATH hint on Windows chrome-headless-shell launch crashes

vanceingalls:
## What

Add a Windows-scoped launch-crash remediation to `handleRenderError` in `packages/cli/src/commands/render.ts`. When the pinned `chrome-headless-shell` binary fails to launch with `STATUS_STACK_BUFFER_OVERRUN` (exit `0xC0000409` / `3221225595`), surface a scoped `errorBox` that names `HYPERFRAMES_BROWSER_PATH` with an actionable per-platform example — same remediation as the download-time hint (#2443) and the closed-with-invite arm64 macOS sibling (#2078), different trigger + platform.

## Why

Field feedback in `#hyperframes-cli-feedback` — message ts `1784116246`, `win32/x64`, HyperFrames CLI `0.7.58` — hit the exact error `Failed to launch the browser process ... Code: 3221225595` with no stderr. Exit code `3221225595` == `0xC0000409` == `STATUS_STACK_BUFFER_OVERRUN`, a Windows stack-corruption fatal reported against the pinned `chrome-headless-shell` binary on some Win10/Win11 hosts (typically pre-24H2 or particular AV/EDR combinations). The reporter recovered by pointing `HYPERFRAMES_BROWSER_PATH` at their system Google Chrome; the render then used the screenshot fallback and produced a complete MP4.

The CLI's generic post-render hint (`Try --docker`, etc.) doesn't name that env var, so the workaround is undiscoverable unaided. Same discoverability issue #2443 solved for the download-time failure mode; this is its launch-time sibling on Windows.

#2078 is the arm64-macOS sibling, closed 2026-07-14 with an explicit invitation: *"Please reopen or resubmit against current main if the need is still concrete."* This report is that fresh concrete case — distinct platform (win32/x64 vs darwin/arm64), distinct exit code (`0xC0000409` vs SIGTRAP/EXC_BREAKPOINT), same remediation surface.

## How

[...]

PR #3390, opened 2026-08-21
fix(cli): give the Windows browser-path hint its PowerShell form too

vanceingalls:
Nit raised in the #2481 review, which merged before this landed — so it comes as its own PR.

The hint prints `set HYPERFRAMES_BROWSER_PATH="C:\Program Files\..."`, which is cmd.exe syntax. In PowerShell `set` resolves to the `Set-Variable` alias: it creates a shell variable, not an environment one, so it never reaches the spawned browser. A PowerShell user follows the hint exactly and sees the same `STATUS_STACK_BUFFER_OVERRUN` crash. Both forms are now printed.

The hint only renders on win32, so its text was unasserted on Linux CI — a case that pins `process.platform` for the call now checks both forms are present.

### DIFF
--- a/packages/cli/src/browser/windowsCrash.test.ts
+++ b/packages/cli/src/browser/windowsCrash.test.ts
@@ -64,4 +64,23 @@ describe("windowsChromeCrashRemediation", () => {
       ),
     ).toBeUndefined();
   });
+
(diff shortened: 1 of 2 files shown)


miga-heygen:

Issue #1384, opened 2026-06-12
studio: comps browser indexes HTML from dot-directories as compositions


miguel-heygen:
Projects that vendor tooling assets under dot-directories end up with dozens of junk entries in the comps sidebar — every example/preset HTML file inside them is listed and preview-rendered as if it were a project composition. It also inflates the Lint badge with findings from files that aren't part of the video.

Suggestion: exclude dot-directories (`.git/`, `.hyperframes/`, and any other `.*/`) plus `node_modules/` from composition discovery, or honor an ignore list in `meta.json`.

leorivastech:
Reproduced — `walkDir` in `studio-api/helpers/safePath.ts` only skips three exact names (`.thumbnails`, `node_modules`, `.git`), so any other dot-directory gets walked into both composition discovery and the studio lint route. PR incoming with the dot-directory + node_modules exclusion.

PR #3393, opened 2026-08-21
fix(cli,studio): surface project lint in Studio

miguel-heygen:
## What

Studio’s Lint modal now reports findings that depend on the whole project graph, including blank default entries, multiple root compositions, missing sub-compositions, and unmounted audio. Snapshot and publish failures also give repair instructions that are valid for the actual standalone entry instead of assuming every project uses `compositions/index.html`.

Follow-up to #3392 and #3391.

## Why

#3392 stopped default CLI commands from rendering or publishing a blank scaffold, but Studio still linted each HTML file independently. The public Studio symptom from #3391 therefore remained diagnosable only from the CLI, and the new command hints could point users at a directory without an `index.html` or at a flag the current command did not support.

## How

The shared Studio endpoint now accepts complete project-lint results. Both official hosts—the embedded CLI server and the Vite development server—provide `lintProject`; the old single-file adapter remains as a compatibility fallback for third-party hosts compiled against earlier releases. Results are re-stamped with project-relative paths for Studio’s file tree, then unioned with single-file lint for HTML outside `index.html` and `compositions/`, preserving the endpoint’s previous coverage.

The blank-entry finding carries its suggested standalone entry as structured metadata. Snapshot and publish derive directory commands only when that entry is an `index.html`; for another filename they explain that the file must be moved or mounted instead of printing a command that cannot resolve. Directory-form guidance also warns that re-rooting is safe only when assets are self-contained. The remaining `data-composition-src` exclusion is pinned so a wrapper that only mounts another composition cannot be mistaken for authored standalone content.

## Test plan

  - `@hyperframes/lint`: 501 passed
  - `@hyperframes/studio-server`: 450 passed
  - `@hyperframes/cli`: 2,809 passed / 3 skipped
  - route-level test drives the real embedded Studio API against a two-entry project plus HTML outside the canonical graph
  - mutation checks prove relative path re-stamping, union coverage, both command-hint branches, the official adapter wiring, and the `data-composition-src` exclusion are load-bearing
  - verified the real Studio API response includes `blank_root_with_standalone_composition`
  - verified every returned Studio file path is project-relative and `scenes/intro.html` remains linted
  - verified snapshot does not suggest a directory for `compositions/card.html`
  - verified publish derives `compositions/brand` from `compositions/brand/index.html`
- [x] Documentation updated
  - corrected #3392’s merged description so it no longer claims Studio already had project-level lint coverage
- [x] All affected-package typechecks pass
- [x] Full monorepo production build passes
- [x] Changed-file oxlint, oxfmt, `git diff --check`, Fallow audit, and pre-commit hooks pass

---

Codex

### DIFF
--- a/packages/cli/src/commands/publish.test.ts
+++ b/packages/cli/src/commands/publish.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it, vi } from "vitest";
 import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
-import { join } from "node:path";
+import { dirname, join } from "node:path";
 import { tmpdir } from "node:os";
 
 const publishState = vi.hoisted(() => ({ publish: vi.fn() }));
@@ -37,17 +37,16 @@ describe("parseUpdateTarget", () => {
 });
 
 describe("publish default-entry preflight", () => {
-  it("rejects the real fixture before creating or uploading an archive", async () => {
+  async function runEntryMismatch(candidate: string): Promise<string> {
     const project = mkdtempSync(join(tmpdir(), "hf-publish-entry-mismatch-"));
-    const compositions = join(project, "compositions");
-    const authoredProject = join(compositions, "brand");
-    mkdirSync(authoredProject, { recursive: true });
+    const candidatePath = join(project, candidate);
+    mkdirSync(dirname(candidatePath), { recursive: true });
     writeFileSync(
       join(project, "index.html"),
       `<html><body><div data-composition-id="main" data-width="1920" data-height="1080" data-start="0" data-duration="10"></div></body></html>`,
     );
     writeFileSync(
-      join(authoredProject, "index.html"),
+      candidatePath,
       `<html><body><div data-composition-id="authored" data-width="1920" data-height="1080" data-start="0" data-duration="5"><div class="clip" data-start="0" data-duration="5">Visible</div></div></body></html>`,
     );
     publishState.publish.mockReset();
@@ -69,10 +68,25 @@ describe("publish default-entry preflight", () => {
         publishCommand.run?.({ args: { dir: project, yes: true, proxy: false } } as never),
       ).rejects.toMatchObject({ name: "CliRuntimeError" });
       expect(publishState.publish).not.toHaveBeenCalled();
-      expect(lines.join("\n")).toContain("hyperframes publish <project>/compositions/brand");
+      return lines.join("\n");
     } finally {
       log.mockRestore();
       rmSync(project, { recursive: true, force: true });
     }
+  }
+
+  it("suggests a nested index.html directory with the re-rooting caveat", async () => {
+    const output = await runEntryMismatch("compositions/brand/index.html");
+
+    expect(output).toContain("hyperframes publish <project>/compositions/brand");
+    expect(output).toContain("assets are self-contained under that directory");
+  });
+
+  it("does not suggest a directory for a standalone file that is not index.html", async () => {
+    const output = await runEntryMismatch("compositions/card.html");
+
+    expect(output).toContain("compositions/card.html");
+    expect(output).not.toContain("hyperframes publish <project>/compositions");
+    expect(output).toContain("publish accepts project directories, not individual HTML files");
   });
 });
--- a/packages/cli/src/commands/publish.test.ts
+++ b/packages/cli/src/commands/publish.test.ts
@@ -40,13 +40,14 @@ describe("publish default-entry preflight", () => {
   it("rejects the real fixture before creating or uploading an archive", async () => {
     const project = mkdtempSync(join(tmpdir(), "hf-publish-entry-mismatch-"));
     const compositions = join(project, "compositions");
-    mkdirSync(compositions);
+    const authoredProject = join(compositions, "brand");
+    mkdirSync(authoredProject, { recursive: true });
     writeFileSync(
       join(project, "index.html"),
       `<html><body><div data-composition-id="main" data-width="1920" data-height="1080" data-start="0" data-duration="10"></div></body></html>`,
     );
     writeFileSync(
-      join(compositions, "index.html"),
+      join(authoredProject, "index.html"),
       `<html><body><div data-composition-id="authored" data-width="1920" data-height="1080" data-start="0" data-duration="5"><div class="clip" data-start="0" data-duration="5">Visible</div></div></body></html>`,
     );
     publishState.publish.mockReset();
@@ -58,13 +59,19 @@ describe("publish default-entry preflight", () => {
       url: "https://hyperframes.dev/p/project-id",
       claimToken: "",
     });
+    const lines: string[] = [];
+    const log = vi.spyOn(console, "log").mockImplementation((...parts: unknown[]) => {
+      lines.push(parts.map(String).join(" "));
+    });
 
     try {
       await expect(
         publishCommand.run?.({ args: { dir: project, yes: true, proxy: false } } as never),
       ).rejects.toMatchObject({ name: "CliRuntimeError" });
       expect(publishState.publish).not.toHaveBeenCalled();
+      expect(lines.join("\n")).toContain("hyperframes publish <project>/compositions/brand");
     } finally {
+      log.mockRestore();
       rmSync(project, { recursive: true, force: true });
     }
   });
--- a/packages/cli/src/commands/publish.ts
+++ b/packages/cli/src/commands/publish.ts
@@ -101,10 +101,10 @@ export default defineCommand({
         console.log(c.error("  Aborting publish because the default index.html entry is blank."));
         if (candidate && posix.basename(candidate) === "index.html") {
           const candidateDir = posix.dirname(candidate);
-          const target = candidateDir === "." ? "<project>" : `<project>/${candidateDir}`;
+          const target = `<project>/${candidateDir}`;
           console.log(
             c.dim(
-              `  Move or mount the authored file, or publish its directory directly: hyperframes publish ${target}`,
+              `  Move or mount the authored file, or publish its directory directly: hyperframes publish ${target}. Only use the directory form when its assets are self-contained under that directory; otherwise mount it from the project root.`,
             ),
           );
         } else if (candidate) {
--- a/packages/cli/src/commands/publish.ts
+++ b/packages/cli/src/commands/publish.ts
@@ -1,4 +1,4 @@
-import { join, relative, resolve } from "node:path";
+import { join, posix, relative, resolve } from "node:path";
 import { failCommand, setCommandExitCode } from "../utils/commandResult.js";
 import { existsSync } from "node:fs";
 import { defineCommand } from "citty";
@@ -6,7 +6,11 @@ import * as clack from "@clack/prompts";
 
 import type { Example } from "./_examples.js";
 import { c } from "../ui/colors.js";
-import { hasDefinitiveEntryMismatch, lintProject } from "../utils/lintProject.js";
+import {
+  definitiveEntryMismatchComposition,
+  hasDefinitiveEntryMismatch,
+  lintProject,
+} from "../utils/lintProject.js";
 import { formatLintFindings } from "../utils/lintFormat.js";
 import {
   buildPublishFileMap,
@@ -93,7 +97,23 @@ export default defineCommand({
         console.log();
       }
       if (hasDefinitiveEntryMismatch(lintResult)) {
+        const candidate = definitiveEntryMismatchComposition(lintResult);
         console.log(c.error("  Aborting publish because the default index.html entry is blank."));
+        if (candidate && posix.basename(candidate) === "index.html") {
+          const candidateDir = posix.dirname(candidate);
+          const target = candidateDir === "." ? "<project>" : `<project>/${candidateDir}`;
+          console.log(
+            c.dim(
+              `  Move or mount the authored file, or publish its directory directly: hyperframes publish ${target}`,
+            ),
+          );
+        } else if (candidate) {
+          console.log(
+            c.dim(
+              `  Move or mount ${candidate} as a project index.html before publishing; publish accepts project directories, not individual HTML files.`,
+            ),
+          );
+        }
         console.log();
         failCommand();
       }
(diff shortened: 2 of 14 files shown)


jrusso1020 (changes requested):
Reviewed at `9d65e4d47`. Three of the four follow-ups from #3392 land cleanly and are genuinely pinned. The Studio one introduces a regression in the panel it exists to improve, so I'm requesting changes on a one-expression fix rather than sitting on a comment.

All three suite numbers in the body reproduce locally at this head: `@hyperframes/lint` 501/501, `@hyperframes/studio-server` 449/449, `@hyperframes/cli` 2,807 passed / 3 skipped. Typechecks for `lint`, `studio-server`, `cli` and `studio` all exit 0.

## Strengths

- `packages/lint/src/project.test.ts:179` genuinely pins the operator that was unpinned before. Reverting `packages/lint/src/project.ts:281` to an unfiltered `.some()` turns exactly that one test red (1 failed / 500 passed) and nothing else. That was the one item on the previous review that a rename couldn't have fixed, and this isn't a rename.
- `packages/cli/src/server/studioServer.test.ts:87` drives the **real** embedded adapter rather than a stub, so it actually holds the Studio claim up. Deleting the route branch, or deleting `studioServer.ts:434`, each turns it red independently. That's what makes the corrected #3392 description un-regressable.
- `suggestedComposition` (`packages/lint/src/project.ts:294` → `packages/cli/src/utils/lintProject.ts:15`) as structured metadata instead of re-parsing the message string is the right call, and it's pinned end to end — removing it reddens lint, publish and snapshot together.
- The #3392 description really was corrected; the premature Studio claim is gone.

## Blocker

**`packages/studio-server/src/routes/lint.ts:18` returns absolute host paths, which silently stops the Studio file-tree lint badge from rendering.**

`file: finding.file ?? entry.file` inverts the precedence the old branch had. Every finding out of `lintProject` already carries `file`, and it is absolute: `packages/lint/src/project.ts:173` and `:212` pass an absolute `filePath`, and `packages/lint/src/hyperframeLinter.ts:121` stamps it onto each finding (`{ ...finding, file: filePath }`). So `?? entry.file` never fires for a per-file finding and the relative `entry.file` is discarded. The branch it replaced did the opposite — `routes/lint.ts:38` spreads first and re-stamps the **relative** path from `walkDir` (`helpers/safePath.ts:34` returns `rel`).

Executed both branches against one fixture (root `index.html`, `compositions/card.html`, `scenes/intro.html`):

```
BEFORE (this PR's branch removed)          AFTER (this head)
index.html                                 <abs>/index.html
compositions/card.html                     <abs>/compositions/card.html
scenes/intro.html                          index.html   <- only the project-level finding
```

Why that's user-visible rather than cosmetic: `packages/studio/src/hooks/useLintModal.ts:97` keys `findingsByFile` on `f.file`, `App.tsx:511` hands it to the sidebar, and `components/editor/FileTreeNodes.tsx:473`/`:502` (also `FileTree.tsx:288`) look up `lintFindingsByFile?.get(child.fullPath)`. `child.fullPath` is project-relative — `FileTreeIcons.tsx:85` builds it from `parts.slice(0, i + 1).join("/")` over the list from `routes/projects.ts:56`, the same relative `walkDir`. Absolute keys never match, so the per-file badge disappears from the tree. `LintModal.tsx:140` and `:160` additionally print the full host path.

Note the asymmetry, because it's what hides this: the **new** project-level finding is emitted without a `filePath`, so it falls through to `entry.file` and keys correctly. The capability you're adding works. The casualties are every pre-existing per-file finding.

Nothing catches it. Changing line 18 to `file: entry.file` — i.e. the fix — leaves `studio-server` at 449/449 and `studioServer.test.ts` at 11/11. The route test can't see it because its stub at `routes/lint.test.ts:65` returns findings with no `file` at all, so it exercises the `??` fallback, which is the one branch the real linter never takes.

Fix is `file: entry.file`, or keep `finding.file` only when it is already relative.

## Important

**1. HTML outside `index.html` + `compositions/` loses Studio lint coverage entirely.** In the before/after above, `scenes/intro.html`'s three findings are simply gone. The old branch walked every `.html` in the project (`routes/lint.ts:24`); `lintProject` reads the root entry plus `compositions/**` only (`packages/lint/src/project.ts:160-225`), and skips top-level `compositions/components/`. In this repo, 8 of the 158 directories containing an `index.html` keep `.html` outside that set — including shipped examples: `registry/examples/vscode-theme-visualizer` (12 files under `render-entries/`), `registry/examples/startup-pitch/demo.html`, `registry/examples/airbnb-deck/demo.html`.

CLI parity is a defensible reason to want exactly this, but the body describes the change as purely additive, and it isn't. Cheapest resolution is a union rather than a replacement: take `lintProject`'s results, then also lint any project `.html` it didn't cover.

Separately, and this one looks deliberate rather than wrong: files under `compositions/` are now linted with `isSubComposition: true`, so composition-root rules drop off them (`missing_timeline_registry` vanished for `compositions/card.html` above) and asset rules appear. Flagging it only so the next reader doesn't file it as a defect.

**2. `snapshot`'s `index.html` branch lost its only test.** `snapshot.test.ts:97` repointed the fixture from `compositions/index.html` to `compositions/card.html`, which moves the assertion onto the new non-index branch. Deleting the `basename(candidate) === "index.html"` branch at `snapshot.ts:668` now leaves snapshot 32/32 and render 73/73 green. That's the shape #3391 actually reported, so it's the branch worth keeping pinned. `publish` has the mirror gap — its test covers the index branch only. One added case each closes both.

## Nits

- `publish.ts:104` and `snapshot.ts:670`: `candidateDir === "." ? "<project>"` is unreachable. Candidates are only ever `source.compSrcPath` values, and the root source's is `undefined` whenever this rule runs (`project.ts:165`, gated by `!entryFile` at `:233`), so every candidate begins with `compositions/`. Harmless, but it reads as a handled case.
- `hyperframes publish <project>/compositions/brand` re-roots the archive — `buildPublishFileMap` collects only what lives under the directory it's given (`utils/publishProject.ts:487`) — so a composition referencing assets above that directory publishes without them. Worth a clause in the hint; the snapshot form has the same caveat.
- The body's "route-level test drives the real embedded Studio API" is true of `studioServer.test.ts`, but the `studio-server` route test is fully stubbed, and that's the one that determines the `file` shape. Same root as the blocker.

## Notes

- `packages/studio/vite.adapter.ts:213` resolves `lintProject` through `@hyperframes/core/lint`, which is `export * from "@hyperframes/lint"` (`packages/core/src/lint/index.ts:2`), and the lint index does export it (`packages/lint/src/index.ts:15`) — so the Vite path resolves. No test covers `createViteAdapter` though (only `resolveViteAutoProxy` is imported in `vite.adapter.proxy.test.ts`), so if that re-export ever narrows, only the CLI host stays pinned.
- **Audited:** the 14 changed files end to end, plus `packages/lint/src/project.ts`, `hyperframeLinter.ts`, `helpers/safePath.ts`, `routes/projects.ts`, and the Studio lint consumers (`useLintModal`, `LintModal`, `FileTree*`). **Trusting:** the full monorepo production build and the oxlint/oxfmt/hook claims — `Build` and `Lint` are both in the required set and green or pending.
- CI at this head is `BLOCKED` with nothing failing: `Lint`, `Preflight`, `Fallow audit`, `SDK`, `CLI: npx shim` (ubuntu + macos), `regression` and `player-perf` green; `Test`, `Build`, `Test: runtime contract`, `CLI smoke`, `Studio: load smoke`, both producer suites and the two windows jobs still pending. So the block is the review gate plus pending, not a red required check.
- Unrelated but closes the open item from the #3392 thread: `main` at `a9ea07ed` finished fully green, including `Test` and `Tests on windows-latest`, the two that were still in flight at merge.

**Verdict:** REQUEST CHANGES
**Reasoning:** `routes/lint.ts:18` silently breaks the existing Studio file-tree lint badge for every per-file finding, verified by executing both branches on one fixture and by the fix passing every suite untouched; the coverage narrowing and the lost snapshot branch are in the same area and cheap to close in the same push. Everything else here is solid and independently mutation-verified.

— Rames Jusso

miguel-heygen:
> `file: finding.file ?? entry.file` inverts the precedence the old branch had.

Addressed at `3eb650d4a`: project results are always re-stamped with `entry.file`. The shared-route fixture now supplies an absolute `finding.file` and requires `index.html`; the real embedded Studio endpoint additionally asserts every returned path is relative, so the file-tree badge contract is pinned rather than inferred.

> HTML outside `index.html` + `compositions/` loses Studio lint coverage entirely.

Addressed by unioning the project results with single-file lint for every uncovered HTML path from the existing safe walk. `scenes/intro.html` is covered at both the shared route and real embedded API levels, while already-covered project files are not linted twice.

> `snapshot`'s `index.html` branch lost its only test. `publish` has the mirror gap.

Addressed with both branches on both commands: snapshot now covers `compositions/index.html` and `compositions/card.html`; publish covers `compositions/brand/index.html` and `compositions/card.html`. Mutation checks confirm disabling snapshot's index branch or publish's non-index branch reddens exactly the corresponding test.

> re-roots the archive ... so a composition referencing assets above that directory publishes without them.

The directory-form snapshot/publish guidance now says it is safe only when assets are self-contained under that directory; otherwise the authored file must be mounted from the project root. The unreachable root-candidate branch was removed.

Fresh validation on the final refactor: studio-server 450/450, CLI 2,809 passed / 3 skipped, all affected typechecks, full monorepo build, changed-file oxlint/oxfmt, `git diff --check`, Fallow audit, and pre-commit hooks pass. The PR body is updated to match the exact behavior and counts.

jrusso1020 (approved):
Re-review at `3eb650d4a` — my `CHANGES_REQUESTED` was pinned to `9d65e4d47`. All three findings are closed, and I re-verified each by mutation rather than by reading the diff.

**The blocker is fixed, and now pinned twice.** `routes/lint.ts:21` re-stamps `file: entry.file` after the spread, so absolute host paths can't reach the client. A real-adapter probe against `createStudioServer` on a three-file fixture (root + `compositions/index.html` + `scenes/intro.html`) returns 7 findings, every path project-relative — `index.html` x3, `compositions/index.html` x1, `scenes/intro.html` x3. And the mutation that stayed green last time now reddens: flipping line 21 back to `finding.file ?? entry.file` fails `studio-server` (1 of 450) *and* `cli/src/server/studioServer.test.ts`. The route test's stub returning an absolute `file` is what makes it able to see this line at all — that was the actual gap, and it's closed.

**The union closes the coverage gap without double-linting.** Probed a project carrying `compositions/components/button.html`, a snippet fragment, `.hyperframes/preset.html` and `scenes/intro.html`: 13 findings across 5 files, **zero duplicate `(file, code)` pairs**, and `.hyperframes/preset.html` correctly absent, so #1384 still holds. Stubbing `lintUncoveredHtml` to return `[]` reddens two studio-server tests plus one CLI test. The dedup key is sound cross-platform too — `walkDir` builds `${prefix}/${entry.name}` and `lintProject` normalizes with `.replace(/\\/g, "/")`, so `coveredFiles.has(file)` can't miss on Windows.

**Both hint branches are pinned, including the discriminator.** Disabling snapshot's `index.html` branch → red; disabling publish's non-index branch → red; and replacing `posix.basename(candidate) === "index.html"` with a bare `candidate` (always-directory form) → red. That third one is the one worth having: it proves the branch *condition* is tested, not just that each branch prints something.

Your numbers all reproduce locally at this head: lint 501/501, studio-server 450/450, CLI 2809 passed / 3 skipped.

**CI.** The red rows are attributable, so I'll write it down rather than wave at it: `regression` failure and `Test` failure belong to runs `32500079134` and `32500079118`, both **cancelled** by the 15:54:19 push, and the `Tests on windows-latest` cancellation to `32500079030`. Fresh runs `32500118717` / `32500118633` / `32500118634` are still in progress. So 5 of `main`'s 8 required contexts are green (Build, Render on windows-latest, Semantic PR title, Test: runtime contract, Typecheck) and three still owe a real result. My approval covers the code at `3eb650d4a`; it does not stand in for those three.

Two things I should have caught on the first pass. Both are live at the previous head as well, so neither is caused by the rework, and neither blocks.

**important — `hevc_preview_codec` is the linter's only info-severity rule, and Studio has no info tier.** It is reachable only through `lintProject` (`packages/lint/src/project.ts:236`), so before this PR it could not reach Studio at all. The route forwards findings unfiltered and `packages/studio/src/hooks/useLintModal.ts:20` collapses severity with `f.severity === "error" ? "error" : "warning"` — so any project containing an HEVC video now carries a permanent *warning* badge on `index.html` in the file tree and the lint modal. The rule's own comment says "Never escalated beyond 'info' — this must not fail lint or check", and the CLI honors that (`lintFormat.ts` drops info unless `verbose`). One line in `collectProjectFindings` settles it, in a file this PR already rewrites. Traced at source; I did not execute this one, which needs an HEVC fixture plus ffprobe on the box.

**important — the Studio lint endpoint now spawns subprocesses.** `lintProject` calls `lintHevcPreviewCodec`, which spawns one `ffprobe` per unique local video, 8 concurrent (`hevcPreviewLint.ts:25`), with an in-run cache only — there is no module-level memo, so nothing carries across requests. And the endpoint isn't hit only on modal open: `useLintModal.ts:72` re-lints in the background 1s after every `refreshKey` bump, which `useFileManager.ts:322,430` and `App.tsx:115` fire on file save and preview reload. That path spawned zero processes before this PR. It is bounded by `findFfBinary` returning null, which is not the common case for Studio users. A probe cache keyed on (path, mtime, size) would close it.

nits:
- Nothing in the repo references `createViteAdapter`, so `vite.adapter.ts:213` is the one adapter path with zero coverage. I checked the chain it leans on — `@hyperframes/core/lint` re-exports `* from "@hyperframes/lint"` and `packages/lint/src/index.ts:14` exports `lintProject` — so this is missing coverage, not a defect.
- `studioServer.test.ts`'s `!finding.file?.startsWith(projectDir)` passes vacuously for a finding with no `file` at all; the `toContainEqual` assertions above it carry the real weight.
- Worth recording so nobody re-files it: registry components and snippet fragments are excluded from `lintProject`'s `results`, so the union lints them through the single-file path *without* `isSubComposition` — the probe confirms both appear. That matches pre-PR behavior exactly, i.e. the suppression `project.ts` applies for the CLI still doesn't apply in Studio. Unchanged by this PR.

**Verdict:** APPROVE
**Reasoning:** the blocker and both coverage gaps are closed and independently verified by probe and mutation at `3eb650d4a`; the two remaining findings predate this rework and neither breaks the feature under review. Not merging — three required contexts are still pending, and that call isn't mine.

— Rames Jusso

PR #3397, opened 2026-08-21
test(engine): give the real-ffmpeg group mix suite a 60s timeout

vanceingalls:
`Tests on windows-latest` failed on #3290 with `Test timed out in 5000ms` at `mix level arithmetic > a group FX chain fully cutting its members leaves an ungrouped track untouched (routing isolation)`. #3290 merged before this fix landed on its branch, so it comes as its own PR.

Every case in that suite shells out to real ffmpeg at least twice — the mix under test plus a reference mix to compare it against. 5s is enough on Linux and not on the Windows runner: a slow host, not a hung mix.

`vi.setConfig({ testTimeout: 60_000 })` rather than a `describe(..., 60_000)` argument — the latter makes the callback a second parameter, and oxfmt then reindents all 170 lines of the suite for one number (354-line diff vs 8).

### DIFF
--- a/packages/engine/src/services/audioMixer.grouping.test.ts
+++ b/packages/engine/src/services/audioMixer.grouping.test.ts
@@ -74,6 +74,17 @@ const track = (id: string, end: number, volume = 1) => ({
   type: "audio" as const,
 });
 
+// Each case here runs at least two real ffmpeg mixes — the one under test plus
+// a reference mix to compare it against — so it is seconds of work, not
+// milliseconds. On the Windows runner `routing isolation` measures ~10s when the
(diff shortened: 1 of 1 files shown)


miga-heygen (approved):
**Approve.** Per-test timeout bump from 30s to 120s on the real-ffmpeg group mix suite — ~12x the healthy Linux time, generous enough to absorb Windows runner variance without masking a genuine stall. Comment explains the rationale clearly. Using per-test arguments rather than `vi.setConfig` is correct since the per-test arg overrides the file-level config. No issues.

— Miga

PR #3401, opened 2026-08-21
feat(audio): open the audio FX, group and mute features to everyone

vanceingalls:
The twelve-PR audio stack (#3274–#3292) is fully merged, and **all three of its canaries are still at 0%** on `main`. So the FX rack, the group rows, mute and solo are in the build and reachable by nobody.

This removes the gates rather than raising the numbers — a canary that gates nothing is a branch every future reader has to evaluate.

## Gone

- The `audio-fx-rack`, `audio-track-mute` and `audio-groups` registry entries.
- The five studio gates they fed:
  - `PropertyPanelFlat.tsx:434` — the Audio FX section
  - `TimelineTrackPlainHeader.tsx:40` — the Mute label
  - `TimelineTrackPlainHeader.tsx:96` — the muted strike-through
  - `TimelineTrackPlainHeader.tsx:114` — the solo button
  - `useTimelineTrackDerivations.ts:161` — the group-row derivation

Each feature now renders on its own precondition — an audio track, a grouped track — exactly as it did for an enrolled user.

## The test flips with it

`canary.test.ts` pinned `audio-fx-rack` as registered and at 0%, which is the opposite of what should hold now. Replaced with a pin that **no `audio-*` canary exists at all**: re-registering one silently re-hides a shipped feature, and nothing else in the tree would say so. Verified it fails when one is added back.

## Why not #3363

That branch carries the same removal, but it is 105 commits and 310 files divergent from `main` now that the stack squash-merged past it (its diff still reads +13883/-1795 across 188 files, all stale). Every other thing on it has landed via the stack, so this is the only unique content left there.

## Verification

core 2489 ✓ · studio 4346 ✓ · `tsc --noEmit` clean on core + studio · `typecheck:runtime` clean · fallow gate exit 0


miguel-heygen (approved):
Exact-head review at `0b3623ff56a0`.

The graduation is complete across the runtime surface, not just the registry:

- `PropertyPanelFlat.tsx:427` now exposes Audio FX whenever the existing audio-section precondition holds.
- `TimelineTrackPlainHeader.tsx:36-112` unconditionally applies audio mute labeling/icons, muted/group-muted strike-through, and solo when the callback exists.
- `useTimelineTrackDerivations.ts:156-162` always resolves group rows and membership; ungrouped tracks still pass through unchanged inside `groupTimelineTracks`.
- `canaryRegistry.ts:80-83` removes all three definitions, so CLI/Studio telemetry and manual overrides no longer publish dead rollout assignments.

I grepped the full tree for the three canary names and `HF_CANARY_AUDIO*`: no live source/test reference remains. The only remaining name is a historical weekly-update entry describing the earlier rollout, which should remain historical. The new registry test correctly fails if an `audio-*` canary is silently reintroduced.

All required checks are green: build, typecheck, runtime contract, Test, regression, and Windows render/tests. `BLOCKED` is only the missing approval gate.

Non-blocking nit: `TimelineTrackPlainHeader.tsx:9` still says the relabel is “Gated ... behind the canary”; delete that stale sentence. The source behavior itself is correct.

**Verdict:** APPROVE
**Reasoning:** Every one of the five authoring/UI gate sites and all three registry entries are removed consistently, while each feature retains the same enrolled-user preconditions and the full required suite is green.

— Magi

miguel-heygen (approved):
Corrected exact-head review at `02d7846be6f4`.

The source review from `0b3623ff` still holds, and I reviewed the two-file delta that landed during submission:

- `TimelineTrackPlainHeader.tsx` removes the stale “Gated behind the canary” sentence.
- `docs/weekly-updates.mdx` keeps the historical rollout context while explicitly saying the canary is retired, the feature is now on for everyone, and `HF_CANARY_AUDIO_FX_RACK` no longer has an effect.

Across the final tree, all five Studio gate sites and all three registry entries are removed. No live source/test reference to `audio-fx-rack`, `audio-track-mute`, `audio-groups`, or `HF_CANARY_AUDIO*` remains. The enrolled behavior is now unconditional on the same audio/group preconditions, and the registry test prevents accidental re-registration.

All required checks are green at the final head.

**Verdict:** APPROVE
**Reasoning:** The final delta only closes the two documentation nits; the canary removal remains complete and behaviorally consistent across registry, Studio UI, derivation, telemetry, and tests.

— Magi

jrusso1020 (commented):
Additive review at exact head `02d7846be6`. @miguel-heygen enumerated the five gate sites, the registry entries and the reference sweep; I re-derived those independently rather than inheriting them, and the three things below are what I'd add.

**Verified at source, so this isn't just convergence:**

- The 0% claim holds. `origin/main:packages/core/src/canaryRegistry.ts` has all three at `percentage: 0` (lines 85/86, 93/94, 103/104).
- Five gate sites is the complete set. Grepping `isCanaryEnabled("audio` on `origin/main` returns exactly `PropertyPanelFlat.tsx:434`, `TimelineTrackPlainHeader.tsx:40/96/114`, `useTimelineTrackDerivations.ts:161`. All five are removed here, and no `isCanaryEnabled("audio*")` call site remains at head.
- The no-group passthrough is real. With no `el.audioGroup`, `resolveGroupMembership` returns empty maps, so the loop in `groupTimelineTracks` pushes every `[trackNum, elements]` tuple in raw ascending order — value-identical to the old off-cohort `{ tracks: rawTracks, groups: [], trackGroupOf: new Map() }`. Only the array identity differs, and the `useMemo` keyed on `[rawTracks]` (`useTimelineTrackDerivations.ts:159-161`) makes that inert. Ungrouped projects see no layout change.
- The ratchet's projection is honest. `CANARIES` is imported from `./canaryRegistry.js` at `canary.test.ts:3` and there is no `vi.mock` of the registry anywhere in core/studio/cli, so `canary.test.ts:441` filters the real array with no name normalization.
- One point in the PR's favour that the body undersells: the mute vocabulary was already half-shipped. `packages/studio/src/hooks/timelineTrackVisibility.ts:224-231` writes `Mute track N` / `Unmute track N` into undo history **ungated**, so today every user clicks a button labeled "Hide track 2" and gets a history entry that says "Mute track 2". This PR closes that mismatch.

---

### important — the safety argument rests on a suite that has never covered this path, and hasn't finished now

At `percentage: 0`, `evaluateCanary` returns `out_of_cohort` at `packages/core/src/canary.ts:120`, before any bucketing. Studio additionally returns `excluded` for `navigator.webdriver` (`packages/studio/src/telemetry/canary.ts:221`), so a Playwright/Puppeteer-driven Studio is off-cohort even at a non-zero percentage. Only an explicit override could ever have reached the enrolled path — and grepping the whole tree for `hf_canary` / `HF_CANARY` outside the canary modules themselves returns **one hit, and it is a comment** (`packages/cli/src/server/telemetryIdentity.ts:151`). No test, workflow or e2e job sets one.

So these five call sites have never executed in the enabled state, on any platform, in any suite. The gated *components* have their own direct unit tests that never consult the gate; the *composition* of gate-on at the call site is new. That makes the required suite load-bearing on this PR rather than a formality — and as of this review, 5 of the 8 required contexts are still `in_progress`: `Typecheck`, `Test`, `Tests on windows-latest`, `Render on windows-latest`, and `regression` (all 9 shards). Green so far: `Build`, `Semantic PR title`, `Test: runtime contract`.

Worth flagging because it's the stated basis of the two approvals above: both say the required suite is green at the final head. Review `4997265715` was submitted at 20:37:52Z and the workflow runs for `02d7846be6` started at 20:38:03Z onward, so at submission time those checks had not started. No failures — just not yet run.

### important — `visibilityButtonLabel` names a state, not an action, and drops the track suffix

`TimelineTrackPlainHeader.tsx:9-12`:

```ts
if (showAsMute) return hidden ? "Muted" : "Mute";
return hidden ? `Show track${suffix}` : `Hide track${suffix}`;
```

The non-audio branch is action-phrased and per-track unique. The audio branch is neither, and `showAsMute` is now `Boolean(isAudioTrack)` for everyone. Two consequences arrive at 100%:

1. A muted audio track's only accessible name is "Muted", so a screen-reader user cannot tell that activating the control will unmute it.
2. With more than one audio track — music plus VO is the normal case — every visibility button carries the same accessible name, where each was previously unique.

`showAsMute` also drives the icon (`visibilityButtonIcon`, line 14-17), so this is the control's whole identity, not just its tooltip. The fix is two lines and the right vocabulary already exists at `timelineTrackVisibility.ts:227-228`:

```ts
if (showAsMute) return hidden ? `Unmute track${suffix}` : `Mute track${suffix}`;
```

That also makes the button and its undo-history entry agree, which is the mismatch this PR is otherwise fixing. Pre-existing code — but that is the point: it becomes reachable here.

### important — the producer of the newly-live layout has no direct test

Measured across `packages/studio/src/**/*.test.{ts,tsx}`: `groupTimelineTracks` 0 files, `useTimelineTrackDerivations` 0, `anchorKey` 0, `isSoloed` 0, `onToggleSolo` 0, `TimelineSoloButton` 0, `PlainTrackHeader` 0. `trackGroupOf` appears in three consumer tests (`useTimelineLogicalRows`, `timelineKeyboardNavigation`, `TimelineLanes`), all of which populate it synthetically.

So the function that now unconditionally reorders rows and emits synthetic fractional anchor rows for every user is untested at the unit level, and the solo control that now appears on every audio track has none either. A table test on `groupTimelineTracks` pinning (a) no-group input → identical order and `groups: []`, and (b) interleaved members → contiguous under `memberTracks[0] - 0.5`, is cheap and pins exactly the invariant both reviewers had to establish by reading.

### nit — the ratchet is prefix-shaped, so a rename escapes it

`canary.test.ts:441` catches a re-registered `audio-fx-rack` but not `audiofx-rack`, `fx-rack` or `audio_fx_rack`. Given the stated intent ("re-registering one of these re-hides a shipped feature"), pinning the three former names is the assertion that matches it — `findCanary` is already imported in this file:

```ts
expect(["audio-fx-rack", "audio-track-mute", "audio-groups"].filter((n) => findCanary(n))).toEqual([]);
```

Keep the prefix check alongside it if you want the family guard too.

### the framing, for the record

The body reads this as removing gates that "gate nothing." A canary at 0% gates *everything* — deleting it moves three features from nobody to everybody in one commit. The registry's own procedure prescribes the other order: ramp `0 → 5 → 25 → 100` (`canaryRegistry.ts:16-18`), then "at 100 and holding, DELETE the entry and the branch it guarded" (`:19-21`). And `evaluateCanary` carries a comment for precisely this hazard — it returns `in_cohort` at `pct >= 100` *ahead of* the exclude and unit-id checks because "at 100 the registry's step 4 says to delete the entry and the guard, so anything still resolving false here would take the new path for the FIRST time at deletion, unstaged" (`packages/core/src/canary.ts:122-127`). Going 0 → deleted skips every rung that comment is defending.

The title is honest about the outcome, so nothing is being hidden. But the right lens is a launch, not a cleanup, and worth confirming that shipping all three simultaneously is intended now rather than ramping the numbers.

**Verdict:** COMMENT
**Reasoning:** I found no defect in the five removals — each verified at source, and the ungrouped passthrough is genuinely behaviour-preserving — but 5 of 8 required contexts are still running and this PR is the first execution of these call sites in the enabled state, so the suite is doing real work here rather than confirming a known-good path. Happy to approve once `Typecheck`, `Test`, `Tests on windows-latest`, `Render on windows-latest` and `regression` land green; the label fix is the one finding I'd want in before merge.

— Rames Jusso

miguel-heygen (changes requested):
Correction to my approvals at exact `02d7846be6f4`: the gate-removal inventory is complete, but I approved before the final-head launch bar had actually run and missed one accessibility defect in code that was previously unreachable at 0%.

**Blocker — `TimelineTrackPlainHeader.tsx:9-11` exposes the wrong accessible button contract to every audio track.** The now-live audio branch returns `"Muted"` when the current track is hidden and `"Mute"` otherwise. Two problems ship together:

1. a hidden track's button names its current state (`Muted`) rather than its action (`Unmute`); and
2. the branch discards `suffix`, so every audio row has the identical accessible name even when several tracks are present.

The canonical action wording already exists in `timelineTrackVisibility.ts:224-231`: use `hidden ? \`Unmute track${suffix}\` : \`Mute track${suffix}\`` for the button too. Add a focused `VisibilityButton`/label test for both states and multiple display rows.

**Important — the newly universal branches have no direct tests.** A canary at 0% returns before bucketing, and Studio excludes webdriver enrollment, so prior CI never exercised these call sites enabled. `groupTimelineTracks`, `useTimelineTrackDerivations`, and `TimelineSoloButton` have no direct test; the new core test only proves registry removal. Please at minimum pin grouped-row derivation (group anchor + contiguous members + ungrouped passthrough) now that it runs for everyone.

**CI correction:** my prior review said all required checks were green at the final head, but the final-head workflows started after that review. At this correction, Build/Typecheck/runtime-contract/title are green while Test, Windows tests/render, and regression are still running. Branch protection still gates merge, but the review body must not claim evidence that did not exist.

The other reviewed facts remain true: all 3 registry entries and all 5 gate readers are removed; FX, mute/solo, and groups now use the enrolled-user preconditions unconditionally; stale docs/comments are corrected.

**Verdict:** REQUEST CHANGES
**Reasoning:** This is the first enabled execution of the 0%-canary UI, and it currently launches duplicate, state-not-action accessible labels on every audio row with no direct group-row regression coverage; final-head CI is also still in flight.

— Magi

jrusso1020 (commented):
Re-review at `f9311d74d` (previous head `02d7846be6`). Both findings from my earlier comment are closed. I re-derived them from the code rather than reading the diff and nodding — a PR that implements my own suggestion is exactly where I'd be least critical.

## Closed, verified at source

**The label fix is right, including the part the diff doesn't show.** `visibilityButtonLabel` receives the *current* state (the click sends `!hidden`, `TimelineTrackPlainHeader.tsx:62`), while `toggleTimelineTrackHidden` receives the *incoming* state (`timelineTrackVisibility.ts:206-211`), so the two branches read inverted — and they now emit the same string for the same click: current `hidden: false` → button "Mute track 2", incoming `hidden: true` → history "Mute track 2". The doc comment's claim about the inverted convention checks out.

**The new tests genuinely execute the newly-live path.** 81 test files already sit in `packages/studio/src/player/components/`, so collection was never in question, and `Test` is green at this head — that's what makes the coverage claim mean something rather than being an untested new file.

**The ratchet is better than what I suggested.** Keeping the family guard alongside the explicit three means a fourth `audio-*` canary still trips it. `retired.filter((name) => findCanary(name))` is a sound projection — `findCanary` returns `undefined` when absent, so the filter keeps only names that came back.

One thing I checked that is *not* a finding: `trackDisplaySuffix(null)` returns `""`, which would reinstate the duplicate-name problem — but `trackDisplayNumber`'s own doc (`timelineTrackDisplay.ts:22`) establishes null is unreachable by construction. Noting it so it doesn't get raised again.

## important — the same mismatch class, one level up: the two display numbers disagree once a group exists

The fix aligns the *wording* of the button and the undo entry. The *number* inside them is computed from two different orderings, and removing `audio-groups` is what makes them diverge.

- **Button:** `TimelineLanes.tsx:180` → `trackDisplayNumber(displayTrackOrder, trackNum)`. `displayTrackOrder` traces through `useTimelineTrackLayout.ts:296` to `trackOrder` at `useTimelineTrackDerivations.ts:172` — `tracks.map(...)` over `groupTimelineTracks` output, which carries synthetic anchor rows (`emitGroupRows` pushes `[info.anchorKey, []]`) and pulls members contiguous.
- **History:** `timelineTrackVisibility.ts:220-221` → `trackDisplayNumber(timelineTrackOrder(timelineElements), track)`. `timelineTrackOrder` (`timelineTrackDisplay.ts:13`) is a plain ascending sort of element-bearing keys, with no anchors.

Both sides take the same `expandedElements`, so the ordering is the whole difference. Using the fixture in your own new test — grouped order `[-0.5, 0, 2, 1]` against ascending `[0, 1, 2]` — track 0 is row 2 in the header and row 1 in history; track 1 is row 4 and row 2. Click mute on the group's first member: the button says "Mute track 2", the history entry records "Mute track 1".

Off-cohort this was impossible. The old branch returned `rawTracks` with `groups: []`, so `trackOrder` was ascending and element-bearing — identical to what the undo path computes. Sub-composition expansion doesn't cause it either, since those fractional keys carry elements and appear on both sides.

Graded important rather than blocking because the callback is unaffected: `onToggle(trackNumber, !hidden)` passes the real track key, so nothing mutes the wrong row — only the announced and recorded row number is wrong.

`timelineTrackDisplay.ts:9` says every user-visible track number, "whether it is rendered by a component or baked into an undo-history label, routes through here". They do route through the same function, with two different orderings, which is the hole in that guarantee. The click site already holds the correct number, so threading it into `toggleTimelineTrackHidden` instead of recomputing looks like the smaller fix.

## nit — residual from my earlier third finding

`TimelineSoloButton` and `PlainTrackHeader`'s muted strike-through are still newly-live surface with no direct test, now that `VisibilityButton` has one. Not worth a round on its own.

## CI at this head

6 of 8 required contexts green. `Tests on windows-latest` and `regression` are still pending — `regression` has no aggregate context yet because shard-2 is running, so its absence is pending, not passing. Nothing red.

**Verdict:** COMMENT
**Reasoning:** Both earlier findings are genuinely closed and re-verified at source, but the group-ordering mismatch above is newly live and unaddressed, and two required contexts haven't finished. My comment doesn't move the gate in either direction: @miguel-heygen's CHANGES_REQUESTED (`4997318533`, submitted at `02d7846be6`) is the live blocker, and what it needs is its author's re-review at this head rather than another commit — the two findings it adopted are the ones now fixed.

— Rames Jusso

vanceingalls:
All four findings are addressed at `3faaffa8b`. Each fix was checked by reverting it and confirming the new test fails, so none of them is a test that passes either way.

## @miguel-heygen's blocker — the accessible label

`visibilityButtonLabel`'s audio branch is now `Unmute track${suffix}` / `Mute track${suffix}`. Both halves of the defect are gone: it names the action rather than the current state, and it keeps the suffix, so two audio rows are distinguishable.

The doc comment above it records why the two branches read inverted — the button gets the current state (`onToggle(trackNumber, !hidden)`), `toggleTimelineTrackHidden` gets the incoming one — which is the thing that makes "use the same wording" look wrong at a glance.

`TimelineTrackPlainHeader.test.tsx` pins both audio states, two rows staying distinct, the visual branch unchanged, and the callback still acting on the real fractional key. Restoring `"Muted" / "Mute"` fails three of them.

## @jrusso1020's ordering mismatch — real, and thanks for catching it

You were right that the *wording* fix left the *number* wrong, and right about which fix is smaller. `onToggleTrackHidden` now carries the display row the clicked control rendered, and `toggleTimelineTrackHidden` prefers it over deriving its own; the fallback stays for the keyboard and programmatic paths, where no group has reordered anything.

Your worked example is the test: `displayNumber: 2` on a track whose own derivation would say 1, asserting the history records `Mute track 2`. It fails if the parameter is ignored.

This is what `timelineTrackDisplay`'s "one owner of what track number does the user see" was already promising — two derivations were the hole in it, not the function.

## Both "no direct tests" findings

- `useTimelineTrackDerivations.test.ts` — ungrouped input stays in raw ascending order with `groups: []` and an empty `trackGroupOf`; an interleaved group's members become contiguous under an anchor at `memberTracks[0] - 0.5` while the ungrouped track between them keeps its place; plus label/volume/mute mirroring and the id fallback.
- The residual nit is closed too: the solo button's presence and pressed state, its absence on a visual track, and the strike-through for both a row's own mute and a group mute, with the title that names which.

## The ratchet

Took your version and kept the family guard beside it: the three retired names pinned individually via `findCanary`, plus the `audio-*` prefix check, so a rename escapes neither.

## On the framing — 0 → deleted, deliberately

Fair challenge, and worth answering rather than leaving implied: shipping all three at once is the intent, decided by the owner, not an oversight of the ramp.

The reasoning is that the ramp buys staged exposure, and these three canaries never had any. At `percentage: 0` for the whole stack, plus Studio's `navigator.webdriver` exclusion, nobody was ever enrolled — no cohort, no telemetry from a real user, no soak. Ramping `0 → 5` now would stage *the same unexercised code*, just more slowly, and the twelve PRs have already merged, so the flag is not gating unreviewed work — only reachability.

That does put the weight on this PR's suite, exactly as you said. Which is why the two untested-surface findings mattered more than the label one: they were the actual gap, and they are the reason this now has direct coverage rather than the reviews' reading standing in for it.

## CI

49 green, 0 failing at `3faaffa8b`. Still running: `Tests on windows-latest` and regression shards 2, 4, 7, 8. I'll flag it if any of them turns.

@miguel-heygen — the two findings your `CHANGES_REQUESTED` named are the ones fixed here, so this needs your re-review at this head rather than another commit.

jrusso1020 (commented):
Re-review at `3faaffa8b0` (previous head `f9311d74da`). The round-2 finding is closed with the right fix, and the nit alongside it is closed too. One new nit, plus a diagnosis of the red check — which isn't yours.

## Closed

**The display-number divergence is fixed at the source rather than papered over.** `displayNumber` now rides the click (`TimelineTrackPlainHeader.tsx:62`) into `toggleTimelineTrackHidden`, which prefers it over its own derivation (`timelineTrackVisibility.ts:227`). I re-walked the call graph rather than trusting the shape of the diff: every production path into `onToggleTrackHidden` goes through `VisibilityButton` — `TimelineTrackPlainHeader.tsx:119` and `TimelineTrackHeader.tsx:494` are the only two, both rendering that component — so the number is always supplied in the UI, and the `??` fallback is defensiveness rather than a second live path.

Pinned at more levels than I asked for: the unit case on `toggleTimelineTrackHidden` itself, plus the third argument asserted at `VisibilityButton`, `TimelineTrackHeader` (`1/6, true, 1` — good that this one uses the fractional sub-comp key), `TimelineLanes` and `Timeline`.

**The earlier nit is closed:** `PlainTrackHeader` now covers solo presence and absence, own-mute strike-through, group-mute strike-through with its title, and the unstruck case.

## nit — one declaration didn't get the new parameter

`useTimelineEditCallbacks.ts:40` still types it as two arguments:

```ts
handleToggleTrackHidden: (track: number, hidden: boolean) => Promise<void> | void;
```

No runtime effect today: it is an unwrapped passthrough (destructured `:117`, assigned to `onToggleTrackHidden` at `:206`), and a two-parameter type is assignable to the three-parameter slot, so the real function still receives `displayNumber`. The cost is that this boundary now understates the contract — wrap it here as `(track, hidden) => …` and the third argument disappears silently, no type error, bug back. `NonNullable<TimelineEditCallbacks["onToggleTrackHidden"]>` would make it carry the contract, which is the shape the very next line already uses for `handleTimelineGroupResize`.

Small accuracy note on the new doc comment: it names "a keyboard path" among the fallback callers, and I could not find one — no keyboard or shortcut handler toggles track visibility today. Better not to imply a caller that doesn't exist.

## The red check is not yours

`Tests on windows-latest` is red at this head (run `32528342406`, attempt 1, completed, not superseded). It isn't this PR:

- The failure is `packages/engine/src/services/audioMixer.grouping.test.ts` › "a group FX chain fully cutting its members leaves an ungrouped track untouched (routing isolation)", and this PR touches no file under `packages/engine`.
- It failed by timeout — `30015ms` against a 30000ms budget.
- The same test passed at your previous head in the same job (run `32526258747`) in `19311ms`, 64% of budget, with the engine code byte-identical between the two heads. That file needed `30561ms` for its 5 tests even in the passing run.

A rerun should clear it. The durable problem is that this test sits at two-thirds of its timeout on a Windows runner — worth its own ticket, not yours to fix here.

Both gates, so the picture is complete: 7 of 8 required contexts are green with only that one red, and on the review axis @miguel-heygen's CHANGES_REQUESTED (`4997318533`) is still recorded against `02d7846be6`, so it needs its author's re-review at this head. The ruleset also sets `require_last_push_approval: true` with `dismiss_stale_reviews_on_push: false`, so a fresh approval at this head is required regardless of which gate clears first.

**Verdict:** COMMENT
**Reasoning:** Everything I raised across both rounds is closed and re-verified at source, and my only open item is a nit. Withholding a stamp because a required check is red and a CHANGES_REQUESTED is live — not because of the code.

— Rames Jusso

miguel-heygen (commented):
Re-reviewed exact head `3faaffa8b00ed2754fa140226beff5b329d8d1f4`.

Both code blockers from my prior review are closed:

- The audio visibility control now exposes the action, not the state: `Mute track N` / `Unmute track N`. The display suffix is preserved, multiple rows have distinct accessible names, and the tests cover both states plus the rendered callback arguments.
- The newly universal group-row path now has direct coverage for an ungrouped passthrough, a group anchor, contiguous grouped members, ordering, and group metadata/defaults.

I also verified the launch ratchet now names all three retired canaries and rejects every `audio-*` registry entry, and the rendered display number is carried into `toggleTimelineTrackHidden` so the button label and undo-history label cannot diverge on synthetic group anchors.

One non-blocking type-contract nit remains: `useTimelineEditCallbacks.ts:40` still declares `handleToggleTrackHidden` with two parameters even though `TimelineEditCallbacks["onToggleTrackHidden"]` now has the optional display-number parameter. Runtime is correct today because it is a direct passthrough, but using `NonNullable<TimelineEditCallbacks["onToggleTrackHidden"]>` would stop a future wrapper from silently dropping the third argument.

The exact-head required suite is 7/8 green. `Tests on windows-latest` failed in untouched `packages/engine/src/services/audioMixer.grouping.test.ts` by 15 ms over its 30 s timeout; the same byte-identical engine test passed on the prior head. This is an unrelated runner-sensitive timeout, but it is still a required red context and the PR body does not document an exception.

**Verdict:** COMMENT
**Reasoning:** The requested code changes are complete and clean at the current head. I am withholding approval only until the unrelated required Windows check reruns green; my earlier CHANGES_REQUESTED therefore remains live for now.

— Magi

PR #3405, opened 2026-08-21
fix(producer): anchor local-font embedding to its url() occurrence

### LINKED ISSUE #3369: Local-font embedding: file:// misclassified, and an unanchored replaceAll can corrupt unrelated URLs
## Describe the bug

Two defects in the local-font embedding pass in `packages/producer/src/services/htmlCompiler.ts`.
The second is the serious one: an unanchored `String.replaceAll` over the entire compiled
document can corrupt unrelated URLs.

## 1. `file:` is not excluded from the local-font pattern

```js
// packages/producer/src/services/htmlCompiler.ts:1658
const LOCAL_FONTFACE_URL_RE = /url\(["']?(?!data:|https?:\/\/)([^"')]+)["']?\)/gi;
```

The negative lookahead skips `data:` and `http(s)://` but not `file:`. So a
`@font-face { src: url("file:///abs/path/font.ttf") }` is classified as a *local project
path* and passed to `resolve(projectDir, localPath)`. Since the captured string starts with
`f`, not `/`, it is treated as relative and resolved into a path like
`<projectDir>/file:/abs/path/font.ttf`, which does not exist.

The subsequent read throws into an empty catch (line 1731):

```js
} catch {
  // File read or compression failed — keep the original path
}
```

so the failure is silent, the rule is left untouched, and the browser then refuses the
`file://` fetch with `net::ERR_FAILED`. The composition renders with the CSS fallback font
and `render` exits `0`.

## 2. The embed replaces by bare substring across the whole document

```js
// packages/producer/src/services/htmlCompiler.ts:1729
result = result.replaceAll(localPath, dataUri);
```

`localPath` here is a bare relative path such as `fonts/testfont.ttf`, with no surrounding
`url(...)` or quotes. Every other replacement in this file anchors on the surrounding
syntax:

```js
// :1310
result = result.replaceAll(`"${url}"`, `"${relPath}"`).replaceAll(`'${url}'`, `'${relPath}'`);
// :1434
(h, url, rel) => h.replaceAll(`url(${url})`, `url(${rel})`),
// :1654
(h, url, relPath) => h.replaceAll(`url(${url})`, `url("${relPath}")`),
```

Line 1729 is the only unanchored one, so it matches that path text *anywhere* in the
compiled HTML — including inside a longer URL that merely ends with the same characters.

Concretely: a document containing both

- `url("fonts/testfont.ttf")` — correctly embedded, and
- `url("file:///abs/path/fonts/testfont.ttf")` — the untouched rule from defect 1

produces, after the global replace, a corrupted second rule of the form
`url("file:///abs/path/data:font/woff2;base64,...")`. This was observed in the compiled
debug output under `.debug/<jobId>/compiled/index.html` while investigating defect 1.

The collision does not require `file://` to be involved. Any two paths where one is a
trailing substring of the other — `img/logo.png` and `assets/img/logo.png`, `a/font.ttf`
and `b/a/font.ttf` — are enough for the shorter one's embed to rewrite text inside the
longer one.

## Expected behavior

An absolute `file://` font src is either embedded or left alone, and either way the
compiler does not modify text belonging to a different URL.

## Actual behavior

The `file://` rule is silently skipped, and if a relative font with a colliding path
suffix is embedded in the same document, the `file://` rule is rewritten into a URL that
is neither the original nor a valid data URI.

## Suggested fix

Both are small and independent:

1. Add `file:` to the lookahead in `LOCAL_FONTFACE_URL_RE`, so an absolute `file://` src is
   not mistaken for a project-relative path. Consider also surfacing the swallowed failure
   at line 1731 — an empty catch on a font read means a composition can silently render
   with the wrong typeface.
2. Replace by matched `url()` occurrence rather than by bare substring, matching what
   lines 1310, 1434 and 1654 already do. The regex at line 1699 already yields match
   positions, so the anchored form is available without new parsing.

Fix 2 matters more than the `file://` case that surfaced it: today the blast radius is any
project where one font path is a suffix of another.

## Environment

```shell
hyperframes  0.8.4
Affected     packages/producer/src/services/htmlCompiler.ts:1658, 1729, 1731
```

miguel-heygen:
## What

Anchor local-font embedding to its `url(...)` occurrence, and stop classifying an absolute
`file://` src as a project-relative path.

Closes #3369.

## Why

The embed step rewrote the whole compiled document with a bare substring replace:

```js
result = result.replaceAll(localPath, dataUri);
```

`localPath` is a plain relative path like `fonts/x.ttf`, with no surrounding `url(` or
quotes. So it also rewrites that text **anywhere else it appears** — including inside a
longer URL whose tail happens to match.

Every sibling rewrite in this file already anchors on the surrounding syntax
(`url(${url})`, `"${url}"`). This one was the outlier.

### The corruption, run literally

```
$ # pre-fix behaviour: html.replaceAll("fonts/x.ttf", dataUri)
  BEFORE FIX -> b { src: url("file:///abs/data:font/woff2;base64,AAAA") }
```

That is neither the original URL nor a valid data URI.

It does not need `file://` to happen. **Any two paths where one is a suffix of the other**
collide — `img/logo.ttf` and `assets/img/logo.ttf` are enough — so the blast radius is any
project with nested asset directories.

## How

Replace by matched `url()` occurrence:

```ts
export function urlOccurrenceRe(localPath: string): RegExp {
  const escaped = localPath.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
  return new RegExp(`url\\((["']?)${escaped}\\1\\)`, "g");
}
```

Exported so the collision can be tested directly — it is invisible in ordinary projects and
trivial to reintroduce, since the naive form passes every test that uses a single font.

Two smaller changes in the same area:

- **`file:` added to `LOCAL_FONTFACE_URL_RE`'s exclusion list**, alongside `data:` and
  `http(s):`. Without it, `url("file:///abs/font.ttf")` was read as project-relative,
  resolved to `<projectDir>/file:/abs/font.ttf`, and the failed read was swallowed.
- **The empty `catch {}` now logs.** A font that cannot be read must not fail the render, but
  a silent skip means the composition renders in a fallback typeface with nothing saying why.

## Test plan

### Regression tests

`htmlCompiler.fontEmbed.test.ts`, five cases:

| case | asserts |
|---|---|
| exact occurrence, quoted / bare / single-quoted | all three forms rewrite |
| longer url whose tail matches | `url("file:///abs/fonts/x.ttf")` survives untouched |
| sibling suffix collision, no `file://` | `assets/img/logo.ttf` survives when `img/logo.ttf` embeds |
| path mentioned in a CSS comment | text outside `url()` is left alone |
| path containing regex metacharacters | `fonts/x+y(1).ttf` is matched literally, not as a pattern |

The last one matters because the fix introduces a regex where there was none — an unescaped
path would be a new bug in place of the old one.

### Suites

```
$ bunx vitest run packages/producer/src/services/htmlCompiler.fontEmbed.test.ts
  Tests  5 passed (5)

$ bun test htmlCompiler.test.ts htmlCompiler.animatedGif.test.ts htmlCompiler.naturalDuration.test.ts
  132 pass, 0 fail, 361 expect() calls
```

(The three existing suites use `bun:test` rather than vitest, so they need `bun test`.)

- `oxlint` + `oxfmt` on both touched files — clean

### DIFF
--- a/packages/producer/src/services/htmlCompiler.fontEmbed.test.ts
+++ b/packages/producer/src/services/htmlCompiler.fontEmbed.test.ts
@@ -1,0 +1,49 @@
+import { describe, expect, it } from "vitest";
+import { urlOccurrenceRe } from "./htmlCompiler.js";
+
+/**
+ * The embed step used `result.replaceAll(localPath, dataUri)` — a bare
+ * substring rewrite over the whole compiled document. These pin the collision
+ * that made unsafe: it is invisible in ordinary projects and easy to
+ * reintroduce, because the naive form passes every test that uses one font.
+ */
+describe("font embed url() anchoring", () => {
+  const DATA_URI = "data:font/woff2;base64,AAAA";
+  const replace = (html: string, path: string) =>
+    html.replace(urlOccurrenceRe(path), `url("${DATA_URI}")`);
+
+  it("rewrites the exact occurrence, quoted or bare", () => {
+    expect(replace(`src: url("fonts/x.ttf")`, "fonts/x.ttf")).toBe(`src: url("${DATA_URI}")`);
+    expect(replace(`src: url(fonts/x.ttf)`, "fonts/x.ttf")).toBe(`src: url("${DATA_URI}")`);
+    expect(replace(`src: url('fonts/x.ttf')`, "fonts/x.ttf")).toBe(`src: url("${DATA_URI}")`);
+  });
+
+  it("does not corrupt a longer url whose tail matches the embedded path", () => {
+    // The reported corruption: embedding `fonts/x.ttf` rewrote the tail of an
+    // untouched absolute rule, producing `url("file:///abs/<data-uri>")`.
+    const html = `a { src: url("fonts/x.ttf") } b { src: url("file:///abs/fonts/x.ttf") }`;
+    const out = replace(html, "fonts/x.ttf");
+    expect(out).toContain(`a { src: url("${DATA_URI}") }`);
+    expect(out).toContain(`b { src: url("file:///abs/fonts/x.ttf") }`);
+  });
+
+  it("does not corrupt a sibling whose path is a suffix of another project path", () => {
+    // No file:// needed. Any two paths where one is a suffix of the other
+    // collide under a bare substring replace.
+    const html = `a { src: url("img/logo.ttf") } b { src: url("assets/img/logo.ttf") }`;
+    const out = replace(html, "img/logo.ttf");
+    expect(out).toContain(`a { src: url("${DATA_URI}") }`);
+    expect(out).toContain(`b { src: url("assets/img/logo.ttf") }`);
+  });
+
+  it("leaves the path alone outside a url() wrapper", () => {
+    const html = `/* see fonts/x.ttf for the source */ a { src: url("fonts/x.ttf") }`;
+    const out = replace(html, "fonts/x.ttf");
+    expect(out).toContain("/* see fonts/x.ttf for the source */");
+  });
+
+  it("treats regex metacharacters in a path literally", () => {
+    const html = `a { src: url("fonts/x+y(1).ttf") }`;
+    expect(replace(html, "fonts/x+y(1).ttf")).toBe(`a { src: url("${DATA_URI}") }`);
+  });
+});
--- a/packages/producer/src/services/htmlCompiler.ts
+++ b/packages/producer/src/services/htmlCompiler.ts
@@ -1668,7 +1668,21 @@ export async function localizeRemoteFontFaces(
   );
 }
 
-const LOCAL_FONTFACE_URL_RE = /url\(["']?(?!data:|https?:\/\/)([^"')]+)["']?\)/gi;
+// `file:` joins data: and http(s): in the exclusion list. Without it an
+// absolute `file:///abs/path/font.ttf` src was read as a project-RELATIVE path,
+// resolved to `<projectDir>/file:/abs/path/...`, and the failed read was
+// swallowed below — leaving the rule untouched for the browser to reject.
+const LOCAL_FONTFACE_URL_RE = /url\(["']?(?!data:|file:|https?:\/\/)([^"')]+)["']?\)/gi;
+
+/**
+ * Match one `url(<path>)` occurrence, with or without quotes, for a literal
+ * path. Exported for tests: the suffix-collision it prevents is invisible in
+ * ordinary projects and easy to reintroduce.
+ */
+export function urlOccurrenceRe(localPath: string): RegExp {
+  const escaped = localPath.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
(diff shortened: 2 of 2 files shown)


terencecho (approved):
**Verdict: LGTM.** Correct fix for the suffix-collision bug, tests cover the three failure modes that matter, and the two adjacent changes (`file:` exclusion + logged catch) are the right shape.

### Fix is correct

`urlOccurrenceRe` uses the standard regex-escape set (`[.*+?^${}()|[\]\\]`) and a `\1` back-reference so opening/closing quote must match. Verified end-to-end against the reported corruption plus a `local()` + `format()` fallback chain and a regex-meta path:

```
input:  a { src: url("fonts/x.ttf") } b { src: url("file:///abs/fonts/x.ttf") }
output: a { src: url("data:font/woff2;base64,AAAA") } b { src: url("file:///abs/fonts/x.ttf") }

input:  @font-face { src: local("X"), url(fonts/y.ttf) format("woff2"), url("fonts/z.ttf") format("truetype"); }
output: @font-face { src: local("X"), url("data:font/woff2;base64,AAAA") format("woff2"), url("fonts/z.ttf") format("truetype"); }
```

`local()` and `format()` siblings survive; only the matched `url()` is rewritten. Multi-`url()` `@font-face` blocks are handled because the extractor loop feeds each `localPath` to the anchored replace in turn — regex-collision-free even across sibling `url()`s where one path is a suffix of another.

### Sibling `.replaceAll` audit — clean

Only three `replaceAll(...url...)` sites in `htmlCompiler.ts`; the other two already anchor via `` `url(${url})` `` (L1468, L1667) or via quotes (L1344). This PR fixes the last outlier. No other bare-substring rewrites of a font path.

### `file:` exclusion

Correctly adds `file:` to `LOCAL_FONTFACE_URL_RE`'s negative lookahead. Prior to the fix, `url("file:///abs/font.ttf")` was silently classified as project-relative, resolved to `<projectDir>/file:/abs/...`, and the read failure was swallowed by the empty `catch`. Now correctly ignored at extraction — and the `catch` logs a `warn`, so future silent-skip regressions surface.

### Test coverage

Five cases including the suffix-collision (`img/logo.ttf` vs `assets/img/logo.ttf` — no `file://` needed) and regex-meta path (`fonts/x+y(1).ttf`). The metachar test matters exactly as the PR body says: the fix introduces a regex where there was none, and an un-escaped path would be a new bug in place of the old.

### Nit (non-blocking)

Whitespace inside `url()` — `url(  "fonts/x.ttf"  )` — isn't matched by either `LOCAL_FONTFACE_URL_RE` (extraction) or `urlOccurrenceRe` (rewrite). Consistent (extraction and rewrite reject together, so no corruption), pre-existing behavior, and CSS authored via `hyperframes` won't hit it in practice.

CI green across all required checks (Producer unit + integration, regression shards 1–9, Windows tests, Typecheck, Preflight, Lint, Format, Fallow).

— Review by tai (pr-review)

PR #3413, opened 2026-08-22
fix(studio): cap the timeline FX popover to its gap, and scroll the list inside

vanceingalls:
## Context — this is a restore, not new work

The audio-groups feature currently on `main` (#3274, #3286, #3287, #3289, #3290, #3291, #3292, and the flag flip #3401) was **re-split out of the branch behind #3363**, which is still open and has never been reviewed. The five feature commits were carried across under new SHAs; the follow-up fix commits in that PR were not.

This stack restores three of those fixes, re-implemented against `main`'s current files (the originals will not cherry-pick — `main`'s `TimelineFxPopover.tsx` was re-authored during the split, and #3363's branch now conflicts with `main` in 55 files including `add/add` on this very file).

**Do not merge #3363 to get these.** After the split it went on to *remove* things `main` shipped — `2199f55c3` removes mute/solo (#3291) and `1b16a4c87` removes the group volume slider and meter (#3290), both breaking changes.

## The bug

The popover grew to whatever the preset list needed. On a short window it ran off the top or the bottom of the viewport and took its footer (`+ effect` / `Open rack`) with it, and nothing scrolled — so presets past the edge were simply unreachable.

## The fix

It caps to the space on whichever side it opens toward, and the preset list scrolls inside that while the footer stays put.

`min-h-0` on the scroller is load-bearing: a flex child defaults to `min-height: auto` and would refuse to shrink, pushing the footer out of the popover instead of scrolling.

`spaceAbove` is named for the cap's benefit. It equals `anchorRect.top`, so **the flip condition itself is unchanged** — this is a rename, not a behaviour change.

## Tests

Four, because the original fix shipped with none — which is exactly how it silently reverted:

- the downward cap
- the upward cap
- the usable-minimum clamp
- the footer being a *sibling* of the scroller, not inside it

Verified all four fail with the cap removed.

## Checks

`tsc --noEmit` clean · studio suite 4376 passed / 0 failed · `oxlint` 0 warnings 0 errors · `oxfmt --check` clean · `fallow audit --base origin/main` no issues.

### DIFF
--- a/packages/studio/src/components/editor/TimelineFxPopover.test.tsx
+++ b/packages/studio/src/components/editor/TimelineFxPopover.test.tsx
@@ -14,6 +14,18 @@ function rect(top: number, bottom: number): DOMRect {
   return { left: 0, top, right: 0, bottom } as DOMRect;
 }
 
+/** Pin the viewport height: every expected number below is derived from it, and
+ *  happy-dom's 768 default is not something this file should silently inherit. */
+function withViewportHeight<T>(value: number, run: () => T): T {
+  const previous = window.innerHeight;
+  Object.defineProperty(window, "innerHeight", { value, configurable: true });
+  try {
+    return run();
+  } finally {
+    Object.defineProperty(window, "innerHeight", { value: previous, configurable: true });
+  }
+}
+
 function dialogOf(host: HTMLElement): HTMLElement {
   const el = host.querySelector('[role="dialog"]');
   if (!el) throw new Error("no dialog");
@@ -112,35 +124,58 @@ describe("TimelineFxPopover", () => {
     expect(onClose).not.toHaveBeenCalled();
   });
 
-  // These four guard the regression that shipped once already: an uncapped
+  // These guard the regression that shipped once already: an uncapped
   // popover grew past the gap it opened into, ran under the timeline chrome and
   // took its footer with it, and nothing scrolled.
   it("caps its height to the space below when it opens downward", () => {
     // spaceBelow 738 > spaceAbove 10, so it opens down: 738 - margin(8) - gap(4).
-    const { host } = mount({ anchorRect: rect(10, 30) });
-    const dialog = dialogOf(host);
+    const dialog = withViewportHeight(768, () =>
+      dialogOf(mount({ anchorRect: rect(10, 30) }).host),
+    );
     expect(dialog.style.top).toBe("34px");
     expect(dialog.style.maxHeight).toBe("726px");
   });
 
   it("caps its height to the space above when it flips upward", () => {
     // spaceBelow 48 < 260 and spaceAbove 700 is larger, so it flips up.
-    const { host } = mount({ anchorRect: rect(700, 720) });
-    const dialog = dialogOf(host);
+    const dialog = withViewportHeight(768, () =>
+      dialogOf(mount({ anchorRect: rect(700, 720) }).host),
+    );
     expect(dialog.style.bottom).toBe("72px");
     expect(dialog.style.maxHeight).toBe("688px");
   });
 
   it("never caps below a usable minimum, however tight the gap", () => {
-    const innerHeight = window.innerHeight;
-    Object.defineProperty(window, "innerHeight", { value: 200, configurable: true });
-    try {
-      // Both gaps are tiny (80 below / 100 above); the cap must not collapse.
-      const { host } = mount({ anchorRect: rect(100, 120) });
-      expect(dialogOf(host).style.maxHeight).toBe("160px");
-    } finally {
-      Object.defineProperty(window, "innerHeight", { value: innerHeight, configurable: true });
-    }
+    // Both gaps are tiny (80 below / 100 above); the cap must not collapse.
+    const dialog = withViewportHeight(200, () =>
+      dialogOf(mount({ anchorRect: rect(100, 120) }).host),
+    );
+    expect(dialog.style.maxHeight).toBe("160px");
+  });
+
+  it("keeps both edges in the viewport when the minimum exceeds the gap", () => {
+    // The case the minimum used to lose: at 200px of viewport (roughly 400% zoom
+    // on a laptop) neither gap can hold 160px, so honouring the floor has to
+    // slide the box in rather than hang its top edge off-screen.
+    const dialog = withViewportHeight(200, () =>
+      dialogOf(mount({ anchorRect: rect(100, 120) }).host),
+    );
+    // bottom:32 + maxHeight:160 puts the box at y = 8..40 — a margin on each side.
+    expect(dialog.style.bottom).toBe("32px");
+    const bottom = Number.parseFloat(dialog.style.bottom);
+    const height = Number.parseFloat(dialog.style.maxHeight);
+    expect(bottom).toBeGreaterThanOrEqual(8);
+    expect(200 - bottom - height).toBeGreaterThanOrEqual(8);
+  });
+
+  it("shrinks below the minimum only when the whole window is shorter", () => {
+    // A floor against a tight gap is not a floor against a tight window: 120px of
+    // viewport cannot hold 160px, and hanging off the edge is worse than short.
+    const dialog = withViewportHeight(120, () =>
+      dialogOf(mount({ anchorRect: rect(60, 80) }).host),
+    );
+    expect(dialog.style.maxHeight).toBe("104px");
+    expect(dialog.style.bottom).toBe("8px");
   });
 
   it("scrolls the preset list and leaves the footer outside the scroller", () => {
--- a/packages/studio/src/components/editor/TimelineFxPopover.test.tsx
+++ b/packages/studio/src/components/editor/TimelineFxPopover.test.tsx
@@ -10,6 +10,16 @@ import { TimelineFxPopover } from "./TimelineFxPopover.js";
 const EMPTY_CHAIN: HfAudioFxChain = { version: 1, nodes: [] };
 const RECT = { left: 0, top: 0, right: 0, bottom: 0 } as DOMRect;
 
+function rect(top: number, bottom: number): DOMRect {
+  return { left: 0, top, right: 0, bottom } as DOMRect;
+}
+
+function dialogOf(host: HTMLElement): HTMLElement {
+  const el = host.querySelector('[role="dialog"]');
+  if (!el) throw new Error("no dialog");
+  return el as HTMLElement;
+}
+
 function byTextButton(host: HTMLElement, text: string): HTMLButtonElement | undefined {
   return Array.from(host.querySelectorAll("button")).find((b) => b.textContent?.includes(text));
 }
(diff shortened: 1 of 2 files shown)


jrusso1020 (commented):
First review, no prior reviews or comments on this PR. Read both files end to end at `f3ccd7fc4a`, plus `useContextMenuDismiss` and the sibling portals for the stack context.

## Strengths

**The cap arithmetic is right on both sides, and it is the kind of thing worth checking rather than trusting.** Downward, `top = anchorRect.bottom + 4` leaves `spaceBelow - 4` of room, and `available = spaceBelow - VIEWPORT_MARGIN - 4` lands exactly one margin short of the viewport edge. Upward, `bottom = innerHeight - anchorRect.top + 4` puts the box's bottom edge at `anchorRect.top - 4`, so `spaceAbove - VIEWPORT_MARGIN - 4` leaves the same margin at the top. Symmetric, no off-by-one in either direction.

**The `spaceAbove` claim is exactly true, so the rename really is free.** `anchorRect.top` is substituted verbatim into the same comparison (`TimelineFxPopover.tsx:32-33`), leaving `openUpward` semantically identical — the flip behaviour is untouched, as the comment says.

**`min-h-0` is correctly identified as the load-bearing part** (`:116-118`). Without it the flex child's `min-height: auto` floor holds the scroller at content height and the footer gets pushed out of the box instead of the list scrolling — which is the original bug wearing a different hat. `shrink-0` on the footer (`:131`) completes the pair.

**The footer test pins structure rather than styling**, which is the assertion that actually matches the bug: `scroller.contains(footer) === false` *and* `dialog.contains(footer) === true`. A class-name check would have passed on the broken markup too.

Listener cleanup is intact (`:85-86`), and the four new tests are genuinely four — the file was pre-existing and the diff adds exactly these cases.

## Nits

**The new tests inherit an ambient viewport height that only one of them pins.** `34px` / `726px` / `72px` / `688px` in the first two cases are all derived from happy-dom's default `window.innerHeight` of 768, and nothing in the file records that dependency. The third case already shows the pattern — `Object.defineProperty` with a restore in `finally` — so pinning it in all three makes the expected numbers derivable from the test itself and immune to a happy-dom default change. That matters more here than it usually would: the whole reason this file exists is to stop this fix reverting quietly, and a test that breaks with confusing arithmetic after a dependency bump is a test people delete.

**Not a finding, just so it is not mistaken for something the cap introduced:** `clampedStyle` is called inline in the style prop (`:112`) and reads `window.innerHeight` at render, so with the popover open across a window resize the cap goes stale exactly as the position already did. Pre-existing, and outside-pointerdown dismissal covers most real interaction.

## Gate

`Tests on windows-latest` is `pending` at this head — the only non-terminal required context; the other 44 are pass or skipping. Worth knowing that this is the same job that hit a 30s timeout in `audioMixer.grouping.test.ts` on an unrelated PR earlier today, so if it comes back red, check the failing test before assuming it is this diff.

**Verdict:** COMMENT
**Reasoning:** The fix is correct where I could verify it — cap arithmetic on both sides, the rename's equivalence, and the flex mechanics — and my only open item is a test-hygiene nit; withholding a stamp because a required check is still pending and I was asked for a review rather than an approval.

— Rames Jusso

miguel-heygen (changes requested):
**Strengths:** `TimelineFxPopover.tsx:33-46` now derives the cap and anchor direction from the same geometry, and `TimelineFxPopover.tsx:113-128` correctly makes the preset list the only flex scroller while keeping the footer fixed.

**Blocker — the minimum defeats the viewport cap:** `packages/studio/src/components/editor/TimelineFxPopover.tsx:43` uses `Math.max(MIN_POPOVER_HEIGHT, available)`. When the chosen gap is smaller than 160px, the dialog therefore stays 160px tall and still leaves the viewport. The new test at `TimelineFxPopover.test.tsx:134-140` actually pins that failure: with `innerHeight=200` and anchor `100..120`, the code opens upward with `bottom=104px` and `maxHeight=160px`, putting the dialog top at `-64px`. The footer/list are again unreachable at the exact short-window boundary this restore claims to fix. The usable minimum cannot override physical space; cap to the actual chosen gap (or reposition the whole dialog within the viewport) and assert both edges remain in-bounds.

**Verdict:** REQUEST CHANGES
**Reasoning:** The flex/scroller structure is right, but the geometry regression test currently preserves an off-screen popover instead of preventing it.

— Magi

jrusso1020 (commented):
Additive to @miguel-heygen's blocker. I had walked that same clamp case and judged it not worth raising — that was a miscalibration on my part, and the finding is real. Two corrections to how it is described, because they change which fix is the right one.

**The failure mode is narrower than "the footer/list are again unreachable".** Working the case through: `bottom: 104px` puts the box's bottom edge at `y = 96`, and a 160px box therefore spans `y = -64 … 96`. The footer is the last flex child with `shrink-0` (`TimelineFxPopover.tsx:131`), so it sits at roughly `y = 57 … 88` — **on screen and clickable**. The scroller above it is visible from `y = 0 … 57` and still scrolls, so every preset remains reachable, just through a ~57px window. What is actually wrong is that the top ~64px of the dialog is off-viewport and the usable list window collapses to about a third of the intended cap. Real, and worth fixing; not the total loss of access the popover had before the cap.

**The trigger is browser zoom, not a 200px window.** The clamp only overflows when the chosen gap is under ~172px (`available = gap - VIEWPORT_MARGIN - 4 < 160`), and reaching that on both sides needs a viewport under roughly 360px tall. No one drags a window that short — but 400% zoom on a 1080p display puts `innerHeight` near 270 CSS px, and Chrome goes to 500%. So this is a high-zoom accessibility path, which is a better argument for fixing it than the synthetic `innerHeight = 200` in the test.

**That points at the second of the two fixes suggested, not the first.** "Cap to the actual chosen gap" reintroduces exactly what `MIN_POPOVER_HEIGHT` exists to prevent — at a 20px gap you get a 20px dialog, which is the vanishing popover in a new costume, and the constant's comment at `:20` is making that trade deliberately. Repositioning the dialog to stay in-bounds keeps both properties: honour the 160px floor, then clamp the resulting box into the viewport (the way `left` is already clamped at `:24-27`) so a floor that exceeds the gap slides the box down instead of off the top. The assertion to add is the one named in the blocker — both edges in-bounds — and it holds under either strategy, so it is worth writing first.

**Verdict:** COMMENT
**Reasoning:** Agreeing with an existing blocker I under-weighted, with a corrected failure mode and a trigger that is reachable via zoom; my only substantive addition is that the shrink-to-gap fix would undo the minimum on purpose while the reposition fix keeps both invariants.

— Rames Jusso

vanceingalls:
Blocker fixed at `589e3f08`.

@miguel-heygen was right that `Math.max(MIN_POPOVER_HEIGHT, available)` defeated the cap, and @jrusso1020 was right about which of the two suggested fixes to take: shrinking to the gap would undo the floor on purpose (a 20px gap → a 20px popover), so the floor is honoured and the resulting box is clamped into the viewport the way `left` already is.

Two parts:

- **Cap the floor by the window**, `innerHeight - 2 * VIEWPORT_MARGIN`. The 160px minimum is a floor against a tight *gap*, not against a tight *window*; below 176px of viewport physical space has to win.
- **Inset the offset** to `innerHeight - height - VIEWPORT_MARGIN`, from either side, so a floor larger than the gap slides the box back in rather than off the top.

The case in the blocker (`innerHeight = 200`, anchor `100..120`) now lands `bottom: 32px` / `maxHeight: 160px` — the box at y = 8..40, one margin on each side. The two ordinary cases are byte-identical (34/726 down, 72/688 up).

On the corrected failure mode: agreed the top ~64px was off-viewport with the footer still clickable through a ~57px window, rather than a total loss of access — and agreed the trigger worth naming is high browser zoom, not a hand-dragged 200px window. The commit message says it that way.

Nit also done: the three geometry tests now pin `window.innerHeight` through one shared `withViewportHeight` helper instead of inheriting happy-dom's 768, so every expected number is derivable from the test.

Two tests added, both failing on the previous arithmetic and passing now:

```
× keeps both edges in the viewport when the minimum exceeds the gap  → expected '104px' to be '32px'
× shrinks below the minimum only when the whole window is shorter    → expected '160px' to be '104px'
```

12 tests in the file pass; the full `packages/studio` suite is 4378 passing on the stack tip.

Not fixed, and still true: `clampedStyle` reads `window.innerHeight` at render, so the cap goes stale across a resize with the popover open, exactly as the position already did. Pre-existing and left alone.

miguel-heygen (approved):
**Strengths:** `TimelineFxPopover.tsx:39-66` now preserves the 160px usable floor when only the chosen gap is tight, caps that floor to physical viewport height, and clamps the resulting top/bottom inset so both edges stay inside the window. The ordinary upward/downward geometry remains unchanged.

`TimelineFxPopover.test.tsx:154-179` directly pins the high-zoom failure and the physically-short-window fallback; the ambient viewport values are now explicit. I also ran the exact restacked tip locally: all 12 popover tests pass, with 29/29 across the affected popover/context-menu/caret surface. All required main-targeting checks at `589e3f08`, including Test and both Windows jobs, are green.

**Nit:** the tight-case comment says the box is at `y=8..40`; with `bottom=32` and height 160 in a 200px viewport, its edges are `y=8..168`. The assertions calculate the right geometry, so this is comment-only.

**Verdict:** APPROVE
**Reasoning:** The prior minimum-vs-viewport blocker is closed without sacrificing the usable floor, and both exact-head tests and full required CI are green.

— Magi

jrusso1020 (commented):
Re-review at `589e3f0825`. **@miguel-heygen's blocker is FIXED**, and fixed in the shape that keeps both invariants rather than trading one for the other.

`clampedStyle` now does two separate things, which is what the case needed:

- **Height** is capped by the viewport as well as the gap — `min(max(MIN_POPOVER_HEIGHT, available), innerHeight - VIEWPORT_MARGIN*2)` (`:41-45`). The floor still wins against a tight *gap*, and now yields to a tight *window*, which is the right precedence: you cannot honour 160px inside 120px of viewport.
- **Position** is inset instead of shrunk (`:50-55`). `inset()` clamps the chosen edge into `[VIEWPORT_MARGIN, innerHeight - height - VIEWPORT_MARGIN]`, and because that upper bound is expressed in the same coordinate as the edge being set, it is correct for `top` and `bottom` alike — the constraint is symmetric, so one helper genuinely covers both branches.

I re-derived every assertion rather than trusting the expected values, and all five are right:

| case | derivation | asserted |
|---|---|---|
| down, 768 | below 738 → avail 726, cap 752 → h 726; top `min(34, 768-726-8=34)` | `top 34` / `726` ✓ |
| up, 768 | above 700 → avail 688 → h 688; bottom `min(72, 72)` | `bottom 72` / `688` ✓ |
| floor, 200 | avail 88 → h `max(160,88)`=160 ≤ 184 | `160` ✓ |
| **new** — floor > gap, 200 | h 160; bottom `min(104, 200-160-8=32)` → 32 ⇒ box at y 8…168 | `bottom 32`, both margins ≥ 8 ✓ |
| **new** — window < floor, 120 | h `min(160, 104)`=104; bottom `min(64, 120-104-8=8)` → 8 | `104` / `bottom 8` ✓ |

The first two land exactly on the `inset` boundary (34 == 34, 72 == 72), so they now double as edge pins for free.

`withViewportHeight` is the right call too — the earlier assertions silently inherited happy-dom's 768 default, and the two new cases only mean anything against a stated viewport.

**nit — one wrong number in a comment.** `TimelineFxPopover.test.tsx`, the new floor-exceeds-gap case: "bottom:32 + maxHeight:160 puts the box at y = 8..40". With `bottom: 32` in a 200px viewport the bottom edge is at y=168 and the top edge at y=8, so it is **y = 8…168**, not 8…40. The assertions below it are correct and the intent ("a margin on each side") holds; only the interval is wrong. Worth fixing because this is a regression test whose comment is the next reader's model of what it pins.

**Verdict:** COMMENT
**Reasoning:** The blocker I under-weighted last round is properly fixed, the floor/viewport precedence is right, and all five expected values check out by derivation; one incorrect comment interval is the only thing left.

— Rames Jusso

PR #3416, opened 2026-08-22
fix(skills): make the carve CLI work against the published core, and honour its own group invariant

vanceingalls:
Found by using the shipped audio-groups/carve feature end to end on a real project (a HyperFrames video built against published `hyperframes@0.8.8`) rather than from inside this repo. Independent of #3413/#3414/#3415 — different files, no shared surface, so it is based on `main` rather than stacked.

## 1. The CLI could not load core at all — for everyone except us

`loadCore` resolved `./audio-carve` and `./audio-fx` with `require.resolve`. Reproduced both ways:

```
# inside this monorepo
require.resolve('@hyperframes/core/audio-carve') -> OK

# against the published package, in a real project
require.resolve('@hyperframes/core/audio-carve') -> ERR_PACKAGE_PATH_NOT_EXPORTED
import('@hyperframes/core/audio-carve')          -> OK, 12 exports
```

The workspace manifest declares a `node` condition, which `require.resolve` matches. The **published** manifest does not:

| | `./audio-carve` |
| --- | --- |
| `packages/core/package.json` `exports` | `{bun, node: ./dist/audioCarve.js, import, types}` |
| `packages/core/package.json` `publishConfig.exports` | `{import: ./dist/audioCarve.js, types}` |

So the script worked in development and failed for the audience it ships to. Its error text made this worse by naming a remedy that cannot work:

> `cannot resolve @hyperframes/core … install or update it: npm i -D @hyperframes/core (the audio-carve export needs a version that ships the carve analysis)`

No install adds a condition the manifest never declares. This cost a full delivery cycle downstream — the carve was recorded as "blocked by a tool defect" and the piece shipped with coarse volume ducking instead.

**Fix:** keep the project anchor (this script lives wherever the skill was installed, so a bare `import(spec)` would resolve against the *skill* directory and find nothing), and fall back to the manifest's declared `import` target when no require-resolvable condition exists. Error message now describes what actually happened.

## 2. It violated the invariant its own SKILL.md sets

`skills/hyperframes-audio/SKILL.md` is unambiguous:

> **A carve against more than one clip id is wrong. Group the clips and carve against the group.** This is an invariant, not a tip.

The script wrote `sources: voices.map((v) => v.id)` unconditionally. So every run against grouped voices emitted output that trips this repo's own `audio_carve_ungrouped_sources` lint rule, and — the reason the invariant exists — a voice added to the group later plays outside the carve's awareness and the bed silently fails to duck under it.

Now: when every detected voice shares one group, record the group. Mixed groups, partial grouping, or no grouping keep the clip ids, so the lint rule still fires on exactly the case it is meant to catch rather than being papered over.

Verified end to end against a real 26-element composition on published core:

- grouped voices → `data-fx-carve.sources = ["voiceover"]`, and `hyperframes check` passes clean
- same composition with the group attributes stripped → `["vo1","vo2","vo3","vo4","vo5"]`

## Notes

`main()` moves behind the standard entry guard so the pure helper can be imported by a test. `node carve.mjs …` is unaffected — re-verified against a real composition after the change.

**Not addressed here, deliberately:** the root cause of (1) is that `publishConfig.exports` omits `node`/`require` for *every* subpath of `@hyperframes/core` (generated by `scripts/package-subpaths.mjs` — `publishedExport()` emits only `{import, types}`, and because the descriptor is read back from the published map the omission is self-perpetuating). That breaks any require-resolving consumer of any published subpath, which is the same class as the known `@hyperframes/sdk` ESM-only issue. It is a packaging-policy change with repo-wide blast radius and belongs in its own PR with a maintainer decision — happy to open it.

## Checks

`node --test` carve suite 6/6 · full `test:skills` **512 passed / 0 failed** · `oxlint` 0/0 · `oxfmt --check` clean · `lint:skills` clean · `check:skill-mirror` OK · `check:tracked-artifacts` OK · `fallow audit --base origin/main` no issues · `skills-manifest.json` regenerated for the changed skill.

### DIFF
--- a/skills-manifest.json
+++ b/skills-manifest.json
@@ -26,8 +26,8 @@
       "files": 121
     },
     "hyperframes-audio": {
-      "hash": "819ab4e70d0f1cc9",
-      "files": 6
+      "hash": "5be0b86606a24f4a",
+      "files": 7
     },
     "hyperframes-cli": {
       "hash": "3fa884269c43d7df",
--- a/skills-manifest.json
+++ b/skills-manifest.json
@@ -26,7 +26,7 @@
       "files": 121
     },
     "hyperframes-audio": {
-      "hash": "a5c5b62882ddad63",
+      "hash": "5564aeda7725f455",
       "files": 7
     },
     "hyperframes-cli": {
--- a/skills-manifest.json
+++ b/skills-manifest.json
@@ -26,7 +26,7 @@
       "files": 121
     },
     "hyperframes-audio": {
-      "hash": "5be0b86606a24f4a",
+      "hash": "0b5e0ac8d731e1d2",
       "files": 7
     },
     "hyperframes-cli": {
--- a/skills-manifest.json
+++ b/skills-manifest.json
@@ -26,7 +26,7 @@
       "files": 121
     },
     "hyperframes-audio": {
-      "hash": "0b5e0ac8d731e1d2",
+      "hash": "a5c5b62882ddad63",
       "files": 7
     },
     "hyperframes-cli": {
--- a/skills/hyperframes-audio/SKILL.md
+++ b/skills/hyperframes-audio/SKILL.md
@@ -261,15 +261,33 @@ A `sources` list naming two or more plain clip ids instead of a group is caught
 by the `audio_carve_ungrouped_sources` lint rule — it still works, but it is the
 version that silently rots when a clip is added.
 
-**The bed must not be a member of the group it carves against.** A group id in
-`sources` resolves to every current member on every analysis, so a bed inside
-that group is handed itself as a voice and carved against its own content —
-which is the "never carve a track against itself" rule arriving one re-analysis
-later, after a first pass that looked correct. Give the bed its own group
-(`music`) and the narration its own (`voiceover`). `carve.mjs` refuses to write
-the group form when it detects this and records clip ids instead, so the lint
-rule points at the arrangement rather than the CLI quietly producing a
-self-carve.
+**Keep the carve group a voice group: no bed, no SFX, no music.** A group id in
+`sources` resolves to every _current_ member on _every_ analysis, so the group
+you name is the group you get later — not the tracks that were measured when it
+was written. Two ways that bites:
+
+- **The bed in its own source group.** It is handed to itself as a voice and
+  carved against its own content — the "never carve a track against itself" rule
+  arriving one re-analysis later.
+- **An SFX or music clip in the voice group.** It enters the sidechain on the
+  next analysis and the bed starts ducking under a whoosh, even though the run
+  that wrote the attribute never measured it.
+
+Both are invisible at the moment the carve is written: the analysis sums the
+voices it detected and never round-trips through group resolution, so the first
+pass is genuinely correct and only the next one is wrong. So give each role its
+own group — `music` for the bed, `voiceover` for the narration, `sfx` for the
+hits — and keep the group named in `sources` holding nothing but voices.
+
+`carve.mjs` refuses to write the group form when it sees either case, records
+clip ids, and says on stderr which member blocked it. The
+`audio_carve_ungrouped_sources` rule then points at the arrangement instead of
+the CLI quietly persisting a wider carve than it measured.
+
+A voice that this run left out is **not** one of these cases and does not block
+the group form: `carve.mjs` only analyses voices that overlap the bed, and
+picking up a clip that plays later without an edit to `sources` is the whole
+reason to name the group.
 
 **One knob.** `strength` is 0..1 and derives everything: how deep to cut, how
 many bands, how wide, how far to favour intelligibility over raw voice energy,
--- a/skills/hyperframes-audio/SKILL.md
+++ b/skills/hyperframes-audio/SKILL.md
@@ -261,6 +261,16 @@ A `sources` list naming two or more plain clip ids instead of a group is caught
 by the `audio_carve_ungrouped_sources` lint rule — it still works, but it is the
 version that silently rots when a clip is added.
 
+**The bed must not be a member of the group it carves against.** A group id in
+`sources` resolves to every current member on every analysis, so a bed inside
+that group is handed itself as a voice and carved against its own content —
+which is the "never carve a track against itself" rule arriving one re-analysis
+later, after a first pass that looked correct. Give the bed its own group
+(`music`) and the narration its own (`voiceover`). `carve.mjs` refuses to write
+the group form when it detects this and records clip ids instead, so the lint
+rule points at the arrangement rather than the CLI quietly producing a
+self-carve.
+
 **One knob.** `strength` is 0..1 and derives everything: how deep to cut, how
 many bands, how wide, how far to favour intelligibility over raw voice energy,
 how far the level may drop, how far under the voice to aim. Those six move
--- a/skills/hyperframes-audio/scripts/carve.mjs
+++ b/skills/hyperframes-audio/scripts/carve.mjs
@@ -26,7 +26,7 @@
 
 import { execFileSync } from "node:child_process";
 import { createRequire } from "node:module";
-import { readFileSync, writeFileSync } from "node:fs";
+import { readFileSync, realpathSync, writeFileSync } from "node:fs";
 import { dirname, resolve } from "node:path";
 import { pathToFileURL } from "node:url";
 
@@ -81,7 +81,7 @@ function fail(message, code = 1) {
  * the skill was installed — a sibling of the composition is what has the
  * dependency.
  */
-async function loadCore(fromDir) {
+export async function loadCore(fromDir) {
   const require = createRequire(pathToFileURL(resolve(fromDir, "package.json")));
 
   /*
@@ -142,7 +142,7 @@ async function loadCore(fromDir) {
 }
 
 /**
- * The `sources` a carve should record for these voices.
+ * The `sources` a carve should record for these voices, on this bed.
  *
  * SKILL.md states the invariant: "A carve against more than one clip id is
  * wrong. Group the clips and carve against the group." Naming the group lets
@@ -155,12 +155,38 @@ async function loadCore(fromDir) {
  * skill's invariant and tripped its own lint rule on every run. When every
  * voice shares one group, record the group. Mixed or ungrouped voices keep
  * their ids, and the lint rule then correctly tells the author to group them.
+ *
+ * The bed has to be part of the decision, because the group form resolves
+ * LATER and wider than it looks. If the bed is itself a member of the voices'
+ * group, `resolveCarveSourceIds` expands that id to every current member on the
+ * next analysis — including the bed — and the bed ends up carved against
+ * itself, which SKILL.md calls a bug rather than a mix choice. This run cannot
+ * see it: `main()` sums the voice list it detected and never round-trips
+ * through group resolution, so the first pass is correct and only the next
+ * re-analysis in Studio is wrong. So decline the group form there and fall back
+ * to clip ids, which is exactly the case `audio_carve_ungrouped_sources` exists
+ * to put in front of the author.
+ *
+ * Only an `<audio>` bed can trip it: group membership is audio-only
+ * (`audioGroupOf`), so `data-audio-group` on a `<video>` bed is ignored by core
+ * and expanding a group can never pull it in.
  */
-export function carveSources(voices) {
+export function carveSources(voices, bed) {
+  const group = sharedVoiceGroup(voices);
+  return group && !bedInVoiceGroup(voices, bed) ? [group] : voices.map((v) => v.id);
+}
+
+/** The one group every voice belongs to, or null if they do not share exactly one. */
+function sharedVoiceGroup(voices) {
   const groups = voices.map((v) => attrOf(v.tag, "data-audio-group"));
   const first = groups[0];
-  const allShareOneGroup = Boolean(first) && groups.every((g) => g === first);
-  return allShareOneGroup ? [first] : voices.map((v) => v.id);
+  return Boolean(first) && groups.every((g) => g === first) ? first : null;
+}
+
+/** The case above: naming this group would put the bed in its own source list. */
+export function bedInVoiceGroup(voices, bed) {
+  const group = sharedVoiceGroup(voices);
+  return Boolean(group) && bed?.kind === "audio" && attrOf(bed.tag, "data-audio-group") === group;
 }
 
 /** Mono float PCM for one media file, via ffmpeg. */
@@ -414,7 +440,16 @@ async function main() {
     : [];
   const lanes = [...carriedLanes, ...carvedLanes];
 
-  const settings = { enabled: true, sources: carveSources(voices), strength: args.strength };
+  const settings = { enabled: true, sources: carveSources(voices, bedEl), strength: args.strength };
+  // Say why the group form was declined, or the lint rule tells the author to
+  // group clips they have already grouped.
+  if (bedInVoiceGroup(voices, bedEl)) {
+    process.stderr.write(
+      `note   bed ${bedEl.id} is in group "${attrOf(bedTag, "data-audio-group")}" with the voices,\n` +
+        `       so sources are clip ids: naming that group would carve the bed\n` +
+        `       against itself on the next analysis. Move the bed to its own group.\n`,
+    );
+  }
   const written =
     ` data-fx-carve="${escapeAttr(JSON.stringify(settings))}"` +
     ` data-fx-chain="${escapeAttr(fxApi.serializeAudioFxChain(chain))}"` +
@@ -451,6 +486,19 @@ async function main() {
 
 // Only run as a CLI. Guarded so the pure helpers above can be unit-tested by
 // importing this module (`skills/**/*.test.mjs`, run by `bun run test:skills`).
-if (process.argv[1] && pathToFileURL(process.argv[1]).href === import.meta.url) {
+//
+// realpath both sides: on macOS /tmp → /private/tmp, and node resolves the main
+// module's symlinks in import.meta.url while argv[1] keeps the invoked spelling —
+// a raw compare silently skips main() when invoked through any symlinked path.
+function isMainModule(importMetaUrl) {
+  if (!process.argv[1]) return false;
+  try {
+    return pathToFileURL(realpathSync(process.argv[1])).href === importMetaUrl;
+  } catch {
+    return false;
+  }
+}
+
+if (isMainModule(import.meta.url)) {
   await main();
 }
--- a/skills/hyperframes-audio/scripts/carve.mjs
+++ b/skills/hyperframes-audio/scripts/carve.mjs
@@ -171,9 +171,9 @@ export async function loadCore(fromDir) {
  * (`audioGroupOf`), so `data-audio-group` on a `<video>` bed is ignored by core
  * and expanding a group can never pull it in.
  */
-export function carveSources(voices, bed) {
+export function carveSources(voices, bed, members = []) {
   const group = sharedVoiceGroup(voices);
-  return group && !bedInVoiceGroup(voices, bed) ? [group] : voices.map((v) => v.id);
+  return group && !groupSourceRefusal(voices, bed, members) ? [group] : voices.map((v) => v.id);
 }
 
 /** The one group every voice belongs to, or null if they do not share exactly one. */
@@ -183,10 +183,49 @@ function sharedVoiceGroup(voices) {
   return Boolean(first) && groups.every((g) => g === first) ? first : null;
 }
 
-/** The case above: naming this group would put the bed in its own source list. */
-export function bedInVoiceGroup(voices, bed) {
+/**
+ * Why naming the voices' shared group would persist something this run did not
+ * analyse — or null when the group is safe to name.
+ *
+ * `members` is every `<audio>` in the composition as `{id, group, nameKind}`,
+ * with `nameKind` from core's `classifyAudioName`, so this and Studio's picker
+ * classify the same way.
+ *
+ * Two refusals, and both exist because the group form resolves LATER and WIDER
+ * than the analysis: `resolveCarveSourceIds` expands a group id to every current
+ * member on every analysis, and `resolveCarveVoices` keeps any audio member with
+ * a src. `main()` meanwhile sums the voice list `detectTracks` returned, so the
+ * first pass looks correct however wrong the persisted attribute is.
+ *
+ *   `bed`   — the bed is a member, so it would be handed to itself as a voice
+ *             and carved against its own content.
+ *   `mixed` — a member classified music or sfx is not a voice this run measured,
+ *             so it would enter the sidechain on the next analysis and duck the
+ *             bed under a whoosh.
+ *
+ * Deliberately NOT a refusal: a member classified `voice` or `unknown` that this
+ * run left out. That is the group form working as designed — `detectTracks` only
+ * takes voices that overlap the bed, and picking up a clip that starts playing
+ * later without an edit to `sources` is the whole reason SKILL.md says to name
+ * the group. Refusing there would collapse the group form into clip ids for
+ * every ordinary narration sequence.
+ */
+export function groupSourceRefusal(voices, bed, members = []) {
   const group = sharedVoiceGroup(voices);
-  return Boolean(group) && bed?.kind === "audio" && attrOf(bed.tag, "data-audio-group") === group;
+  if (!group) return null;
+  if (bed?.kind === "audio" && attrOf(bed.tag, "data-audio-group") === group) {
+    return { group, reason: "bed", ids: [bed.id] };
+  }
+  const analysed = new Set(voices.map((v) => v.id));
+  const strays = members
+    .filter(
+      (m) =>
+        m.group === group &&
+        !analysed.has(m.id) &&
+        (m.nameKind === "music" || m.nameKind === "sfx"),
+    )
+    .map((m) => m.id);
+  return strays.length > 0 ? { group, reason: "mixed", ids: strays } : null;
 }
 
 /** Mono float PCM for one media file, via ffmpeg. */
@@ -320,7 +359,7 @@ function detectTracks(html, given, classify, overlaps) {
         `  name one with --voice`,
     );
   }
-  return { bed, voices: usable };
+  return { bed, voices: usable, all };
 }
 
 const startOf = (tag) => {
@@ -335,12 +374,21 @@ async function main() {
   const { carve: carveApi, fx: fxApi } = await loadCore(args.core ? resolve(args.core) : compDir);
 
   const html = readFileSync(compPath, "utf-8");
-  const { bed: bedEl, voices } = detectTracks(
-    html,
-    args,
-    carveApi.classifyAudioName,
-    carveApi.clipsOverlap,
-  );
+  const {
+    bed: bedEl,
+    voices,
+    all: media,
+  } = detectTracks(html, args, carveApi.classifyAudioName, carveApi.clipsOverlap);
+  // Group membership + name classification for every audio track, so the source
+  // decision can see what the group will resolve to later and not just what this
+  // run analysed.
+  const members = media
+    .filter((el) => el.kind === "audio")
+    .map((el) => ({
+      id: el.id,
+      group: attrOf(el.tag, "data-audio-group"),
+      nameKind: carveApi.classifyAudioName(el.id, unescapeAttr(attrOf(el.tag, "src") ?? "")),
+    }));
   const bedTag = bedEl.tag;
   const bedSrc = attrOf(bedTag, "src");
   process.stdout.write(
@@ -440,14 +488,24 @@ async function main() {
     : [];
   const lanes = [...carriedLanes, ...carvedLanes];
 
-  const settings = { enabled: true, sources: carveSources(voices, bedEl), strength: args.strength };
+  const settings = {
+    enabled: true,
+    sources: carveSources(voices, bedEl, members),
+    strength: args.strength,
+  };
   // Say why the group form was declined, or the lint rule tells the author to
   // group clips they have already grouped.
-  if (bedInVoiceGroup(voices, bedEl)) {
+  const refusal = groupSourceRefusal(voices, bedEl, members);
+  if (refusal) {
     process.stderr.write(
-      `note   bed ${bedEl.id} is in group "${attrOf(bedTag, "data-audio-group")}" with the voices,\n` +
-        `       so sources are clip ids: naming that group would carve the bed\n` +
-        `       against itself on the next analysis. Move the bed to its own group.\n`,
+      refusal.reason === "bed"
+        ? `note   bed ${bedEl.id} is in group "${refusal.group}" with the voices, so\n` +
+            `       sources are clip ids: naming that group would carve the bed\n` +
+            `       against itself on the next analysis. Move the bed to its own group.\n`
+        : `note   group "${refusal.group}" also holds ${refusal.ids.join(", ")}, which this run\n` +
+            `       did not analyse (music/sfx by name), so sources are clip ids: naming\n` +
+            `       the group would pull them into the sidechain on the next analysis.\n` +
+            `       Move them out of the voice group.\n`,
     );
   }
   const written =
--- a/skills/hyperframes-audio/scripts/carve.mjs
+++ b/skills/hyperframes-audio/scripts/carve.mjs
@@ -171,7 +171,7 @@ export async function loadCore(fromDir) {
  * (`audioGroupOf`), so `data-audio-group` on a `<video>` bed is ignored by core
  * and expanding a group can never pull it in.
  */
-export function carveSources(voices, bed, members = []) {
+export function carveSources(voices, bed, members) {
   const group = sharedVoiceGroup(voices);
   return group && !groupSourceRefusal(voices, bed, members) ? [group] : voices.map((v) => v.id);
 }
@@ -190,6 +190,13 @@ function sharedVoiceGroup(voices) {
  * `members` is every `<audio>` in the composition as `{id, group, nameKind}`,
  * with `nameKind` from core's `classifyAudioName`, so this and Studio's picker
  * classify the same way.
+ *
+ * Required, deliberately not defaulting to `[]`. With an empty list the `mixed`
+ * refusal below cannot fire, so a call that forgot the argument would return the
+ * group form and restore the exact behaviour this function exists to prevent —
+ * silently, because the first CLI pass is correct either way and only a later
+ * Studio re-analysis is wrong. A missing argument throws on `members.filter`
+ * instead.
  *
  * Two refusals, and both exist because the group form resolves LATER and WIDER
  * than the analysis: `resolveCarveSourceIds` expands a group id to every current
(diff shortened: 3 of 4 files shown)


jrusso1020 (changes requested):
First review, no prior reviews or comments. Read `carve.mjs` end to end at `a7168775b7`, plus `packages/core/package.json`'s two export maps, `SKILL.md`, the new test, and the three sibling skill scripts that solve the same entry-guard problem.

Confirming the framing first: this is based on `main` and shares no files with #3413/#3414/#3415, so it is genuinely independent of those two open blockers rather than a fourth rung.

## Strengths

**The bug-1 diagnosis is exactly right, and the export tables in the body match the source.** `./audio-carve` is `{bun, node: ./dist/audioCarve.js, import: ./src/audioCarve.ts, types}` in the workspace manifest and `{import: ./dist/audioCarve.js, types: ./dist/audioCarve.d.ts}` under `publishConfig`. So `require.resolve` matches on `node` in the monorepo and has no resolvable condition at all once published — which is precisely the "worked in development, failed for the audience it ships to" split you describe. Naming a remedy that cannot work in the old error text is a fair thing to call out; no install adds a condition the manifest never declares.

**The fallback's anchor assumption needed checking and it holds.** Node does *not* auto-export `./package.json` once a package declares `exports`, so `require.resolve("@hyperframes/core/package.json")` could easily have been a second `ERR_PACKAGE_PATH_NOT_EXPORTED` inside the handler for the first one. It is explicitly present in both the workspace map and `publishConfig.exports`, and the published `./audio-carve` entry is a plain string under `import`, so the `typeof entry === "string" ? … : entry?.import ?? entry?.default` extraction lands on a real path.

**Rethrowing anything that is not `ERR_PACKAGE_PATH_NOT_EXPORTED` is the right narrow catch.** A missing install still reports as a missing install, and inside the monorepo an unbuilt `dist/` does not quietly fall through to `./src/audioCarve.ts`, which plain node could not have imported anyway.

**`carveSources` (`:157-163`) is correct on every branch I could construct, and the six tests cover exactly those branches.** `attrOf` ends in `?? null` (`:195`), so a missing attribute and `data-audio-group=""` both land on clip ids — which is what keeps `audio_carve_ungrouped_sources` firing on the case it exists for instead of being papered over. The two branches most likely to have been fudged, different groups and partial grouping, are both pinned.

**Deferring the `publishConfig` root cause is the right call and it is deferred honestly** — naming `publishedExport()` and the read-back that makes the omission self-perpetuating is a diagnosis, not a shrug. Worth opening.

**And the tests actually ran here:** `Test: skills`, `Skills: manifest in sync` and `Skills: project-native lint + mirror` all pass, with the heavier lanes correctly path-filtered out. `files: 6 → 7` matches the one added file.

## Blocker — the new entry guard silently skips `main()` through any symlinked path, and the repo already documents this exact trap

`carve.mjs:454`:

```js
if (process.argv[1] && pathToFileURL(process.argv[1]).href === import.meta.url) {
```

`process.argv[1]` keeps the spelling the caller typed; `import.meta.url` is derived from the **realpath**, because Node resolves the main module's symlinks by default (that is what `--preserve-symlinks-main` turns off). So whenever the invoked path differs from its realpath, this compares unequal, `main()` never runs, and the CLI **writes nothing and exits 0** — a silent success that produces no carve.

Three places in this repo already solve this, with the same comment on each:

- `skills/hyperframes-core/scripts/lib/frame-packets-core.mjs:183` — exports `isMainModule(importMetaUrl)`
- `skills/pr-to-video/scripts/preflight.mjs:42`
- `skills/pr-to-video/scripts/project-dir.mjs:82`

> realpath both sides: on macOS /tmp → /private/tmp, and node resolves the main module's symlinks in import.meta.url while argv[1] keeps the invoked spelling — **a raw compare silently skips main() when invoked through any symlinked path.**

Two reachable triggers, both ordinary. On macOS `/tmp` is a symlink to `/private/tmp`, so any invocation with a scratch project under `/tmp` breaks. And `SKILL.md:304` documents the entry point as `node <SKILL_DIR>/scripts/carve.mjs`, so any install that places or links the skill through a symlinked path breaks too.

The uncomfortable part is that this is the same failure class as bug 1 in this very PR — correct where it was developed, broken for the audience it ships to — and it is a regression, since `await main()` previously ran unconditionally.

The new test cannot catch it: importing the module *should* leave `main()` unrun, so a correct guard and this one are indistinguishable from the test's side. The importable-helper goal is verified; the CLI path is not.

Fix is one line — realpath the left side, or import the existing helper rather than adding a fourth copy:

```js
return pathToFileURL(realpathSync(process.argv[1])).href === import.meta.url;
```

`realpathSync` needs adding to the `node:fs` import at `:29` (currently `readFileSync, writeFileSync`), and the `try/catch { return false }` the siblings wrap it in is worth keeping for a deleted or unreadable `argv[1]`.

**Verdict:** REQUEST CHANGES
**Reasoning:** Both fixes are correct and well-evidenced and the carve-sources logic is right on every branch, but the entry guard added to make the helper testable turns the CLI into a silent no-op on macOS `/tmp` and on symlinked skill installs — a regression against a trap this repo already documents and solves in three other scripts.

— Rames Jusso

miguel-heygen (changes requested):
Additive to Rames Jusso at the same exact head.

**Confirmed blocker:** I installed the real published `@hyperframes/core@0.8.8` in a clean project. Direct invocation reaches both import-only subpaths and proceeds to analysis; invoking the same `carve.mjs` through a symlink prints nothing and exits 0. So the `process.argv[1]` / `import.meta.url` guard finding at `skills/hyperframes-audio/scripts/carve.mjs:454` is reproduced, not inferred.

**Separate blocker — group collapse can persist the bed as its own carve source:** `carve.mjs:159-163` decides from the detected voices alone. If the music bed and voice are both members of group `mix`, the helper returns `["mix"]`, and line 417 writes that source onto the bed. The initial DSP pass still analyzes only `voices`, so it appears correct. But the next Studio re-analysis calls `resolveCarveSourceIds`; `packages/core/src/audioGroups.ts:121-124` expands `mix` to every member, including the bed itself. The bed then participates in its own sidechain, violating `SKILL.md:229-230` (never carve a track against itself). A group containing any other non-voice audio has the same delayed widening. Pass the bed/full membership context into the decision and collapse only when the group is a safe voice source; add a bed-in-same-group regression.

**Important test gap:** the six new tests cover only `carveSources`. The published-manifest loader and the CLI entry path are both central fixes and neither is exercised. A fixture package with import-only exports plus a symlinked CLI invocation would pin both regressions without depending on npm.

**Verdict:** REQUEST CHANGES
**Reasoning:** The export fallback is correct, but the new entry guard can silently no-op and the group-source rewrite can make the bed its own carve source on the next analysis.

— Magi

jrusso1020 (commented):
Additive follow-up, and a **retraction of my own praise**.

My first review said, in bold: "`carveSources` (`:157-163`) is correct on every branch I could construct." @miguel-heygen constructed the branch I didn't — the **bed** being a member of the same group as the voices — and it's a real defect. Withdrawing that line, because it reads as clearing the helper this PR's second fix is built on, and I don't want it weighed against a blocker that holds.

I verified the mechanism end to end rather than taking it on description, and every link confirms it:

1. `carveSources(voices)` (`carve.mjs:157-163`) reads only `voices`. It never looks at `bedTag`, so the bed's own `data-audio-group` is invisible to it.
2. The result is written onto the **bed** — `data-fx-carve` on `bedTag` (`carve.mjs:417`, written at `:447`). The carve target and the group member are the same element.
3. `resolveCarveSourceIds(doc, ids)` (`packages/core/src/audioGroups.ts:112`) expands a group id to **all current members**. It cannot exclude the host: it receives `(doc, ids)` and no host element.
4. `resolveCarveVoices(doc, sources)` (`useFxCarve.ts:121`) filters only on *exists* / `tagName === "AUDIO"` / *has a `src`*. Also no host exclusion, also no host parameter.
5. So on the next Studio analysis (`useFxCarve.ts:454`) the bed lands in its own voice list, and `measureCarve` measures the bed against itself — at level-matching strengths the duck envelope is then fighting the bed's own content, not speech.

Two things to add to the report:

**Scope.** This needs an `<audio>` bed, which is the ordinary case. A `<video>` bed is immune twice over — group membership is audio-only (`MEMBER_SELECTOR` + `audioGroupOf`, `audioGroups.ts:147`), and step 4 would skip it anyway. Worth stating so the fix isn't scoped wider than the bug.

**Why the PR's own verification can't catch it — this is the part I'd flag hardest.** The commit says `node carve.mjs` was "verified against a real composition," and that verification is sound and still passes. `main()` sums the voice list from `detectTracks` *directly* (`mixCarveSources`, `:~395`) and never round-trips through group resolution, so the CLI's first pass is genuinely correct and the artifact an author inspects right after running it looks right. The corruption appears only on the next re-analysis in Studio, against a file the CLI already declared good. A fixture test over the CLI alone would stay green.

On fix shape: the resolver is the wrong place for it. Neither `resolveCarveSourceIds` nor `resolveCarveVoices` takes the host element, so "make the resolver skip the target" is a signature change on shared core. `carveSources` is where the bed is already in scope at the call site — passing it the bed and declining the group form when the bed is a member of it (falling back to clip ids, which is exactly the case `audio_carve_ungrouped_sources` exists to flag) keeps the fix local and keeps the lint rule meaningful.

My `CHANGES_REQUESTED` from the entry-guard blocker stands unchanged; this is a second, independent blocker on the same PR.

**Verdict:** COMMENT (standing CHANGES_REQUESTED unchanged)
**Reasoning:** Retracting a positive claim of mine that a co-reviewer falsified, with the mechanism verified at each link, plus the reason this one survives the PR's stated end-to-end check.

— Rames Jusso

vanceingalls:
Both blockers fixed at `9a4ee305`.

**Entry guard** (`carve.mjs`). Realpath the left side. This repo already documents and solves the same trap in three scripts, so the canonical comment is carried over verbatim:

```js
// realpath both sides: on macOS /tmp → /private/tmp, and node resolves the main
// module's symlinks in import.meta.url while argv[1] keeps the invoked spelling —
// a raw compare silently skips main() when invoked through any symlinked path.
```

A local copy rather than importing `isMainModule` from `hyperframes-core/scripts/lib/frame-packets-core.mjs`: skills install independently (`npx skills add --skill hyperframes-audio`), and `hyperframes-audio` has no other cross-skill import, so reaching into a sibling skill would make the CLI depend on an install that may not be there. A fourth copy is the lesser evil; the shared comment is the marker.

Thanks for reproducing it against the real published `@hyperframes/core@0.8.8` rather than leaving it inferred.

**Bed as its own carve source.** Real, and the retraction is appreciated but the finding is @miguel-heygen's. Fixed at the call site, not the resolver — as @jrusso1020 says, neither `resolveCarveSourceIds` nor `resolveCarveVoices` receives the host, so "make the resolver skip the target" is a signature change on shared core.

`carveSources(voices, bed)` now declines the group form when the bed is a member and records clip ids, which is exactly what `audio_carve_ungrouped_sources` exists to raise. Two additions beyond the minimum:

- **A stderr note saying why.** Without it the author sees the lint rule telling them to group clips they *have* grouped. The note names the group and says to move the bed out of it.
- **Scoped to `<audio>` beds**, per the scope point: group membership is audio-only, so a `<video>` bed cannot be pulled in by an expansion and declining there would be a false positive. There is a test for that case.

Deliberately **not** the full-membership scan the blocker also offered. A non-voice non-bed member sitting in the voice group is an authoring error the skill already forbids, and full membership would decline the legitimate case of a grouped VO clip that does not overlap the bed. The bed is the element being written to, so it is the one that has to be excluded.

`SKILL.md` now states the constraint next to the group invariant it belongs to — the bed must not be a member of the group it carves against, and why the first pass looks correct anyway.

**Test gap closed** — six added, `carveSources` alone is no longer the only thing covered:

```
✓ a bed inside the voices' group keeps clip ids, so it is never its own source
✓ a bed in a DIFFERENT group leaves the group form alone
✓ an ungrouped bed leaves the group form alone
✓ a VIDEO bed is immune — group membership is audio-only
✓ the CLI still runs when invoked through a symlinked path
✓ loadCore honours an import-only export map, as the published core has
```

The last two are the fixture pair asked for. The symlink test spawns `node <symlinked-dir>/carve.mjs` with no args and asserts exit 1 with `--comp is required` on stderr — the broken guard's symptom is a silent exit 0, so no-args is the discriminator (`--help` exits 0 either way). The loader test builds a tmp project with `node_modules/@hyperframes/core` declaring `import`-only exports plus dist stubs, so it exercises the `ERR_PACKAGE_PATH_NOT_EXPORTED` fallback without npm; `loadCore` is exported for it.

Verified both new blocker tests fail on the previous head and pass now:

```
not ok 7  - a bed inside the voices' group keeps clip ids, so it is never its own source
not ok 11 - the CLI still runs when invoked through a symlinked path
# tests 12  # pass 10  # fail 2      <- at a7168775
# tests 12  # pass 12  # fail 0      <- at 9a4ee305
```

Whole skills suite: 518 passing.

On the deferred `publishConfig` root cause — agreed it is worth opening, and it is still deferred; this PR keeps the consumer-side fallback.

miguel-heygen (changes requested):
**Closed from the previous round:** `carve.mjs:491-500` now realpaths the entry path and the symlink CLI regression pins it; `loadCore` has an import-only fixture matching the published manifest; and `carveSources(voices, bed)` correctly refuses the self-bed group case. Those three fixes are sound, and `Test: skills` / skill-manifest checks pass at this head.

**Remaining blocker — the other widening case from my original review is unchanged:** `skills/hyperframes-audio/scripts/carve.mjs:174-176` still knows only the detected voices and the bed. If `vo1` and `vo2` share group `voiceover` with a third audio member classified as SFX or music, detection correctly excludes that third member, but `carveSources` returns `["voiceover"]`. On the next Studio analysis, `resolveCarveSourceIds` expands the group to all current members and `resolveCarveVoices` filters only to audio-with-src, so the SFX/music member enters the sidechain. The first CLI pass again looks correct because it analyzes the detected voice list directly. My prior review explicitly called out both the bed and any non-voice member; this follow-up covers only the bed.

Pass the current full membership/classification into this decision and collapse only when the existing group is a safe voice source (or deliberately fall back to ids with the same diagnostic), then add a grouped-voices-plus-SFX regression. The self-bed, different-bed-group, ungrouped-bed, and video-bed tests at `carve.test.mjs:67-90` do not exercise this case.

**Verdict:** REQUEST CHANGES
**Reasoning:** The entry, loader, and self-bed blockers are closed, but a valid mixed-content audio group still widens the persisted carve beyond what the CLI analyzed on the next Studio pass.

— Magi

jrusso1020 (commented):
Re-review at `9a4ee3056b`. **Both blockers are FIXED.** One new finding at a sibling writer, below — it is not a blocker on this PR.

**1. The entry guard (mine) — FIXED, using the pattern that already existed.** `isMainModule(import.meta.url)` at `:490-500` realpaths `argv[1]`, returns `false` when `argv[1]` is absent, and swallows a throwing `realpathSync` into `false`; `realpathSync` is added to the `node:fs` import at `:29`. Same signature and same comment as `frame-packets-core.mjs:183`, so the four copies now agree.

**2. The self-carve (@miguel-heygen's) — FIXED, and correctly scoped.** `carveSources(voices, bed)` declines the group form via `bedInVoiceGroup`, which gates on `bed?.kind === "audio"` — matching core, where membership is audio-only, so a `<video>` bed is not falsely declined. Falling back to clip ids is the right landing spot because it re-arms `audio_carve_ungrouped_sources` on an arrangement that genuinely needs the author, and the new stderr note (`:445-451`) is what stops that lint message from reading as nonsense to someone who *has* grouped their clips.

**The symlink test is the right shape** — worth saying, because I claimed last round that a test could not catch this. It could; it just could not be an import test. Spawning `carve.mjs` through a symlinked dir and asserting `status === 1` + `/--comp is required/` works precisely because the broken guard's signature is *exit 0 with no output*, so the assertion fails on the bug and passes on the fix. It also reproduces on Linux, not just macOS, since the divergence is `argv[1]`-spelling vs realpath rather than anything `/tmp`-specific. And `parseArgs` failing first is what keeps the test hermetic — no core needed.

**The `loadCore` fixture genuinely exercises the fallback**, which is the easy thing to get wrong: with only `{import}` on the subpath, `require.resolve` cannot match under the require condition, so the `ERR_PACKAGE_PATH_NOT_EXPORTED` branch is the only way to the marker assertions.

---

**important (NOT blocking this PR) — the same self-carve is reachable from Studio, which this fix does not cover.** The invariant is now enforced at the CLI writer; there is a second writer with the identical failure mode:

- `collectCarveCandidates` (`useFxCarveGrouping.ts:90-131`) excludes the bed as an *individual* row (`others` filters `a.id !== element.id`, `useFxCarve.ts:343-345`) — but it offers a **group** row whenever any member is in `others` (`:112-129`). So if the bed and a voice share a group, that group is offered, and its `(N)` label counts the bed, since `:113-119` maps `memberIds` with no host exclusion.
- Selecting it writes `sources: [groupId]` unchanged — `withAutoGroupedSources` returns early when a source is already a group id (`:54`).
- Next analysis: `resolveCarveSourceIds` expands it, `resolveCarveVoices` keeps the bed, and the bed is carved against itself. Identical to what `carveSources` now refuses to write.

The auto-group path itself is safe (`mintGroupId` mints a fresh id and `assignGroup` applies it only to the selected clips, so the bed cannot be a member). It is specifically the *pre-existing group* case that is unguarded. `bedInVoiceGroup` is exported, so the same predicate can be reused at the picker — either filtering such a group out of the candidate list or refusing it on selection.

I am deliberately not blocking on it: this PR's scope is the CLI, it strictly improves the situation, and the new SKILL.md paragraph is careful to say "`carve.mjs` refuses" rather than claiming the tool prevents it everywhere. But it should not sit long, because the doc now states the rule normatively while one of the two writers still lets a user walk into it.

**Checked, not findings:**
- *A non-bed, non-voice member widening the group* (e.g. an SFX clip the CLI's classifier excludes from `voices`, or a member that does not overlap the bed). Still expands at analysis time, so the first pass and the re-analysis can differ — but that is the group form working as designed, and carving a bed against an author-declared group member is a mix choice, not a self-reference. Materially different from the bed case; not worth guarding.
- *`--preserve-symlinks-main`* inverts the guard (that flag keeps `import.meta.url` at the invoked spelling, so realpathing `argv[1]` mismatches). Shared with all three existing copies, opt-in, and consistency across the four is worth more than special-casing it here — noting it only so nobody "fixes" one copy out of step.
- *The fixture pins `./package.json` as exported*, so it does not cover a future core that stops exporting it. That path degrades into the outer `catch` and a clear `cannot load @hyperframes/core` failure rather than an unhandled throw, so it is fine.

**Verdict:** COMMENT — my `CHANGES_REQUESTED` (`4999953165`) is resolved in substance; I have no remaining objection to this code. I am not converting it to an approval because stamps on this repo are James's call, not mine, so please read the red badge as "satisfied, awaiting his stamp" rather than as an open objection.
**Reasoning:** Both blockers are properly fixed with tests that would actually catch the regressions, and the one new finding lives in a file this PR does not touch.

— Rames Jusso

jrusso1020 (commented):
Correcting my own previous review ([5000023036](https://github.com/heygen-com/hyperframes/pull/3416#pullrequestreview-5000023036)) rather than adding a new finding. That review said every blocker was fixed. **@miguel-heygen's remaining blocker is well-founded and my clearance of it was wrong.** I had considered that same branch and graded it non-blocking; one check on the base falsifies the reasoning I used.

**What I got wrong.** I read the group form's late resolution as pre-existing, intended behavior — the feature `resolveCarveSourceIds` documents ("adding a fourth voice to a group already named in a carve's `sources` picks it up on the next analysis"). It isn't pre-existing. On the base, `carve.mjs:357` was:

```js
const settings = { enabled: true, sources: voices.map((v) => v.id), strength: args.strength };
```

Clip ids, unconditionally, every time. So before this PR the resolved set and the analyzed set could never diverge — there was no group id for `resolveCarveSourceIds` to expand. **The group form, and therefore every widening it can cause, is introduced here.** That makes this a new divergence in this diff, not an existing one I was declining to relitigate.

**Why it's the same contract, not a second concern.** `carveSources` (`:174-176`) decides from the detected voices alone — `sharedVoiceGroup` maps over `voices` and never looks at the named group's full membership. The invariant the fix needs is *every current member of the named group is one of the analyzed voices*; the bed is one way to violate it, a member the analysis never selected is another. The doc comment at `:160-170` already states the general principle — "the group form resolves LATER and wider than it looks", `main()` "never round-trips through group resolution, so the first pass is correct and only the next re-analysis in Studio is wrong" — and then guards one instance of it. Identical silent signature: correct artifact on the run the author inspects, wrong one pass later.

**One refinement to the repro, which changes what the test has to look like.** Classification alone does not exclude an SFX from `voices`. `detectTracks` (`:308-315`) takes *plays over the bed* as sufficient — `why: kindOf(el) === "voice" ? "name looks like a voice" : "plays over the bed"` — so an SFX-named clip overlapping the bed is already in the analyzed set and there is no divergence. The reachable case is a group member `detectTracks` did **not** select, most plainly one that does not overlap the bed's span. A regression fixture that groups an SFX *over* the bed will pass while broken; it needs a grouped member outside the bed's window.

**Test gap, for the record.** The ten `carveSources` cases cover shared/single/ungrouped/different/partial/empty-attr, bed-in-group, bed-in-another-group, ungrouped bed and a `<video>` bed. None covers a non-bed member of the voices' group, which is why the surviving branch reads as covered.

Everything else in [5000023036](https://github.com/heygen-com/hyperframes/pull/3416#pullrequestreview-5000023036) stands: the `isMainModule` guard now matches `frame-packets-core.mjs:183`, the subprocess-through-symlink test is the right shape (and disproves my earlier claim that no test could catch that bug), the `bed?.kind === "audio"` gate correctly spares a `<video>` bed, and CI is green — 37 distinct checks at `9a4ee305`, zero non-green, so `BLOCKED` here is the reviewer gate alone.

My own `CHANGES_REQUESTED` ([4999953165](https://github.com/heygen-com/hyperframes/pull/3416#pullrequestreview-4999953165)) is pinned to the old head `a7168775` and is stale — read @miguel-heygen's review at `9a4ee305` as the live gate, not mine. Not restamping either way; approvals here are James's call.

**Verdict:** COMMENT
**Reasoning:** Withdrawing my clearance of the mixed-membership branch — the group form is new in this PR, so the divergence it introduces is in scope, and @miguel-heygen's blocker should be treated as live.

— Rames Jusso

vanceingalls:
Second branch of the blocker fixed at `6f0b634e`. Taking the first of the two options — membership + classification in the collapse decision — rather than deriving the first pass from the resolved group: analysing whatever the group happens to hold would make the CLI measure clips it classified as non-voice, which treats the arrangement problem as a licence to sidechain them.

`groupSourceRefusal(voices, bed, members)` replaces `bedInVoiceGroup` and returns `{group, reason, ids}` or `null`, so the decision and the stderr note come from one place. `members` is every `<audio>` in the composition as `{id, group, nameKind}`, with `nameKind` from core’s `classifyAudioName` — same classifier Studio’s picker uses, so the two cannot disagree. `detectTracks` now returns the media list it already built.

Two refusals:

- `bed` — the bed is a member of the group.
- `mixed` — a member classified `music` or `sfx` that this run did not analyse. Your case: `vo1`/`vo2` in `voiceover` with a `whoosh` alongside them → `sources: ["vo1","vo2"]`, plus a note naming `whoosh`.

**Classification, not membership, is what makes this safe, and that distinction is load-bearing.** A member classified `voice` or `unknown` does *not* refuse. `detectTracks` only analyses voices that overlap the bed, so an outro line starting after the bed ends is routinely a group member the run did not measure — and covering it on a later analysis without editing `sources` is the entire reason SKILL.md says to name the group. Refusing on "any member this run did not analyse" would collapse the group form into clip ids for every ordinary narration sequence. `unknown` follows detection's own loose-in-the-safe-direction rule, since detection already treats an unknown name as a possible voice.

Tests: six added, 18 in the file.

```
✓ an sfx member of the voices' group blocks the group form          <- your regression
✓ a music member of the voices' group blocks it too
✓ groupSourceRefusal names which member blocked the group, and why
✓ a voice member this run did NOT analyse keeps the group form      <- must not refuse
✓ an unclassifiable member keeps the group form
✓ an sfx member of a DIFFERENT group is irrelevant
```

Verified by ablation rather than by assertion — with the `mixed` branch stubbed to `return null` and everything else unchanged, exactly those first three fail (`# tests 18 # pass 15 # fail 3`) and the three must-not-refuse cases stay green. Whole skills suite 524 passing.

The stderr note now names the blocking member for either reason. SKILL.md states both refusals and the voice-member exemption next to the group invariant.

@jrusso1020 — this lands on the case you checked and dismissed, so worth flagging the disagreement explicitly rather than letting the commit imply consensus. Your read (an author-declared group member is a mix choice, not a self-reference) is a fair one and I did weigh it. What tipped it: the CLI persists an attribute whose meaning at write time and at re-analysis time differ, and the author has no way to see that from the output. Classification-gating keeps your objection's substance — the non-overlapping voice case, which is the one that would have hurt — while closing the case where the run demonstrably did not measure what it persisted. Also noting your separate `useFxCarveGrouping` finding is real and untouched here; it wants its own PR, and this PR's SKILL.md wording stays scoped to "`carve.mjs` refuses" for exactly that reason.

miguel-heygen (approved):
**Closed:** `carve.mjs:174-228` now makes the group decision from detected voices, the bed, and every current audio member classified through the same core classifier. It refuses the group only for the two actual widening hazards — the bed itself or an unanalyzed `music`/`sfx` member — while preserving the designed late-expansion behavior for non-overlapping `voice` and `unknown` members.

The new tests at `carve.test.mjs:98-169` pin both mixed-member refusals, the returned diagnostic shape, and all three exemptions. The symlink entry and import-only loader regressions from the earlier rounds remain covered. All exact-head checks at `6f0b634e`, including Test: skills, the full matrix, regression shards, and Windows, are green.

**Verdict:** APPROVE
**Reasoning:** The final mixed-membership branch now uses the complete current group context without collapsing the intended future-voice behavior, and every prior blocker has an executable regression.

— Magi

jrusso1020 (commented):
Re-review at `6f0b634e`. **The mixed-membership blocker is fixed**, and one thing I told you last round was wrong — that first, because it was about the test you were being asked to write.

**Correcting myself.** In [5000033888](https://github.com/heygen-com/hyperframes/pull/3416#pullrequestreview-5000033888) I said classification alone does not exclude an SFX from the analysed set, quoting the `why: kindOf(el) === "voice" ? … : "plays over the bed"` ternary, and concluded a grouped-SFX-over-the-bed fixture would pass while broken. That's wrong. `detectTracks` filters classification **before** overlap:

```js
const plausible = all
  .filter((el) => el.id !== bed.id && kindOf(el) !== "music" && kindOf(el) !== "sfx")  // :340
  .filter(overlapping);                                                                //  :341
```

An SFX never enters the pool regardless of whether it overlaps, so the `"plays over the bed"` label only ever describes an `unknown`-classified clip. @miguel-heygen's original framing was right and my refinement of it was not; the constraint I put on your fixture was unnecessary, and your own test comment at `carve.test.mjs:99-104` states the mechanism correctly. I read a line and inferred from its label instead of reading the filter above it.

**Why the fix is right, and not just sufficient.** `groupSourceRefusal` (`:213-229`) refuses a stray that is `!analysed.has(m.id)` **and** classified `music` or `sfx` — which makes the refusal set *exactly* the set `plausible` removes from the analysis by classification. Nothing wider, nothing narrower. That equivalence is what makes it a real invariant rather than a second patch: the group form is declined precisely when it would resolve to something classification had already ruled out as a voice, and the writer now sees what the resolver will see later.

The carve-out is the part I'd have got wrong. A `voice` or `unknown` stray stays permitted (`carve.test.mjs:118-142`), and `unknown` belongs on that side because `plausible` itself treats `unknown` as a voice candidate — so the refusal is loose in the same direction as detection, not loose in a new one. Refusing there would have collapsed the group form into clip ids for any narration sequence with a later clip, which is the feature SKILL.md asks for. The comment at `:203-210` says so explicitly, so the next reader can't mistake it for an oversight.

Rest of the verification:
- Bed branch still evaluated first (`:216-218`) and still gated on `bed?.kind === "audio"`, so a `<video>` bed is still spared — `carve.test.mjs:93-95` unchanged and still passing.
- `detectTracks` returns `all` (`:362`); `members` is built from `media.filter((el) => el.kind === "audio")` with `unescapeAttr` on `src` (`:385-392`), matching how `kindOf` composes id + src inside `detectTracks`. Same classifier, same inputs.
- `bedInVoiceGroup` was removed as an export — grepped the tree at this head, zero remaining references.
- The stderr note now names the blocking member ids and distinguishes the two reasons (`:498-510`), pinned by `carve.test.mjs:149-165`. That matters for the case the note exists for: an author who *did* group their clips reading `audio_carve_ungrouped_sources`.
- SKILL.md (`:261-290`) documents both refusals and the voice-left-out carve-out, and claims only what the code does — `carve.mjs` refuses, records clip ids, says which member blocked it. `skills-manifest.json` hash bumped for the 7-file skill.

**important — not blocking, and not a defect at this head: the wiring is the one link no test covers, and the default hides a break.** `carveSources` and `groupSourceRefusal` both default `members = []` (`:174`, `:213`), and with `[]` the `mixed` refusal cannot fire. Every unit test passes `members` explicitly; the only code that *builds* it is `main()` (`:385-392`), which no test exercises — the symlink test only reaches the usage error, and a real run needs ffmpeg. So a refactor that drops the third argument at `:493` silently restores the pre-fix behaviour with the whole suite green. That is the same signature as the bug itself: first pass correct, artifact wrong, nothing red. Cheapest fix is dropping both `= []` defaults so a missing argument throws on `members.filter` instead of quietly returning the group form — costs updating the nine legacy call sites at `carve.test.mjs:43-95` to pass `[]`, which also documents that those cases predate the membership check.

**CI:** 46 distinct checks at `6f0b634e` (full matrix, since this rung's base is `main`) — 45 green, `Tests on windows-latest` still running when I looked, nothing failed. `Render on windows-latest` passed.

Not stamping. My `CHANGES_REQUESTED` ([4999953165](https://github.com/heygen-com/hyperframes/pull/3416#pullrequestreview-4999953165)) is pinned to `a7168775`, three heads back, and every finding in it is satisfied in substance — read it as satisfied-awaiting-stamp, not as an open objection. I'm not converting it to an approval myself because approvals here are James's call, so if it's what's holding the gate, that's the route to clear it.

**Verdict:** COMMENT
**Reasoning:** The blocker I withdrew my clearance on is properly fixed at the writer with the right invariant and honest scoping; the one remaining item is a test-coverage gap on the wiring, not a defect in it.

— Rames Jusso

vanceingalls:
Fixed at `721e81f38` — the wiring gap @jrusso1020 flagged as the one remaining item.

Both `= []` defaults are gone, so omitting `members` throws on `members.filter` instead of quietly returning the group form. The nine cases that predate the membership check now pass `[]` explicitly, which also documents that they are about the bed and the group attributes alone rather than membership.

Added the test that closes the gap, and verified it discriminates: restoring the defaults fails it, keeping them removed passes. 19 tests, all green.

The reasoning is worth restating because it is the whole point of the fix — this had the same signature as the bug the argument exists to prevent. `main()` sums the detected voice list directly, so the first CLI pass is correct however wrong the persisted attribute is, and only a later Studio re-analysis reads the widened `sources`. Nothing goes red. And `main()` is the only code that builds `members`, which no test reaches: the symlink test stops at the usage error, and a real run needs ffmpeg. So a dropped argument was invisible from every direction.

---

On the two open items from this round, neither of which is in this diff:

**The Studio picker's pre-existing-group case.** @jrusso1020's trace is right and I have not touched it here. `collectCarveCandidates` offers a group row when any member is in `others`, so a bed sharing a pre-existing group with a voice is offered, and selecting it writes `sources: [groupId]` unchanged. `bedInVoiceGroup` was the natural predicate to reuse — but that export is gone as of the mixed-membership fix, so the reuse now means `groupSourceRefusal` and the `members` list the picker would have to build. Worth its own PR rather than widening this one's scope, and I would rather not have a second writer land a half-shaped version of the same guard.

**`--preserve-symlinks-main`.** Agreed on leaving it: the flag inverts all four copies identically, so consistency across them beats special-casing one. Noted so nobody "fixes" a single copy out of step.

@jrusso1020 @miguel-heygen — your `CHANGES_REQUESTED` reviews are pinned to earlier heads (`a7168775` and the pre-`6f0b634e` round). Every finding in both is satisfied in substance at this head, and the remaining Studio-picker item is out of this PR's scope by both your reads. A re-review at `721e81f38` is what the gate needs.

jrusso1020 (commented):
Re-review at `721e81f3`. The one item I raised is closed, and closed better than I asked.

**`members` is now required on both functions** (`:174`, `:220`), with the reasoning written into the doc comment (`:194-200`) rather than left for the next reader to reconstruct. What makes it right is the scoping, which is easy to misread as partial: the throw only fires when the voices share a group *and* the bed branch doesn't return first — and those are exactly the inputs where `members` changes the answer. On `!group` the result is clip ids regardless; on the bed refusal it declines regardless. So it throws precisely on the outcome-changing path and stays quiet where omitting the argument is harmless. The fixture at `:229-238` is that path specifically (group-sharing voices, non-member bed), not an arbitrary one.

**I ran it rather than reasoning about it.** Extracted `carve.mjs` + `carve.test.mjs` at this head and ran `node --test`: **19/19 pass**. Then the probe that matters — restoring both `= []` defaults and re-running: **18 pass, 1 fail, and the failure is test 19 only**. So the pin is sensitive to exactly the regression it names and nothing else; it isn't a test that would stay green through the thing it's guarding.

*nit, and the wrong number is mine.* `carve.test.mjs:75` says "The nine cases above". There are **six** tests above that comment, and **ten** call sites pass `[]` in total — the six above plus the four bed cases below it. "Nine" came from my last review, where I wrote "the nine legacy call sites at `carve.test.mjs:43-95`"; I miscounted and it propagated into the comment. Simplest fix is to drop the count ("The cases above pass `[]` explicitly because…"), since it rots on the next added case either way.

**Don't push for that nit alone**, though — fold it into a commit you're making anyway. `main` here runs `dismiss_stale_reviews_on_push: false` with `require_last_push_approval: true`, so @miguel-heygen's approve at `6f0b634e` still exists but no longer satisfies the gate now that `721e81f3` is the last push; another push restarts that. Trading a fresh approval cycle for a comment's word count isn't worth it.

**CI:** 45 distinct checks at `721e81f3`, nothing failed, 9 still running when I looked — including `Test` and seven `regression-shards`. I'm not calling it green; the unit tests I verified myself above, and the shards are the part still outstanding.

**Gate, unchanged and still mine:** `reviewDecision` is `CHANGES_REQUESTED` on the strength of my `4999953165` at `a7168775`, four heads back, with nothing in it unaddressed. `latestReviews` still doesn't show it. Clearing it needs an approval I'm not authorised to give, so it stays James's call — and per the rule above, that approval has to land at the current head to count.

**Verdict:** COMMENT
**Reasoning:** The required-argument guard is correctly scoped and independently verified sensitive; what's left is one miscounted word in a comment that I introduced, and a stamp that isn't mine.

— Rames Jusso

miguel-heygen (approved):
**Current-head delta verified:** `carve.mjs:174` and `:220` remove both empty defaults, so omitting `members` cannot silently bypass mixed-membership refusal. The guard is scoped correctly by short-circuiting: no-group inputs return ids without consulting members, and the self-bed refusal returns before `members.filter`; a missing argument throws only on the shared-group path where membership can change the persisted answer.

`carve.test.mjs:229-238` pins that exact outcome-changing path. The mutation result is discriminating: restoring both defaults yields 18 pass / 1 fail, while the current head is 19/19. The previous symlink, published-loader, self-bed, and music/SFX widening regressions remain covered.

**Nit:** `carve.test.mjs:75` says nine cases above; there are six above and ten total passing `[]`. Drop the count next time the file moves; do not repush just for it.

No current-head check is failing; Windows and three regression shards are still running and remain merge gates.

**Verdict:** APPROVE
**Reasoning:** The final wiring regression is now fail-closed and mutation-pinned at exact head `721e81f3`; all code findings are resolved, with CI still enforcing the remaining in-flight jobs.

— Magi

PR #3417, opened 2026-08-22
fix(studio): stop the grouping dialog opening off the bottom of the window

vanceingalls:
Stacked on #3415. Fourth in the restore series — see #3413 for why these fixes were stranded.

Reported as: **"the grouping button didn't work — I clicked it and nothing happened."**

## It was opening. You couldn't see it.

The group-creation dialog positioned itself raw:

```jsx
style={{ position: "fixed", left: anchorRect.left, top: anchorRect.bottom + 4 }}
```

No flip, no clamp, no cap. The timeline lives at the **bottom** of the studio window, so `anchorRect.bottom + 4` is at or past the viewport's bottom edge — the dialog rendered off-screen. It was the only floating surface in this feature with no viewport handling at all; the FX popover beside it has had a clamp since it was written.

Measured in the test: an anchor at `top: 760` in a 768px-tall window. Before, `top: 780px` — 12px below the bottom edge. Now it flips to `bottom: 12px`.

The clamp moves to `components/editor/floatingPanelPosition.ts` and **both** surfaces use it, so the next floating surface can't drift the same way. Popover behaviour is unchanged — same numbers, now parameterised (`width`, `preferredHeight`, `minHeight`).

## A second silent path on the same click

`TimelineTrackHeader` offers the grouping pointer when `clipCount > 1`, but the write bails returning `[]` when fewer than two clips **resolve** against the expanded-rows list:

```js
if (elements.length < 2) return []; // createAudioGroupAndAssignMembers
```

Two different counts: one decides whether the button appears, the other whether anything happens. That bail is right for the carve picker (genuinely nothing to group) and invisible for a button someone just pressed — and it's reachable, because `useExpandedTimelineElements` filters to top-level rows when the playhead isn't inside a sub-composition, so sub-comp audio can resolve to nothing.

A request for 2+ clips that resolves to fewer now reports instead of failing silently, and logs both id sets for diagnosis. `showToast` was already wired into that hook and unused on this path.

## Tests

Two, with a realistic bottom-of-window anchor: the upward flip, and the right-edge clamp. Both verified to fail against the raw positioning.

Worth noting why this shipped: the **existing** `group-pointer` test passes either way. happy-dom reports an all-zero rect for an unlaid-out button, so the dialog landed at `top: 4px` — on screen, and nothing like the real app. A geometry test that never sets a geometry proves nothing.

## Incidental

`useAudioGroupCarveAssignment` moves to its own module. It isn't track-visibility work, and `timelineTrackVisibility.ts` was at 597 of the studio 600-line ceiling — adding the diagnostic would have broken the `filesize` gate. Now 534 + 90.

## Checks

`tsc --noEmit` clean · full studio suite **4378 passed / 0 failed** · `oxlint` 0/0 · `oxfmt --check` clean · `fallow audit --base origin/main` passes (its 3 duplication findings are inherited, all in files/blocks this PR didn't introduce) · every touched file under the 600-line limit.


vanceingalls:
Superseded by #3421 — closing this.

This PR was written on top of #3413–#3415 and went stale when they merged: it now conflicts with `main`, and more importantly it added a **second** positioning helper (`floatingPanelPosition.ts`) at the same time the #3413/#3414 reviews were extracting `floatingPanel.ts` for the same job. Landing both would have left two parallel implementations of viewport clamping.

#3421 is the same fix rebuilt on `main`, using `resolveFloatingPanelPosition` — the helper `RenderQueue` and `propertyPanelColor` already use — so there is one implementation rather than two. Same two regression tests, both still verified to fail against the raw positioning.

Also dropped on the way: the toast for the grouping write's silent `elements.length < 2` bail, and the `useAudioGroupCarveAssignment` extraction it forced to stay under the 600-line cap. The bail is real in code but I could not reach it from the UI, and #3421 says so rather than shipping a message for a branch I cannot demonstrate.

PR #3425, opened 2026-08-22
fix(producer): fall back to a temp dir when the font cache is unwritable

### LINKED ISSUE #3412: Unwritable font cache directory aborts the render instead of degrading
## Describe the bug

A render fails outright when the font **cache** directory cannot be created. The cache is an
optimisation — it stores Google Fonts downloads between runs — but failing to create it is
fatal, so a first render dies with a raw `mkdir` error.

Reported from the field on 0.8.8, on a first-ever render:

```
EPERM: operation not permitted, mkdir '<home>/.cache/hyperframes/fonts/inter'
```

The remediation the CLI offers for this failure is `Try --docker for containerized
rendering`, which does not address an unwritable cache root.

## Steps to reproduce

Platform-independent — any unwritable cache root reproduces it:

```bash
mkdir -p /tmp/rocache && chmod 500 /tmp/rocache
# a composition using any Google font (e.g. Inter)
HYPERFRAMES_FONT_CACHE_DIR=/tmp/rocache/nested npx hyperframes render --output renders/o.mp4
```

## Actual behavior

```
✗  Render failed

   EACCES: permission denied, mkdir '/tmp/rocache/nested'
```

Same shape as the field report (`EACCES` vs `EPERM` are the same class here: the cache root
is not writable).

## Expected behavior

An unwritable cache should degrade, not abort. The font can still be fetched per-run, or
cached to a temp directory. The render only needs the font bytes; where they are stored
between runs is an optimisation detail.

## Root cause

`packages/producer/src/services/deterministicFonts.ts:612`:

```ts
function fontCacheDir(slug: string): string {
  const dir = join(resolveFontCacheRoot(), slug);
  if (!existsSync(dir)) {
  }
  return dir;
}
```

The `mkdirSync` is unguarded, and this runs during compile, so the throw propagates and
fails the render.

## Notes

- `HYPERFRAMES_FONT_CACHE_DIR` exists and would let an affected user redirect the cache, but
  it appears in **no** user-facing guide or reference page (`grep -rlI
  HYPERFRAMES_FONT_CACHE_DIR docs` matches only changelog / weekly-update / plan files). A
  user hitting this has no documented way out.
- Worth fixing the remediation hint at the same time: suggesting Docker for a permissions
  error on the host cache sends people the wrong way.

miguel-heygen:
## What

An unwritable font cache directory no longer aborts the render. It falls back to a temp
directory and warns once.

Closes #3412.

## Why

The font cache holds Google Fonts downloads between runs. It is an **optimisation** — the
bytes are still fetchable without it — but failing to create it was fatal, because
`fontCacheDir` called `mkdirSync` unguarded during compile and the throw propagated out of
the render.

A first-time user on 0.8.8 lost their very first render to it:

```
EPERM: operation not permitted, mkdir '<home>/.cache/hyperframes/fonts/inter'
```

Two things made it worse than the bug. The CLI's remediation was `Try --docker for
containerized rendering`, which does not address an unwritable host directory. And
`HYPERFRAMES_FONT_CACHE_DIR` — which would have rescued them — appears in **no** user-facing
guide page.

## How

`fontCacheDir` now catches the failure and falls back to one temp root per process, the same
shape the Lambda cache root a few lines above already uses:

```ts
fallbackFontCacheRoot ??= mkdtempSync(join(tmpdir(), "hyperframes-fonts-"));
```

A run still de-duplicates its own downloads; it just cannot reuse them next time. The
one-time warning names `HYPERFRAMES_FONT_CACHE_DIR`, so the escape hatch is discoverable at
the moment it is needed.

## Test plan

**End to end, before and after.** Pointing `HYPERFRAMES_FONT_CACHE_DIR` at a path under a
`chmod 500` parent, rendering a composition that requests Inter.

Before:

```
✗  Render failed
   EACCES: permission denied, mkdir '.../rocache/nested'
```

After:

```
[WARN] [Compiler] Font cache is not writable (EACCES: permission denied, mkdir '...').
       Caching to a temporary directory for this run instead; fonts will be re-downloaded
       next time. Set HYPERFRAMES_FONT_CACHE_DIR to a writable path to keep them.
  █████████████████████████  100%  Render complete
```

**New regression test** `deterministicFonts-unwritableCache.test.ts` drives
`injectDeterministicFontFaces` with the cache root under a read-only parent. Verified it
fails on the unfixed code rather than assuming:

```
EACCES: permission denied, mkdir '/var/folders/.../hf-font-locked-nSYU6R/nested'
(fail) unwritable font cache > still resolves fonts instead of aborting the render
```

All `deterministicFonts` suites green: **61 pass across 8 files**.

## Not covered

- The misleading `Try --docker` remediation is unchanged. It is emitted generically for
  render failures, so fixing it is a separate change about how that hint is chosen.
- `HYPERFRAMES_FONT_CACHE_DIR` is still absent from the docs site. Now surfaced in the
  warning, but a real docs page would be better.

### DIFF
--- a/packages/producer/src/services/deterministicFonts-unwritableCache.test.ts
+++ b/packages/producer/src/services/deterministicFonts-unwritableCache.test.ts
@@ -1,0 +1,68 @@
+/**
+ * The font cache holds Google Fonts downloads between runs. It is an
+ * optimisation: the bytes are still fetchable without it, so losing the place
+ * to keep them is a slower render, not a failed one.
+ *
+ * It used to be fatal. `fontCacheDir` called `mkdirSync` unguarded during
+ * compile, so an unwritable cache root threw straight out and a first-time user
+ * lost their very first render to
+ * `EPERM: operation not permitted, mkdir '<home>/.cache/hyperframes/fonts/inter'`.
+ */
+
+import { afterAll, beforeAll, describe, expect, it } from "bun:test";
+import { chmodSync, mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+let lockedRoot: string;
+let prevCacheEnv: string | undefined;
+
+const FACE_URL = "https://fonts.gstatic.com/s/inter/v1/inter.woff2";
+const FACE_BYTES = "INTER_BYTES";
+
+beforeAll(() => {
+  prevCacheEnv = process.env.HYPERFRAMES_FONT_CACHE_DIR;
+  lockedRoot = mkdtempSync(join(tmpdir(), "hf-font-locked-"));
+  chmodSync(lockedRoot, 0o500);
+  // A path *under* a read-only parent: creating it is what fails.
+  process.env.HYPERFRAMES_FONT_CACHE_DIR = join(lockedRoot, "nested");
+});
+
+afterAll(() => {
+  if (prevCacheEnv === undefined) delete process.env.HYPERFRAMES_FONT_CACHE_DIR;
+  else process.env.HYPERFRAMES_FONT_CACHE_DIR = prevCacheEnv;
+  chmodSync(lockedRoot, 0o700);
+  rmSync(lockedRoot, { recursive: true, force: true });
+});
+
+const fetchImpl = (async (input: unknown) => {
+  const url = String(input);
+  if (url.startsWith("https://fonts.googleapis.com/")) {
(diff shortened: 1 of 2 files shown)


jrusso1020:
Closing as superseded by #3572, which provides the temporary font-cache fallback with a unique directory and once-only warning. The regression test here remains available for a focused coverage follow-up. Thanks for the contribution.

PR #3428, opened 2026-08-22
docs: say one paused ROOT timeline, not "timelines must be paused"

miguel-heygen:
## What

The four compressed copies of the timeline contract now say what the skill says: **one**
paused timeline, the registered root, and don't nest your own paused sub-timelines into it.

## Why

`skills/hyperframes-core/SKILL.md` has always been correct:

> Each composition registers **exactly one** `gsap.timeline({ paused: true })` ... Don't
> manually nest sub-timelines into the host.

Every compressed copy dropped both halves:

> GSAP timelines must be paused and registered on `window.__timelines`.

Plural, no cap, no warning against nesting. An author following *that* writes a paused
timeline per scene and combines them with `.add()` — exactly the shape that renders black,
because a paused child never advances when the root is seeked. Every frame is the t=0 state,
and `lint`, `check` and `validate` all pass, none of them looking at pixels.

This is the drift the skill-catalog sync rule exists to prevent: a compressed copy asserting
something the skill does not.

## How

Four surfaces, all previously carrying the same sentence:

- `CLAUDE.md`
- `AGENTS.md`
- `packages/cli/src/templates/_shared/CLAUDE.md`
- `packages/cli/src/templates/_shared/AGENTS.md`

The two template files are written into every `hyperframes init` project and must stay
byte-identical — verified with `diff -q` after the edit.

## Relationship to #3427

[#3427](https://github.com/heygen-com/hyperframes/pull/3427) makes the runtime repair this
shape, so nobody following the old wording is stuck. This PR stops it being authored in the
first place. Either alone is a partial fix: the runtime one leaves the misleading guidance in
place, and this one leaves every existing composition broken.

## Not covered

- `docs/` prose is untouched; a grep for the old sentence across `docs/`, `README.md` and
  `docs/guides/` returns nothing, so these four were the whole set.
- No lint rule. #3427's runtime repair cannot be evaded, so a static rule would only
  duplicate it for the subset it can see.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -80,7 +80,7 @@ skills/                 → AI agent skill definitions
 - **Package manager**: bun (not pnpm, not npm for workspace operations)
 - **Commit format**: Conventional commits (`feat:`, `fix:`, `docs:`, `refactor:`, `test:`)
 - **TypeScript**: Avoid `any` and `as T` assertions. Prefer type guards and narrowing.
-- **Compositions**: HTML files with `data-*` attributes. Clips need `class="clip"`. GSAP timelines must be paused and registered on `window.__timelines`.
+- **Compositions**: HTML files with `data-*` attributes. Clips need `class="clip"`. Each composition registers **exactly one** paused GSAP timeline on `window.__timelines`. Don't nest your own paused sub-timelines into it: a paused child never advances when the root is seeked, so the render comes out blank.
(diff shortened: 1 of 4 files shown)


miguel-heygen:
Closing this stale branch. Current main still has four compressed timeline instructions that omit the one-root contract. The runtime repairs registered sibling timelines, but not arbitrary paused timelines manually nested by an author. A small fresh documentation correction remains worthwhile.

miguel-heygen:
Narrow verified successor: https://github.com/heygen-com/hyperframes/pull/3670

PR #3432, opened 2026-08-23
Fix Windows console window for chrome-headless-shell

### LINKED ISSUE #3430: [Windows] chrome-headless-shell.exe opens a visible console window on every render
### Describe the bug

On Windows, every call to `npx hyperframes render` causes a visible CMD/console 
window to flash open and close for each shot rendered. When rendering a batch 
(e.g. 78 shots), 78 console windows open sequentially, making the machine 
completely unusable during the render. This happens even when HyperFrames is 
called from a GUI application with no terminal involved, confirming the window 
is being created inside HyperFrames itself.

### Link to reproduction

Not applicable — this bug reproduces on every Windows machine with any render call,  including a blank project created with:  npx hyperframes init repro --non-interactive --example blank npx hyperframes render ./repro --format png  No custom code is needed to trigger it. The console window appears on all Windows  installs regardless of project content.

### Steps to reproduce

1. On Windows 11, run: npx hyperframes init repro --non-interactive --example blank
2. Run: npx hyperframes render ./repro --format png
3. Observe a CMD/console window flash open and close during render
4. Run a batch of renders (multiple shots) — one console window opens per shot

### Expected behavior

No console window should appear. Chrome headless should run silently in the background.

### Actual behavior

A console window opens for every shot rendered, showing the path:
C:\Users\<username>\.cache\hyperframes\chrome\chrome-headless-shell\win64-152.0.7977.30\chrome-headless-shell-win64\chrome-headless-shell.exe

Suppressing the window from the calling side (Python subprocess CREATE_NO_WINDOW + 
STARTUPINFO) has no effect because the window is opened by chrome-headless-shell.exe 
which is spawned inside HyperFrames' own Node.js code.

Suspected fix: add `windowsHide: true` to the Node.js child_process.spawn() call 
that launches chrome-headless-shell.exe.

### Environment

```shell
hyperframes doctor

  ✓ Version          0.8.10 (latest)
  ✓ Node.js          v22.19.0 (win32 x64)
  ✓ CPU              16 cores · AMD Ryzen 7 6800H with Radeon Graphics @ 3194MHz
  ✓ Memory           13.7 GB total · 3.0 GB available
  ✓ Chrome           cache: C:\Users\username\.cache\hyperframes\chrome\chrome-headless-shell\win64-152.0.7977.30\chrome-headless-shell-win64\chrome-headless-shell.exe
  ✓ FFmpeg           N-120424-g03b9180fe3-20250731
  ✓ Node environment non-TTY
```

aryansk:
Fixes #3430

### Problem

On Windows, every `npx hyperframes render` flashed a visible CMD/console window for each shot (78 windows for 78 shots). `chrome-headless-shell.exe` is a console-subsystem binary, so Node's `child_process.spawn` shows a console window unless `windowsHide: true` (`CREATE_NO_WINDOW`) is set. `puppeteer-core`'s `launch` spawns the browser without `windowsHide` by default, so the window was visible on every launch, even from GUI apps with no terminal.

### Change

Add `windowsHide: true` to all `puppeteer.launch` sites so the browser is launched hidden on Windows:

- `packages/engine/src/services/browserManager.ts` (pooled browser for `hyperframes render`)
- `packages/cli/src/capture/*`, `packages/cli/src/beats/headlessAnalyzer.ts`, `packages/cli/src/commands/*`, `packages/studio/vite.browser.ts`, `packages/producer/src/parity-harness.ts`

On non-Windows platforms `windowsHide` is ignored.

### Validation

- `git diff --check` clean
- Change is isolated to `puppeteer.launch` options; existing `chrome-headless-shell` path resolution and launch-arg tests expected to be unaffected

### DIFF
--- a/packages/cli/src/beats/headlessAnalyzer.ts
+++ b/packages/cli/src/beats/headlessAnalyzer.ts
@@ -140,6 +140,8 @@ export async function analyzeBeatsHeadless(audioBytes: Buffer): Promise<Headless
   const browser = await ensureBrowser();
   const chrome: Browser = await puppeteer.default.launch({
     headless: true,
+      // @ts-ignore -- windowsHide not in PuppeteerLaunchOptions but forwarded to spawn on Windows (see #3430)
+      windowsHide: true,
     executablePath: browser.executablePath,
     args: ["--no-sandbox", "--disable-dev-shm-usage", "--autoplay-policy=no-user-gesture-required"],
   });
--- a/packages/cli/src/capture/captureCompositionFrame.ts
+++ b/packages/cli/src/capture/captureCompositionFrame.ts
@@ -182,6 +182,8 @@ export async function openSettledCompositionPage(
   try {
     chromeBrowser = await puppeteer.default.launch({
       headless: true,
+      // @ts-ignore -- windowsHide not in PuppeteerLaunchOptions but forwarded to spawn on Windows (see #3430)
+      windowsHide: true,
       executablePath: browser.executablePath,
       args: buildChromeArgs(
         { ...viewport, captureMode: "screenshot" },
--- a/packages/cli/src/capture/index.ts
+++ b/packages/cli/src/capture/index.ts
@@ -128,6 +128,8 @@ export async function captureWebsite(
   const puppeteer = await import("puppeteer-core");
   const chromeBrowser = await puppeteer.default.launch({
     headless: true,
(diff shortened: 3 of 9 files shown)


jrusso1020:
Closing this approach as superseded by the Puppeteer update in #3394; windowsHide is not a supported Puppeteer launch option. This does not establish that every Windows case in #3430 is resolved, so that issue should remain available for reproduction. Thanks for the contribution.

PR #3433, opened 2026-08-23
fix(cli): keep the preview server alive on an uncaught exception

### LINKED ISSUE #3373: Studio preview server dies on any uncaught exception, surfacing as "Connection lost"
## Describe the bug

The Studio preview server shares the one-shot CLI's process-wide `uncaughtException`
handler, which calls `process.exit(1)`. For a `render` invocation that is correct — the
command is over anyway. For a long-running server it means **any** uncaught error anywhere
in the process tears down the server, dropping every open SSE connection at once.

Every Studio client watching a render then shows:

```
Connection lost. Is the render server running?
```

which blames the user's setup for something that was a crash inside our own process.

## The chain, verified in source

**1. The handler is process-wide and fatal** — `packages/cli/src/cli.ts:393`:

```ts
process.on("uncaughtException", (error) => {
  if ((error as NodeJS.ErrnoException).code === "EPIPE") { ... }
  if (isRenderSucceeded()) {
    exitAfterPostRenderTermination("uncaughtException", "uncaught_exception", error);
  }
  exitAfterCliFailure("uncaught_exception", error);   // -> process.exit(1)
});
```

**2. The only escape hatch is never set by the server.** `markRenderSucceeded()` is called
from exactly two places, both inside the one-shot render command —
`packages/cli/src/commands/render.ts:748` and `:913`. The `preview` subcommand that hosts
Studio never calls it, so `isRenderSucceeded()` is permanently `false` there and every
uncaught error takes the `exitAfterCliFailure` branch.

**3. Studio reports the resulting socket drop as a connectivity problem** —
`packages/studio/src/components/renders/useRenderQueue.ts:259` sets
`error: "Connection lost. Is the render server running?"` when the `EventSource` errors on a
job still in `rendering`.

Note the asymmetry with the normal failure path: a render job that fails *cleanly* is caught
in `packages/cli/src/server/studioServer.ts` and delivered as a graceful `status: "failed"`
SSE event carrying the real error text. Only a process death produces the generic
"Connection lost" wording. So the message a user sees tells you which of the two happened,
and this one means we exited.

## Why this matters beyond the wording

A one-shot command exiting on an uncaught exception loses nothing. A server doing it loses
every concurrent job and every connected client, and the users who hit it are told to check
whether they started the server — which they did.

It also makes the failure look reproducible from the user's side ("happens every time")
while being invisible from ours, because the thing that died left no per-job error behind.

## What is verified and what is not

**Verified:** the handler, the missing exemption for the preview server, the `process.exit(1)`,
and the Studio error string. Those are facts about the code as it stands.

**Not verified:** that this is what produced any particular field report. Chasing the
suspected trigger — a browser target dying mid-capture — did not reproduce: **0 failures in
48 trials** of a 1050-frame composition across `--workers 1 --low-memory-mode`,
`--no-low-memory-mode` with default workers, and `--workers 4`, on both macOS and Linux. By
rule of three that puts the true crash rate under roughly 6%, consistent with "rare" and not
with "reliably reproducible". So no live escape-path evidence was captured, and no specific
unawaited listener has been identified.

That does not change the recommendation, because the defect here is structural rather than
dependent on any one trigger.

## Suggested fix

Do not let a render job's uncaught error kill the process that is also serving the SSE
connection reporting on it. Either:

- isolate the render-job promise chain in the preview server so its failures cannot reach
  the global handler, or
- exempt the preview server from the fatal branch the same way `isRenderSucceeded()` already
  exempts a completed CLI render, logging and failing the affected job instead of exiting.

The second is closer to the existing design and reuses machinery that is already there.

Either way the outcome should be that a crashed render fails *that job* with its real error,
and every other client stays connected.

## Environment

```shell
hyperframes  0.8.4
Affected     packages/cli/src/cli.ts:393
             packages/cli/src/commands/render.ts:748,913 (only markRenderSucceeded callers)
             packages/studio/src/components/renders/useRenderQueue.ts:259
```

rajanpanth:
Fixes #3373

### Problem

The preview server shares the one-shot CLI's process-wide `uncaughtException` handler, which ends in `process.exit(1)`:

```ts
process.on("uncaughtException", (error) => {
  if ((error as NodeJS.ErrnoException).code === "EPIPE") handleStreamEpipe();
  if (isRenderSucceeded()) {
    exitAfterPostRenderTermination("uncaughtException", "uncaught_exception", error);
  }
  exitAfterCliFailure("uncaught_exception", error);   // -> process.exit(1)
});
```

For `render` that is correct: the command is over anyway and the exit code is the result. For `preview`, which hosts Studio, it means **any** uncaught error anywhere in the process tears the server down and drops every open SSE connection at once. Studio then shows every connected client:

```
Connection lost. Is the render server running?
```

which blames the user's setup for a crash inside our own process.

The only escape hatch, `isRenderSucceeded()`, is set exclusively by the render command (`commands/render.ts:753` and `:920`), so in the preview server it is permanently `false` and every uncaught error took the fatal branch.

### Fix

A small server-mode sentinel, mirroring the existing `render-success-state.ts` shape:

- `utils/server-mode.ts` holds `markServerMode()` / `isServerMode()`, plus a test-only reset.
- `commands/preview.ts` calls `markServerMode()` at the single point where every remaining path is long-running, immediately after the short-lived `--status`, `--stop` and `--list` branches have returned and before dispatching to dev, local-studio or embedded mode.
- The `uncaughtException` handler checks it after the EPIPE and post-render branches, so neither of those changes behaviour. In server mode it reports the error to telemetry and stderr and returns, leaving the process running.

The error is not swallowed: it still goes through `emitCliErrorEvent`, and the stack is written to stderr so the crash stays visible for diagnosis.

### Scope

I deliberately left `unhandledRejection` alone. It already does not exit, so it does not cause the reported teardown, and changing how it sets `commandFailed` / `process.exitCode` would alter behaviour beyond this issue. Happy to follow up if you would like the server exempted there too.

I also did not touch the Studio wording in `useRenderQueue.ts:259`. As the issue notes, that message is now a genuine signal that the process died, and with this change it should stop appearing for uncaught errors, which seemed more useful than rewording it.

### Testing

Two cases added to `cli.lifecycle.test.ts`, alongside the existing EPIPE and post-render ones:

- a one-shot command still exits 1 on an uncaught exception
- in server mode the process stays up, `process.exit` is not called, and the report reaches stderr

They invoke the CLI's own registered listener directly rather than emitting on `process`, which would also reach vitest's handler and fail the run.

Reverting the `isServerMode()` guard while keeping the tests fails the server case, so it covers the change rather than restating current behaviour.

`bunx oxlint` and `bunx oxfmt --check` are clean on all four files. On the wider CLI suite I get 401 passing (up from 399 with the two added), with 15 test *files* failing to collect both before and after my change on a fresh `bun install` without a workspace build, so that set looks unrelated.

### DIFF
--- a/packages/cli/src/cli.lifecycle.test.ts
+++ b/packages/cli/src/cli.lifecycle.test.ts
@@ -68,6 +68,49 @@ describe("CLI lifecycle", () => {
     expect(order).toEqual(["cli_error", "flush"]);
   });
 
+  it("exits on an uncaught exception for a one-shot command", async () => {
+    const exitSpy = vi.spyOn(process, "exit").mockImplementation((() => undefined) as never);
+    try {
+      mockInitCommand(vi.fn());
+      mockTelemetry({});
+      process.argv = ["node", "cli.ts", "init", "--json"];
+      await import("./cli.js");
+
+      emitUncaught(new Error("boom"));
+
+      // A one-shot command is over anyway, so the exit code is the result.
+      expect(exitSpy).toHaveBeenCalledWith(1);
+    } finally {
+      exitSpy.mockRestore();
+    }
+  });
+
+  it("keeps a long-running server alive on an uncaught exception", async () => {
+    const exitSpy = vi.spyOn(process, "exit").mockImplementation((() => undefined) as never);
+    const stderr = vi.spyOn(process.stderr, "write").mockImplementation(() => true);
+    try {
+      mockInitCommand(vi.fn());
+      mockTelemetry({});
+      process.argv = ["node", "cli.ts", "init", "--json"];
+      await import("./cli.js");
+
+      const serverMode = await import("./utils/server-mode.js");
+      serverMode.markServerMode();
+
+      emitUncaught(new Error("boom"));
+
+      // Exiting would drop every open SSE connection at once, which Studio
+      // reports to every client as "Connection lost. Is the render server
+      // running?" — so the server stays up and the error is reported instead.
+      expect(exitSpy).not.toHaveBeenCalled();
+      expect(stderr).toHaveBeenCalledWith(expect.stringContaining("server kept running"));
+      serverMode._resetServerModeForTests();
+    } finally {
+      stderr.mockRestore();
+      exitSpy.mockRestore();
+    }
+  });
+
   it("hands queued events to flushSync even after finalizeCli has run", async () => {
     const flushSync = vi.fn();
     const trackCommandResult = vi.fn();
@@ -182,6 +225,15 @@ describe("CLI lifecycle", () => {
     }
   });
 });
+
+function emitUncaught(error: Error): void {
+  // Invoke the CLI's own listener directly. Emitting on `process` would also
+  // reach vitest's handler and fail the run.
+  const listeners = process.listeners("uncaughtException");
+  const handler = listeners[listeners.length - 1];
+  if (!handler) throw new Error("no uncaughtException listener registered");
+  handler(error, "uncaughtException");
+}
 
 function mockInitCommand(run: () => void): void {
   vi.doMock("./commands/init.js", () => ({
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -105,6 +105,7 @@ import type { ArgsDef, CommandDef } from "citty";
 import { getRunId } from "./telemetry/runId.js";
 import { reportCommandFailure, trackCommandFailures } from "./utils/command-failure-tracking.js";
 import { isRenderSucceeded } from "./utils/render-success-state.js";
+import { isServerMode } from "./utils/server-mode.js";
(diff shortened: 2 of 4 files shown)


miga-heygen:
Closing per repo owner's decision after deep review.

The structural observation is real (preview server inherits the CLI's `process.exit(1)` handler), but continuing after `uncaughtException` is explicitly unsafe per Node.js docs — the process may be in an inconsistent state (open file handles, half-written state, corrupted in-memory caches). The preview server holds SSE connections, file watchers, and build state, making this particularly risky.

Additionally, no concrete reproduction was found — every async path in the preview server appears properly guarded. The right fix is either (a) finding and fixing the specific uncaught exception that's actually crashing preview, or (b) process-level supervision for general resilience.

The PR is well-written and the analysis is solid — the issue just needs a different approach.

— Miga

Issue #3372, opened 2026-08-20
Video extraction failures are swallowed, surfacing as a misleading coverage abort


miguel-heygen:
## Describe the bug

A render aborts with a coverage error that names the wrong subsystem, because the real
failure — a per-source video extraction that produced zero frames — is swallowed by default
and never surfaces.

The reported symptom:

```
Video "bg-blur" captured 0 of expected 36 frames (coverage 0.0%, threshold 95.0%).
check/snapshot may pass while the encoded MP4 renders this clip blank — aborting render
to prevent shipping a wrong MP4. Set HF_VIDEO_COVERAGE_THRESHOLD=0 to disable this gate.
```

The coverage gate is not at fault. It is the only component still talking by the time the
clip arrives with no frames.

## Steps to reproduce

Cold cache, one local file, one clip, default flags. The only unusual thing is a source
whose **audio outlasts its video** — which is what screen recorders, trimmed exports, and
image-plus-music renders routinely produce.

```bash
npx --yes hyperframes@0.8.4 init proj --example blank --non-interactive
cd proj && mkdir -p assets

# video stream 3s, audio 8s -> container reports duration 8s
ffmpeg -y -f lavfi -i "testsrc=size=640x360:rate=30:duration=3" \
       -f lavfi -i "sine=frequency=440:duration=8" \
       -c:v libx264 -pix_fmt yuv420p -c:a aac assets/tail.mp4
```

Set the root to `data-duration="1.2"` and use one clip whose media-start lands in the
audio-only tail:

```html
<div class="clip" data-start="0" data-duration="1.2" data-track-index="0">
  <video id="bg-blur" src="assets/tail.mp4" data-media-start="5"></video>
</div>
```

```bash
npx --yes hyperframes@0.8.4 render --output out.mp4
```

## Expected behavior

The render fails naming the actual cause: extraction produced no frames because the
requested media-start is past the end of the video stream.

## Actual behavior

```
ffprobe container duration:     8.000000
ffprobe video stream duration:  3.000000

totalFramesExtracted":0
minVideoFrameCoverageRatio":0
captured 0 of expected 36 frames (coverage 0.0%, threshold 95.0%)
```

Reproduces every time on a fresh cache. Note the message is byte-identical to the field
report, including the frame count.

## Root cause

`resolveVideoExtractionPolicy()` in
`packages/producer/src/services/render/stages/extractVideosStage.ts:138` defaults
`failureMode` to `"off"`:

```ts
[...]

PR #3434, opened 2026-08-23
fix(engine): treat a sentineled cache entry with no frames as a miss

rajanpanth:
Addresses the cache half of #3372.

### Problem

`lookupCacheEntry` decides a hit purely on the presence of the `.hf-complete` sentinel:

```ts
const complete = existsSync(join(dir, COMPLETE_SENTINEL));
return { entry: { dir, keyHash }, hit: complete };
```

The sentinel records that extraction *finished*, not that the frames survived. `rehydrateCacheEntry` then takes whatever `readdirSync` returns, with no check against `totalFrames`, so an entry whose frames were removed by any per-file cleanup keeps its sentinel and rehydrates as a hit with zero frames.

The clip reaches the coverage gate with no frames, scores `0/expected`, and the render aborts with a message about capture coverage. As the issue notes, this is a plausible explanation for field reports of "happens every time" with no user-discoverable fix: the poison is on disk, not in the composition, so nothing the user changes about the project clears it.

### Fix

An entry counts as a hit only when it carries the sentinel **and** still holds at least one frame file:

```ts
const complete = existsSync(join(dir, COMPLETE_SENTINEL)) && hasFrameFiles(dir);
```

An emptied entry is a miss, so the next render re-extracts and the project recovers by itself.

`hasFrameFiles` matches on `FRAME_FILENAME_PREFIX` only, without the format suffix. A hit has to be usable whatever extension the frames carry, and `lookupCacheEntry` has no format context at that point. An unreadable entry directory also counts as a miss rather than throwing, since re-extracting is always safe.

### Scope

This is deliberately only the second of the three defects in #3372.

I have not touched the primary one, `resolveVideoExtractionPolicy()` defaulting `failureMode` to `"off"`. The issue calls the remedy there a judgement call between flipping the default to `"enforce"` and surfacing the error separately, and that is your decision to make rather than mine to guess at. Nor have I touched the third, clamping media-start against the playable video-stream duration.

Per the issue, the coverage gate is left alone.

Worth noting the two remaining defects are independent of this one: a first render against a video-less tail still fails with the same coverage message. What this change fixes is that the failure no longer becomes permanent for the project.

### Testing

In `extractionCache.test.ts`:

- a sentineled entry whose frame is then removed goes from hit to miss
- a `.png` frame is honoured as well as `.jpg`, covering the format-agnostic count

I also had to amend one existing test. `"hits after ensureCacheEntryDir + markCacheEntryComplete"` created a directory containing only the sentinel and asserted a hit, which is exactly the state this change reclassifies. It now writes a frame file before marking the entry complete, so it still tests that a genuinely complete entry hits. Flagging it explicitly since it is a behaviour change to an existing assertion rather than a new case.

Reverting `extractionCache.ts` while keeping the tests fails the new zero-frame case, so it covers the change rather than restating current behaviour.

`bunx oxlint` and `bunx oxfmt --check` are clean on both files. `extractionCache.test.ts` passes at 33 tests. Across `packages/engine/src/services/` I get 183 passing (up from 181 with the two added) and the same 9 pre-existing failures in `browserManager.test.ts` and `parallelCoordinator-peerAbort.test.ts`, which fail identically on a clean checkout without a workspace build.

### DIFF
--- a/packages/engine/src/services/extractionCache.test.ts
+++ b/packages/engine/src/services/extractionCache.test.ts
@@ -227,11 +227,36 @@ describe("lookupCacheEntry / markCacheEntryComplete", () => {
   it("hits after ensureCacheEntryDir + markCacheEntryComplete", () => {
     const first = lookupCacheEntry(tmpRoot, base(sourceFile));
     ensureCacheEntryDir(first.entry);
+    writeFileSync(join(first.entry.dir, "frame_00001.jpg"), "x", "utf-8");
     markCacheEntryComplete(first.entry);
 
     const second = lookupCacheEntry(tmpRoot, base(sourceFile));
     expect(second.hit).toBe(true);
     expect(second.entry.dir).toBe(first.entry.dir);
+  });
+
+  it("treats a sentineled entry with no frames as a miss", () => {
+    const first = lookupCacheEntry(tmpRoot, base(sourceFile));
+    ensureCacheEntryDir(first.entry);
+    writeFileSync(join(first.entry.dir, "frame_00001.jpg"), "x", "utf-8");
+    markCacheEntryComplete(first.entry);
+    expect(lookupCacheEntry(tmpRoot, base(sourceFile)).hit).toBe(true);
+
+    // Any per-file cleanup can empty the directory while leaving the sentinel.
+    // Serving that as a hit rehydrates zero frames, so every later render of
+    // the project fails identically at the coverage gate.
+    rmSync(join(first.entry.dir, "frame_00001.jpg"));
+
+    expect(lookupCacheEntry(tmpRoot, base(sourceFile)).hit).toBe(false);
+  });
+
+  it("counts frames of any format when deciding a hit", () => {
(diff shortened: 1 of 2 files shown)


miga-heygen (approved):
## Review

**Approve.** Real bug, correct fix.

The permanent-poison path is clear: `lookupCacheEntry` checks only the `.hf-complete` sentinel, not frame presence. External cleanup (OS temp reaper, disk pressure, manual) can remove frames while leaving the sentinel. Every subsequent render hits the zero-frame entry, the coverage gate aborts, and the user has no way to recover without manually finding the cache directory.

The fix requires at least one `frame_`-prefixed file alongside the sentinel. Can't check frame count at this layer (no `totalFrames` context — that's derived at extraction time), and partial frame loss is correctly handled downstream by the coverage gate. The modified existing test was asserting the exact buggy behavior (sentinel-only = hit).

Scope is right — the other two #3372 defects are genuinely independent. Ship it.

— Miga

PR #3267, opened 2026-08-13
fix(catalog): sync tab labels with the sliding indicator

AksharP5:
## What

The Tabs Slide Indicator starts with the pill under "Plan," but marks "Debug" as active. This keeps every visible part of each label aligned with the indicator before, during, and after the slide.

Fixes #3266.

## Why

The component permanently styled the second button white even though the indicator starts under the first button. Its inactive gray also measured 4.4:1 against the track, just below WCAG AA.

Animating the labels between white and gray fixed the endpoints but introduced low-contrast intermediate colors during the slide. At 0.71s, the exposed text measured about 1.82:1 against the light track.

## How

The pill position is now the single source of truth for both movement and label contrast. A hard-stop text gradient keeps the portion over the pill white and the portion over the light track `#6f6f78`, including while the indicator crosses a glyph. No label color is interpolated.

The generated Catalog MDX and public preview payload are refreshed from the corrected component source.

## Test plan

- [ ] Unit tests added/updated
- [x] Manual testing performed
- [x] Documentation updated (generated Catalog MDX and payload)

Validated with:

- `bun run build`
- `bun run lint:registry-items tabs-slide-indicator`
- `bunx oxfmt --check` on all changed files
- `npx hyperframes lint` against the standalone demo
- `npx hyperframes check --samples 15 --at-transitions --at 0.5,0.605,0.71,0.815,0.92` against the standalone demo
- `npx hyperframes keyframes --shot` for the indicator timeline
- Exact-frame pixel inspection at 0.71s

The full build passes. The exact transition samples report zero lint, runtime, or layout findings. At the midpoint, white on the pill measures 17.72:1 and `#6f6f78` on the track measures 4.53:1.

## Before and after

The original recording shows the endpoint correction:

https://github.com/user-attachments/assets/102ee20b-ff3e-4f99-bec8-7b9133e0fbd7

The exact midpoint shows the transient contrast fix:

Before and after at 0.71 seconds

PR #3437, opened 2026-08-23
fix(registry): make the tabs active label follow the sliding indicator

### LINKED ISSUE #3266: Tabs Slide Indicator shows the wrong active label before sliding
## Describe the bug

The Tabs Slide Indicator starts with the pill under "Plan," but its label styling marks "Debug" as active. "Plan" is gray on the dark pill, while "Debug" is white on the light track. The labels only match the indicator after it slides to "Debug."

## Link to reproduction

https://hyperframes.heygen.com/catalog/components/tabs-slide-indicator

Source reproduction: https://github.com/heygen-com/hyperframes/blob/main/registry/components/tabs-slide-indicator/demo.html

## Steps to reproduce

1. Open the Tabs Slide Indicator catalog page.
2. Play or restart the preview.
3. Look at the initial state before the indicator moves.

## Expected behavior

The white active label follows the dark indicator from "Plan" to "Debug," and inactive labels meet WCAG AA contrast.

## Actual behavior

"Debug" remains white before the indicator reaches it, and the inactive gray measures 4.4:1 against the track instead of the required 4.5:1.

## Environment

Hosted HyperFrames Catalog and current `main` at `9ba528914`.

## Additional context

The component permanently applies its active color to the second button even though the indicator starts at the first button. The installable component and catalog demo share the same state mismatch.

rajanpanth:
Fixes #3266

### Problem

The component pinned its active colour to a fixed tab:

```css
.hf-transition-tabs-slide-indicator button:nth-of-type(2) {
  color: #fff;
}
```

The pill starts under the *first* tab, so before the slide "Plan" was gray on the dark pill and "Debug" was white on the light track. The two only agreed once the pill arrived at "Debug".

The inactive gray `#71717a` also measured **4.40:1** against the `#f4f4f5` track, just under the 4.5:1 AA requirement.

### Fix

The pill now carries a clipped copy of the label row:

```html
<span class="indicator"
  ><span class="indicator-labels" aria-hidden="true"
    ><button tabindex="-1">Plan</button>...</span
  ></span
><button>Plan</button>...
```

`.indicator` gets `overflow: hidden`, and the copy counters the pill's own translate:

```css
transform: translateX(calc(var(--hf-tab-x) * var(--hf-tab-step, 1) * -1));
```

That keeps the copy aligned with the buttons underneath at every frame, so the white text is exactly the part of the row the pill covers. It crosses over mid-slide instead of being pinned to one tab, and it is driven by the same `--hf-tab-x` the timeline already animates, so it stays deterministic under seek. The documented timeline recipe is unchanged.

Inactive gray is now `#6b6b73`, **4.80:1**.

### One consequence worth flagging

`.indicator` needed `z-index: 2` to sit above the `z-index: 1` buttons, because its `transform` makes it a stacking context and the label copy cannot otherwise paint above the row. The copy has to live inside `.indicator` since that is the element the timeline sets `--hf-tab-x` on, and inheritance only goes downward.

So the pill's `box-shadow` now paints over the track rather than behind the labels. I think this reads correctly, a raised pill casting onto what is below it, but it is a visual change beyond the reported bug and easy to revert if you would rather not have it. The alternative I considered was animating `left` instead of `transform`, which keeps the old layering exactly but trades a composited property for a layout one.

### Verification

Measured in the browser against the demo, at several points of the slide:

| `--hf-tab-x` | pill position | white label | copy-to-row drift |
| --- | --- | --- | --- |
| `0px` | tab 1 | Plan | 0.00px |
| `43px` | mid-slide | Plan + Debug, split at the pill edge | 0.00px |
| `86px` | tab 2 | Debug | 0.00px |

`elementFromPoint` at each label centre confirms the pill's subtree is topmost exactly where the pill sits and the gray row elsewhere.

Contrast computed with the WCAG relative-luminance formula: `#71717a` 4.397:1, `#6b6b73` 4.805:1, white on the `#18181b` pill 17.72:1.

The `labels` variable rebuilds both rows, checked with four labels: 4 row buttons, 4 copies, 0.00px drift, copies at `tabIndex -1`. The copy is `aria-hidden` and `pointer-events: none` so it adds nothing to the accessibility tree or the tab order.

`node scripts/lint-registry-items.mjs` passes at 372 items, 0 errors. Both `demo.html` and the installable snippet carry the same change, including the demo's duplicated head style block.

### DIFF
--- a/registry/components/tabs-slide-indicator/demo.html
+++ b/registry/components/tabs-slide-indicator/demo.html
@@ -54,6 +54,8 @@
         background: #18181b;
         transform: translateX(var(--hf-tab-x));
         box-shadow: 0 12px 26px rgba(0, 0, 0, 0.18);
+        overflow: hidden;
+        z-index: 2;
       }
       .hf-transition-tabs-slide-indicator button {
         position: relative;
@@ -61,10 +63,19 @@
         height: 36px;
         border: 0;
         background: transparent;
-        color: #71717a;
+        color: #6b6b73;
         font-weight: 600;
       }
-      .hf-transition-tabs-slide-indicator button:nth-of-type(2) {
+      .hf-transition-tabs-slide-indicator .indicator-labels {
+        position: absolute;
+        top: 0;
+        left: 0;
+        display: grid;
+        grid-template-columns: repeat(3, 86px);
+        transform: translateX(calc(var(--hf-tab-x) * -1));
+        pointer-events: none;
+      }
+      .hf-transition-tabs-slide-indicator .indicator-labels button {
         color: #fff;
       }
     </style>
@@ -88,7 +99,12 @@
     { "id": "labels", "type": "string", "role": "content", "label": "Labels", "description": "The tab names, separated by commas. Two or more names rebuild the row.", "default": "Plan,Debug,Ask" }
   ]'
         >
-          <span class="indicator"></span><button>Plan</button><button>Debug</button
+          <span class="indicator"
+            ><span class="indicator-labels" aria-hidden="true"
+              ><button tabindex="-1">Plan</button><button tabindex="-1">Debug</button
+              ><button tabindex="-1">Ask</button></span
+            ></span
+          ><button>Plan</button><button>Debug</button
           ><button>Ask</button>
         </div>
       </div>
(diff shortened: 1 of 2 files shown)


rajanpanth:
Closing: main picked up #3267, which syncs the tab labels with the sliding indicator using an overlay approach, so this gradient-clip variant is redundant now. Happy the behavior is fixed either way.

PR #3448, opened 2026-08-23
fix(studio): unify audio IDs and group state

vanceingalls:
Part 5 of 12 replacing #3439. Base: `wa-26d-lint-audio-groups`.

## Why

Studio previously mixed DOM ids, render ids, file-qualified ids, and sub-composition child state. This slice establishes one translation boundary before edit and UI behavior lands.

## What

- centralize runtime audio-id and timeline group lookup
- mirror group state onto both flat elements and expanded DOM children
- make selection/reveal requests survive the intended id-space crossing
- extract narrow store helpers and the timeline element patch contract
- introduce reveal state while retaining the old expanded-group representation until PR 8 changes the UI contract

## Verification

- `playerStore.test.ts` (62 tests)
- oxfmt and oxlint on the compatibility bridge

Stack: #3447 ← **#3448** → #3449

### DIFF
--- a/packages/core/src/audio/audioFxWorklets.test.ts
+++ b/packages/core/src/audio/audioFxWorklets.test.ts
@@ -136,23 +136,125 @@ describe("the worklet processors themselves", () => {
       return crossings / ((s.length - start) / SR);
     }
 
-    it("at semitones: 0, mix: 1 reproduces the input, delayed by exactly one grain/2", async () => {
+    // This assertion used to be that the output equalled the input DELAYED by
+    // grain/2 — the measurement was right and was written down as the contract.
+    // But the grain delay is there to shift pitch, and at semitones: 0 nothing
+    // is being shifted: the node degenerated into a pure 50 ms delay of the
+    // signal, plus a head of silence while the ring filled, under a label that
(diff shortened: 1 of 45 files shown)


miga-heygen (dismissed):
## Review

The group info cache redesign is the highlight: MutationObserver + revision counter means a forgotten `invalidateGroupInfoCache` call costs a re-scan rather than a wrong answer — belt to the explicit invalidator's braces. The attribute filter is narrow (only group-relevant attrs), and the `childList` filter correctly skips runtime-churned nodes (SplitText, typewriters) by checking for actual group/member elements.

`runtimeAudioId` bridges the composite-key / bare-DOM-id space boundary cleanly. `isTimelineIgnoredElement` prevents phantom group bus rows. `canHideSelections` closes the last path to accidentally muting audio via "Hide all". `TimelineElementPatch` as an explicit allow-list is the right constraint. `audioGroup*` on `DomClipChild` solves the sub-comp membership gap.

Tests cover every new path: boundary crossing, cache invalidation (explicit + observer), group exclusion, hide refusal, automation sync for groups.

No issues. Ship it.

— Miga

PR #3449, opened 2026-08-23
fix(studio): make audio-group edits transactional

vanceingalls:
Part 6 of 12 replacing #3439. Base: `wa-26e-studio-audio-state`.

## Why

Creating a bus and editing its attributes is one transaction across the live DOM, source files, optimistic state, undo, and sub-composition ownership.

## What

- create groups in the owning composition and assign members in their own files
- resolve patch targets before optimistic writes
- mirror group volume, label, FX, automation, and hidden state after edits
- keep visibility edits, group deletion, and carve auto-grouping consistent
- prevent invalid or blocked timeline edits from partially applying

## Verification

Stack: #3448 ← **#3449** → #3450

### DIFF
--- a/packages/core/src/audio/audioFxWorklets.test.ts
+++ b/packages/core/src/audio/audioFxWorklets.test.ts
@@ -136,23 +136,125 @@ describe("the worklet processors themselves", () => {
       return crossings / ((s.length - start) / SR);
     }
 
-    it("at semitones: 0, mix: 1 reproduces the input, delayed by exactly one grain/2", async () => {
+    // This assertion used to be that the output equalled the input DELAYED by
+    // grain/2 — the measurement was right and was written down as the contract.
+    // But the grain delay is there to shift pitch, and at semitones: 0 nothing
+    // is being shifted: the node degenerated into a pure 50 ms delay of the
+    // signal, plus a head of silence while the ring filled, under a label that
+    // reads "Unchanged pitch".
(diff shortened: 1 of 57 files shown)


miga-heygen (dismissed):
## Review

Three real bug fixes in one well-scoped PR:

1. **Transactional group creation**: captures prior group state, restores on save failure. Emits `<hf-audio-group>` element (not just membership attrs) so later writes have a DOM target. Group ID validation + `escapeAttr` close injection.

2. **Store mirroring**: `syncStoredGroupAttribute` writes to BOTH `elements` and `domClipChildren` in one store notification — fixes the "muted group whose button stayed Mute" symptom for flat AND sub-comp members. `mirroredGroupVolume` correctly reads null/empty as unity (matching core), closing the `Number(null)===0` parse divergence.

3. **ID space boundary**: `useAudioGroupCarveAssignment` now resolves DOM ids via `runtimeAudioId`, not store keys — fixes the silent match-nothing → carve-pointing-at-nothing bug. `resolveGroupSourceFile` climbs past the sub-comp root to the host that names the file — fixes "Unable to patch element in index.html" for sub-comp groups.

`persistElementAttribute` now resolves the target BEFORE the optimistic patch and reads the unwind value from the FILE, not the already-patched live DOM — closing two failure classes at once.

No issues. Ship it.

— Miga

PR #3450, opened 2026-08-23
fix(studio): keep preview state synchronized

vanceingalls:
Part 7 of 12 replacing #3439. Base: `wa-26f-studio-group-editing`.

## Why

Preview hydration and DOM edit sessions must agree on project/session identity, expanded children, and message ordering before the timeline UI consumes them.

## What

- decompose timeline sync hydration and callback routing
- preserve audio-group state for expanded sub-compositions
- keep DOM edit sessions stable across HMR and provider remounts
- route visibility, attribute, delete-member, and playback updates through one sync path
- remove stale group-level message handling while retaining its compatibility module until PR 11

## Verification

- all relevant TypeScript projects pass cumulatively
- `fallow audit --base main --fail-on-issues`

The final stack tip preserves the verified #3439 replacement and includes the review fixes landed across the stack.

Stack: #3449 ← **#3450** → #3451

PR #3451, opened 2026-08-23
fix(studio): align audio rows, automation lanes and headers

vanceingalls:
Part 8 of 12 replacing #3439. Base: `wa-26g-studio-preview-sync`.

## Why

Group rows, track headers, automation lanes, and their store representation share prop and row-layout contracts. Splitting this slice produced real TypeScript failures, so it is the one intentionally larger PR in the stack.

## What

- render groups expanded by default with indented member rows
- unify one-line headers, right-anchored controls, lane disclosure, and keyboard focus
- show group and carve automation lanes with honest read-only affordances
- reach the shared FX preset shelf from track and group headers
- audition at a span where the selected audio is actually sounding
- retain `STRIP_H` and solo state as compatibility bridges until their explicit removal in PR 11

## Verification

- group bus strip (7), group row (3), and track header (31) focused tests
- oxfmt and oxlint on compatibility files

Stack: #3450 ← **#3451** → #3452


miga-heygen (commented):
Reviewed the full diff (43 files, +2740/-886). This is a dense but well-organized slice — synthetic "group as element" pattern for automation lanes, one-line header layout unification, and the audition-transport extraction are all clean, well-tested seams. Comments below are grouped by severity; nothing here blocks, but a few are worth a follow-up commit.

## Correctness

- **`useAuditionTransport` / `TimelineFxButton` mute-borrow** (`TimelineFxButton.tsx`): the borrow/return sequencing (`borrowedMute.current = props.isMuted === true` on entry, read-before-reset on exit) is correct and matches the new tests. Good catch on reading `isMuted` on the way *in* rather than re-deriving it on the way out (the stale-prop trap called out in the comment is real).
- **`auditionStart`** (`useAuditionTransport.ts`): wrap-to-first / stay-inside-span / jump-to-next logic checks out against the test cases, including the "past everything" wrap case.
- **`collapsedGroupIds` inversion** (`keyframeSlice.ts`, `useTimelineTrackDerivations.ts`): inverting the set so "expanded" is the unmarked default is the right fix for the "grouping vanishes the tracks" bug, and `createTimelineResetState()` was updated so it resets per-composition — easy to have missed.
- **`isCarveLane`**: prefix-matches `fx.${node.id}.` (with trailing dot), so `n1` vs `n10` collisions are correctly avoided.
- No obvious off-by-one in the index math for `TimelineGroupLaneLabels`/`AutomationLaneHeaderRow` — the fix (using the raw `.map()` index rather than a manually incremented counter) keeps a null-returning entry from shifting everything below it, which was the actual prior bug.

## Nits worth fixing (style/lint, not correctness)

- `TimelineGroupRow.tsx`: a `const GROUP_LANE_ACCENT = "#3CE6AC"` declaration is sandwiched between two import statements:
  ```
  import { TRACK_H } from "./timelineLayout";
  ...
  const GROUP_LANE_ACCENT = "#3CE6AC";
  import { LABEL_COL_W } from "./timelineLayout";
  import { useTimelineEditContextOptional } from "../../contexts/TimelineEditContext";
  ```
  Works at runtime (imports hoist), but will likely trip `import/first`-style lint rules and is just confusing to scan. Move the const below all imports.
- Same file: `TRACK_H` and `LABEL_COL_W` end up imported from `"./timelineLayout"` in two separate `import` statements instead of one — worth merging (`import/no-duplicates`).
- `TimelineTrackHeader.tsx`: `groupableClipIds.every((id) => id !== null)` followed by `onGroupClips(groupableClipIds as string[], ...)` — since `.every` already proves no `null`s remain, prefer a type predicate (`(id): id is string => id !== null`) over the `as string[]` cast so the compiler enforces it rather than trusting the cast.

## Performance (minor, worth a follow-up)

- `TimelineTrackHeader.tsx` now computes `groupAutomatedTargets` via a fresh `groupAutomationLanes([...])` call (which parses `automation`/`fxChain` JSON) plus four `trackElements.find(...)` scans, all inline in the component body with no memoization. This header already re-renders every playhead tick during playback (it takes `currentTime` for the keyframe-value readouts), so this adds JSON-parsing + array-scan work per audio track per frame. Given it only depends on `trackElements`/`activeKey`/`groupAutomationRaw`/`groupFxChainRaw` — none of which change every frame — it's a good `useMemo` candidate. Not urgent at current track counts, but will show up in profiling once compositions get audio-heavy.

## Architecture / SSOT

- The `groupAutomationElement` "group wearing a clip's shape" pattern is a good call — it avoids a second parallel lane-rendering path for groups, and the comments explain the `start: 0` / composition-duration choice clearly (ties to the groups doc's "composition time" clock decision).
- `LaneToggleButton` extraction out of `LayerDisclosureRow` and reuse in `TimelineGroupHeader`/`TimelineTrackHeader` is a solid dedup — one control, one set of a11y semantics, instead of three near-identical buttons.
- `trackHeaderLabelRows.tsx` split (to stay under the 600-line file cap) moved code verbatim without duplicating logic — clean extraction, nothing left stale in the original file.
- Mute/solo removal from `TimelineGroupHeader` and `TimelineTrackPlainHeader` is a deliberate, well-flagged step in the stack (store-side `soloed`/`toggleSolo` and `STRIP_H` explicitly kept as "compatibility bridges" until PR 11 per the description) — consistent, not an accidental half-removal.
- One thing I couldn't verify from the diff alone: `PlayerControls.tsx`'s `previewIframeRef` prop and `SoloBanner` were removed, but no caller of `<PlayerControls>` is in this PR's file list. Presumably CI's cumulative TypeScript check (mentioned in the PR description) already caught any caller still passing that prop — just flagging so it's confirmed rather than assumed, since a stale caller elsewhere wouldn't show up in this diff.

## Testing

Coverage is thorough and the tests read as regression tests for real bugs (the group-row-throws-without-provider case, the dialog-positioning-off-screen case, the collapsed-group-dead-space case, the read-only-lane-hides-handles case). Nothing to add here.

## Type safety

No unsafe casts of note beyond the one `as string[]` called out above. The discriminated union on `TimelineFxButtonProps` (`variant: "chain" | "group-pointer"`) is used correctly to narrow `props` before accessing variant-specific fields like `isMuted`/`auditionSpans`.

— Miga

miguel-heygen (dismissed):
Stamp review at exact head `f6a0f9d6`.

**Audited:** group/lane expansion state, synthetic group automation element path, audition transport borrowing, whole-track grouping guard, PlayerControls caller after prop removal, and current CI.  
**Trusting:** Miga's full 43-file UI/layout pass and the large body of focused regression tests; I spot-checked the seams they identified rather than duplicating their complete review.

The correctness-sensitive paths hold: collapsed-group inversion keeps expanded as the default, group lanes reuse the existing element-shaped automation pipeline, carve matching keeps the trailing-dot collision guard, and whole-track grouping withholds the action when any member lacks a resolvable DOM id. The sole `<PlayerControls>` caller no longer supplies the removed iframe prop.

All reported checks at this stacked head are terminal with no failures; player-perf, preview parity, preview-regression, and regression aggregate are green.

Non-blocking cleanup remains exactly as Miga noted: move `GROUP_LANE_ACCENT` below imports/merge the duplicate timelineLayout imports, replace the `as string[]` with a type predicate, and consider memoizing group automation parsing on playback-heavy compositions.

**Verdict:** APPROVE
**Reasoning:** The current head is behaviorally coherent at the timeline/group/lane boundaries, its caller contract is consistent, and the available stacked CI is green; remaining items are maintainability/performance follow-ups.

— Magi

vanceingalls:
Addressed the timeline review feedback:

- imports are normalized
- whole-track group IDs narrow through a type predicate without an assertion
- group metadata, automated targets, and automation rows are memoized across playhead-driven renders
- simplified the memoization after the static complexity gate flagged the first implementation

The two focused timeline suites pass (34 tests), Studio typechecks, and fallow reports zero gated complexity/dead-code findings.

vanceingalls:
Follow-up CI fix pushed: timelineLayout.ts is now exactly 600 lines, satisfying the cumulative Studio source-size gate without changing behavior. The local replica of the check passes.

vanceingalls:
CI follow-up: the TimelineFxButton regression test landed in this PR before both implementation pieces it exercises. This layer now owns useFxAudition().storedChain() and the popover applies presets to that stored base, so an auditioned preset cannot leak into the saved chain. Focused test: 12/12 passing. The resulting tip tree is unchanged from the full Studio run (4,464 passing); cascaded #3452–#3455.

miguel-heygen (approved):
Changed-head re-review at c187af00280f954c83704669e7c331e7a46f257c.\n\nThe original reviewed PR commit is patch-identical after restack: range-diff reports f6a0f9d6 = e2a80314. I audited the four follow-up commits and their complete five-file delta.\n\nThe follow-ups close the prior maintenance/performance notes without changing the timeline contract: TimelineGroupRow.tsx now has module-scoped, deduplicated imports; TimelineTrackHeader.tsx:126-195 memoizes group metadata and parsed lane rows; and the grouping guard now narrows the full id list with a real type predicate before calling the write boundary. The layout-only STRIP_H change is behavior-neutral.\n\nThe new persistence fix is sound. useFxAudition.ts:35-64 keeps the document chain in auditionBase, and TimelineFxPopover.tsx:110-115 derives the clicked preset from that stored base before clearing audition state and persisting. Hovering preset A then clicking preset B can no longer save A+B.\n\nImportant, non-blocking coverage gap: c187af00 adds that concrete data-corruption regression without adding its exact witness. TimelineFxPopover.test.tsx:67-90 tests apply and audition separately, but never hover/focus A, click B, then assert the persisted chain is original+B and excludes A. That mutation-sensitive case should be added; the implementation itself is correct.\n\nCurrent CI is fully green: all required contexts, nine regression shards, preview parity/regression, player perf, lint/format/fallow, Studio viewport/load, and both Windows lanes pass.\n\n**Verdict:** APPROVE\n**Reasoning:** The previously approved patch is unchanged, every follow-up delta is coherent, and the new audition/persist repair correctly preserves the stored-chain boundary; only a non-blocking exact regression test remains.\n\n— Magi

PR #3452, opened 2026-08-23
fix(studio): harden carve and FX rack behavior

vanceingalls:
Part 9 of 12 replacing #3439. Base: `wa-26h-studio-timeline-ui`.

## Why

The rack should reveal the requested parameter safely, audition without persisting preview state, and never make a bed measure or carve itself.

## What

- prevent reciprocal, self-referential, and double-fired carve analysis
- keep a bus out of carve-bed roles and expand group sources at analysis time
- escape reveal selectors, route lane-label requests to the correct rack row, and honor repeated requests for an already-open row
- separate signal-path and summary derivation from rack rendering
- share audition and levelling behavior across rack entry points

## Review note

GitHub renders `useFxCarve.ts` as binary because this PR replaces three literal NUL bytes inherited from its base with the source escape `"\0"`. The resulting TypeScript is ordinary text; review that file from the head version or with a local diff if GitHub does not display it.

## Verification

Stack: #3451 ← **#3452** → #3453


miga-heygen (commented):
Reviewed the full diff (including the `useFxCarve.ts` change directly against the API blobs, since `gh pr diff` renders it as binary — see below). Overall this is a well-reasoned hardening pass; the carve self-reference/reciprocal/double-fire fixes are correct and well tested. Two things worth fixing before merge, plus some smaller notes.

## Correctness issues

**1. `useFxCarve.ts` diffs as binary on GitHub — and it's hiding the PR's most important logic.**
`gh pr diff` / GitHub's UI render this file as `Binary files a/... and b/... differ`, so nobody reviewing on GitHub actually sees the line-level diff for the file that contains the new `autoCarveInFlight` guard, `CARVE_ABORTED` handling, and the bed-self-exclusion fix. Root cause: the **base** blob has a literal raw NUL byte embedded in a string literal (`before.sources.join("<0x00>")` — an actual control byte, not the `\0` escape sequence), at 3 spots. This PR replaces those with the proper `"\0"` escape sequence, which is a reasonable choice (safer join delimiter than `" "` for de-duping id arrays), but because git's binary-detection looks at *either* side of the diff, the presence of the raw byte in the old blob makes the whole pair "binary" regardless of the fix.
  - Good: the fix itself is correct and probably was needed (a plain-space join can collide across different id sets).
  - Bad: this means the diff was effectively unreviewable through normal GitHub tooling for this PR (and likely for whatever PR introduced the raw byte originally). Worth calling out explicitly in the PR description so reviewers know to pull the branch locally, and worth a quick check that no toolchain (tsc/esbuild/swc) choked on the raw byte before this fix landed.
  - I manually diffed the base/head blobs out-of-band to review this file; see notes below on the logic itself, which looks correct (self-carve exclusion, `autoCarveInFlight` ref latch, `CARVE_ABORTED` short-circuit all check out against the stated bugs and the new tests in `useFxCarveGrouping.test.ts`).

**2. Reveal-and-scroll can silently no-op (or fire stale) when the target surface is already open.**
In `propertyPanelFxSection.tsx`:
```js
if (where.kind === "node") setOpenNode(where.index);
if (where.kind === "eq") setOpenEq(where.eqId);
if (where.kind === "carve") setCarveOpen(true);
...
pendingRevealRef.current = revealTarget ?? null;
```
```js
useEffect(() => {
  const target = pendingRevealRef.current;
  if (target && scrollRevealedRowIntoView(rootRef.current, target, chain)) {
    pendingRevealRef.current = null;
  }
}, [openNode, openEq, carveOpen, collapsedRuns]);
```
If the reveal target's surface is **already** open/expanded (e.g. clicking "reveal" twice on the same node, or on a param whose row is already visible), `setOpenNode`/`setOpenEq`/`setCarveOpen` are called with the *same* value they already hold. React bails on re-running effects whose deps didn't change (via `Object.is`), so this scroll effect simply never fires — the click looks dead. Worse, `pendingRevealRef.current` stays set with the stale target and will only get consumed whenever one of those four values *happens* to change for an unrelated reason later (e.g. the author opens a different node), at which point the panel scrolls to the **earlier, no-longer-relevant** target instead of the one just clicked.

Since `revealNonce` changes on every reveal request even when the target surface state doesn't, the fix is to key the scroll effect on the nonce (or `consumedRevealNonce`) rather than solely on the derived open-state:
```js
}, [consumedRevealNonce, openNode, openEq, carveOpen, collapsedRuns]);
```
By the time this effect runs, the row will already be mounted (either newly, or because it was already open), so this is safe.

Also worth noting: `audioFxRevealTarget.test.ts` thoroughly covers the pure resolver, but there's no test exercising this render+effect wiring in `FxSection`, so this class of regression wouldn't be caught by CI.

## Minor / nits

- `useFxCarve.ts` top of file has an orphaned doc comment left over from the extraction:
  ```
  /**
   * Rate the carve source is decoded at. Analysis is self-consistent because it
   * reads the decoded buffer's own rate, so this only has to be a sane audio rate.
   */
  ```
  `DECODE_SAMPLE_RATE` moved to `useFxCarveNodes.ts` (which has its own, still-accurate copy of this comment), but this now-dangling comment sits above `carveAction` in the old file and describes nothing there. Should be deleted.
- Setting `pendingRevealRef.current` during the render-phase block (rather than in an effect) is a bit unusual — legitimate under React's "adjust state during render" pattern for the `useState` calls, but mutating a ref as a side effect during render is technically outside the pure-render contract. Low risk here since it's idempotent for a given `revealNonce`, but flagging in case a future Strict Mode / compiler pass gets stricter about this.

## What looks solid

- `useFxCarveGrouping.ts`: `excludedFor`, `carveBedRoles`, and the group-expansion fix in `carvesAgainst`/`carverAgainst` all directly address the bugs described in the comments, and the new tests (`useFxCarveGrouping.test.ts`) cover exactly those regressions (member offered its own group, group offered itself, bus misclassified as a bed, reciprocal carve through group expansion).
- `useFxAudition.ts`'s `storedChain()` fix for the "hover preset A, click preset B" double-persist bug is correct and matches the new `applyPreset` call site (`clearAudition()` before `mutate()`, so no stale revert races the just-landed write).
- `audioFxSignalPath.ts` / `audioFxSummary.ts` (group IN/OUT labels, "in Voiceover" summary) are clean, well-tested, pure functions — nothing to flag.
- `useFxLevelling.ts` swapping its inline audition-transport logic for the shared `useAuditionTransport()` hook is a good SSOT cleanup, not a regression.
- The `autoCarveInFlight` ref latch correctly prevents the described cross-await double-fire (two effects racing across the `resolveNextCarveSettings` await when auto-grouping mutates the DOM mid-flight), and deliberately leaves manual `setCarve` calls unlatched per its own comment — reasonable tradeoff, documented.

Nice test coverage overall on the pure logic modules; the two issues above are both in the stateful/effect-driven glue code, which is exactly where this style of thorough unit testing doesn't reach.

vanceingalls:
Addressed the FX-rack review feedback:

- repeated same-target reveal requests now scroll even when the owning surface is already open
- pending reveal state is request-keyed instead of mutating a ref during render
- added a two-request regression test
- removed the orphaned decode-rate comment
- documented the inherited literal-NUL/binary-diff issue in the PR body

The focused rack suite passes (96 tests), Studio typechecks, and the production build is green.

vanceingalls:
Follow-up CI fix inherited from the preceding rung: timelineLayout.ts is now exactly 600 lines, so this PR's cumulative source-size check passes locally.

vanceingalls:
CI follow-up (corrected after the adjacent PR completed): the TimelineFxButton regression test begins in #3451, so the complete production fix now lands there, one layer earlier than this PR. This PR and #3453–#3455 were rebased onto that fix. Focused test: 12/12 passing; final tip tree matches the full Studio run with 4,464 passing tests.

miga-heygen (commented):
Re-reviewed after the two follow-up commits (`e3b9db4` reveal fix, `a816541` hardening). Both flagged items are resolved or reasonably justified:

**1. `useFxCarve.ts` binary diff — justified, not newly broken.**
Confirmed by pulling both blobs directly from the API: the base blob (from `wa-26h-studio-timeline-ui`) still contains 3 raw NUL bytes (`file` reports it as `data`), while the head blob in this PR is clean UTF-8 text with the NULs replaced by the `"\0"` escape. Git's binary heuristic is comparing against the base side, which is inherited/pre-existing and outside this PR's diff — so the opaque rendering can't be fixed from within #3452 alone. The PR body now documents this explicitly with a workaround (review via API blobs / local diff), which is the right call given the constraint. Content-wise the change itself is correct.

**2. Reveal-and-scroll bug — fixed, and well tested.**
The effect no longer depends solely on `[openNode, openEq, carveOpen, collapsedRuns]` — it now also keys on `pendingReveal`, which is a freshly-identitied object set on every accepted reveal request (guarded by `revealNonce !== consumedRevealNonce`). That guarantees the effect re-runs even when the target surface is already open and the `setOpenX` calls are no-ops. Added regression test `"scrolls again when the same open parameter is revealed twice"` directly exercises this: renders with `revealNonce=1`, asserts one scroll, re-renders with `revealNonce=2` on the same target, asserts a second scroll. This is exactly the repro from the original review — good coverage.

Nice work tightening this up. No blocking issues from my side.

— Miga

miguel-heygen (dismissed):
Changed-head stamp at `e3b9db4e`.

**Audited:** reveal request lifecycle/effect dependencies and regression, the carve-source blob repair/documentation, and current-head workflow attempts.  
**Trusting:** Miga's full FX-rack/carve/audition review and out-of-band base/head blob comparison for `useFxCarve.ts`.

The reveal defect is closed: each accepted nonce creates a fresh `pendingReveal` identity, the scroll effect depends on that request, and the test proves two reveals of an already-open parameter scroll twice. The raw-NUL binary classification comes from the inherited base blob; the head replaces those bytes with valid escaped delimiters and documents how to review the otherwise opaque diff.

The first current-head preview/Windows runs were cancelled when a same-head rerun superseded them, which made the old aggregate appear failed. Every latest workflow attempt for this exact SHA is terminal success, including CI, preview parity/regression, Windows, player perf, CodeQL, and regression.

**Verdict:** APPROVE
**Reasoning:** The stateful reveal path now has a request-level witness and regression, the binary-diff limitation is inherited and independently audited, and the latest exact-head checks are green.

— Magi

miguel-heygen (approved):
Changed-head re-review at dd9342e523f30e2b2a507ea26468c9e208cdb852.\n\nThe two #3452 commits are patch-identical to the previously approved pair: range-diff reports a816541b = 66bb83c8 and e3b9db4e = dd9342e5. The reveal request lifecycle, repeated-reveal regression, carve self/reciprocal guards, raw-NUL source repair, and FX-rack extraction therefore transfer exactly.\n\nThe four tree differences between the old and current heads are inherited from the lower-stack solo-bridge repair in core runtime files, not changes in this PR. I independently reviewed and approved that exact lower-rung patch on #3445 at aa785328; it keeps every intermediate rung functional and is removed later by #3454.\n\nCurrent CI is fully green, including all required contexts, nine regression shards, preview parity/regression, player perf, lint/format/fallow, Studio viewport/load, and both Windows lanes.\n\nThe non-blocking exact-test gap for the inherited hover-A then click-B persistence fix remains recorded on #3451, where that change belongs; it does not alter this PRs prior verdict.\n\n**Verdict:** APPROVE\n**Reasoning:** #3452s own reviewed patches are unchanged after restack, the only inherited delta was independently audited at its owning rung, and the complete current-head matrix is green.\n\n— Magi

PR #3454, opened 2026-08-23
feat(studio,core)!: remove solo and the group meter

vanceingalls:
Part 11 of 12 replacing #3439. Base: `wa-26j-studio-property-panel`.

## Why

Mute/solo and the group level strip were deliberately removed from the final design. Delaying deletion until every consumer was migrated keeps every earlier PR compiling and makes the breaking removal reviewable in one place.

## What

- remove track/group solo state, controls, bridge hook, and predicates
- remove the group level meter, bus strip, and level store
- remove the temporary `STRIP_H` and solo compatibility bridges
- update the main-deletion guard for these intentional removals

## Verification

Stack: #3453 ← **#3454** → #3455


jrusso1020 (commented):
Reviewed at `87a7b5a0`. **No blockers.** A deletion is the one shape where reading the diff proves the least, so I ran the removal audit against the whole tree at this head rather than against the diff: the tarball at `87a7b5a0`, every removed export grepped repo-wide.

## The audit

Thirteen exported symbols leave in this PR — `AudioSoloSlice`, `GroupLevelReading`, `STRIP_H`, `TimelineGroupBusStrip`, `TimelineSoloButton`, `createAudioSoloSlice`, `groupLevels`, `isAudibleUnderSolo`, `isGroupHalfLitUnderSolo`, `parseGroupLevelsMessage`, `useAudioSoloBridge`, `useGroupLevel`, `useSoloBannerText`. At this head, the only hits outside the deleted files are the allowlist entries themselves and one comment (below). No module is still imported by path. The three imports left in `audioGroups.ts` are all still used two or three times each, and `timelineLayout.ts` still uses `AUTOMATION_LANE_H` — so the characteristic leftovers of an incomplete deletion, an import whose last user just left and a constant whose only caller just left, are both absent. `ALLOWED_DELETIONS` is keyed on exact paths and consumed with `Map.has(path)`, so the allowlist permits exactly these eight files and nothing under them.

Residual `solo` identifiers in `packages/studio/src` and `packages/core/src` are all unrelated — element ids named `"solo"` in z-order and GSAP-patch tests, plus one comment in `TimelineTrackPlainHeader.test.tsx` that correctly describes the removal as history.

## 1. One stale comment, and it is load-bearing for a fixture

`packages/studio/src/player/components/TimelineTrackHeader.test.tsx:772-774`:

```
// The set is pushed straight into the runtime, which compares it against
// `el.id`. A store key here matches nothing, `isAudibleUnderSolo` returns
// false for every element, and soloing silences the whole preview.
```

That is the only surviving mention of `isAudibleUnderSolo` anywhere in the tree, and it is a comment rather than code — but it is the paragraph explaining *why* the fixture uses `domId` rather than a store key. With solo gone, the next person to touch that fixture has a justification pointing at a function that no longer exists, and no way to tell whether the shape still matters. Worth deleting or rewriting in the same PR that removes what it describes.

## 2. The guard this PR edits has not run at this head

`scripts/check-no-main-deletions.mjs` is invoked from exactly one place — `.github/workflows/ci.yml:50`, the `Reject accidental file deletions` step of the `Detect changes` job. `ci.yml` filters `branches: [main]`, and the four workflow runs at `87a7b5a0` are `regression` (x2), `preview-regression` and `Player perf`. No `ci.yml` run.

The trap is that all three of those workflows define their own `Detect changes` job, so the rollup shows **`Detect changes ✅`** — from workflows that do not carry the deletion step. A reviewer reading the check names has every reason to think the allowlist was exercised. It has not been, so the eight new entries are unverified until the retarget to `main`. Not a defect in the change; just the one claim I would not make on this head's CI.

## 3. The allowlist reasons cite a closed PR

All eight reasons read `#3439 deliberately removes …`. #3439 was closed at 21:27Z today and this stack replaces it, so the justification that lands permanently in `main` points at a PR nobody will find the discussion in. Cheap to make it cite this PR or the `wa-26` stack instead — these strings are the whole audit trail for a deletion, and they outlive the branch by design.

## Carry-forward from #3444, verified at *this* head

Worth flagging here rather than only on the lower rung, because this PR changes what it costs. At `87a7b5a0`, `packages/core/src/audioGroups.ts:117` still reads:

```ts
const groupId = el?.getAttribute?.(HF_AUDIO_GROUP_ATTR);
```

while `resolveAudioGroups` eight lines below prefers `AUDIO_GROUP_RENDER_ID_ATTR` and falls back to the author id. So the instance-blind resolve I raised on #3444 is not fixed anywhere later in the stack. Once solo is removed here, `isMemberGroupHidden` is the *only* group-level audibility gate left in the runtime — `runtime/media.ts:342` and `runtime/init.ts:195` both compute audibility as `el.closest("[data-hidden]") !== null || isMemberGroupHidden(...)`. A one-line fix on #3444 is still the right home for it; it just stops being a corner case after this rung.

## For the #3453 hold

`packages/studio/src/hooks/useAudioSoloBridge.ts` is still present at #3453's head (`1e187cd1`, confirmed 200 from the contents API) and is deleted here with an explicit allowlist reason. So the "dead code left behind" half of the hold on #3453 is a deliberate deferral to this rung rather than an oversight — worth knowing before anyone asks for a fix-up commit on #3453 that this PR would then have to un-delete.

## CI

Of `main`'s eight required contexts, only `regression` is present at `87a7b5a0`, and it is green along with all nine regression shards and the five perf jobs. `Build`, `Test`, `Typecheck`, `Test: runtime contract`, both windows jobs and `Semantic PR title` have never run, for the same `branches: [main]` reason as the rest of the stack — self-repairing on retarget, since `edited` is in `ci.yml`'s `types:`. For a PR that deletes 598 lines across the studio player and store, `Typecheck` is the gate carrying most of the signal, and it is one of the seven.

vanceingalls:
Addressed the review follow-ups:

- removed the stale isAudibleUnderSolo fixture comment
- deletion reasons now cite #3453 for the bridge moved down-stack and #3454 for the remaining solo/meter removal
- reran the deletion-guard suite (8 tests) and the guard against the cumulative tip

The cumulative file-size and deletion gates now pass locally as well.

jrusso1020 (commented):
Re-review at `7d99717033`. My earlier review (`5003558797`) was pinned to `87a7b5a019`, which the restack replaced; the branch is now `0be1bb5e` + `7d997170` *"keep removal rationale current"*.

Both things I raised are closed, and the deletion audit re-runs clean at the new head.

## Closed: the citations now name this PR

Seven `ALLOWED_DELETIONS` reasons moved from `#3439` — closed 2026-08-23 — to `#3454`. In a deletion allowlist the citation *is* the audit trail and it outlives the branch, so this was the half worth fixing.

## Closed: the stale comment is gone

`TimelineTrackHeader.test.tsx:772-774` no longer explains a fixture by reference to `isAudibleUnderSolo`, a function this PR removes. The surrounding `aria-level="2"` rationale is kept, which is the part that still describes live behaviour.

## The solo-bridge deletion moved to #3453, and there is no duplicate entry

`useAudioSoloBridge.ts` is out of this PR's file list; its allowlist entry now sits in the base with the reason *"#3453 removes the obsolete solo bridge after its last consumer leaves"*. Deletion and permission are owned by one rung, which resolves the split I flagged.

## The audit, re-run rather than carried

Seven files removed here, eleven exported symbols leaving with them — `AudioSoloSlice`, `GroupLevelReading`, `STRIP_H`, `TimelineGroupBusStrip`, `TimelineSoloButton`, `createAudioSoloSlice`, `groupLevels`, `isAudibleUnderSolo`, `isGroupHalfLitUnderSolo`, `parseGroupLevelsMessage`, `useGroupLevel`. Grepped repo-wide at this head: **every hit is an `ALLOWED_DELETIONS` entry naming the deleted path.** Zero live references, zero orphaned imports.

And I ran the guard rather than reporting that it had not run. `check-no-main-deletions.mjs` is invoked only from `ci.yml`, which filters `branches: [main]` — so on a stacked rung it still never fires, and the green `Detect changes` in the rollup comes from `regression` / `preview-regression` / `Player perf`, each of which defines its own job by that name. What the guard will actually be asked at merge time is `main...HEAD`, so I asked it that: the cumulative stack removes exactly **8** files against `main`, I drove the shipped `classify()` and `ALLOWED_DELETIONS` with that list, and it returns **zero unagreed deletions**. The guard passes.

## One residue, filed on #3445

`packages/core/src/runtime/window.d.ts:45` still declares `setAudioSolo?: (ids: readonly string[]) => void` with a docblock describing solo as a live feature. After the whole stack lands it is the only surviving mention of solo in the tree — no implementation, no caller. **No rung in the stack edits `window.d.ts`.**

My round-1 audit here missed it because the sweep was keyed on *exported symbols*, and this is an interface member rather than an export. I have filed the detail on #3445, which is the rung that removes the implementation and therefore where the declaration goes stale — but this PR is the natural place to carry the six-line deletion if you would rather keep the solo removal in one commit.

vanceingalls:
Fixed at fd4bb23dc. This rung now retires the complete runtime solo bridge together with the Studio solo UI, and removes the stale `setAudioSolo` member from `packages/core/src/runtime/window.d.ts`. The final tree is unchanged from the previously verified tip except for that obsolete declaration.

jrusso1020 (commented):
**Re-review at `fd4bb23dc4`** (prior review `5003703365` @ `7d99717033`).

Both of my findings are closed here.

The PR grew by five files — `runtime/init.ts`, `runtime/media.ts`, `runtime/webAudioTransport.ts`, `runtime/webAudioTransport.test.ts`, `runtime/window.d.ts` — so this rung now owns the whole solo removal instead of inheriting a half-removed one from #3445. That split is the right way round: #3453 takes the consumers (the bridge and the render sites), this takes the producer.

`window.d.ts:40-45` is gone, which was my second finding. It is the only line of the merged end state that differs from what I reviewed last round.

**Verified:**

- The removal is complete on this rung — 25 removed lines mentioning solo in `webAudioTransport.ts`, zero added. `soloGain`, `connectThroughSolo`, `setSolo`, `_soloed`, the `media.ts` fallback parameter and the `init.ts` writer all leave together.
- Nothing is stranded. I traced the chain at #3453's head first: `useAudioSoloBridge.ts` is deleted there, nothing renders `TimelineSoloButton`, and `audioSoloSlice.ts` never touched the runtime — so by the time this rung lands there is no consumer left to orphan.
- The deletion-guard result from my last review carries by construction rather than by assumption. The recursive trees at the old and new stack tops hold the same 6,971 paths with one blob differing (`window.d.ts`, modified rather than deleted), so the `main...top-of-stack` deletion set is unchanged: 8 deletions, allowlist covers all 8, zero unagreed.

No blockers.

jrusso1020 (dismissed):
Approving at `fd4bb23dc4`, the exact head of my re-review above — that review is the audit; this is the stamp <@vanceingalls> asked for in Slack.

Scope of this approval, stated so it is checkable: it covers `fd4bb23dc4bc7029682b41d4c98c7e0bb43fc24c` only. `main` runs `require_last_push_approval: true`, so if this head moves the PR returns to REVIEW_REQUIRED rather than carrying my stamp forward — re-ping me and I will re-review at the new head.

miguel-heygen (approved):
Independent exact-head pass at fd4bb23dc4bc7029682b41d4c98c7e0bb43fc24c.\n\n**Audited:** all 17 changed paths, with a repository-wide live-symbol sweep for the removed solo and meter APIs.\n**Trusting:** none of the changed production paths; deleted test bodies were checked through their owning production deletions and the current required matrix.\n\nThe removal is complete at this rung: Studio consumers are already gone, the runtime writer/transport stage and Window declaration leave together, the compatibility predicates are removed, and no live import or call site remains. The deletion allowlist names every intentionally deleted path without widening the guard.\n\nAll current required checks are green.\n\n**Verdict:** APPROVE\n**Reasoning:** Solo and group-meter removal is consolidated, reference-clean, and guarded as an intentional breaking deletion with no stranded runtime surface.\n\n— Magi

PR #3455, opened 2026-08-23
docs(audio): document grouped audio and its guardrails

vanceingalls:
Part 12 of 12 replacing #3439. Base: `wa-26k-remove-solo-meter`.

## Why

Documentation, skills, manifest, and local Studio-server guidance should describe the code only after the runtime and authoring contract is complete.

## What

- teach `/hyperframes-audio` the submix-bus model and attributes
- update routing guidance and the generated skills manifest
- retain the historical audio-group handoff as implementation context
- correct local Studio server invocation and top-level contributor guidance

## Verification

- final tree preserves the verified #3439 replacement and includes the review fixes landed across the stack

Stack: #3454 ← **#3455**


jrusso1020 (commented):
Reviewed at `b0436d62`. **No blockers.** Docs match what the stack actually ships, and the one machine-checkable claim on the PR passes — I ran it rather than trusting it.

## Ran the manifest check by hand, because CI has not

`skills-manifest.json` is maintained by a lefthook **pre-commit** hook (`lefthook.yml:19`) and verified in CI by `bun packages/cli/scripts/gen-skills-manifest.ts --check` at `.github/workflows/ci.yml:397`. `ci.yml` filters `branches: [main]`, so on this base that check has never run — and this is a PR whose entire risk surface is "did the generated fingerprint keep up with the hand-edited skills."

Ran the shipped `--check` against the tree at `b0436d62`:

```
✓ skills-manifest.json is in sync (20 skills)
```

Both edited hashes match, and `files: 17` / `files: 7` are right since the PR edits existing skill files rather than adding any. Worth stating precisely here because §1 of the handoff doc this same PR commits records that **34 of 69 commits were made with `--no-verify`** — that flag is exactly what skips the hook keeping this file honest, so "the manifest is fine" was a claim with a real way to be false. It is fine.

## 1. The handoff doc's own banner points at a closed PR

`plans/audio-groups-session-handoff.md:3-8` opens with:

> §1's "nothing is pushed, no PR exists" and §8's open-items list are superseded by that PR's description.

where "that PR" is **#3363**, closed 2026-08-22. Its successor #3439 closed 2026-08-23. So the doc's one instruction for finding current state routes a reader through two closed PRs, and it lands in `main` permanently that way. Repointing it at the `wa-26` stack (or at #3455) is a one-line fix.

Same shape as the eight `#3439 deliberately removes …` reasons in `ALLOWED_DELETIONS` on #3454 — worth fixing in both places with the same edit, since in both cases the citation *is* the audit trail and it outlives the branch.

## 2. What I checked and did NOT find, so nobody re-derives it

A 493-line internal session handoff — worktree paths, `--no-verify` counts, a list of files over the repo's own line cap — landing in a public repo is the obvious thing to flag, so I checked the base before writing it up. `plans/webaudio-stack-handoff.md` is already on the base branch and already carries both the `--no-verify` discussion and `~/src/wt` worktree paths, alongside six other `plans/` entries. The convention is established and already public; this PR is following it, not introducing it. Nothing to change.

Also verified clean: no `solo` or level-meter references survive anywhere in `skills/hyperframes-audio/`, so the docs are consistent with the removal one rung down rather than describing a feature that no longer exists.

## 3. The `data-hidden` row is the contract #3444's resolve gap breaks

Not a defect in this PR, but this is where the guarantee gets written down, so it is where it is worth naming. The new bus table says:

| `data-hidden` | drops every member from the mix |

Confirmed at *this* head (`packages/core/src/audioGroups.ts:117`) that `isMemberGroupHidden` still resolves the bus with `el.getAttribute(HF_AUDIO_GROUP_ATTR)` — the author id — while `resolveAudioGroups` a few lines below prefers `AUDIO_GROUP_RENDER_ID_ATTR` and falls back to it. On a document where a sub-composition carrying a bus is inlined twice, instance 1's render id *is* its author id, so muting either instance resolves to instance 1's bus. "Drops every member" then holds for the first instance and inverts for the rest.

The reason to mention it on the docs rung rather than only on #3444: this table is what an agent reads, and an agent has no way to discover the discrepancy — it will author `data-hidden`, be told it worked, and be wrong. The fix still belongs on #3444; this is the sentence to re-read once it lands.

## 4. `studioServer.ts` — good change, and the right shape

Turning a `Cannot find module …/renderOrchestrator.js` into "Studio dev-mode rendering requires bun … Restart the studio with `bun run studio`" is the correct trade: the failure was previously indistinguishable from a broken checkout, and the comment records the actual mechanism (`.js` specifiers naming `.ts` files, plus Vite's own `#!/usr/bin/env node` shebang) rather than just the symptom. Nothing that worked stops working — the dev-mode branch imports producer *source*, which Node could never resolve, so the only behaviour change is which error you get. Guarding on `process.versions.bun` rather than on a version string is the durable spelling.

## CI

Same stack position: of `main`'s eight required contexts only `regression` is present at `b0436d62`, green alongside the regression shards and perf jobs. `Build`, `Test`, `Typecheck`, `Test: runtime contract`, both windows jobs and `Semantic PR title` have never run — `branches: [main]`, self-repairing on retarget via `edited` in `ci.yml`'s `types:`. On this rung the practical consequence is the manifest check above, which is why I ran it here rather than waiting. `Perf: ${{ matrix.shard }}` showing with the matrix variable unexpanded is a job-level `if:` that was false, not a job still queued.

vanceingalls:
Addressed the handoff feedback: its historical banner now points directly to the replacement stack (#3444–#3455) instead of the closed predecessor PR. The deletion audit trail one rung down was updated at the same time.

I also fixed the cumulative CI failures inherited here: all changed Studio source files are now at or below the 600-line gate, and all eight intentional deletions are accepted by the path-specific guard. Full build, typechecks, fallow, formatting, lint, and 387 focused tests pass locally.

jrusso1020 (commented):
Re-review at `3a09474b37`. My earlier review (`5003567032`) was pinned to `b0436d623e`, which the restack replaced; the branch is now `b60b2306` + `3a09474b` *"point handoff at replacement stack"*.

## Closed: the handoff banner no longer routes through closed PRs

`plans/audio-groups-session-handoff.md` now reads *"That branch was later replaced by the 12-PR stack **#3444–#3455**"*, so the doc's one instruction for finding current state points at open work instead of at #3363 (closed 08-22) and #3439 (closed 08-23). Same edit landed on #3454's allowlist reasons.

## Re-ran the manifest gate at this head, not the old one

The head moved, so the old result does not transfer. Against the tree at `3a09474b37`:

```
✓ skills-manifest.json is in sync (20 skills)
```

Still the only machine-checkable claim on the PR, still not covered by CI on this base (`ci.yml:397` sits behind the same `branches: [main]` filter), and still worth running by hand because §1 of the handoff doc this PR commits records **34 of 69 commits made with `--no-verify`** — the flag that skips the lefthook pre-commit hook maintaining that file.

## Retracting a caveat: the `data-hidden` sentence is now true

I noted last time that this PR documents `data-hidden` on a bus as *"drops every member from the mix"*, and that #3444's `isMemberGroupHidden` bug made that false for a twice-inlined sub-composition — the preview kept the muted instance's members audible while the render dropped them. #3444 at `62b6d35b` fixes exactly that, with a test on the two-instance fixture. The documentation and the behaviour now agree, so the caveat is withdrawn rather than left standing.

## Also verified clean at this head

No `solo` or level-meter references survive in the docs or skills the PR edits. The one solo leftover anywhere in the tree is `packages/core/src/runtime/window.d.ts:45`, which no rung in the stack touches — detail on #3445, and this rung could equally carry the deletion since it is the one that documents the audio surface.

Nothing else open from me.

jrusso1020 (commented):
**Re-review at `2c762fa446`** (prior review `5003703408` @ `3a09474b37`).

Restack only — all 8 files in the PR's own file list are blob-identical across `3a09474b37` → `2c762fa446`, checked by blob sha per path, so the prior review transfers exactly.

The one machine-checkable claim carries too, and by construction rather than assumption: the recursive trees at the two stack tops hold the same 6,971 paths with exactly one blob differing (`runtime/window.d.ts`, which is not a skill file), so my run of the manifest gate at the previous head still holds — `✓ skills-manifest.json is in sync (20 skills)`. Worth re-deriving rather than assuming, since that gate lives in the `main`-only `ci.yml` and cannot run on a stacked rung.

No blockers.

jrusso1020 (dismissed):
Approving at `2c762fa446`, the exact head of my re-review above — that review is the audit; this is the stamp <@vanceingalls> asked for in Slack.

Scope of this approval, stated so it is checkable: it covers `2c762fa446f407de67d646b2e3a3307bbfc2895b` only. `main` runs `require_last_push_approval: true`, so if this head moves the PR returns to REVIEW_REQUIRED rather than carrying my stamp forward — re-ping me and I will re-review at the new head.

miguel-heygen (approved):
Independent exact-head pass at 2c762fa446f407de67d646b2e3a3307bbfc2895b.\n\nThe skill/router/README descriptions agree on the submix-bus contract and no removed solo or meter surface is still advertised. The Studio-server guard is correctly limited to dev-mode source loading: packaged producer imports remain unchanged, while Node receives an actionable failure before attempting a TypeScript-source import it cannot resolve. The generated skill manifest is current, and its required CI check is green at this exact head.\n\n**Verdict:** APPROVE\n**Reasoning:** Documentation, routing guidance, manifest metadata, and local Studio invocation now match the final runtime behavior without widening production behavior.\n\n— Magi

PR #3459, opened 2026-08-24
fix(core): prevent cross-origin Web Audio capture from silencing audio

### LINKED ISSUE #3458: createMediaElementSource silently mutes cross-origin media that has no CORS opt-in
# createMediaElementSource silently mutes cross-origin media that has no CORS opt-in

**Package:** `@hyperframes/core` (runtime) — reproduced on 0.8.10 and 0.8.12; 0.8.2 is unaffected (it has no `createMediaElementSource` call site).

## Summary

The runtime routes every playing `<audio data-start>` through
`AudioContext.createMediaElementSource()`. When the element's media is
cross-origin and the element has no `crossorigin` attribute, the Web Audio spec
requires the captured output to be **zeroed**:

> If the media element's resource has a different origin than the document and
> the element does not satisfy the CORS-cross-origin check, the node outputs
> silence.

So the composition looks perfectly healthy and is completely silent:

- the element keeps playing, `currentTime` advances, visuals animate
- no exception is thrown, nothing is logged
- `paused === false`, `muted === false`, `volume === 1`

This is the common shape for any composition whose audio lives on a third-party
CDN (podcast enclosures, most media hosts), because such CDNs usually don't send
`Access-Control-Allow-Origin`, and adding `crossorigin="anonymous"` without it
makes the element fail to load at all.

## Measurement

Same composition, same runtime, same player, only the audio origin differs.
Peak measured on the master bus with an `AnalyserNode`:

| audio | capture | master-bus peak |
| --- | --- | --- |
| cross-origin, no `crossorigin` attribute | succeeds | **0** |
| cross-origin, `crossorigin="anonymous"` + `Access-Control-Allow-Origin: *` | succeeds | 0.366 |
| same-origin | succeeds | 0.366 |

## Repro

Two origins: A serves the page, B serves the audio with **no** CORS headers.

```html
<!-- served from origin A, cross-origin audio from origin B -->
<!doctype html>
<html>
  <head>
    <script src="https://cdn.jsdelivr.net/npm/@hyperframes/core@0.8.12/dist/hyperframe.runtime.iife.js"></script>
  </head>
  <body>
    <div data-composition-id="demo" data-width="1080" data-height="1920" data-duration="8">
      <audio id="demo-audio" src="https://origin-b.example/tone.wav" data-start="0" data-duration="8" data-volume="1"></audio>
    </div>
  </body>
</html>
```

Play it: the timeline runs, the element plays, there is no sound. Serve the same
file from origin A instead, or add `Access-Control-Allow-Origin: *` on B plus
`crossorigin="anonymous"` on the element, and the audio is heard.

## The existing escape hatch can't be used reliably

`postMessage({source: "hf-parent", type: "control", action: "set-web-audio-media-disabled", disabled: true})`
does produce the right behaviour — but it can only ever *race* the problem:

- the capture happens during the runtime's own bootstrap play
- `createMediaElementSource` is permanent for the lifetime of the element, so a
  message that arrives after the first capture changes nothing (a later capture
  attempt on that element throws `InvalidStateError`)
- the runtime's control-message listener registers during its DOMContentLoaded
  bootstrap, while the player's `ready` probe polls for globals the runtime IIFE
  sets at parse time — so a message sent on `ready` can land in the gap and be
  dropped

In practice, sending it from the embedder on `ready`, and again on the runtime's
own `{source: "hf-preview", type: "ready"}` broadcast, still left roughly 4 out
of 5 page loads silent.

## Suggested fixes

1. **Guard before capturing.** Skip `createMediaElementSource` for an element
   whose `currentSrc` is cross-origin and which has no `crossorigin` attribute,
   and fall through to the existing native-playback path — the same path the
   runtime already takes when the capture fails or when there are no active
   sources. A diagnostic would make the trade-off visible ("routing X natively:
   capture would be silent without CORS").

   This alone fixes the class of bug: the runtime never chooses a path that is
   guaranteed to be silent.

2. **Make the switch declarative.** A player attribute (e.g.
   `web-audio-media="off"`) or a value the runtime reads at bootstrap would let
   an embedder opt out without having to win a message race. Today the only
   route is a control message whose delivery window is not under the embedder's
   control.

3. *(optional)* A per-element opt-out, e.g. `data-native-audio` on the media
   element, for compositions that mix same-origin FX-processed audio with a
   third-party track.

## Workaround, for anyone hitting this

Install a tiny script ahead of the runtime that makes
`createMediaElementSource` throw a `SecurityError` for exactly those elements —
cross-origin, no `crossorigin` attribute. The runtime already handles a failed
capture gracefully (it keeps no active source, and its transport tick then plays
the element natively), so the result is audible playback with no ordering
assumption. It is a no-op for same-origin and CORS-enabled media, which keep the
full Web Audio path.

desenmeng:
## What

Prevent `AudioContext.createMediaElementSource()` from silently taking over and zeroing cross-origin `<audio>` elements that were loaded without a `crossorigin` opt-in.

- Classify each audio element before Web Audio capture as `web-audio`, `decode-only`, or explicitly `native`.
- Keep the existing fetch/decode fallback for automatically bypassed cross-origin media, preserving the FX graph when the server permits CORS.
- Add `data-native-audio` as a per-element escape hatch for redirects and other cases URL inspection cannot settle.
- Surface preview-only diagnostics through `hyperframes check`, including processing that native fallback cannot reproduce.
- Document the new attribute and add regression coverage across core runtime, transport, and CLI checks.

## Why

The Web Audio specification requires a `MediaElementAudioSourceNode` created from media that fails the CORS-cross-origin check to output silence. The call succeeds and does not throw, so the existing `try/catch` cannot recover. Creating the node also permanently reroutes the element away from native output for its lifetime.

The result is a composition whose timeline and visuals continue normally while its audio is completely silent, with no error or diagnostic. Guarding before node creation preserves audible native playback instead.

Fixes #3458.

## How

The runtime now uses a pure route classifier at discovery and scheduling time:

- Same-origin, CORS-opted-in, and non-HTTP(S) sources keep the existing Web Audio path.
- Cross-origin HTTP(S) sources without a `crossorigin` attribute skip `createMediaElementSource()` but still attempt `fetch` plus `decodeAudioData`; native playback remains the final fallback.
- `data-native-audio` skips both Web Audio capture and decode.

Source selection follows `currentSrc`, then `src`, then `<source>` candidates. The transport also rechecks the route immediately before the irreversible capture call. Diagnostics are emitted during preview discovery so `hyperframes check` can observe them without playback, but are suppressed during export rendering, where audio is mixed offline. The existing non-unit-rate fail-closed policy remains limited to `data-fx-chain` and `data-automation`.

## Test plan

Verified locally:

- Core focused runtime tests: 183 passed
- CLI `checkBrowser` tests: 16 passed
- Core and CLI typechecks passed
- Changed-file formatting and lint checks passed

### DIFF
--- a/docs/reference/html-schema.mdx
+++ b/docs/reference/html-schema.mdx
@@ -126,6 +126,7 @@ Audio has no visual lifecycle.
 | `data-playback-rate` | Video, audio, nested composition | Playback multiplier from `0.1` to `5` |
 | `data-volume` | Video and audio | Static volume from `0` to `1` |
 | `data-has-audio="true"` | Video | Declares that the video contributes audio |
+| `data-native-audio` | Audio | Keeps the element on the browser's own output instead of routing it through Web Audio. The runtime already does this automatically for cross-origin media with no `crossorigin` opt-in — the Web Audio spec makes such a source silent — so this is the escape hatch for the cases a URL cannot settle, such as a same-origin path that redirects to a CDN. Native output cannot carry `data-fx-chain`, `data-automation`, `data-audio-group`, or a `data-volume` above `1`; the runtime reports what it dropped, and `hyperframes check` surfaces it as a `web_audio_bypass` finding. |
 
 Video and audio may omit `data-duration` when their intrinsic duration is known
 and the whole remaining source should play.
--- a/packages/cli/src/utils/checkBrowser.test.ts
+++ b/packages/cli/src/utils/checkBrowser.test.ts
@@ -436,6 +436,53 @@ it("surfaces the runtime's media-proxy-unavailable console.info line as its own 
   );
 });
 
+it("surfaces the runtime's web-audio-bypass console.info line as its own info finding", async () => {
+  // The whole complaint in #3458 is that nothing was reported. The runtime
+  // emits this from media DISCOVERY, not from playback scheduling, precisely
+  // because `check` seeks and never calls play() — a diagnostic raised from
+  // the transport would never reach this scraper.
+  vi.spyOn(Date, "now").mockReturnValue(100);
+  mountCanvasFixture();
+  const page = fakePage();
+  const bypassMessage = fakeConsoleMessage(
+    "info",
+    '[hyperframes] runtime_web_audio_bypass: "https://cdn.example.com/track.mp3" ' +
+      "(cross_origin_no_cors): Web Audio capture withheld; the track plays through native " +
+      "HTMLMediaElement output. Native playback cannot reproduce: fx-chain — proxy or download " +
+      "the asset to a same-origin URL to keep it.",
+  );
+  const authorInfo = fakeConsoleMessage("info", "debug runtime_web_audio_bypass lookalike");
+  page.on = vi.fn(
+    (event: string, handler: (message: ReturnType<typeof fakeConsoleMessage>) => void) => {
+      if (event === "console") {
+        handler(bypassMessage);
+        handler(authorInfo);
+      }
+    },
+  );
+  installSessionMock(page);
+
+  const result = await runBrowserCheck(
+    PROJECT,
+    { ...DEFAULT_CHECK_OPTIONS, samples: 1, contrast: false },
+    { kind: "none" },
+    runAuditGrid,
+  );
+
+  expect(result.runtimeFindings).toContainEqual(
+    expect.objectContaining({
+      code: "web_audio_bypass",
+      severity: "info",
+      message: bypassMessage.text(),
+    }),
+  );
+  // Prefix-anchored, so a composition author's own console.info that merely
+  // mentions the code is not promoted into a finding.
+  expect(result.runtimeFindings.some((finding) => finding.message === authorInfo.text())).toBe(
+    false,
+  );
+});
+
 it("elevates and deduplicates WebGPU validation warnings while preserving ordinary warnings", async () => {
   vi.spyOn(Date, "now").mockReturnValue(100);
   mountCanvasFixture();
--- a/packages/cli/src/utils/checkBrowser.ts
+++ b/packages/cli/src/utils/checkBrowser.ts
@@ -264,6 +264,11 @@ export async function captureFindingCrops(
 // `console.info` from a composition author's own script must not.
 const MEDIA_PROXY_MARKER_PREFIX = "[hyperframes] runtime_media_proxy_";
 const MEDIA_PROXY_UNAVAILABLE_MARKER = "[hyperframes] runtime_media_proxy_unavailable";
+// `reportWebAudioMediaRoute` (packages/core/src/runtime/webAudioRoute.ts) uses
+// the same code-in-the-console-line contract. It is emitted from the media
+// DISCOVERY phase rather than from playback scheduling, precisely so this
+// scraper can see it — `check` seeks, it never plays.
+const WEB_AUDIO_BYPASS_MARKER = "[hyperframes] runtime_web_audio_bypass";
 const WEBGPU_RUNTIME_FAILURE =
   /\b(?:GPUValidationError|GPUOutOfMemoryError|GPUInternalError)\b|WebGPU uncaptured error|(?:destroyed\b.*\b(?:GPU )?(?:resource|buffer|texture)\b.*\bsubmit)|(?:(?:GPU )?(?:resource|buffer|texture)\b.*\bdestroyed\b.*\bsubmit)/i;
 
@@ -288,6 +293,20 @@ function pushRuntimeDraft(drafts: RuntimeDraft[], draft: RuntimeDraft): void {
     return;
   }
   drafts.push({ ...draft, count: 1 });
+}
+
+/**
+ * The finding code for a runtime-emitted `console.info` line, or null for the
+ * ordinary info logging a composition author's own script produces. Matching is
+ * prefix-anchored on the stable diagnostic codes the runtime deliberately embeds
+ * in the text, so a line that merely mentions one is not promoted.
+ */
+function runtimeInfoFindingCode(text: string): string | null {
+  if (text.startsWith(WEB_AUDIO_BYPASS_MARKER)) return "web_audio_bypass";
+  if (!text.startsWith(MEDIA_PROXY_MARKER_PREFIX)) return null;
+  return text.includes(MEDIA_PROXY_UNAVAILABLE_MARKER)
+    ? "media_proxy_unavailable"
+    : "media_proxy_fallback";
 }
 
 function wireRuntimeListeners(page: Page, drafts: RuntimeDraft[], currentTime: () => number): void {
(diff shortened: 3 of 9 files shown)


miga-heygen (commented):
Reviewed the diff (`docs/reference/html-schema.mdx`, `packages/cli/src/utils/checkBrowser.ts(.test.ts)`, `packages/core/src/runtime/init.ts`, `webAudioRoute.ts` (new), `webAudioTransport.ts(.test.ts)`, `init.test.ts`) plus the un-diffed `decodeAudioElement`/`scheduleWebAudioForActiveClips` call sites to check a perf hypothesis below.

## Summary
Fixes #3458: `AudioContext.createMediaElementSource()` silently outputs silence (no throw) for cross-origin `<audio>` with no `crossorigin` opt-in, and permanently steals the element's native output on the way. The fix introduces a pure `classifyWebAudioMediaRoute` classifier (`web-audio` / `decode-only` / `native`) consulted at three points — discovery (`loadedmetadata` + immediate call), schedule time in `init.ts`, and as an enforcement backstop inside `WebAudioTransport.acquireMediaElementSource` — plus a `data-native-audio` escape hatch and a `hyperframes check`-visible `web_audio_bypass` finding.

## Correctness
- Verified the classifier's origin/candidate-selection logic against the HTML resource-selection algorithm (`currentSrc` > `src` > `<source>` children) — matches spec priority and is well tested, including the "browser already committed via currentSrc" and "selection still unsettled, be conservative across all `<source>` candidates" cases.
- `crossorigin` enumerated-attribute handling (presence, not value, is the opt-in) is correct and tested against `anonymous`/`use-credentials`/empty/garbage values.
- `data-native-audio` correctly short-circuits both Web Audio capture *and* decode (a decode success would otherwise re-mute the element via a buffer source) — this is enforced independently in `webAudioTransport.ts`, not just `init.ts`, so a direct caller (studio/player) can't reopen the one-way door. Good defense-in-depth, tested (`webAudioTransport.test.ts`).
- The non-unit-playback-rate fail-closed mute rule is correctly scoped to stay `web-audio`-route-only (`route.kind === "web-audio" && hasProcessing`) so it doesn't newly mute grouped/above-unity-volume tracks on the new decode-only route — good catch on scope creep, and it's explicitly tested against regressing pre-#3458 behavior.
- Chased a perf hypothesis: since "decode-only" is now a mainline path for ordinary cross-origin audio (not just a rare fallback), does `webAudio.decodeAudioElement()` get called repeatedly (e.g. once per animation frame)? Confirmed no — `decodeAudioElement` has an internal `_bufferCache`/`_failedSrcs` keyed by `src`, and `scheduleWebAudioForActiveClips()` (the only caller of this scheduling path) is invoked only on discrete transitions (play, playback-rate change, hidden-audio dirty), not per rAF tick. No redundant fetch/decode risk.
- Render-mode gating (`isRenderMode()` checked before the diagnostic and before the latch) matches the documented rationale — render mixes offline and reproduces the FX chain, so reporting a "bypass" there would be a false claim, and it's confirmed the once-per-element latch isn't consumed by a render-mode call (tested).

## Type safety
- `RuntimeJson` recursively permits `string[]`, so `lostProcessing: string[]` in the diagnostic `details` typechecks against `Record<string, RuntimeJson>`. No unsafe casts introduced; `getAttr`/`hasAttr` guard against non-`Element` inputs defensively.

## Architecture / SSOT
- `classifyWebAudioMediaRoute` is called at up to three points (discovery in `init.ts`, schedule loop in `init.ts`, and `acquireMediaElementSource` in `webAudioTransport.ts`). This reads as intentional defense-in-depth (comment: "this stays the enforcement point so a direct caller cannot reopen the one-way door") rather than duplicated decision logic — it's the same pure function called from multiple call sites, not reimplemented, and `reportWebAudioMediaRoute`'s `WeakSet` latch keeps the diagnostic single-fire regardless of how many times it's (re)classified. No SSOT violation.
- `nativeUnexpressibleProcessing()` is intentionally wider (adds `audio-group`, `above-unity-gain`) than the schedule loop's own `hasProcessing` check (`fx-chain`/`automation` only) — the PR comments explicitly call out *why* these answer different questions (diagnostic = "what's lost", fail-closed rule = "is losing FX worse than silence"). Documented rather than accidental drift.

## Edge cases
- Non-http(s) schemes (`blob:`, `data:`, `file:`) correctly fall through to the pre-existing `web-audio` behavior rather than guessing — reasonable, avoids new regressions on schemes the origin check can't meaningfully judge.
- `<video>` is correctly excluded (`reportWebAudioRoute` checks `instanceof HTMLAudioElement`) since only `<audio>` reaches `createMediaElementSource` per the existing capture path — tested.
- One minor question, not a blocker: `reportWebAudioMediaRoute` returns early (before adding to the `diagnosedElements` latch) when `route.kind === "native"` and nothing is lost. That's correct behavior (no spam, "honoring a request isn't a finding"), but it does mean that specific case recomputes `nativeUnexpressibleProcessing()` on every call for the lifetime of the element instead of latching a no-op result — cheap, so not worth blocking on, just noting for anyone chasing a future perf ticket.

## Tests / docs
- Coverage is thorough: classifier unit tests, transport-level enforcement tests, CLI scraper prefix-anchoring test (verifies a composition author's own `console.info` that merely mentions the code isn't promoted into a finding), and an end-to-end `init.test.ts` suite covering the full cross-origin/native/same-origin/video matrix plus the fail-closed-rule regression guard.
- Docs (`html-schema.mdx`) accurately describe the new attribute and match `nativeUnexpressibleProcessing()`'s actual dropped-feature list (fx-chain, automation, audio-group, above-unity volume).

## Verdict
Solid, well-reasoned fix for a nasty silent-failure class of bug (spec-mandated silence instead of a throw). Didn't find correctness, type-safety, or SSOT issues worth blocking on — the multi-call-site classification is deliberate defense-in-depth, and the perf concern I chased down doesn't materialize since decode is cached by src and the scheduling path only runs on discrete state transitions.

— Miga

miga-heygen (changes requested):
## Changes Requested

**Remove the `data-native-audio` escape hatch — the automatic detection is sufficient.**

The `classifyWebAudioMediaRoute` classifier and `isCorsSilenced` origin check already cover the core bug (#3458): cross-origin `<audio>` elements without a `crossorigin` attribute are correctly routed to `decode-only` or `native` fallback. That's the fix, and it's solid.

The `data-native-audio` attribute handles a narrow edge case (same-origin URLs that redirect to a cross-origin CDN at request time), but URL inspection can't detect that scenario anyway — and authors who hit it can use the standard HTML solution: add `crossorigin` to the `<audio>` element, assuming the CDN sends CORS headers.

**What to remove:**
- The `HF_NATIVE_AUDIO_ATTR` constant and all references to `data-native-audio`
- The `authored_opt_out` reason branch in `classifyWebAudioMediaRoute`
- The `nativeUnexpressibleProcessing` function (only used for the native opt-out diagnostic)
- The `web_audio_bypass` finding in `hyperframes check` (tied to the attribute)
- All test cases specific to `data-native-audio` behavior
- The doc-table row describing the attribute

**Keep everything else** — the automatic origin detection, the `decode-only` route, the transport-layer enforcement, the `WeakSet` diagnostic latch, and the existing tests for cross-origin behavior are all good.

— Miga

desenmeng:
@miga-heygen Thanks — makes sense. I’ll remove `data-native-audio`, the `authored_opt_out` route, and the related tests/docs.

One thing I want to clarify before I update the branch: `web_audio_bypass` and `nativeUnexpressibleProcessing()` also cover the automatic `cross_origin_no_cors` path; they’re not only used by the authored opt-out. You also mentioned keeping the `WeakSet` diagnostic latch.

Should I keep the automatic bypass diagnostic and its `hyperframes check` finding, or remove the diagnostic/CLI finding entirely and leave only the routing and enforcement?

jrusso1020:
Closing as superseded by #3481: automatic cross-origin Web Audio routing is implemented without requiring the data-native-audio escape hatch. Thanks for the contribution.

PR #3461, opened 2026-08-24
fix(cli): assign w{index} ids when loading JSON transcripts

### LINKED ISSUE #3442: transcribe never assigns word `id` on the JSON path, so per-word caption overrides have nothing to key on
## Summary

`media-use/audio/references/transcribe.md:43` states the word `id` (`w0`, `w1`, …) "is added during normalization for stable references in caption overrides."

For the JSON transcripts the CLI itself produces, it never is. Per-word caption overrides have nothing to key on.

## Environment

- hyperframes 0.8.10, installed globally via npm
- macOS 15, Apple M5 Pro
- whisper-cpp `small.en` via Homebrew `whisper-cli`
- Independently reproduced on WSL2 / Ubuntu

## Measured

`hyperframes transcribe probe.wav --model small.en` on 1.7 s of speech:

```json
[
  { "text": "The",         "start": 0.01, "end": 0.16 },
  { "text": "pipeline",    "start": 0.16, "end": 0.6  },
  { "text": "is",          "start": 0.6,  "end": 0.68 },
  { "text": "proven",      "start": 0.82, "end": 1.04 },
  { "text": "end-to-end.", "start": 1.04, "end": 1.74 }
]
```

Zero `id` fields across all 5 words.

## Where it is and is not assigned

`loadTranscript` (dist/cli.js:97787) branches on file type:

| input | `id` assigned? |
|---|---|
| `.srt` | yes — `id: w.id ?? "w" + i` |
| `.vtt` | yes — `id: w.id ?? "w" + i` |
| JSON, whisper-cpp shape | **no** — `parseWhisperCpp` emits `{text,start,end}` only |
| JSON, openai shape | **no** — `parseOpenAI` emits `{text,start,end}` only |
| JSON, flat array | **`id: w.id ?? ""`** — empty string, not `"w" + i` |

The flat-array branch is the one that looks like the documented normalization and is not: a missing id becomes `""`, so every word ends up sharing the same empty key.

Since the branch is chosen by **file format** rather than by caller, going through the skill-driven flow instead of the CLI does not change the result — it is the same `loadTranscript`.

## Workaround (verified, no code change)

Round-trip through SRT with `--preserve-cues` on **both** legs:

```bash
hyperframes transcribe audio.wav --model small.en                          # -> transcript.json, no ids
hyperframes transcribe transcript.json --to srt -o w.srt --preserve-cues
hyperframes transcribe w.srt --preserve-cues                               # -> transcript.json WITH ids
```

Result: 5 entries, word-level timings preserved, `id` `w0`…`w4` present.

**`--preserve-cues` is required on the export leg too.** Without it the export groups all 5 words into a single cue, and the word granularity is gone before the id assignment ever runs — the re-import then yields one entry, `id: "w0"`, text `"The pipeline is proven end-to-end."`.

The CLI help describes `--preserve-cues` purely in import terms ("Keep each transcript entry as its own caption cue"), which gives no hint that it is load-bearing on export.

## Suggested fixes

1. Assign `"w" + i` in `parseWhisperCpp` / `parseOpenAI`, or in the flat-array branch of `loadTranscript` instead of defaulting to `""`.
2. Correct `transcribe.md:43` — as written it promises ids on the path that never produces them.
3. Document `--preserve-cues` as applying to export as well as import.

santhiprakash:
Fixes #3442

## What

`loadTranscript` now assigns `id: w{i}` on the JSON branches (whisper-cpp, OpenAI, words-json), matching what the srt/vtt branches already do. The words-json branch also stops defaulting missing ids to `""` — it repairs them to `w{i}` instead.

## Why

`audio/references/transcribe.md` documents that word ids (`w0`, `w1`, …) are "added during normalization for stable references in caption overrides", but every JSON path dropped them: `parseWhisperCpp` and `parseOpenAI` never set `id`, and the flat words-json branch defaulted it to `""`, collapsing every word onto the same empty key. All engines funnel through `loadTranscript` — `transcribeAudio` rewrites `transcript.json` from its output — so CLI-produced transcripts ship without ids and per-word caption overrides have nothing to key on (#3442 has the full breakdown, including the SRT round-trip workaround users currently need).

## How

Single normalization point: after the existing format-specific parsing in `loadTranscript`, map `id: w.id || \`w{i}\`` over the words. `||` rather than `??` so the empty-string ids written by older CLIs into words-json files are repaired on load, not preserved. SRT/VTT branches keep their existing assignment; the `Word.id` doc comment already describes exactly this behavior.

## Test plan

- [x] `packages/cli`: `vitest run src/whisper/` — 104/104 pass, including updated expectations (whisper-cpp / openai / words-json now carry `w0…` ids) and a new regression test that preserves existing ids (`keep-me`) while repairing empty (`"" → w1`) and missing (`→ w2`) ones.
- [x] Full `packages/cli` vitest suite — same 20 pre-existing environmental failures as clean `upstream/main` (macOS-only telemetry tests on Linux, unbuilt workspace artifacts); zero new failures. Verified by stash/re-run on the clean tree.
- [x] `tsc --noEmit` in `packages/cli` — clean. `oxlint` + `oxfmt --check` + `fallow audit` via lefthook pre-commit — all green.

This change is AI-assisted; the bug analysis, fix, and tests were verified against the reproduction in #3442.

### DIFF
--- a/packages/cli/src/whisper/normalize.test.ts
+++ b/packages/cli/src/whisper/normalize.test.ts
@@ -103,8 +103,8 @@ describe("loadTranscript", () => {
     const { words, format } = loadTranscript(path);
     expect(format).toBe("whisper-cpp");
     expect(words).toEqual([
-      { text: "Hello,", start: 0, end: 0.55 },
-      { text: "world.", start: 0.6, end: 1.25 },
+      { text: "Hello,", start: 0, end: 0.55, id: "w0" },
+      { text: "world.", start: 0.6, end: 1.25, id: "w1" },
     ]);
   });
 
@@ -142,8 +142,8 @@ describe("loadTranscript", () => {
     const { words, format } = loadTranscript(path);
     expect(format).toBe("openai");
     expect(words).toEqual([
-      { text: "Hello", start: 0, end: 0.5 },
-      { text: "world", start: 0.6, end: 1.2 },
+      { text: "Hello", start: 0, end: 0.5, id: "w0" },
+      { text: "world", start: 0.6, end: 1.2, id: "w1" },
     ]);
   });
 
@@ -205,7 +205,7 @@ Short format
     expect(words[0]?.text).toBe("Bold and italic");
   });
 
-  it("passes through normalized word arrays", () => {
+  it("assigns w{index} ids to normalized word arrays", () => {
     const input = [
       { text: "Hello", start: 0.0, end: 0.5 },
       { text: "world", start: 0.6, end: 1.2 },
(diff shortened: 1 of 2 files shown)


miga-heygen:

santhiprakash:
Rebased onto current main (`d66cd6dcc`) as suggested. The underlying bug is still on main — the words-json loader at `packages/cli/src/whisper/normalize.ts` still falls back to `id: w.id ?? ""` while the SRT/VTT paths assign `w${i}`.

Verification on the rebased head `c1f87c64b`:

- `cd packages/cli && bunx vitest run src/whisper/normalize.test.ts` — 41/41 passing

PR #3462, opened 2026-08-24
docs(changelog): weekly digest 2026-08-17–2026-08-24

jrusso1020:
## Description

Weekly digest for **August 17 to August 24, 2026**, covering 126 non-merge commits and 15 releases (v0.7.110 through v0.8.12).

Generated with `bun run changelog:weekly`, then rewritten for publication. The generator surfaced 10 bullets from 126 commits and weighted them toward whatever landed last, so this is a rewrite from the commit index rather than a polish of its draft.

**The headline is grouped audio.** A voiceover, a music bed and a sound effect can live in named groups, each with its own volume, effects chain and mute. The rollout finished by deleting the three canaries (`audio-fx-rack`, `audio-track-mute`, `audio-groups`) rather than raising their percentages, so the rack, group rows and mute are on for everyone.

Two things the version numbers would have told you wrong, both checked at HEAD rather than inferred:

- **v0.8.0 is not the headline.** The minor bump's only substantive commit is `chore(catalog): remove two AI UI items` (#3317). It marks the removal of the AI generation canvas and AI prompt flow, not a feature launch.
- **Solo and the group meter shipped and were removed inside the same window.** #3290 and #3291 added them; #3454 removed them as the week's one breaking change. `useGroupLevel.ts` and `TimelineGroupBusStrip` are gone at HEAD, and there is no `isSolo` / `soloTrack` / `onSolo` anywhere in `packages/studio/src` or `packages/core/src`. The digest reports them as removed instead of listing them as features.

Group volume *does* survive, as `data-volume` on the group element, and mute works by dropping the member from the mix rather than writing volume 0.

## Testing

- `npx mint validate` from `docs/` passes (`success build validation passed`).
- `node scripts/check-docs-snippet-motion.mjs` and `node scripts/check-tracked-artifacts.mjs` both exit 0.
- Every sha in the digest was machine-verified before the file was written: resolves to 40 hex chars, prefix-matches its 9-char display form, is an ancestor of `origin/main`, and has a commit date inside `2026-08-17T00:00:00Z .. 2026-08-25T00:00:00Z`. The same guard blocks em-dashes, en-dashes in prose, and any leftover `TODO` / style-bar marker. It was negative-tested on all three failure classes so a pass means something.
- All 126 window commits carry a unique inline `(#N)`, so PR mapping needed no merge-ancestry fallback.
- Feature claims confirmed present at HEAD: `audioGroups.ts` (`data-volume`, mute-by-drop, `resolveCarveSourceIds`), `audio/audioFxWorklets.ts`, `commands/normalize-audio.ts`, `previewLifecycle.ts` + `previewLifecycleOutput.ts`, `safePath` used by `htmlBundler`. Confirmed absent: the three retired canaries (with `canary.test.ts` asserting no audio canary returns) and `head_leaked_text`.
- Only `docs/weekly-updates.mdx` is committed. The generator also writes `updates/weekly/` and `updates/social/`; both are left uncommitted.

The video embed points at `weekly-changelog-aug17-24.mp4` following the convention of the last four entries. The file is not uploaded yet, so that URL 404s until the render lands; I will comment the CDN URL on this PR once it is up.

### DIFF
--- a/docs/weekly-updates.mdx
+++ b/docs/weekly-updates.mdx
@@ -13,6 +13,136 @@ For exact versioned release notes, see the [Changelog](/changelog).
 {/* New weekly digest entries are prepended by `bun run changelog:weekly --from YYYY-MM-DD --to YYYY-MM-DD --write`. */}
 
 <Update
+  label="Week of August 17, 2026"
+  description="Weekly digest - August 17, 2026 - August 24, 2026"
+  tags={["Weekly update", "Highlights"]}
+>
+
+<Frame>
+  <DocsVideo
+    title="HyperFrames video: Weekly Changelog Aug17 24"
+    src="https://static.heygen.ai/hyperframes/changelog-videos/weekly-changelog-aug17-24.mp4"
+  />
+</Frame>
+
+Grouped audio shipped. A voiceover, a music bed and a sound effect can now live in named
+groups, each with its own volume, its own effects chain and its own mute. The rollout
+finished by deleting three feature gates rather than raising their percentages, so the
+effects rack, the group rows and mute are on for everyone. Two controls did not survive
+the week: solo and the group meter both shipped and were then removed. Lint got smaller
+on purpose, the CLI preview grew a real lifecycle, and Windows stopped opening stray
+console windows. Fifteen releases shipped, v0.7.110 through v0.8.12. The minor bump to
+v0.8.0 was not the headline: it marks the removal of two catalog components.
+
+## Features
+
+- **Audio groups.** Give an audio element a group and it joins a bus. The group carries a
+  `data-volume`, an effects chain, and a mute. Preview and render build the same graph, so
+  what you hear is what ships. Mute works by dropping the member from the mix rather than
+  setting its volume to zero, which keeps a muted bus out of the export instead of writing
+  silence into it. ([524036715](https://github.com/heygen-com/hyperframes/commit/524036715043c4685d0fd41eaccb5addf5ce31c7), [#3278](https://github.com/heygen-com/hyperframes/pull/3278), [602cf53cf](https://github.com/heygen-com/hyperframes/commit/602cf53cf6d556711dc02592baad5f7e6088e748), [#3287](https://github.com/heygen-com/hyperframes/pull/3287), [99f42be04](https://github.com/heygen-com/hyperframes/commit/99f42be04ce6bce535600c42a8eb694fed8ebc9e), [#3289](https://github.com/heygen-com/hyperframes/pull/3289), [acfa7c55a](https://github.com/heygen-com/hyperframes/commit/acfa7c55a2f3dcdb070e304bfb8e7b819d2ef3cf), [#3286](https://github.com/heygen-com/hyperframes/pull/3286))
+- **The effects rack and its presets are on for everyone.** The twelve-PR audio stack had
+  landed with all three of its canaries at zero percent, which meant the rack, the group
+  rows and mute were in the build and reachable by nobody. The gates were removed rather
+  than raised, so each feature now renders on its own precondition. `audio-fx-rack`,
+  `audio-track-mute` and `audio-groups` no longer exist, and a test asserts that no audio
+  canary comes back. ([0e9a4f371](https://github.com/heygen-com/hyperframes/commit/0e9a4f371d4cff4e853640ea87caf76814af5278), [#3401](https://github.com/heygen-com/hyperframes/pull/3401), [43c4e6935](https://github.com/heygen-com/hyperframes/commit/43c4e6935ef34468b09b71f49a1977b83781e9a1), [#3274](https://github.com/heygen-com/hyperframes/pull/3274), [6a92d2140](https://github.com/heygen-com/hyperframes/commit/6a92d21401974441a6a9b311c78e40df522e209d), [#3292](https://github.com/heygen-com/hyperframes/pull/3292))
+- **Group volume and mute on the timeline row.** The group row carries a volume control and
+  a mute, and the timeline groups member tracks under a disclosure you can split open.
+  ([5fd84c395](https://github.com/heygen-com/hyperframes/commit/5fd84c395b52d8533c924ea31628e71e7e5bc3f6), [#3290](https://github.com/heygen-com/hyperframes/pull/3290), [0d26072e6](https://github.com/heygen-com/hyperframes/commit/0d26072e6cb3f5b901891c26396a0f6c1e0d8853), [#3291](https://github.com/heygen-com/hyperframes/pull/3291))
+- **A carve aims at voiceover groups.** Ducking music under narration now targets the
+  voiceover group rather than a single clip, including when there are several of them.
+  ([485c037dc](https://github.com/heygen-com/hyperframes/commit/485c037dcf2602b2b92fc9297a8dec8f965e79dc), [#3288](https://github.com/heygen-com/hyperframes/pull/3288))
+- **Pitch shift, as the fifth effect worklet.** A granular shifter joins the chain, and the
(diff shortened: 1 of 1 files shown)


jrusso1020:
Video is live, so the embed committed in this PR now resolves.

**CDN:** https://static.heygen.ai/hyperframes/changelog-videos/weekly-changelog-aug17-24.mp4

Verified rather than assumed:

| | |
|---|---|
| local file | 34,784,921 bytes |
| S3 object | 34,784,921 bytes |
| CDN `content-length` | 34,784,921 bytes |
| CDN | `HTTP/2 200`, `content-type: video/mp4`, `x-cache: Miss from cloudfront` |
| duration | 53.0s, 1080x1080, 1590 frames |

`x-cache: Miss` rather than `Hit` is the part that matters, since a Hit would mean CloudFront served a stale object. Invalidation `I7X1AD25LZ5Q7KMDFDHUEVRHQF`.

Gates, all green at the rendered head:
- `hyperframes check` (repo-local CLI, not `npx`): 0 errors, layout 0 issues, contrast 72/72 WCAG AA. Two warnings remain, `composition_file_too_large` and `timeline_track_too_dense`, both inherent to the single-doc master the build spec mandates.
- Seam gate: 5/5 seams, 0 fail, 0 warn, all cut-the-curve LEFT with matched speeds.
- Frames: captions confirmed on all six scene midpoints (rail max-luma 236-237 against a control band at 41-50 measured in the same pass), and all six green moments verified by a before/after differential across their trigger.

Two defects the automated gates could not see, both caught by reading frames:

1. **The Windows cards rendered off-canvas.** I had positioned them at `top: 640/660` inside a card that itself sits at `top: 330`, putting them at absolute y≈970 to 1088 on a 1080 canvas, clipped at the edge and colliding with the caption rail. `hyperframes check` reported `Layout 0 issues` throughout, because every theme slide carries `data-layout-allow-overlap` exactly as the build spec instructs, which disables collision detection on precisely the slides that hold the mocks. Fixed by moving both inside the terminal card's own box, then re-rendered.
2. **My own green-moment threshold was wrong**, not the video. An absolute `>= 250` pixel floor flagged three of six moments; small green type at 19-23px legitimately paints 60-240 pixels. The differential across each trigger (0 to 101, 0 to 237, 0 to 63) is the real measurement.

Every mock is built from verified source rather than invented: the terminal shows the real `--json` flag and the real `PreviewLifecyclePayload` keys (`schemaVersion` / `operation` / `ok` / `result.state|mode|ready`), and the lint scene names `head_leaked_text`, the one rule removal I could confirm by name. I deliberately did not invent seven rule names to fill the panel.

— Rames Jusso

jrusso1020:
**Correcting my own explanation above.** I said `data-layout-allow-overlap` is why `check` missed the off-canvas cards. That was wrong, and the real mechanism is more useful to know. Verified against `packages/cli/src/utils/layoutAudit.ts` at `main`:

An off-canvas element raises `canvas_overflow`, which `allow-overlap` does not touch (that marker suppresses `content_overlap`). Three separate things kept it quiet, and the first is the one that actually did it:

1. **Sampling density.** The layout audit took **9 samples across 53.0s**, one every ~5.9s. The card was on screen for **2.86s** (44.62 to 47.48), so the odds of any sample landing on it were about **49%**. None did, which is why no row appeared at all, not even info.
2. **Persistence tier.** `applyPersistenceTier:273` caps any issue seen in `occurrences <= 1` at **info**. Escalating needs 2 samples, i.e. roughly **11.8s** of continuous visibility at this sample rate. A ~3s element mathematically cannot exceed info, at any breach size.
3. **Breach floor.** Even with enough samples, `isCanvasBreachHeldLarge:297` requires the breach to clear `min(container.w, container.h) * 0.05`, which is **54px** on a 1080 canvas. Mine was **8px** (the card ran 970 to 1088; the sibling console card breached 0px). And `canvas_overflow` tops out at **warning**, never error.

So `check` behaved exactly as designed. The generalization worth keeping: **a geometry defect on an element visible for less than ~12s cannot exceed `info` severity regardless of how far off-canvas it is**, so for anything short-lived the layout gate is not a weaker version of frame-reading, it is silent by construction. That is the argument for treating frames as a separate instrument rather than a redundant one.

Thanks to Home for catching that my `allow-overlap` reasoning was off; the floor figure in that hand-off was 27px, which is the shape of the rule but not this constant.

— Rames Jusso

jrusso1020:
**Correcting my correction. Home's original 27px was right, and I misattributed it.**

I said the operative constant was `min(w,h) * 0.05` = 54px and that sampling was the dominant cause. Both wrong for this defect. Verified in `packages/cli/src/commands/layout-audit.browser.js` at `95e1ac9f`:

Line 1155 states the ownership split outright: *"Painted panels breaching the canvas: text is canvas_overflow's, media is frame_out_of_frame's, panels were nobody's."* My card was a `.glass` painted panel, so the code is **`panel_out_of_canvas`**, not `canvas_overflow`. Its tolerance is `max(24, min(w,h) * 0.025)` = **27px** at 1080 (`:1150-1151`), and that floor is passed **as the tolerance** to `overflowFor`, so an 8px breach returns falsy and **no issue is generated at any sample time**.

So the causal chain is one step, not three:

| | |
|---|---|
| code | `panel_out_of_canvas` (painted panel, per the `:1155` ownership comment) |
| tolerance | 27px |
| my breach | 8px → not detected, ever |
| had it been detected | area 35,360px² vs hero floor 116,640px² → `info` |

**Sampling was never operative here**, because the issue never existed to be sampled. That part of my previous comment overstated a real property into a cause it didn't have. The property itself is still true and still worth knowing: the layout audit places 9 midpoints across 53.0s and `applyPersistenceTier:273` caps `occurrences <= 1` at `info`, so a short-lived instance of any *detected*, persistence-tiered code cannot exceed `info`. `panel_out_of_canvas` is in that set (`:203`). It just wasn't what hid this one.

One more thing I got wrong in the original comment: the skip flag in this function is `data-layout-allow-**overflow**` (`hasAllowOverflowFlag`), which is a different marker from `data-layout-allow-**overlap**`. So neither marker was involved.

Three passes on one question, and the through-line is that I twice cited a real constant for the wrong code. The check that would have settled it the first time is the one-line ownership comment at `:1155`, not the arithmetic. The conclusion that survives all of it is unchanged and is the only part that matters operationally: **an 8px breach on a painted panel is invisible to `check` by design, so frame-reading is a separate instrument, not a redundant one.**

— Rames Jusso

miguel-heygen (approved):
Exact-head review at `fb23d8d379849d78d8ec2d4e450b62d440589699`.

**Audited:** the full 130-line weekly entry, every linked commit/PR pair, the release/window counts, current-main mergeability, and the attached/CDN video across the full timeline including the repaired scene 4. Nothing skipped.

Independent source checks: 83 commit links (82 unique) all prefix-match their 40-char SHA, are ancestors of the digest base, fall inside Aug 17–24 UTC, and carry the linked PR number. The base has exactly 126 non-merge window commits and 15 release tags, `v0.7.110` through `v0.8.12`. `v0.8.0` contains only the release commit plus #3317. Audio canaries/solo/meter absence and surviving group volume/mute/carve code reconcile with the prose. The one-file branch merges cleanly onto current `main`.

Video evidence: attached and CDN bytes have the same SHA-256 (`0b19c14e…`); CDN is HTTP 200 / video/mp4 / 34,784,921 bytes. ffprobe reports H.264 + AAC, 1080×1080, 30fps, 53.0s; audio is present (mean -18.8 dB, max -0.5 dB). I inspected a 9-frame full-film sheet, the outro, and four scene-4 frames: captions remain visible, the Windows cards stay inside their panel and above the caption rail, and the visual themes match the digest.

No blocking findings. Current-head docs validation, CodeQL, format, and path-relevant checks are green.

**Verdict:** APPROVE
**Reasoning:** The publication claims reconcile to immutable source, the embed resolves to the reviewed artifact, and the rendered video is visually and technically clean at the repaired head.

— Magi

PR #3466, opened 2026-08-24
feat(lint): flag late fromTo tweens that flash the authored state before starting

miguel-heygen:
## Problem

A `tl.fromTo()` positioned after the timeline start renders with GSAP's default `immediateRender: false`. If the target's authored state is visible, the element sits fully rendered from t=0, then **jumps to the from-vars when the tween finally starts** — a visible flash on every playback and every cold seek.

Real-world casualty: a typewriter composition whose code lines were supposed to type in one at a time. Every line was fully readable from frame 0, then vanished and slid back in when its tween fired. The composition passed all existing gates.

## Fix

New lint rule `gsap_fromto_flash_before_start` in `packages/lint`. It fires when a `fromTo()`:

- starts after the timeline start (past a small epsilon, so zero-flash-window tweens at t=0 stay exempt), and
- targets an element whose authored state is **not** already hidden, and
- has state-changing from-vars (non-identity transforms, or `opacity`/`autoAlpha` ≠ 1).

Exemptions, each covered by a test:

- authored hidden state: CSS `opacity: 0`, a standalone `gsap.set()`, or a timeline `set`-to-hidden at the timeline start
- identity from-vars (`tl.fromTo(el, { x: 0 }, ...)`) — nothing to jump to
- explicit `immediateRender: true` — GSAP applies the from-vars at build time and holds them

The fixHint offers the two real fixes: hide the authored state (CSS `opacity: 0` + destination `opacity: 1`, or a position-0 `tl.set`), or `immediateRender: true`.

## Testing

- 8 new tests in `packages/lint/src/rules/gsap.test.ts` covering each trigger and exemption (170 pass in the file)
- full `packages/lint` suite: 517 pass; the 8 failures in `hyperframeLinter.test.ts` (`vi.unstubAllGlobals is not a function`) reproduce on `main` untouched and are unrelated

### DIFF
--- a/packages/lint/src/rules/gsap.test.ts
+++ b/packages/lint/src/rules/gsap.test.ts
@@ -3139,3 +3139,117 @@ describe("SVG draw-on rules", () => {
     expect(finding).toBeUndefined();
   });
 });
+
+describe("gsap_fromto_flash_before_start", () => {
+  const comp = (script: string, body = "", style = "") => `
+<html><body>
+  <div data-composition-id="c1" data-width="1920" data-height="1080">${body}</div>
+  ${style ? `<style>${style}</style>` : ""}
+  <script src="https://cdn.jsdelivr.net/npm/gsap@3/dist/gsap.min.js"></script>
+  <script>
+    window.__timelines = window.__timelines || {};
+    const tl = gsap.timeline({ paused: true });
+    ${script}
+    window.__timelines["c1"] = tl;
+  </script>
+</body></html>`;
+
+  it("errors when a late fromTo jumps from a non-identity from-var on a visible element", async () => {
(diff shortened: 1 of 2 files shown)


miguel-heygen:
Closing after checking current main and GSAP 3.15.0. A default late fromTo applies its from-state when the paused timeline is first seeked to zero, so the proposed rule would flag healthy animations. Its tl.set-at-zero recommendation also conflicts with the existing initial-hide warning. No replacement rule is warranted from this patch.

Audit baseline: `723cd0d7853e2ef5cc4f1b2fba12521ed5d9d5bc`. Closing at the owner’s request as part of the pre-September backlog audit; branch history is preserved.

PR #3478, opened 2026-08-25
fix(cli): hide detached Windows spawns

### LINKED ISSUE #3476: [Windows] Three `detached: true` spawns missing `windowsHide` — telemetry flushSync opens a console window on nearly every CLI exit
### Describe the bug

On Windows, console windows pop up during ordinary CLI use — not just during renders. This is a **follow-up to #3379** (ffmpeg, fixed) and is **distinct from #3430** (chrome-headless-shell, still open): those two cover console-subsystem binaries, whereas this one is about `detached: true` spawns of **Node itself**.

Node's docs are explicit about the mechanism:

> On Windows, setting `options.detached` to `true` makes it possible for the child process to continue running after the parent exits. **The child will have its own console window.**

`child_process.spawn()` defaults `windowsHide` to `false`, so any `detached: true` spawn opens a visible console window on Windows unless `windowsHide: true` is passed.

### Root cause

`packages/cli` v0.8.13 has **four** `detached: true` spawns. Three are missing `windowsHide: true`:

| Source | What it spawns | `windowsHide` |
|---|---|---|
| `src/telemetry/transport.ts` (`flushSync`) | `node -e "fetch(...)"` → PostHog | ❌ missing |
| `src/utils/openBrowser.ts` (`openBrowser`) | user's browser | ❌ missing |
| `src/commands/previewLifecycle.ts` | background preview server | ❌ missing |
| `src/utils/autoUpdate.ts` | update check | ✅ present |

`autoUpdate.ts` already gets this right, which makes the other three look like oversights rather than intent.

### The worst offender: telemetry `flushSync`

`flushSync()` is the **process-exit** flush path. Because it can't `await` an HTTP POST during shutdown, it hands the payload to a detached Node process to deliver after the parent dies:

```js
function flushSync() {
  const payload = buildPayload(eventQueue);
  if (payload == null) return;
  eventQueue = [];
  try {
    const child = spawn(
      process.execPath,
      ["-e", `fetch(${JSON.stringify(`${POSTHOG_HOST}/batch/`)},{...})`],
      { detached: true, stdio: "ignore" }   // <-- no windowsHide
    );
    child.unref();
  } catch {}
}
```

The strategy itself is sound — the only thing missing is `windowsHide: true`.

Impact is amplified by how the CLI is actually used. Agent-driven workflows invoke many short-lived commands (`check`, `lint`, `keyframes`, `capture`, `snapshot`, `beats`, `preview`), and **each invocation that exits with queued events opens its own console window**. My `~/.hyperframes/config.json` shows `commandCount: 1350`, which is a lot of windows. Unlike #3430 this isn't render-only — it fires on nearly every command, including read-only ones.

### Steps to reproduce

1. On Windows, ensure telemetry is on (default): `telemetryEnabled: true` in `~/.hyperframes/config.json`
2. Run any short command, e.g. `npx hyperframes check .` or `npx hyperframes lint .`
3. Watch the desktop/taskbar as the command exits — a console window flashes open
4. Run several commands in sequence; you get one window per exit

Setting `HYPERFRAMES_NO_TELEMETRY=1` (or `DO_NOT_TRACK=1`) makes the flashing stop, which isolates the telemetry spawn as the cause.

For the other two sites: `hyperframes preview` **without** `--foreground` spawns the detached background server (one window), and without `--no-open` spawns the browser via `options.browserPath` (one more).

### Expected behavior

No console window appears for any of these background spawns.

### Actual behavior

A console window opens per detached spawn — in the telemetry case, on nearly every CLI invocation.

### Suggested fix

Add `windowsHide: true` at the three call sites, matching what `autoUpdate.ts` already does:

```js
// src/telemetry/transport.ts
const child = spawn(process.execPath, ["-e", script], {
  detached: true,
  stdio: "ignore",
  windowsHide: true,
});

// src/utils/openBrowser.ts
const child = spawn7(options.browserPath, args, {
  detached: true,
  stdio: "ignore",
  windowsHide: true,
});

// src/commands/previewLifecycle.ts
child = spawn17(execPath, args, {
  detached: true,
  stdio: ["ignore", logFd, logFd],
  env: process.env,
  windowsHide: true,
});
```

`windowsHide` is a no-op on macOS/Linux, so it's safe to apply unconditionally — same rationale as the #3379 fix.

Given that #3379, #3430, and this issue are all the same class of defect, it may be worth a small shared spawn helper that applies `windowsHide: true` by default, plus a lint rule that flags a bare `detached: true` without it.

### Environment

```
hyperframes  0.8.13 (latest)
Node.js      v24.15.0 (win32 x64)
OS           Microsoft Windows 10 Home 10.0.19045 (build 19045)
Install      npx (_npx cache)
```

### Additional context

Verified by static analysis of the shipped `dist/cli.js` in `hyperframes@0.8.13` — all four `detached: true` sites were inspected directly, along with every `windowsHide` occurrence. All ffmpeg/ffprobe spawns correctly carry `windowsHide: true` with the explanatory comment added by the #3379 fix, so that regression has not returned.

Related: #3379 (ffmpeg, closed/fixed) · #3430 (chrome-headless-shell, open)

kvnloo:
## Summary

- pass `windowsHide: true` to telemetry exit flush, custom browser launch, and background preview detached spawns
- cover each real spawn path with a behavior regression test
- match the existing auto-update spawn behavior

Fixes #3476

## Verification

- `bun run --filter @hyperframes/cli test -- src/telemetry/client.test.ts src/utils/openBrowser.test.ts src/commands/previewLifecycle.test.ts` (61 passed)
- `bun run --filter @hyperframes/cli typecheck`
- `bun run lint`
- `bun run format:check`
- `bun run build`
- `git diff --check`
- sabotage check: removing `windowsHide` from `openBrowser` made its new regression test fail

The full `bun run test` is not clean in this environment. Studio has 97 failures caused primarily by unavailable `localStorage`, and the full CLI run has 3 unrelated `skillsManifest.test.ts` lock-pruning failures. The focused tests for all changed paths pass.

Windows real-device validation has not been performed.

## Scope

This does not change ffmpeg handling from #3379 or chrome-headless-shell handling from #3430. It does not add a shared spawn helper or lint rule.

## AI assistance

Implemented and verified with Hermes Agent using Codex. I reviewed the resulting diff and test evidence.

### DIFF
--- a/packages/cli/src/commands/previewLifecycle.test.ts
+++ b/packages/cli/src/commands/previewLifecycle.test.ts
@@ -355,6 +355,11 @@ describe("background preview lifecycle", () => {
 
     expect(result).toMatchObject({ type: "started", port: 3210, pid: 4321 });
     expect(unref).toHaveBeenCalledOnce();
+    expect(spawn).toHaveBeenCalledWith(
+      "/usr/bin/node",
+      expect.any(Array),
+      expect.objectContaining({ detached: true, windowsHide: true }),
+    );
     expect(existsSync(previewSessionPath(projectDir, stateHome))).toBe(true);
   });
 
--- a/packages/cli/src/commands/previewLifecycle.ts
+++ b/packages/cli/src/commands/previewLifecycle.ts
@@ -33,6 +33,7 @@ type SpawnPreview = (
     detached: boolean;
     stdio: ["ignore", number, number];
     env: NodeJS.ProcessEnv;
+    windowsHide: boolean;
   },
 ) => SpawnResult;
 
@@ -219,6 +220,7 @@ function spawnDetachedPreview(
         detached: true,
         stdio: ["ignore", logFd, logFd],
         env: process.env,
+        windowsHide: true,
       },
     );
   } finally {
--- a/packages/cli/src/telemetry/client.test.ts
+++ b/packages/cli/src/telemetry/client.test.ts
@@ -140,13 +140,14 @@ describe("telemetry queue delivery", () => {
     const [execPath, args, opts] = spawnMock.mock.calls[0] as unknown as [
       string,
       string[],
-      { detached: boolean },
+      { detached: boolean; windowsHide?: boolean },
     ];
     expect(execPath).toBe(process.execPath);
     expect(args[0]).toBe("-e");
     expect(args[1]).toContain("render_complete");
     expect(args[1]).toMatch(/[0-9a-f-]{36}/); // event uuid rides along
     expect(opts.detached).toBe(true);
+    expect(opts.windowsHide).toBe(true);
 
     // Queue handed to the child — nothing left for a regular flush.
     const fetchMock = vi.fn(() => Promise.resolve(new Response("")));
--- a/packages/cli/src/telemetry/transport.ts
+++ b/packages/cli/src/telemetry/transport.ts
@@ -155,7 +155,7 @@ export function flushSync(): void {
         "-e",
         `fetch(${JSON.stringify(`${POSTHOG_HOST}/batch/`)},{method:"POST",headers:{"Content-Type":"application/json"},body:${JSON.stringify(payload)},signal:AbortSignal.timeout(${FLUSH_TIMEOUT_MS})}).catch(()=>{})`,
       ],
-      { detached: true, stdio: "ignore" },
+      { detached: true, stdio: "ignore", windowsHide: true },
     );
     // Let the parent exit without waiting for the child
     child.unref();
--- a/packages/cli/src/utils/openBrowser.test.ts
+++ b/packages/cli/src/utils/openBrowser.test.ts
@@ -1,9 +1,31 @@
-import { describe, it, expect } from "vitest";
+import { describe, it, expect, vi } from "vitest";
+
+const spawnMock = vi.hoisted(() =>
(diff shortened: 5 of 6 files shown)


Bortlesboat (commented):
Checked `89205665b1a70cea061c43705ffb44b234655ab1` on native Windows with Node 24.14.0 and Bun 1.3.14. All **61 focused tests** pass: 55 across `previewLifecycle.test.ts` and `openBrowser.test.ts`, plus six in `telemetry/client.test.ts`. Oxlint and oxfmt also pass for all six changed files.

For a fresh checkout, the telemetry suite needs the parser/lint packages and core's generated runtime modules first. After building `@hyperframes/parsers`, `@hyperframes/lint`, and `@hyperframes/core`, it collects and passes without source edits. The initial missing-module errors were checkout preparation, not failures of this patch.

I checked all three modified spawn paths: the added option retains their detached behavior, stdio routing, and existing `unref()` calls, and the preview injection type includes the new option. No issue found within this change.

This supplies Windows test/build evidence, but the focused tests mock process spawning. Actual visible console behavior when launching a real preview or browser remains unverified; I did not repeat the full monorepo suite.

miga-heygen:
Closing — stale draft PR. Please reopen if still relevant.

PR #3479, opened 2026-08-25
fix(producer): trip DE parallel-router circuit breaker on stalls and hangs

### LINKED ISSUE #3441: HF_DE_PARALLEL_ROUTER is on by default and hangs native WSL capture at frame 0 with no output; its circuit breaker cannot fire on a hang
## Summary

The `drawElement` parallel-router experiment (`HF_DE_PARALLEL_ROUTER`) is enabled by default. On WSL2 it hangs native capture at frame 0/300 for ~60 s with **zero console output** — no error, no warning, no progress.

The per-install circuit breaker that should retire the experiment after a bad render never armed across 5 consecutive failed renders. Reading `dist/cli.js` from the published 0.8.10 bundle, that appears structural rather than incidental: **the breaker latches only when the router `"reverted"`, and a hang produces no outcome to latch on.**

## Environment

- hyperframes 0.8.10, installed globally via npm
- WSL2 / Ubuntu on Windows 11
- 300-frame mp4 composition using drawElement capture
- Code analysis below is against the same 0.8.10 npm bundle, read on macOS

## Default-on is unconditional

```js
// dist/cli.js:129066
function isDeParallelRouterEnabled(env) {
  const raw = env.HF_DE_PARALLEL_ROUTER?.trim().toLowerCase();
  if (raw === void 0 || raw === "") return true;   // unset => ON
  return !(raw === "false" || raw === "0" || raw === "off" || raw === "no");
}
```

`shouldPreferParallelDrawElement` (dist/cli.js:129071) gates on worker count, output format, drawElement usage and a RAM floor. There is no install-age, render-count or trial-cohort term — every install with >1 worker doing an mp4 drawElement render is routed.

## Why the safety net never armed

```js
// dist/cli.js:145071
function maybeConsumeDeParallelRouterTrial(routerActive, job, quiet) {
  if (!routerActive) return;
  const outcome = resolveDeParallelRouterOutcome(job);
  if (outcome === void 0) return;                  // <-- a hang exits here
  ...
  const fired = outcome === "reverted";            // <-- only a graceful revert latches
  if (fired) { config.deParallelRouterTrialFired = true; ... }
}
```

`resolveDeParallelRouterOutcome` reads `job.perfSummary?.drawElement?.parallelRouter ?? job.errorDetails?.observability?.capture.deParallelRouter`. A render that stalls and is killed populates neither, so `outcome` is `undefined` and the function returns **before writing the latch**.

Both call sites — dist/cli.js:144855 in the `catch`, and :144868 after success — require `executeRenderJob` to return or throw. A stalled process the user interrupts reaches neither.

**Net effect:** the breaker covers the failure mode that already recovered on its own (verify-fail → revert → continue), and is inert against the one that does not (indefinite silent hang). The experiment stays armed for the next render, indefinitely.

There is a second path on which it cannot arm: when `options.manageDeParallelRouterBreaker !== true`, `deParallelRouterActive` is set to `false`, and the in-source comment at dist/cli.js:144809-144812 states the router still runs but the trip is "neither enforced nor recorded."

## Reproduction

```bash
unset HF_DE_PARALLEL_ROUTER
cd <project> && hyperframes render > /tmp/stall.log 2>&1; echo $?
```

Redirect rather than pipe — a pipe reports the tail's exit status, not the render's.

**Observed:** stalls at frame 0/300 for ~60 s, no console errors, 5/5 renders failed.

**With `HF_DE_PARALLEL_ROUTER=0`:** the same composition renders to completion — 306.6 KB, `beginFrameStalled: false`, verified by extracting and viewing frames rather than trusting the exit code.

**Not a Chrome or WSL capability problem:** raw `chrome-headless-shell --screenshot` works normally in the same environment (9,887 B, exit 0).

**Config after 5 failed renders:** `~/.hyperframes/config.json` contained no `deParallelRouterTrialFired` key.

## Suggested fixes

1. Latch the breaker on *any* non-success of a routed render, not only on `"reverted"`.
2. Arm a pre-render latch that a successful render clears, so a killed process still leaves evidence behind.
3. Emit something during the stall — 60 s of silence gives the user nothing to search for, which is what made this take five renders to localize.

## Possibly related

#3423 reports a drawElement capture path failing at frame 0 where a documented fallback did not trigger. Different flag and a hard error rather than silence, so likely distinct, but the "documented fallback did not happen" shape is the same.

miga-heygen:
## Summary
- Root cause: the per-worker capture calls in `captureFrameRange` (`parallelCoordinator.ts`) take no abort signal of their own and only checked `signal.aborted` *before* starting each frame — a no-op once a worker is already awaiting an in-flight call. On WSL2, the native drawElement/BeginFrame capture call can hang indefinitely at frame 0 with no error. The DE parallel-router's existing stall watchdog (`captureStreamingStage.ts`) does fire `stallController.abort()` after `HF_DE_STALL_MS`, but that abort had no way to reach a worker already wedged inside a hung capture call — so `executeParallelCapture`'s `Promise.all` waited forever, the render hung indefinitely, and the CLI's circuit breaker (which only runs after `executeRenderJob` settles) never got a chance to trip.
- Fix: race each per-frame capture call against the signal actually firing (`raceAgainstAbort`), the same "can't cancel, only race" pattern already used by the sequential capture path. Once the watchdog's abort is observed, the wedged worker rejects, `executeParallelCapture` settles, and the existing pinned-fallback retry / `"reverted"` outcome / circuit-breaker machinery (already correct) runs end to end.
- Also widen the CLI breaker's trip condition (`maybeConsumeDeParallelRouterTrial`) from the literal string `"reverted"` to "not a clean routed success", so any future non-success outcome the observability layer records also latches the breaker instead of silently falling through.
- Added a regression test (`parallelCoordinator-peerAbort.test.ts`) that simulates a worker wedged inside a capture call that never settles; confirmed it hangs/times out without the fix and passes with it.

Closes #3441

## Test plan
- [x] `vitest run packages/engine/src/services/parallelCoordinator.test.ts packages/engine/src/services/parallelCoordinator-peerAbort.test.ts` — 51 passed, including new stall regression test
- [x] `vitest run packages/cli/src/commands/render.test.ts` — 73 passed (all DE parallel-router breaker tests)
- [x] `vitest run packages/producer/src/services/renderOrchestrator.test.ts` — 195 passed
- [x] `bun test packages/producer/src/services/render/stages/captureStreamingStage.test.ts` — 11 passed
- [x] `tsc --noEmit` clean on `packages/engine` and `packages/cli`
- [x] Verified the new test times out (reproducing the hang) when the fix is reverted, and passes with it applied

### DIFF
--- a/packages/cli/src/commands/render.ts
+++ b/packages/cli/src/commands/render.ts
@@ -1314,8 +1314,9 @@ function persistDeParallelRouterTrialFired(): boolean {
 
 /**
  * After a trial-armed render, persist that the router's OWN bet actually
- * failed — its self-verify/generic-failure safety net fired
- * (`deParallelRouter === "reverted"`) — or that the render-count backstop
+ * failed — its self-verify/generic-failure safety net fired (recorded as
+ * anything other than a clean `"routed"`, e.g. `"reverted"`, or a stall/hang
+ * outcome — heygen-com/hyperframes#3441) — or that the render-count backstop
  * (`DE_PARALLEL_ROUTER_TRIAL_MAX_RENDERS`) was reached, so it's never
  * enabled again for this install. A clean "routed" (the render succeeded
  * with no fallback) does NOT consume the trial by itself — the whole point
@@ -1352,11 +1353,19 @@ function maybeConsumeDeParallelRouterTrial(
   const config = readConfigFresh();
   const renderCount = (config.deParallelRouterTrialRenderCount ?? 0) + 1;
   config.deParallelRouterTrialRenderCount = renderCount;
-  // Trip ONLY on an actual fallback. The old trial also tripped at a
-  // 25-render exposure cap, which was sampling logic: bound how long an
-  // experiment force-enables itself. Under a shipped default that would
-  // switch the feature off behind the user's back after 25 good renders.
-  const fired = outcome === "reverted";
+  // Trip on any recorded non-success, not only the literal string
+  // "reverted". The old trial also tripped at a 25-render exposure cap,
+  // which was sampling logic: bound how long an experiment force-enables
+  // itself. Under a shipped default that would switch the feature off
+  // behind the user's back after 25 good renders — so a clean "routed" (no
+  // fallback needed) must NOT trip. But narrowing the positive check to the
+  // single string "reverted" (heygen-com/hyperframes#3441) meant any other
+  // non-success signal the observability layer might ever record — a stall,
+  // a timeout, a future outcome value — would silently fall through to "not
+  // fired" instead of tripping. `outcome` is `undefined`-filtered above, so
+  // by this point it is a real recorded outcome; the only one that means
+  // "no fallback happened" is "routed" itself.
+  const fired = outcome !== "routed";
   if (fired) {
     config.deParallelRouterTrialFired = true;
     // Latch BEFORE attempting persistence — the decision holds for this
--- a/packages/engine/src/services/parallelCoordinator-peerAbort.test.ts
+++ b/packages/engine/src/services/parallelCoordinator-peerAbort.test.ts
@@ -61,4 +61,57 @@ describe("executeParallelCapture peer abort", () => {
       rmSync(root, { recursive: true, force: true });
     }
   });
+
+  // heygen-com/hyperframes#3441: a worker wedged INSIDE a native capture call
+  // (WSL2 hangs the very first drawElement/BeginFrame call at frame 0 with no
+  // error) must actually be unstuck once the caller's `signal` aborts — e.g.
+  // the DE parallel-router stall watchdog in `captureStreamingStage.ts` firing
+  // after HF_DE_STALL_MS. Before this fix, `captureFrameRange` only checked
+  // `signal.aborted` BEFORE starting each frame's capture call, which is a
+  // no-op for a call that is already in flight and never settles — the abort
+  // had no way to reach it, so `executeParallelCapture`'s `Promise.all` (and
+  // therefore the whole render) hung forever, and the CLI circuit breaker
+  // (which only runs after `executeRenderJob` settles) never got a chance to
+  // trip.
+  it("rejects promptly when the signal aborts while a worker is wedged inside a capture call that never settles", async () => {
+    const root = mkdtempSync(join(tmpdir(), "hf-stall-abort-"));
+    // Simulates the native capture call hanging indefinitely (never resolves,
+    // never rejects) — exactly the WSL2 shape from the field report.
+    const captureFrame = vi.fn(() => new Promise<void>(() => {}));
+    const closeCaptureSession = vi.fn().mockResolvedValue(undefined);
+    vi.doMock("./frameCapture.js", () => ({
+      createCaptureSession: vi.fn(
+        async () => ({ workerId: 0, browserConsoleBuffer: [] }) as unknown as CaptureSession,
+      ),
+      initializeSession: vi.fn(async () => {}),
+      captureFrame,
+      captureFrameToBuffer: vi.fn(),
+      captureFrameToBufferPipelined: vi.fn(),
+      closeCaptureSession,
+      getCapturePerfSummary: vi.fn(() => ({ frames: 0 })),
+    }));
+
+    try {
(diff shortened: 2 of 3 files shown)


miguel-heygen (approved):
**Verdict:** APPROVE

**Reasoning:** The stall watchdog can now actually release a worker already awaiting each of the three per-frame capture entrypoints. The stage keeps its watchdog controller separate from the parent cancellation signal, rewrites a watchdog abort into a non-cancellation stall error, and therefore reaches the existing pinned fallback / `reverted` outcome / CLI latch. Listener cleanup and late settlement of the abandoned capture promise are handled. I found no code blocker at this exact head.

The final Windows check was still pending at review time and remains an independent landing gate. No merge action.

— Magi

terencecho (approved):
Concur with @magi-bot APPROVE at `68028bf9`. Independent re-verification at head:

**End-to-end mechanism traced clean:** stall watchdog (`captureStreamingStage.ts:664-669` — separate controller from parent, load-bearing since `shouldRetryViaPinnedFallback` returns false on `isCancellation`) → `workerSignal = AbortSignal.any([parent, stall])` (`parallelCoordinator.ts:924-926`) → `raceAgainstAbort` at all three entrypoints (`parallelCoordinator.ts:539,586,591`) → workers reject → `captureFailure` classifies non-fatal cancel → `captureStreamingStage` rewrites to stall (`captureStreamingStage.ts:739-745`, gated on `abortSignal?.aborted !== true`) → renderOrchestrator's `shouldRetryViaPinnedFallback` returns true → `syncCapturePlan` flips to `"reverted"` → CLI sees `"reverted"` → breaker trips. `lastProgressAt` is initialized at `captureStreamingStage.ts:672` BEFORE `executeParallelCapture` — frame-0 covered, no off-by-one.

**Windows CI resolved** since @magi-bot's review: `Render on windows-latest` SUCCESS 04:43:48Z, `Tests on windows-latest` SUCCESS 04:50:54Z. Landing gate is clear.

**Three non-blocking nits (worth a follow-up commit, not a re-review):**

1. **Regression test hits the wrong branch.** `parallelCoordinator-peerAbort.test.ts:76-116` wedges `captureFrame` (`onFrameBuffer=undefined`, `parallelCoordinator.ts:591`), but the WSL2 field-report path is `captureFrameToBuffer` (`parallelCoordinator.ts:586`, `onFrameBuffer` provided). `captureStreamingStage.ts:633,719` always passes `onFrameBuffer`, and Linux/WSL2 is BeginFrame-mode so `workerEncodeEnabled=false` (gated in `frameCapture.ts:1115-1118` on `syncToPaintEvent=true, beginFrameTimeTicks=0` — macOS-hw-GPU-only). The `raceAgainstAbort` wrapping is identical by inspection so the fix works, but a future refactor of the `captureFrameToBuffer` branch could silently regress WSL2 while the test still passes. One-line change: add an `onFrameBuffer: async () => {}` argument to flip the test to the actual path.

2. **CLI-side "widen" at `render.ts:1355-1368` is inert for the WSL2 bug.** The stall path lands `deParallelRouter="reverted"` at `renderOrchestrator.ts:3400` via `syncCapturePlan` after `replanAfterFailure`, and `observability.ts:110-111` types the field as strictly `"routed" | "reverted"` — so `resolveDeParallelRouterOutcome` already returns exactly `"reverted"` under both old and new logic. The widen is future-defense against enum growth, not part of the fix; the commit message and inline comment overstate its role. Worth re-wording so a future stall-not-tripping-breaker debugger doesn't misplace the load-bearing bit (which is the `syncCapturePlan → "reverted"` handoff, not the CLI predicate).

3. **Pipelined-abort path can drop an unhandled rejection.** `parallelCoordinator.ts:538-547`: an abort during the pipelined await throws at 538-541, so the `encodeResult.catch(() => {})` guard at 547 never runs. If `captureFrameToBufferPipelined` later resolves with a value whose inner `encodeResult` later rejects (worker crash / `cleanupDrawElementWorkerEncode` teardown, per the pre-existing comment at `captureStreamingStage.ts:492-495`), that surfaces as an unhandled rejection. macOS-hw-GPU-only, so the WSL2 fix isn't affected — cosmetic teardown noise.

None of the above blocks landing.

— Review by tai (pr-review)

PR #1615, opened 2026-06-20
fix(engine): preserve AAC start time during MP4 mux

miguel-heygen:
## Problem

HyperFrames MP4 renders with audio can show a black first frame in players that honor MP4 edit lists, including QuickTime, Safari, and preview/chat players. The first decoded video frame is real content, but the final muxed container can start the video stream around 21ms after audio, so those players render the leading empty edit as black.

## What this fixes

- Copies HyperFrames' already-mixed AAC sidecar when muxing MP4 and MOV outputs instead of re-encoding it during final mux.
- Uses an explicit `audioCodec: "aac"` mux contract from HyperFrames callers, keeps `.aac` as a fast compatibility fallback, and probes unknown-extension sidecars with `extractAudioMetadata` before deciding copy vs transcode.
- Removes the distributed pad-path asymmetry: short audio now generates only the missing silent AAC tail and concat-copies source AAC + silence instead of re-encoding the whole mixed source.
- Keeps non-AAC MP4/MOV inputs on the existing AAC transcode path.
- Keeps WebM on the existing Opus transcode path.
- Adds mocked engine coverage for the mux command contract and codec-probe fallback, producer pad/trim coverage for the no-source-reencode pad plan, and ffmpeg-backed distributed assemble coverage for both trim and pad paths asserting final audio and video streams start at zero.

## Root cause

`processCompositionAudio` in `audioMixer.ts` writes the mixed render audio as AAC. That first AAC encode owns the encoder-priming interval. Before this PR, `muxVideoWithAudio` re-encoded that AAC sidecar during final MP4 mux with `-c:a aac`; that second AAC encode introduced another priming interval, and ffmpeg preserved timeline alignment by writing an empty video edit-list entry (`media_time: -1`) before the first video packet. Players that render that empty edit show black for the first ~21ms.

The distributed pad path had the same structural risk one step upstream: padding short `audio.aac` used `apad` over the whole source with `-c:a aac`, then the final mux copied that newly encoded AAC through. That path now preserves the single-priming-origin invariant by encoding only generated silence and concat-copying it after the already mixed AAC.

The engine-level mux helper also previously made the AAC copy decision from a filename-only `.aac` check. That was correct for current HyperFrames sidecars but fragile for renamed, extensionless, or AAC-in-container inputs. Unknown extensions now probe codec metadata before choosing whether to copy.

This avoids the priming interval at the source. It does not suppress edit lists with `-use_editlist 0` or mask the symptom after muxing.

### Local checks

[...]

PR #3488, opened 2026-08-25
fix(engine): stop passing -avoid_negative_ts make_zero in muxVideoWithAudio

### LINKED ISSUE #3487: `-avoid_negative_ts make_zero` in muxVideoWithAudio shifts video +21ms and re-introduces the black first frame from #1615 (regression since v0.7.107)
## Summary

Since **v0.7.107**, MP4 renders with audio take the `-avoid_negative_ts make_zero` branch in `muxVideoWithAudio` again on the common path (audio needs no pad/trim → `operation === "copy"`). The flag destroys the AAC priming edit list on the audio track, shifts the entire video timeline forward by one AAC frame (~21 ms), and writes an **empty edit (`media_time = -1`) on the video track**. Players and thumbnailers that honor edit lists (QuickTime, Safari, WhatsApp previews, `requestVideoFrameCallback`) find no video sample at t=0 and show a **black first frame** — the exact symptom #1615 fixed in v0.6.116.

```text
video  start_time=0.020996   <- should be 0.000000
audio  start_time=0.000000
elst (video): media_time=-1, then media_time=0   <- empty edit: nothing presented at t=0
elst (audio): media_time=0                        <- priming trim destroyed
```

## How the regression happened

1. `0e541673e` (#627, shipped v0.4.45) added `-avoid_negative_ts make_zero` to the mux path — described in the commit itself as "belt-and-suspenders" against negative DTS sneaking back in. The concrete bug reported there (negative DTS from B-frames freezing some players) was fixed in the same commit at the source, with `-bf 0` on the encoder.
2. #1615 (v0.6.116, title: "fix(engine): preserve AAC start time during MP4 mux") diagnosed the black-first-frame symptom (`video start_time=0.021029`, empty video edit / `media time: -1`) and fixed it by removing the second AAC encode.
3. `59c56d325` (#2531) added the comment that is still in `chunkEncoder.ts` today: *"`make_zero` discards that edit and shifts copied video forward by one AAC frame (~21ms), creating a visible first-frame offset."*
4. `9289551e9` (v0.7.72, subject: "scope M4A priming preservation to trims") narrowed the skip to `config?.preserveAudioPrimingEditList === true`. Callers originally set that only for `operation === "trim"`; a same-release follow-up (`4b116b988`) broadened them to `operation !== "copy"` so pad also preserves. Either way, the **copy** path — which is dominant because the mixer already pads every input to the exact composition duration — still runs `make_zero`. Safe at the time: the mixer wrote raw ADTS `audio.aac`, which has no edit list to destroy.
5. **#3200 / `dc4383113` (v0.7.107) changed the mixer artifact from `audio.aac` to `audio.m4a`**, precisely so the container would carry the priming edit list — but the v0.7.72 condition was not revisited. Callers (`assembleStage.ts`, `distributed/assemble.ts`) still set `preserveAudioPrimingEditList: normalizeResult.operation !== "copy"`, so `make_zero` runs on virtually every render and discards the very edit list #3200 introduced.

A unit test currently pins the buggy behavior (the "keeps negative-timestamp repair for an M4A without a known priming edit list" case in `packages/engine/src/services/chunkEncoder.test.ts` asserts the flag IS passed).

## Why `make_zero` is never needed for MP4

ffmpeg's default `-avoid_negative_ts auto` means *"enabled when required by the target format"* (`libavformat/mux.c`). The mp4/mov muxers set `AVFMT_TS_NEGATIVE` (`libavformat/movenc.c`), so `auto` resolves to **disabled**: MP4 expresses offset starts natively via edit lists. Passing `make_zero` overrides that deliberate per-format opt-out. (mpegts does not set the flag — that, along with `-ss -c copy` cuts and concat-demuxer prep, is where `make_zero` belongs.)

It is also **not idempotent**: every re-encode + `make_zero` pass stacks another priming interval. Our delivery upscale turned 21 ms into 42 ms — a full frame at 24 fps.

And the scenario the flag guards against doesn't materialize: muxing B-frame video with genuine leading negative DTS into mp4 comes out correct with the default (`start_time 0.000`, clean edit list) and broken with `make_zero` (both tracks shifted off zero plus the empty edit). Related ffmpeg-side report of this bug class: https://trac.ffmpeg.org/ticket/10380

## Minimal reproduction

(ffmpeg 8.1.2)

```bash
# video-only mp4, no B-frames (like the hyperframes encoder output)
ffmpeg -f lavfi -i testsrc2=duration=2:rate=24 -c:v libx264 -bf 0 -pix_fmt yuv420p video.mp4
# AAC in m4a at 48 kHz — carries a priming edit list, like the mixer artifact since v0.7.107
ffmpeg -f lavfi -i sine=frequency=440:duration=2:sample_rate=48000 -c:a aac -ar 48000 audio.m4a

# the copy path, as muxVideoWithAudio builds it today
ffmpeg -i video.mp4 -i audio.m4a -c:v copy -c:a copy -movflags +faststart -avoid_negative_ts make_zero broken.mp4
# identical, flag omitted (ffmpeg default `auto` = disabled for mp4)
ffmpeg -i video.mp4 -i audio.m4a -c:v copy -c:a copy -movflags +faststart ok.mp4

ffprobe -v error -select_streams v:0 -show_entries stream=start_time -of default=nw=1 broken.mp4   # start_time=0.020996
ffprobe -v error -select_streams v:0 -show_entries stream=start_time -of default=nw=1 ok.mp4       # start_time=0.000000
```

(`ffprobe -v trace broken.mp4` also shows video `media time: -1` then `media time: 0`, and audio `media time: 0` — priming edit destroyed. The source `audio.m4a` itself has `media time: 1024`.)

## Suggested fix

Drop `-avoid_negative_ts` from `muxVideoWithAudio` for mp4/mov outputs entirely (keep `-bf 0` at the encoder, which addresses negative DTS at its source) — or at minimum invert the gate's polarity: skip the flag by default and apply it only for output formats that actually require timestamp shifting. Happy to open a PR if that helps.

## Impact

Every mp4 render with audio on the no-pad/trim path ships with the offset (v0.4.45–v0.6.115, and again since v0.7.107). In our production corpus (rendered on 0.6.13), 121 of 121 finals probe at `video start_time=0.021` and all of them show a black poster frame when shared to WhatsApp.

---

*Investigated and written with Claude Code (Claude Fable 5); the reproduction and the ffmpeg-source claims were verified by hand before filing.*

mincua:
Fixes #3487

## What

`muxVideoWithAudio` no longer passes `-avoid_negative_ts make_zero` — in any mode, for any container. One flag removed from the mux path; the `preserveAudioPrimingEditList` option that used to gate it is deprecated and ignored rather than deleted.

## Why

The flag is actively harmful on the dominant mux path (an AAC sidecar copied into mp4), and it is not needed on any of the others.

**The defect.** ffmpeg's default is `-avoid_negative_ts auto`, i.e. *"enabled when required by the target format"*. The mp4/mov muxers set `AVFMT_TS_NEGATIVE` (`libavformat/movenc.c`), so `auto` already resolves to **disabled** for the containers this function writes — mp4 expresses an offset start natively through an edit list. Passing `make_zero` explicitly overrides that deliberate per-format opt-out. On the copy path it discards the AAC priming edit list the sidecar encode created, shifts the copied video forward by one AAC frame (~21 ms), and makes the muxer write an **empty video edit at t=0** (`media time: -1`). Players and thumbnailers that honor edit lists — QuickTime, Safari, WhatsApp previews, `requestVideoFrameCallback` — find no video sample at t=0 and render a **black first frame**.

**The regression chain.** The flag was added in v0.4.45 as belt-and-suspenders against negative DTS, a problem the same commit had already fixed at the source with `-bf 0` on the encoder. #1615 (v0.6.116) diagnosed the black-first-frame symptom and fixed it by dropping a second AAC encode. Then:

1. **v0.7.72** narrowed the skip to `config?.preserveAudioPrimingEditList === true`, so the **copy** path — dominant, because the mixer already pads every input to the exact composition duration — ran `make_zero` again. Harmless at the time: the mixer wrote raw ADTS `audio.aac`, which has no edit list to destroy.
2. **v0.7.107** (#3200) changed the mixer artifact from `audio.aac` to `audio.m4a`, precisely so the container would carry the priming edit list — but the v0.7.72 gate was not revisited. The callers still passed `preserveAudioPrimingEditList: operation !== "copy"`, so `make_zero` ran on virtually every render and destroyed the very edit list #3200 had introduced.

**Why omitting the flag is correct rather than merely convenient.** For mp4/mov, omitting it *is* the flag's own documented behavior — `auto` → disabled — so this restores ffmpeg's per-format default instead of imposing a new policy. Genuine leading negative DTS is still handled: the priming interval is carried by the audio track's edit list, and the encoder keeps `-bf 0`, which removes negative DTS at the source. `make_zero` also isn't idempotent — every re-encode pass through it stacks another priming interval (our delivery upscale turned 21 ms into 42 ms, a full frame at 24 fps). And the scenario it was guarding against does not materialize: muxing B-frame video with real leading negative DTS into mp4 comes out correct with the default and broken with `make_zero`.

## How

- `packages/engine/src/services/chunkEncoder.ts` — drop the `copiesContainerizedAac` condition and the `args.push("-avoid_negative_ts", "make_zero")` it guarded. Replaced by a comment recording *why* the flag must not be reintroduced here, so the next person reading the file doesn't re-derive it as a missing safety net.
- `MuxVideoWithAudioOptions.preserveAudioPrimingEditList` is part of the exported engine API, so it stays, marked `@deprecated` and documented as a no-op, rather than being removed. **No breaking change** for external callers; removal can happen in a future major.
- `packages/producer/.../render/stages/assembleStage.ts` and `packages/producer/.../distributed/assemble.ts` — the two internal callers stop passing the option. In `distributed/assemble.ts` the `normalizedAudio` object existed only to carry that boolean alongside the path, so it collapses back to a plain `normalizedAudioPath: string | null`.
- Tests: the case that pinned the buggy behavior (`"keeps negative-timestamp repair for an M4A without a known priming edit list"`, asserting the flag **is** passed) is inverted into `"never repairs negative timestamps for an M4A sidecar (regression #3487)"`. A `not.toContain("-avoid_negative_ts")` assertion is added to the re-encode case and to the per-container loop, so the flag can't come back for any container without a test failing.

**Deliberately left alone:** `buildEncoderArgs` and `streamingEncoder` still pass `-avoid_negative_ts make_zero` for **video-only** output. Those chunks are intermediates consumed by the concat/mux stages, not delivered mp4/mov, they carry no audio priming edit list to destroy, and the flag's original negative-DTS rationale still applies there. Changing them is out of scope for this fix and would need its own evidence.

## Test plan

**1. Caller-shaped ffmpeg repro** — a 30 fps h264 mp4 plus a 48 kHz AAC `.m4a` sidecar, muxed with exactly the argv `muxVideoWithAudio` builds, with and without the flag. `ffprobe`:

| | with `make_zero` (before) | without (this PR) |
| --- | --- | --- |
| video `start_time` | `0.021029` | `0.000000` |
| video `elst` | `media time: -1, dur 323` **+** `media time: 0, dur 30720` | `media time: 0, dur 30720` |
| audio `elst` | `media time: 0` (priming trim destroyed) | `media time: 1024` (priming preserved) |
| audio `duration` | `2.021333` | `2.000000` |

The empty leading video edit and the offset both disappear, and the audio keeps its 1024-sample priming edit.

**2. Minimal repro from the issue** (24 fps, both 44.1 kHz and 48 kHz) — `start_time` `0.023031` / `0.020996` with the flag, `0.000000` without; the `media time: -1` empty edit is present in each broken case and absent in each fixed one.

**3. Full render through the fixed pipeline** — a real 24 fps mp4 with audio out of the producer: video `start_time=0.000000`, audio `start_time=0.000000`, one edit per track (`media time: 0`), no empty edit.

**4. Unit + static checks** (ffmpeg 8.1.2, bun 1.3.14, macOS):

```
bun run --filter @hyperframes/engine test          68 files passed, 1 skipped — 1638 passed, 3 skipped
  └ src/services/chunkEncoder.test.ts              85 passed
bun run --filter @hyperframes/producer test:unit   40 files — 598 passed, 0 failed
bun run --filter @hyperframes/engine typecheck     exit 0
bun run --filter @hyperframes/producer typecheck   exit 0
bun run lint                                       0 warnings, 0 errors (2930 files)
bun run format:check                               all matched files correctly formatted
```

- [ ] Documentation updated (if applicable) — no docs page covers this flag; the deprecation note lives in the JSDoc on `MuxVideoWithAudioOptions`.

---

*Investigated, written, and tested with Claude Code (Claude Fable 5).*

### DIFF
--- a/packages/engine/src/services/chunkEncoder.test.ts
+++ b/packages/engine/src/services/chunkEncoder.test.ts
@@ -402,8 +402,6 @@ describe("muxVideoWithAudio audio codec handling", () => {
       "copy",
       "-movflags",
       "+faststart",
-      "-avoid_negative_ts",
-      "make_zero",
       ...renderProvenanceArgs("/tmp/output.mp4"),
       "-r",
       "30",
@@ -424,7 +422,7 @@ describe("muxVideoWithAudio audio codec handling", () => {
     });
   });
 
-  it("keeps negative-timestamp repair for an M4A without a known priming edit list", async () => {
+  it("never repairs negative timestamps for an M4A sidecar (regression #3487)", async () => {
     const { spawn, calls } = createSpawnSpy();
     vi.resetModules();
     vi.doMock("child_process", () => ({ spawn }));
@@ -442,13 +440,15 @@ describe("muxVideoWithAudio audio codec handling", () => {
     await flushMuxCodecResolution();
     expect(calls).toHaveLength(1);
     expect(calls[0]!.args).toContain("copy");
-    expect(calls[0]!.args).toContain("-avoid_negative_ts");
+    // `make_zero` would discard the sidecar's AAC priming edit list, shift
+    // the copied video forward ~21ms and leave an empty video edit at t=0.
+    expect(calls[0]!.args).not.toContain("-avoid_negative_ts");
 
     emitClose(calls[0]!.proc, 0);
     await expect(muxPromise).resolves.toMatchObject({ success: true });
   });
 
-  it("preserves a known M4A priming edit list instead of shifting copied video", async () => {
+  it("ignores the deprecated preserveAudioPrimingEditList option", async () => {
     const { spawn, calls } = createSpawnSpy();
     vi.resetModules();
     vi.doMock("child_process", () => ({ spawn }));
@@ -459,7 +459,7 @@ describe("muxVideoWithAudio audio codec handling", () => {
       "/tmp/audio.duration-normalized.m4a",
       "/tmp/output.mp4",
       undefined,
-      { audioCodec: "aac", preserveAudioPrimingEditList: true },
+      { audioCodec: "aac", preserveAudioPrimingEditList: false },
       { num: 30, den: 1 },
     );
 
@@ -582,6 +582,7 @@ describe("muxVideoWithAudio audio codec handling", () => {
     expect(calls[0]!.args[calls[0]!.args.indexOf("-c:a") + 1]).toBe("aac");
     expect(calls[0]!.args).toContain("-b:a");
     expect(calls[0]!.args).toContain("+faststart");
+    expect(calls[0]!.args).not.toContain("-avoid_negative_ts");
 
     emitClose(calls[0]!.proc, 0);
     await expect(muxPromise).resolves.toMatchObject({ success: true });
@@ -629,6 +630,10 @@ describe("muxVideoWithAudio audio codec handling", () => {
       if (ext !== ".webm") await flushMuxCodecResolution();
       const call = calls[calls.length - 1]!;
       expect(call.args).not.toContain("-shortest");
+      // Same for every container we mux into: ffmpeg's `auto` default is
+      // already `disabled` for mp4/mov, and forcing `make_zero` breaks the
+      // AAC priming edit list (#3487).
+      expect(call.args).not.toContain("-avoid_negative_ts");
       emitClose(call.proc, 0);
       await muxPromise;
     }
--- a/packages/engine/src/services/chunkEncoder.ts
+++ b/packages/engine/src/services/chunkEncoder.ts
@@ -78,7 +78,13 @@ export interface MuxVideoWithAudioOptions extends Partial<
    * depend on the file extension alone.
    */
   audioCodec?: "aac";
-  /** Preserve a priming edit list known to have been created by AAC re-encoding. */
+  /**
+   * @deprecated No longer used. `-avoid_negative_ts` is never passed for
+   * mp4/mov muxing (ffmpeg's `auto` default already resolves to `disabled`
+   * for those containers), so the AAC priming edit list is preserved
+   * unconditionally and this flag has no effect. See issue #3487. Kept for
+   * source compatibility; it will be removed in a future major.
+   */
   preserveAudioPrimingEditList?: boolean;
   /** Hard cap copied audio to the already-encoded video's exact duration. */
 }
@@ -693,14 +699,18 @@ export async function muxVideoWithAudio(
       args.push("-c:a", "aac", "-b:a", "192k", "-movflags", "+faststart");
     }
   }
-  const copiesContainerizedAac =
-    !isWebm && shouldCopyAudio && config?.preserveAudioPrimingEditList === true;
-  // PTS bases can diverge during mux and reintroduce negative DTS. See
-  // buildEncoderArgs for the full reasoning on why that breaks playback.
-  // A freshly encoded M4A is the exception: its edit list already hides the
-  // AAC priming packet. `make_zero` discards that edit and shifts copied video
-  // forward by one AAC frame (~21ms), creating a visible first-frame offset.
-  if (!copiesContainerizedAac) args.push("-avoid_negative_ts", "make_zero");
+  // No `-avoid_negative_ts` here, in any mode. ffmpeg's default is `auto`,
+  // which the mp4/mov muxers (AVFMT_TS_NEGATIVE) already resolve to
+  // `disabled` — the correct behavior for the containers this function
+  // writes. Passing `make_zero` explicitly overrides that default and, on the
+  // dominant audio-copy path, discards the AAC priming edit list the sidecar
+  // encode created: the video start_time shifts forward one AAC frame
+  // (~21ms) and the muxer writes an empty video edit at t=0, which
+  // edit-list-honoring players (QuickTime/Safari) show as a black first
+  // frame. See issue #3487. The video-only encoder args (buildEncoderArgs)
+  // still pass the flag deliberately — those chunks are consumed as raw
+  // elementary output, not as a delivered mp4/mov.
+  //
   // Re-assert provenance here: this stage re-muxes into the delivered
   // container, and the mp4 muxer drops the encode stage's tags without the
   // use_metadata_tags flag that appendRenderProvenanceArgs adds.
--- a/packages/producer/src/services/distributed/assemble.ts
+++ b/packages/producer/src/services/distributed/assemble.ts
@@ -302,10 +302,7 @@ export async function assemble(
     }
 
     // ── 3. Audio: pad-or-trim then mux ────────────────────────────────────
-    let normalizedAudio: {
-      path: string;
-      preserveAudioPrimingEditList: boolean;
-    } | null = null;
+    let normalizedAudioPath: string | null = null;
     if (audioPath !== null && existsSync(audioPath)) {
       const paddedAudioPath = join(workDir, "audio-padded.m4a");
       const padTrimResult = await padOrTrimAudioToVideoFrameCount({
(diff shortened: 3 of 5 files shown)


miga-heygen (commented):
## R1 at `a4bea52` — APPROVE (no merge action from me)

**Root cause analysis is correct.** For mp4/mov containers, ffmpeg's `auto` mode for `-avoid_negative_ts` already resolves to `disabled` (the mp4/mov muxers set `AVFMT_TS_NEGATIVE`). Passing `make_zero` explicitly overrides that default and, on the audio-copy path, discards the AAC priming edit list the sidecar encode created — shifting the video `start_time` forward by one AAC frame (~21ms) and writing an empty video edit at `t=0`, which edit-list-honoring players (QuickTime/Safari) render as a black first frame.

**Verified clean across all mux paths:**

1. **Video-only chunk encoders** (`buildEncoderArgs` in `chunkEncoder.ts:416`, `buildStreamingArgs` in `streamingEncoder.ts:372`) correctly retain `make_zero` — those produce raw elementary output consumed by the concat/mux pipeline, not delivered mp4/mov containers.

2. **WebM muxing** — already skipped `make_zero` before this PR (the old conditional was guarded by `!isWebm`). The new container-agnostic assertion at `chunkEncoder.test.ts:630` now covers all formats uniformly.

3. **Audio re-encode path** (non-copy, when original was not AAC) — previously the only path that set `preserveAudioPrimingEditList: true`. Removing `make_zero` unconditionally is still correct here: the M4A sidecar's re-encoder also creates a priming edit list, so the fix preserves it for both copy and re-encode.

4. **All callers of `muxVideoWithAudio`** updated: `assembleStage.ts` (in-process render) and `distributed/assemble.ts`. Both verified — no other production call sites exist (checked via grep).

**`preserveAudioPrimingEditList` deprecation** — clean pattern: `@deprecated` JSDoc, option kept on the interface for source compatibility, value no longer read. All production callers already stop passing it in this PR.

**Test coverage is strong:**
- Negative assertion on `-avoid_negative_ts` for every container format (mp4, mov, WebM — line 630)
- Renamed test cases clearly document the regression narrative
- Deprecated-option test verifies no-op behavior

**One non-blocking note:**
`streamingEncoder.ts:370-371` references "See chunkEncoder for the full explanation; same playback compatibility class." The chunkEncoder comment has now changed from explaining the flag's use to explaining its *removal* in the mux path. The cross-reference could confuse a future reader who follows it from the streaming encoder (which still uses the flag) and reads about why the mux path removed it. Consider updating the streamingEncoder comment to be self-contained: "Raw chunk output is not a delivered container — `make_zero` is safe here (no edit list to discard)."

— Miga

jrusso1020:
Closing as already shipped in #3505: audio muxing preserves AAC priming by omitting make_zero, with both internal callers updated. Thanks for the contribution.

Issue #646, opened 2026-05-06
fix(core): scoped getElementById fails with duplicate element IDs across sub-compositions


learntimes:
## Bug Summary

When multiple sub-compositions share the same element IDs (e.g. `<canvas id="gl-canvas">`), the scoped `document.getElementById()` proxy in `wrapScopedCompositionScript()` returns `null` for all but the first sub-composition, causing scripts to crash.

## Root Cause

In [`compositionScoping.ts`](https://github.com/heygen-com/HyperFrames/blob/master/packages/core/src/compiler/compositionScoping.ts), the scoped `getElementById` proxy is implemented as:

```typescript
if (prop === "getElementById") {
  return function(id) {
    var found = target.getElementById(id);  // global DOM lookup → returns FIRST match
    return found && __hfContains(found) ? found : null;  // containment check
  };
}
```

The issue: `target.getElementById(id)` always returns the **first** element with that ID in the entire document. When multiple sub-compositions are inlined (each with their own `<canvas id="gl-canvas">`, `<div id="s1">`, etc.), the containment check `__hfContains(found)` rejects elements belonging to other compositions, returning `null`.

### Why `querySelector`/`querySelectorAll` work correctly

The scoped `querySelectorAll` uses `__hfQueryAll`, which fetches ALL matching elements and filters by containment:

```typescript
var __hfQueryAll = function(selector) {
  // ...
  return Array.prototype.filter.call(
    window.document.querySelectorAll(/* ... */),
    function(node) { return __hfContains(node); }
  );
};
```

This correctly handles ID collisions by filtering from the full result set.

## Reproduction

1. Create a parent composition with two sub-compositions that each use WebGL (e.g. `flash-through-white` and `cinematic-zoom`):
   ```html
   <!-- index.html -->
   <div data-composition-src="compositions/scene1.html"
        data-composition-id="comp-a" data-duration="4"
        data-width="1920" data-height="1080" data-start="0"></div>
   <div data-composition-src="compositions/scene2.html"
        data-composition-id="comp-b" data-duration="4"
        data-width="1920" data-height="1080" data-start="4"></div>
   ```
2. Both sub-compositions contain `<canvas id="gl-canvas">` and call `document.getElementById("gl-canvas")`.
3. Render: `hyperframes render . -o out.mp4 --no-browser-gpu`

**Expected:** Both sub-compositions render their WebGL content.

**Actual:** The second sub-composition fails with:
```
[...]

PR #3494, opened 2026-08-26
fix(core): namespace SVG ids during composition inline to prevent cross-scene collisions

### LINKED ISSUE #3490: Nested compositions that reuse SVG ids bind url(#…) / <use href="#"> to the first scene after inline
### Describe the bug

Two nested scenes that each define their own `#clip`, `#shape`, and `#fx` are legal per file and pass `hyperframes check`. After inline, the assembled document has duplicate IDs. Browser-native references (`clip-path="url(#clip)"`, `<use href="#shape">`, CSS `filter: url(#fx)`) resolve to the first match in document order, so later scenes clip to nothing or paint the earlier scene's colour.

`getElementById` was scoped in #646. Media pipeline collisions were fixed with render keys in #3340. Native `url(#…)` / `href="#…"` were not rewritten, and lint does not flag generic duplicate ids — only (some) media ones.

Catalog blocks already hardcode ids such as `url(#tracing-beam-glow)`, so two instances of the same block hit this without the author duplicating anything by hand.

Ask: rewrite `url(#)` / `href="#"` (and the matching `id`s) per composition on inline, **or** lint generic duplicate ids the way media ids are now covered. Rewrite is the one that makes nested catalog blocks work without unique-prefixing every SVG id.

### Link to reproduction

https://github.com/ArcadeHQ/hyperframes-repros/tree/patch/duplicate-svg-ids

### Steps to reproduce

1. Clone https://github.com/ArcadeHQ/hyperframes-repros and check out `patch/duplicate-svg-ids`.
2. Run `npx hyperframes@latest check` — 0 errors.
3. Run `npx hyperframes@latest snapshot --at 1,3 --no-end --describe false`
4. Open `snapshots/frame-00-at-1s.png` and `snapshots/frame-01-at-3s.png`.

### Expected behavior

- `t=1s` (scene-a): all four rows red
- `t=3s` (scene-b): all four rows blue

### Actual behavior

- `t=1s`: all four rows red (ok)
- `t=3s`: (1) control blue; (2) clip-path absent; (3) `<use>` red (scene-a's shape); (4) CSS filter red (scene-a's flood)

### Environment

```
hyperframes doctor

  ✓ Version          0.8.14 (latest)
  ✓ Memory           16.0 GB total · 4.2 GB available
  ✓ Disk             365.8 GB free
  ✓ Frames cache     /var/folders/bl/c4hk6r7n34g1d81txd99gsgr0000gn/T/hyperframes-extract-cache-501 · 365.8 GB free at /var/folders/bl/c4hk6r7n34g1d81txd99gsgr0000gn/T/hyperframes-extract-cache-501 · default
  ✓ Archive extractor unzip
  ✓ Environment      non-TTY
  ✗ Docker           Not found
```

### Additional context

Related, not the same: #646 (scoped `getElementById`), #3340 (media render keys). #556 was CSS attribute-selector scope across sibling instances.

miga-heygen:
## Summary
- Prefix SVG element ids with the composition's document-unique runtime id during sub-composition inline, so `<clipPath id="clip">`, `<symbol id="shape">`, `<filter id="fx">`, etc. no longer collide once two nested scenes are merged into one render/preview document.
- Rewrite every same-document reference to a renamed id to match: DOM attributes (`href`, `xlink:href`, any `url(#id)` value — `clip-path`, `filter`, `mask`, `fill`, `stroke`, `marker-start/mid/end`, including inside `style`) and the composition's own extracted `<style>` text (both `#id` selectors and `url(#id)` declaration values).
- Renamed elements keep their original id on `data-hf-authored-id` (the same attribute #646 added for the composition root), so an inline script's own `document.getElementById(originalId)` keeps resolving via the existing `__hfGetElementById` scoping shim — the fix doesn't regress the already-fixed getElementById scoping.
- Extracted `selectorIdTokens.ts` from `compositionScoping.ts`'s existing single-id selector scan so the new many-id rewrite reuses the same guarded-region (quote/bracket) logic instead of a second copy.

### Why not the media-id / getElementById approach?
`getElementById` was already scoped per composition in #646, and media pipeline ids got a parallel `data-hf-render-id` attribute in #3340 — both work by adding a side-channel attribute without touching the real `id`. That doesn't work here: `url(#id)` and `href="#id"` are resolved by the **browser's native SVG/CSS engine**, which always binds to the first element in document order carrying that literal `id` attribute. No JS proxy can intercept native resolution, so the `id` attribute itself has to become document-unique.

Closes #3490

## Test plan
- [x] `packages/core/src/compiler/svgIdNamespacing.test.ts` (new, 13 tests): unit coverage for id renaming, `url()`/`href` rewriting across `clip-path`/`filter`/`mask`/`fill`/`stroke`/`marker-*`/`style`, `xlink:href`, and CSS selector/declaration rewriting.
- [x] `packages/core/src/compiler/inlineSubCompositions.test.ts` (new suite): end-to-end repro of the issue — two sibling scenes reusing `#clip`/`#shape`/`#fx` get distinct non-colliding ids that still resolve correctly; the same catalog block used twice in one scene is disambiguated; `getElementById(originalId)` still resolves for an inline script after rename.
- [x] `bun run typecheck` — clean.
- [x] `oxlint` — clean.
- [x] Full existing suites re-run clean: `compositionScoping.test.ts` (50), `htmlBundler.test.ts` (58, exercises `inlineSubCompositions` end-to-end via the preview bundler) — no regressions.

## Review follow-up (compatibility)

Two review blockers changed the semantics from "rename every natively referenced SVG id" to the following:

- **Collision-driven, not just demand-driven.** `namespaceCollidingSvgIds` runs once per assembled document, after every instance is inlined, and renames an id only when the merged document would otherwise carry it more than once. Ids that never collide — including every single-composition project — are byte-for-byte unchanged, so `svg.querySelector("#shape")` / `document.querySelector("#shape")` / `gsap.to("#shape")` keep working exactly as before. Among colliding elements one keeps the authored id (the first in document order that cannot be renamed — top-level document content or a JS-only id — else the first in document order); native `url(#id)` / `href="#id"` resolution binds to that keeper, and every other renamable duplicate gets a document-unique id with its references rewritten. The JS-only gate from the previous revision is kept; its residual (two instances sharing a JS-only id) is documented in `collectNativelyReferencedIds`.
- **Renamed ids stay reachable from author scripts.** The composition script runtime rewrites `#authoredId` to `:is(#authoredId, [data-hf-authored-id="authoredId"])` (a superset, so it still matches an untouched instance) in the scoped `document.querySelector/querySelectorAll`, the GSAP proxy (`gsap.to`, `timeline.to`, `gsap.utils.toArray/selector`), GSAP **array** targets (string entries now resolve through the scoped lookup), and `Element.prototype.querySelector/querySelectorAll` — patched once per document and only when at least one id was actually renamed (discovered from `[data-hf-authored-id][id]` in the compiled DOM), so documents without collisions run untouched natives.
- **Escaped CSS ids.** The shared selector scanner now decodes CSS identifier escapes (`\.` and `\HEX ` forms, per CSS Syntax Level 3) before matching and emits replacements through a `CSS.escape`-equivalent, so `#fx\.1 { … }` for `id="fx.1"` becomes `#scene-b--fx\.1 { … }` and `url(#fx.1)` becomes `url(#scene-b--fx.1)`. The guarded-segment regex also lost its ambiguous `[^\]]` branch (CodeQL backtracking alert).

**Not covered by the selector shim (documented in code):** `Document.prototype` is deliberately not patched — an unscoped document-wide lookup by a third-party library (e.g. `anime({ targets: "#id" })`) cannot know which instance it means and resolves to the first in document order, exactly as before this PR; `Element.prototype.matches/closest` and script selectors using hex escapes at runtime are not rewritten; references a script injects at runtime (`el.setAttribute("clip-path", "url(#x)")` as the only reference) are not pre-scanned.

### Test plan (follow-up)
- [x] `svgIdNamespacing.test.ts` (24): single composition byte-for-byte unchanged with native refs; two non-colliding compositions untouched; JS-only colliding ids kept; keeper/rename rules incl. parent-document and JS-only keepers; nested-scope exclusion; id minting against taken ids; authored-id idempotency; escaped selector and hex-escape rewriting.
- [x] `selectorIdTokens.test.ts` (10, new): CSS identifier decoding (character/hex escapes, `\r\n`, U+FFFD), `CSS.escape` algorithm parity and round-trip, whole-token and guarded-region behaviour.
- [x] `inlineSubCompositions.test.ts` (35): end-to-end regressions that execute the compiled scoped scripts in a real DOM — (i) single composition, native + script references to the same element; (ii) two colliding compositions, native refs and every script lookup form (`document.querySelector`, `svg.querySelector`, `getElementById`, GSAP array/`toArray`) resolve to their own element; (iii) `id="fx.1"` styled via `#fx\.1` and referenced via `filter="url(#fx.1)"` keeps its rule after rename.
- [x] `compositionScoping.test.ts` (54): Element.prototype shim installs only when an id was renamed; element/document/GSAP-array lookups; escaped authored id in a script selector.
- [x] `htmlBundler.test.ts` (58) unchanged and green; `bun run typecheck` (core + producer) clean; `oxlint`, `oxfmt`, and `fallow audit` clean.
- [x] Windows test job: `inlineSubCompositions.test.ts` no longer imports `./htmlBundler` (whose static esbuild import cannot load in a jsdom realm on Windows); it carries a local mirror of the duplicate-host runtime-id assignment instead.


vanceingalls (commented):
R1 adversarial — SVG ID collision namespace pass

*Summary:* DOM walk over `svg[id]`/`svg [id]` renames each declared id to `${sanitizedNamespace}--${originalId}`, records the authored id on `data-hf-authored-id`, and rewrites same-element attribute refs via a `url(#...)` regex + suffix `:href` detector. `<style>` text goes through a postcss pass + a shared `selectorIdTokens` scan (extracted from the existing root-id rewrite so both callers share one quote/bracket state machine).

*Blockers (P0/P1):*
- (none)

*Concerns (non-blocking):*
- *ARIA id-refs not rewritten* — `aria-labelledby`, `aria-describedby`, `aria-controls`, `aria-owns`, `aria-flowto` all carry bare id lists (no `#`), and `<svg><title id=\"chart-title\">` referenced by `aria-labelledby=\"chart-title\"` is standard SVG a11y. Post-rename the id is `scene-a--chart-title`; the aria attr still says `chart-title` and either resolves to another scene's title in the merged doc or dangles. Not a first-order render bug (matches the ticket's `url(#)`/`href` scope) but worth a followup — same class of native-resolver reference that motivated this PR.
- *`[id=\"foo\"]` attribute selectors in `<style>` are silently skipped* — `selectorIdTokens.ts:markUnguardedOffsets` deliberately masks out bracket regions (correct for `#` disambiguation), so `rewriteSvgIdReferencesInCss` never touches `[id=\"clip\"]`. Rare in author CSS, but any composition that reaches for the attribute form instead of `#clip` will silently stop matching after rename. Worth a test asserting current behavior + a doc note.
- *Idempotency not guarded* — a second `namespaceSvgIds(root, ns)` call double-prefixes (`scene-a--scene-a--clip`) AND overwrites `data-hf-authored-id` with the already-namespaced id, destroying the original. `inlineSubCompositions` looks single-pass per instance so this is latent, but no `if (el.hasAttribute(SVG_AUTHORED_ID_ATTR)) skip` guard exists to catch a future re-entry (e.g. nested inline of a fragment that was itself pre-inlined).
- *No golden-frame / render-pixel test* — three unit tests confirm the id map is correct, but nothing asserts \"two scenes reusing `#fx` actually paint their own filter after the compiler ran end-to-end.\" The repro in #3490 is visual; the failure mode is silent misresolution. A single regression-shard scene mirroring the ticket repro would lock the fix in against future refactors of the URL regex.
- *`sanitizeNamespaceSegment` folds special chars to `-` without a collision guard* — `foo!` and `foo?` both become `foo-`, so any two runtime ids that differ only in special chars produce the same prefix. Runtime ids from `assignBundledRuntimeCompositionIds` are compiler-generated and almost certainly alphanumeric, but the sanitizer is a public-ish contract; a defensive test asserting the runtime-id shape would prevent surprise later.

*Verified clean:*
- URL fragment coverage: `url(#id)`, `url(\"#id\")`, `url('#id')`, `url( #id )`, and all funcref attrs (`clip-path`, `filter`, `mask`, `fill`, `stroke`, `marker-start/mid/end`) reach the same regex via the generic attribute walk — no per-attr whitelist to drift.
- Fragment-`href`: `isHrefAttrName` catches `href` and every `foo:href` suffix (covers `xlink:href` regardless of DOM impl), and `rewriteHrefValue` short-circuits on non-`#` values so `https://example.com/#clip` and asset URLs stay untouched (explicit test).
- Longer-id-shadowing: `selectorIdTokens.ts` sorts candidates longest-first and gates on `isSelectorNameChar` boundary, so `#clip` in the id map never eats `#clip2`. Explicit test in both suites.
- No-SVG fast path: `querySelectorAll(\"svg [id], svg[id]\")` returns empty → `idMap.size === 0` → early return, no attribute walk.
- Anonymous-host guard: empty `namespace` → early return with empty map, no mutation. Matches the same guard `scopeCssToComposition` and `wrapScopedCompositionScript` apply — consistent with existing scoping-primitive contract.
- `getElementById(originalId)` for author scripts: `data-hf-authored-id` is set on every renamed element; `__hfGetElementById` shim (from #646) already checks this attr as fallback. Explicit test asserts `scene-a` root still resolves `symbol` by authored id `shape` after rename.
- `SVG_AUTHORED_ID_ATTR` shares the exact string constant (`\"data-hf-authored-id\"`) with `AUTHORED_ROOT_ID_ATTR` in `compositionScoping.ts` — the shim's fallback path already reads it, so no third attr introduced.
- Shared scanner extraction: `replaceAuthoredRootIdSelectors` is now a thin wrapper over `replaceSelectorIdTokens`; behavior preserved (single-form → one-element candidate list), one state machine to maintain instead of two.
- Perf: single tree walk + O(N) attr scan + one postcss parse per composition — no visible O(N²).

*CI:* Preflight/lint/format/typecheck/unit/producer-integration/SDK/perf-drift/parity/fps/load/scrub/preview-parity/Fallow all green. `regression-shards` shards 1-9, `Smoke: global install`, `Render/Tests on windows-latest`, `Test`, `Analyze (javascript-typescript)` still pending — the render-parity signal is exactly what would exercise the visual repro so worth waiting on before merge.

Signature: — Via

terencecho (commented):
**COMMENT** — R1 SVG id namespacing @ `75b84322586c06a4f8b9c84415a9e9a8465639a6`. No P0/P1 blockers; three orthogonal follow-ups. Holding off APPROVE only because regression-shards (esp. shard-5 which contains `style-17-prod`) and preview-parity are still in-flight and are the load-bearing signals for both this fix AND the regression it patches.

**Concurring with @vibook-bot's R1** (unchanged at head):
- ARIA id-refs (`aria-labelledby` / `-describedby` / `-controls` / `-owns` / `-flowto`) still not tracked. Same class as `url(#)` / `href="#"` — native browser resolution — worth a follow-up.
- `[id="foo"]` attribute-selector inside `<style>` intentionally skipped by the bracket-guard mask.
- No idempotency guard on `SVG_AUTHORED_ID_ATTR` — a re-entry double-prefixes and clobbers the original.
- No golden-frame test locking the #3490 repro; leaning on regression-shards + preview-parity to catch it.

**Independent concerns (differentiated):**

1. **Regression-fix trade-off worth documenting in-code.** The `75b8432` pivot to "only rename ids that have a native `url()`/`href` reference" is the right call for `style-17-prod` (GSAP → `document.querySelector` bypasses the scoped shim, so a renamed id is unreachable). But it introduces a residual cross-comp collision path: two composition instances that both animate the SAME JS-only id (e.g. two catalog scenes each doing `tl.to("#cut-1")`) now BOTH keep `id="cut-1"` → merged doc has two `id="cut-1"` → scene B's GSAP targets scene A's element — the exact document-order-resolution bug this PR fixes, now residual for the JS-only subset. Sound trade-off (less common than the always-broken JS case), but the module doc should call it out so a future reader doesn't broaden the pre-scan and silently re-break `style-17`. Suggest a comment near `collectNativelyReferencedIds` naming this residual class.

2. **Test gap paired with #1.** No test asserts the two-comps-sharing-JS-only-id case — either as the "we accept this collision" contract or a scoped mitigation. Would lock in the current design.

3. **Runtime-injected `url(#id)` isn't pre-scanned.** A `<script>` that later does `el.setAttribute("clip-path", "url(#foo)")` as the ONLY reference to `#foo` misses the pre-scan → `foo` isn't renamed → same-class collision if duplicated across comps. Unlikely in HF composition authoring, but a known blind spot.

**Verified clean at `75b8432`:**
- `url(#..)` (bare/quoted/single-quoted/spaced), all funcref attrs (`clip-path`/`filter`/`mask`/`fill`/`stroke`/`marker-*`), inline `style` attr, and `<style>` text content — all routed through the same `URL_HASH_REF_RE`. No per-attr whitelist to drift.
- `href` / `foo:href` (covers `xlink:href` regardless of DOM impl); non-fragment hrefs (`https://example.com/#clip`) untouched, explicit test.
- Pre-scan scope is per-comp (`namespaceSvgIds(innerRoot ?? contentDoc, ns)`) — no cross-comp poisoning path.
- Multi-instance same-src block: `assignBundledRuntimeCompositionIds` gives each instance a distinct runtime id → distinct namespace. End-to-end test locks it.
- `data-hf-authored-id` reuses the exact string constant from `compositionScoping.ts` — `__hfGetElementById` fallback already checks it, no third attr introduced.
- Shared `selectorIdTokens.ts` scanner: `replaceAuthoredRootIdSelectors` is now a thin wrapper — one state machine to maintain, longest-first sort preserved (`#clip` never eats `#clip2`).
- `<style>` extraction pipeline: sub-comp `<style>` textContent (including SVG-inline `<style>`) is extracted via `plan.styleSources` and rewritten by `rewriteSvgIdReferencesInCss` before `scopeCssToComposition` scopes it. Ordering is correct (rename → asset URL rewrite → composition scope).

**CI:** many required checks in-progress at snapshot; nothing failing. `regression-shards` (esp. shard-5 containing `style-17-prod`) and `preview-parity` are the load-bearing signals for both the primary fix AND the regression it patches — worth waiting on before merge. Happy to bump to APPROVE once those settle green.

— Review by tai (pr-review)

jrusso1020 (changes requested):
Validated against current main c98d6fbac with the public #3490 fixture in Chrome: the collision reproduces, and this patch restores the missing/incorrect native SVG paint. However, I cannot recommend merging at 75b8432 because renaming also breaks working author code.

- **Blocker — selector compatibility:** packages/core/src/compiler/svgIdNamespacing.ts:203-205 renames any natively referenced ID, even when scripts also use it. With `<svg><path id="shape"/><use href="#shape"/></svg>`, `svg.querySelector("#shape")` now returns null, even in a single composition. Preserving the authored attribute/getElementById shim does not preserve native Element selector APIs or GSAP selector arrays. Add a regression covering native and script references to the same element and preserve those lookups before landing. This is the mixed-use case beyond the previously noted JS-only collision.
- **Blocker — escaped CSS IDs:** svgIdNamespacing.ts:221-225 searches raw IDs and emits unescaped replacement selectors. For `id="fx.1"`, `#fx\.1` stops matching after rename. Match decoded CSS IDs and emit valid escaped selectors; cover an actual styled element whose ID is also referenced natively.

The original issue is worth fixing, but a partial selector shim still introduces regressions. Keeping this PR open for a complete compatibility fix rather than publishing the partial successor explored during this audit.

**Verdict:** REQUEST CHANGES
**Reasoning:** The native rendering improvement is demonstrated, but valid existing selector-based author code and escaped-ID styling regress.

— Codex

miga-heygen:
Addressed both blockers at `0f4830bd0cd85830b3eb6de39fa98f45030a223f` (two new commits on top of `75b8432`, no rewrite): `2b85e607c` (selector scanner: CSS identifier decoding + `CSS.escape` equivalent) and `0f4830bd0` (collision-driven renaming + selector runtime).

**Blocker 1 — selector compatibility**

- Renaming is now collision-driven on the assembled document, not per file: `packages/core/src/compiler/svgIdNamespacing.ts:375` (`namespaceCollidingSvgIds`) builds an id census over the whole document and `:323` (`planRenames`) renames an element only when its id appears more than once; one colliding element keeps the authored id (first unrenamable in document order, else first in document order) so native `url(#)`/`href="#"` resolution still binds to it. A single composition — the `<svg><path id="shape"/><use href="#shape"/></svg>` repro — comes out with its ids byte-for-byte unchanged and no runtime shim installed. The pass moved to after the inline loop in `inlineSubCompositions.ts:460-490`, with nested hosts excluded from their parent's scope. The JS-only gate is kept and its residual is documented at `svgIdNamespacing.ts:198`.
- For genuine collisions, the composition script runtime (`compositionScoping.ts`) rewrites `#authoredId` → `:is(#authoredId, [data-hf-authored-id="authoredId"])` (a superset, so it still matches an untouched instance): `:269` generalised token rewriter, `:340` renamed-id discovery from `[data-hf-authored-id][id]` (cached per document), `:389` applied in the scoped `document` proxy / GSAP proxy path, `:535` GSAP **array** targets now resolve string entries through the scoped lookup, and `:371` installs the same rewrite on `Element.prototype.querySelector/querySelectorAll` once per document — only when at least one id was actually renamed (`:648`). I chose the prototype patch over leaving `element.querySelector` uncovered because it is the only way to reach arbitrary-Element lookups without compiler↔runtime plumbing; it is gated so collision-free documents run untouched natives. `Document.prototype` is deliberately not patched: an unscoped document-wide lookup by a third-party library cannot know which instance it means and resolves to the first in document order, exactly as before this PR. `matches`/`closest` and runtime selectors written with hex escapes are also outside the shim — all of this is stated in the code comments and the PR body.
- Regressions (executing the compiled scoped scripts in a real DOM): `inlineSubCompositions.test.ts:851` (single composition, native + script refs, `svg.querySelector`/`document.querySelector`/`getElementById` all resolve), `:877` (two colliding compositions — native refs and every lookup form incl. GSAP array/`toArray` resolve to their own element), `compositionScoping.test.ts:1078-1135` (shim gating, element/document/GSAP-array lookups, escaped authored id), `svgIdNamespacing.test.ts:45` (byte-for-byte single composition).

**Blocker 2 — escaped CSS ids**

- `selectorIdTokens.ts:74` (`decodeCssIdentifierAt`) reads the `#…` token as a CSS identifier per CSS Syntax Level 3 (`\.` and `\HEX ` escapes, one trailing whitespace, U+FFFD for out-of-range) before comparing with the raw attribute id; `:129` (`escapeCssIdentifier`) is the `CSS.escape` algorithm (Node has no `CSS` global; the repo only used `CSS.escape` in browser runtime code) and `svgIdNamespacing.ts:410` emits replacements through it, so `#fx\.1 { … }` → `#scene-b--fx\.1 { … }`. `url(#fx.1)` fragments keep their original quoting and only swap the id (the prefix is restricted to `[A-Za-z0-9_-]`, so a valid fragment stays valid). The runtime rewriter uses `CSS.escape` when available and matches both the raw and escaped spelling.
- Also fixed the CodeQL alert on `selectorIdTokens.ts:154` by removing the ambiguous `[^\]]` branch from the guarded-segment regex.
- Regressions: `inlineSubCompositions.test.ts:934` (`id="fx.1"` styled via `#fx\.1` and referenced via `filter="url(#fx.1)"` in two colliding compositions — the rewritten rule selector matches the renamed element in a real selector engine, the untouched instance's stylesheet is unchanged, and the author's escaped selector resolves through the scoped proxy), `svgIdNamespacing.test.ts:330-341` (escaped and hex-escaped selectors), `selectorIdTokens.test.ts` (decode/escape parity and round-trip).

**CI follow-up** — head is now `04aa0f931e9184a5fc03072061757a52d6fb6738` (one more commit, `04aa0f931`, tests only): the required `Tests on windows-latest` job was failing on this branch since `31274556e` because `inlineSubCompositions.test.ts` imported `assignBundledRuntimeCompositionIds` from `./htmlBundler`, which statically imports esbuild, and esbuild refuses to load inside a jsdom test realm on Windows (`new TextEncoder().encode('') instanceof Uint8Array is incorrectly false`). The test now carries a small local mirror of that id assignment (`<id>__hf<n>` for duplicate hosts); no behaviour change. `regression` (all 9 shards, incl. shard-5 `style-17-prod`) passed on `0f4830bd0`.

Local results: `svgIdNamespacing` 24, `selectorIdTokens` 10, `inlineSubCompositions` 35, `compositionScoping` 54, `htmlBundler` 58 — all passing; `typecheck` (core, producer), `oxlint`, `oxfmt`, `fallow audit` clean.

— Miga

jrusso1020 (changes requested):
Re-reviewed `04aa0f931e9184a5fc03072061757a52d6fb6738`. The 88 focused scoping/SVG/selector tests pass and all eight required CI contexts are green. The original direct-query, selector-array, and ordinary escaped-ID cases are covered, but two executable regressions remain.

- **[P2] Preserve CSS selector grammar in the runtime shim** — `packages/core/src/compiler/compositionScoping.ts:354-355` registers both raw and escaped spellings of every renamed ID. If an authored ID is `foo.bar`, the valid compound selector `#foo.bar` (id `foo`, class `bar`) gets rewritten as the literal ID `foo.bar`. I executed the compiled wrapper with both elements present: `svg.querySelector("#foo.bar")` returned `scene--foo.bar` instead of `foo`. Only the escaped `#foo\.bar` spelling should target that literal ID. Decode selector tokens rather than matching arbitrary raw ID strings, and cover both selectors side by side.

- **[P2] Preserve inherited SVG references across nested scopes** — `packages/core/src/compiler/svgIdNamespacing.ts:236-240` / `inlineSubCompositions.ts:460-463` exclude child subtrees from parent rewriting, but a child with no local definition does not inherit the parent rename map. Actual `inlineSubCompositions` repro: parent defines/uses red `#paint`; nested child uses `fill="url(#paint)"`; a later top-level SVG defines blue `#paint`. The parent becomes `parent--paint`, while the child remains `url(#paint)` and binds to blue instead of its previous red definition. Preserve the original reference target when a nested child does not shadow it, including extracted child CSS.

**Verdict:** REQUEST CHANGES
**Reasoning:** The earlier cases improved, but these two cases still silently change which element or SVG resource author code references.

— Codex

PR #3503, opened 2026-08-26
fix(engine): preserve source frame identity above 99,999

### LINKED ISSUE #3502: Renders above 99,999 frames silently show the wrong source-video frames (5-digit frame filenames + lexicographic sort)
### Describe the bug

Any render whose frame count exceeds **99,999** silently pairs timeline positions with the
wrong extracted source-video frames. At 30 fps that is everything longer than **55:33**;
at 60 fps, everything longer than 27:46.

There is no error and no warning. The run reports `Render complete`, the output has the
correct duration and frame count, and the file plays. **The timeline is unaffected** —
overlays, text and animations stay correctly timed to the audio; only the frames taken
from the `<video>` source are mismapped. To a viewer this looks like heavy flicker plus
audio drift, which sends you looking in entirely the wrong place.

**Cause.** The extractor writes frames with 5-digit zero padding and the reader lists them
with a plain lexicographic `.sort()`, then uses the *list position* as the frame index:

```js
// writer — ffmpeg output pattern
const framePattern = `${FRAME_FILENAME_PREFIX}%05d.${format}`;

// reader
readdirSync(outputDir)
  .filter((f) => f.startsWith(FRAME_FILENAME_PREFIX) && f.endsWith(suffix))
  .sort()                                        // <-- lexicographic
  .forEach((file, index) => framePaths.set(index, join(outputDir, file)));  // <-- position as index
```

Once the count passes 99,999 the names need six digits and the order breaks apart right
after `frame_10000`. The six-digit names do not form one contiguous block: every five-digit
name is a prefix of ten six-digit names, and `.` (0x2E) sorts before any digit, so the two
groups interleave. Taken verbatim from a live extraction directory:

```
frame_09999.jpg, frame_10000.jpg,
frame_100000.jpg … frame_100009.jpg,
frame_10001.jpg,                      <-- five-digit name slips back in here
frame_100010.jpg … frame_100019.jpg,
frame_10002.jpg, …
```

In `dist/cli.js` this `%05d` + `.sort()` pairing occurs at **three** call sites — fresh
extraction, `extractedFramesFromDirectory`, and cache reuse (`rehydrateCacheEntry`) — so
`--frames-cache-dir off` is not a workaround. Note the inconsistency: the *capture* side of
the same build already uses six digits (`frame_${String(index).padStart(6, "0")}`,
`frame_%06d`). Only source-frame extraction is still on five.

### Link to reproduction

https://github.com/tetsuDFgit/hyperframes-frame-order-repro

### Steps to reproduce

1. Clone the repo above.
2. `PRODUCER_ENABLE_CHUNKED_ENCODE=true npm run render` — without that flag the run dies
   at the encode stage on `ffmpegEncodeTimeout` before producing a file, which is a
   separate problem. The composition is a full-bleed
   `<video>` whose source has its own 1-based frame number burned into every frame,
   420 s at 240 fps = **100,800 frames**, just above the threshold. A second, GSAP-driven
   readout prints the timeline frame number in the corner.
3. In a correct render the two numbers are identical on every frame. Read them at output
   frame 3,000 (they match) and at 12,000 (the burned-in number reads 11,199).
   `tools/verify.sh <output-frame>` checks any position automatically with ffmpeg alone.
4. To confirm the threshold rather than the duration is what matters, drop
   `data-duration` to 400 (96,000 frames) and re-render: every frame matches again.

### Expected behavior

Every output frame shows the source frame belonging to that timeline position — the
burned-in number matches the output frame number for the whole file. If the extracted
frames cannot be mapped reliably, the render fails loudly instead of producing a corrupt
file.

### Actual behavior

Output frames 1–10,000 are correct. From output frame 10,001 the picture jumps to material
from elsewhere in the source and never recovers. No error, no warning, correct duration,
correct frame count, correctly timed overlays.

The sorted order predicts exactly which source frame lands where. For the 58:29.85
composition this was measured on (105,296 frames at 30 fps, hence 5,297 six-digit names),
reproducing the sort gives, for output position `p`:

| position `p` | source frame shown | note |
|---|---|---|
| 1 – 10,000 | `p` | correct — exactly 10,000 of 105,296 |
| 10,001 – 15,826 | six-digit names interleaved with `frame_10001` … `frame_10529` | 530 separate runs, offset drifts down from +89,999 |
| 15,827 – 105,296 | `p − 5,297` | constant offset |

Verified against the rendered file by locating each rendered frame in the source (32x18
grayscale signature, mean/variance normalised to cancel the pc/tv range difference):

```
position    predicted   found       verdict
   3,000        3,000     3,000     exact
   9,000        9,000     9,000     exact
   9,999        9,999     9,998     exact (-1)
  16,000       10,703    10,703     exact
  30,000       24,703    24,704     exact (+1)
  60,000       54,703    54,704     exact (+1)
  90,000       84,703    84,704     exact (+1)
 105,000       99,703    99,705     exact (+2)
```

Every measured point matches the reproduced sort exactly. Sample points inside the
interleaved band (10,001–15,826) were inconclusive in my run because my search window was
too narrow; running the sort in isolation resolves them (position 12,000 takes
`frame_101818`, position 15,826 takes `frame_104816`).

A 44-minute project (79,301 frames) on the same machine and the same version renders
correctly, as do short probe renders (1,200 and 12,000 frames) — consistent with the
threshold being the 100,000th filename rather than duration, worker count or capture mode.

### Environment

```shell
hyperframes doctor

  ✗ Version          0.7.107 → 0.8.4 available
                     Run: hyperframes upgrade
  ✓ Node.js          v24.15.0 (win32 x64)
  ✓ CPU              22 cores · Intel(R) Core(TM) Ultra 9 185H @ 3072MHz
  ✓ Memory           31.4 GB total · 16.1 GB available
  ✓ Disk             422.2 GB free
  ✓ Frames cache     <tmpdir>\hyperframes-extract-cache-u · 422.2 GB free · default
  ✓ Environment      non-TTY
  ✓ FFmpeg           ffmpeg 8.1.1-full_build-www.gyan.dev
  ✓ FFprobe          ffprobe 8.1.1-full_build-www.gyan.dev
  ✓ Chrome           cache: chrome-headless-shell win64-152.0.7928.2
  ✓ Docker           Docker version 29.2.1, build a5c7197


Observed on 0.7.107. I did not upgrade the affected project, but I unpacked the published
**0.8.4** tarball and compared `package/dist/cli.js`: the extraction pattern (`%05d`) and
the lexicographic `.sort()` are **byte-identical** there, so 0.8.4 is affected as well.

Render flags: `-q high -f 30`, `PRODUCER_ENABLE_CHUNKED_ENCODE=true`.
Composition: 3509.85 s, 30 fps, one 1920x1080 H.264 source video, 27 HTML overlays.
```

### Additional context

„Same render, two positions: the overlay stays right, the picture does not."
Image

**Measured on the linked reproduction** (`-q draft`, 100,800 frames), comparing each
rendered frame against the source frame that belongs there and against the one the sort
puts there instead:

```
output frame   3,000    expected  33.94 dB   |  sort predicts frame   3,000  33.94 dB  (same frame)
output frame  10,000    expected  32.56 dB   |  sort predicts frame  10,000  32.56 dB  (same frame)
output frame  12,000    expected  16.13 dB   |  sort predicts frame  11,199  35.20 dB
output frame 100,800    expected  12.72 dB   |  sort predicts frame  99,999  33.98 dB
```

**Suggested fix**

- Pad extraction filenames to the width actually required (or simply `%06d` / `%09d`), and
- sort numerically rather than lexicographically — or better, derive the index from the
  filename instead of from the list position, so ordering can never silently drift.

A guard would also help: if the number of files found does not match the expected frame
count, or if any filename is longer than the pattern implies, fail loudly instead of
rendering a corrupt file. (The existing coverage gate does not catch this case — the file
count is exactly right, only the order is wrong.)

**Workaround**

Keep every render at or below 99,999 frames. For longer material, split the source at a
keyframe, render the halves separately, concatenate the video with `-c:v copy` and mux the
full-length audio once, so no seam artefacts are introduced.

**How this was missed at first**

Checking only that overlays appear at the right times will not catch this — that is how the
broken file initially passed acceptance here. Acceptance needs a second leg: compare the
rendered picture against the source master frame by frame.

**Related**

Not the same as #1047 (stale frames in `snapshot` with multi-composition projects) or
#1043 (black frames when the source is shorter than `data-duration`), though #1043 touches
the same area of the source-video pipeline. #2770 (`fix(producer): align frame coverage
with extraction rounding`, merged 2026-07-26) tightens the expected-frame-count model in
the same code area but addresses off-by-one coverage, not filename ordering.

miguel-heygen:
## Summary

Long renders now keep extracted source-video frames aligned with the timeline after frame 99,999 instead of silently interleaving six-digit filenames into lexical order.

Frame readers derive identity from the numeric FFmpeg ordinal and fail on malformed, zero, duplicate, or non-contiguous sequences. The existing `%05d` writer remains unchanged for cache compatibility, while the shared validation covers fresh extraction, superset slices, and cache reuse.

Fixes #3502.

## Test plan

- `bun run --cwd packages/engine test` (1,649 passed, 3 skipped)
- `bun run --cwd packages/engine typecheck`
- `bun run --cwd packages/engine build`
- Oxlint, Oxfmt, Fallow audit, and `git diff --check`

---

Codex

### DIFF
--- a/packages/engine/src/services/extractedFrameIndex.test.ts
+++ b/packages/engine/src/services/extractedFrameIndex.test.ts
@@ -1,0 +1,81 @@
+import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { basename, join } from "node:path";
+import { afterEach, describe, expect, it } from "vitest";
+
+import {
+  ExtractedFrameSequenceError,
+  extractedFrameIndex,
+  framePathsFromDirectory,
+} from "./extractedFrameIndex.js";
+
+const roots: string[] = [];
+
+function frameDir(): string {
+  const root = mkdtempSync(join(tmpdir(), "hf-extracted-frame-index-"));
+  roots.push(root);
+  return root;
+}
+
+function seed(root: string, ...files: string[]): void {
+  for (const file of files) writeFileSync(join(root, file), file);
+}
+
+afterEach(() => {
+  for (const root of roots.splice(0)) rmSync(root, { recursive: true, force: true });
+});
+
+describe("extractedFrameIndex", () => {
+  it("derives frame identity across the five-to-six-digit boundary", () => {
+    expect(extractedFrameIndex("frame_99999.jpg", "jpg")).toBe(99_998);
+    expect(extractedFrameIndex("frame_100000.jpg", "jpg")).toBe(99_999);
+  });
+
+  it("refuses malformed, zero, and wrong-format frame candidates", () => {
+    expect(() => extractedFrameIndex("frame_bad.jpg", "jpg")).toThrow(ExtractedFrameSequenceError);
+    expect(() => extractedFrameIndex("frame_00000.jpg", "jpg")).toThrow(
+      ExtractedFrameSequenceError,
+    );
+    expect(() => extractedFrameIndex("frame_00001.png", "jpg")).toThrow(
+      ExtractedFrameSequenceError,
+    );
+  });
+});
+
+describe("framePathsFromDirectory", () => {
+  it("maps by the filename ordinal instead of directory or lexical position", () => {
+    const root = frameDir();
+    seed(
+      root,
+      ...Array.from({ length: 10 }, (_, index) => `frame_${index + 1}.jpg`).reverse(),
+      "notes.txt",
+    );
+
+    const paths = framePathsFromDirectory(root, "jpg");
+
+    expect(paths.size).toBe(10);
+    expect(basename(paths.get(8)!)).toBe("frame_9.jpg");
+    expect(basename(paths.get(9)!)).toBe("frame_10.jpg");
+  });
+
+  it("fails loudly when two filenames claim the same numeric frame", () => {
+    const root = frameDir();
+    seed(root, "frame_1.jpg", "frame_00001.jpg");
+
+    expect(() => framePathsFromDirectory(root, "jpg")).toThrow(/duplicate.*frame index 0/i);
+  });
+
+  it("fails loudly instead of shifting later frames across a gap", () => {
+    const root = frameDir();
+    seed(root, "frame_00001.jpg", "frame_00003.jpg");
+
+    expect(() => framePathsFromDirectory(root, "jpg")).toThrow(/missing.*frame index 1/i);
+  });
+
+  it("fails on frame-prefixed malformed candidates but ignores unrelated files", () => {
+    const root = frameDir();
+    seed(root, "frame_00001.jpg", "frame_bad.jpg", "notes.txt");
+
+    expect(() => framePathsFromDirectory(root, "jpg")).toThrow(/invalid.*frame filename/i);
+  });
+});
--- a/packages/engine/src/services/extractedFrameIndex.ts
+++ b/packages/engine/src/services/extractedFrameIndex.ts
@@ -1,0 +1,55 @@
+import { readdirSync } from "node:fs";
+import { join } from "node:path";
+
+export type ExtractedFrameFormat = "jpg" | "png";
+
+export const FRAME_FILENAME_PREFIX = "frame_";
+
+export class ExtractedFrameSequenceError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = "ExtractedFrameSequenceError";
+  }
+}
+
+export function extractedFrameIndex(file: string, format: ExtractedFrameFormat): number {
+  const match = new RegExp(`^${FRAME_FILENAME_PREFIX}(\\d+)\\.${format}$`).exec(file);
+  if (!match) {
+    throw new ExtractedFrameSequenceError(`Invalid extracted frame filename: ${file}`);
+  }
+  const ordinal = Number.parseInt(match[1]!, 10);
+  if (!Number.isSafeInteger(ordinal) || ordinal < 1) {
+    throw new ExtractedFrameSequenceError(`Invalid extracted frame ordinal: ${file}`);
+  }
+  return ordinal - 1;
+}
+
+export function framePathsFromDirectory(
+  outputDir: string,
+  format: ExtractedFrameFormat,
+): Map<number, string> {
+  const suffix = `.${format}`;
+  const indexed = new Map<number, string>();
+  for (const file of readdirSync(outputDir)) {
+    if (!file.startsWith(FRAME_FILENAME_PREFIX) || !file.endsWith(suffix)) continue;
+    const index = extractedFrameIndex(file, format);
+    if (indexed.has(index)) {
+      throw new ExtractedFrameSequenceError(
+        `Duplicate extracted frame index ${index}: ${indexed.get(index)} and ${file}`,
+      );
+    }
+    indexed.set(index, join(outputDir, file));
+  }
+
+  const ordered = new Map<number, string>();
+  for (let index = 0; index < indexed.size; index += 1) {
+    const path = indexed.get(index);
+    if (!path) {
+      throw new ExtractedFrameSequenceError(
+        `Missing extracted frame index ${index} in ${outputDir}`,
+      );
+    }
+    ordered.set(index, path);
+  }
+  return ordered;
+}
--- a/packages/engine/src/services/extractionCache.test.ts
+++ b/packages/engine/src/services/extractionCache.test.ts
@@ -25,6 +25,7 @@ import {
   partialCacheEntryDir,
   publishCacheEntry,
   readKeyStat,
+  rehydrateCacheEntry,
   type CacheKeyInput,
 } from "./extractionCache.js";
 
@@ -73,6 +74,42 @@ describe("extractionCache constants", () => {
 
   it("uses a dotfile sentinel so ls-without-A hides it", () => {
     expect(COMPLETE_SENTINEL.startsWith(".")).toBe(true);
+  });
+});
+
+describe("rehydrateCacheEntry frame identity", () => {
+  it("fails loudly when a complete cache has a frame-number gap", () => {
+    const { tmpRoot } = makeCacheRoot();
+    try {
+      writeFileSync(join(tmpRoot, "frame_00001.jpg"), "one");
+      writeFileSync(join(tmpRoot, "frame_00003.jpg"), "three");
+
+      expect(() =>
+        rehydrateCacheEntry(
+          { dir: tmpRoot, keyHash: "a".repeat(64) },
+          {
+            videoId: "video-gap",
+            srcPath: "/video.mp4",
+            fps: 30,
+            format: "jpg",
+            metadata: {
+              durationSeconds: 1,
+              videoStreamDurationSeconds: 1,
+              width: 1920,
+              height: 1080,
+              fps: 30,
+              videoCodec: "h264",
+              hasAudio: false,
+              isVFR: false,
+              hasAlpha: false,
+              colorSpace: null,
+            },
+          },
+        ),
+      ).toThrow(/missing.*frame index 1/i);
+    } finally {
+      removeCacheRoot(tmpRoot);
+    }
   });
 });
 
--- a/packages/engine/src/services/extractionCache.ts
+++ b/packages/engine/src/services/extractionCache.ts
@@ -46,9 +46,10 @@ import {
 } from "node:fs";
 import { join } from "node:path";
 import type { VideoMetadata } from "../utils/ffprobe.js";
+import { FRAME_FILENAME_PREFIX, framePathsFromDirectory } from "./extractedFrameIndex.js";
 
 /** Filename prefix for extracted frames. Shared with the extractor. */
-export const FRAME_FILENAME_PREFIX = "frame_";
+export { FRAME_FILENAME_PREFIX } from "./extractedFrameIndex.js";
 
 /** Sentinel filename written after a cache entry is fully populated. */
 export const COMPLETE_SENTINEL = ".hf-complete";
@@ -508,14 +509,7 @@ export function rehydrateCacheEntry(
   options: RehydrateOptions,
 ): RehydratedFrames {
   const framePattern = `${FRAME_FILENAME_PREFIX}%05d.${options.format}`;
-  const framePaths = new Map<number, string>();
-  const suffix = `.${options.format}`;
-  const files = readdirSync(entry.dir)
-    .filter((f) => f.startsWith(FRAME_FILENAME_PREFIX) && f.endsWith(suffix))
-    .sort();
-  files.forEach((file, idx) => {
-    framePaths.set(idx, join(entry.dir, file));
-  });
+  const framePaths = framePathsFromDirectory(entry.dir, options.format);
   return {
     videoId: options.videoId,
     srcPath: options.srcPath,
--- a/packages/engine/src/services/videoFrameExtractor.ts
+++ b/packages/engine/src/services/videoFrameExtractor.ts
@@ -6,7 +6,7 @@
  * Videos are replaced with <img> elements during capture.
  */
 
-import { copyFileSync, existsSync, linkSync, mkdirSync, readdirSync, rmSync } from "fs";
+import { copyFileSync, existsSync, linkSync, mkdirSync, rmSync } from "fs";
 import { isAbsolute, join, posix, resolve, sep } from "path";
 import { parseHTML } from "linkedom";
 import {
@@ -55,6 +55,7 @@ import {
   type CacheEntry,
   type CacheFrameFormat,
 } from "./extractionCache.js";
+import { framePathsFromDirectory } from "./extractedFrameIndex.js";
 
 export interface VideoElement {
   id: string;
@@ -802,13 +803,7 @@ export async function extractVideoFramesRange(
     );
   }
 
-  const framePaths = new Map<number, string>();
-  const files = readdirSync(videoOutputDir)
-    .filter((f) => f.startsWith(FRAME_FILENAME_PREFIX) && f.endsWith(`.${format}`))
-    .sort();
-  files.forEach((file, index) => {
-    framePaths.set(index, join(videoOutputDir, file));
-  });
+  const framePaths = framePathsFromDirectory(videoOutputDir, format);
   if (framePaths.size === 0 && duration > 0) {
     throw new VideoSourceExtractionError(
       "zero_output",
@@ -1180,13 +1175,6 @@ type SupersetGroupPlan = {
   members: SupersetMemberPlan[];
 };
 
-function extractedFrameFileNames(outputDir: string, format: CacheFrameFormat): string[] {
-  const suffix = `.${format}`;
-  return readdirSync(outputDir)
-    .filter((file) => file.startsWith(FRAME_FILENAME_PREFIX) && file.endsWith(suffix))
-    .sort();
-}
-
 function extractedFramesFromDirectory(
   work: PreparedExtraction,
   outputDir: string,
@@ -1194,10 +1182,7 @@ function extractedFramesFromDirectory(
   fps: number,
 ): ExtractedFrames {
   const framePattern = `${FRAME_FILENAME_PREFIX}%05d.${work.format}`;
-  const framePaths = new Map<number, string>();
-  extractedFrameFileNames(outputDir, work.format).forEach((file, index) => {
-    framePaths.set(index, join(outputDir, file));
-  });
+  const framePaths = framePathsFromDirectory(outputDir, work.format);
   return {
     videoId: work.video.id,
     srcPath,
--- a/packages/producer/src/services/distributed/rebuildExtractedFrames.test.ts
+++ b/packages/producer/src/services/distributed/rebuildExtractedFrames.test.ts
@@ -178,6 +178,34 @@ describe("rebuildExtractedFramesFromPlanDir", () => {
     }
   });
 
+  it("orders mixed-width dense-v1 filenames by numeric ordinal", () => {
+    const planDir = mkdtempSync(join(tmpdir(), "hf-rebuild-frames-v1-mixed-width-"));
+    try {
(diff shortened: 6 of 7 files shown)


vanceingalls (commented):
R1 adversarial review — the in-scope fix is solid; flagging one sibling-path scope gap.

## Fix — correctness

The parse-by-ordinal helper (`extractedFrameIndex.ts`) is the right shape. Points I verified adversarially:

- **Boundary math** — `frame_99999.jpg` → index 99_998 and `frame_100000.jpg` → index 99_999 tests pin down the exact off-by-one that would otherwise re-introduce the bug on refactor. `Number.isSafeInteger` guards ludicrously wide digit runs; `< 1` rejects zero-based writers.
- **Backward compat** — the writer is unchanged (`frame_%05d.${format}` — ffmpeg's `%05d` grows past 5 digits, doesn't wrap). New reader accepts 5-digit *and* 6-digit filenames since it parses the digit run rather than the padded width. Existing caches on disk keep working.
- **Missing-frame detection** (`framePathsFromDirectory` lines 44–52) iterates `0..indexed.size-1` and requires every slot filled — a hole at index N with a later frame present raises "Missing extracted frame index N" rather than silently truncating. Covered by test.
- **Duplicate detection** — `frame_1.jpg` + `frame_00001.jpg` both parse to index 0 and throw. Extractor's own writer always pads to 5 so this is defensive, but the assertion is right.
- **Sentinel/other files** — the `startsWith(FRAME_FILENAME_PREFIX) && endsWith(suffix)` filter leaves `.hf-complete` and stray files alone.

Tests hit semantics (specific index values, specific error strings), not just presence — mutation-resistant.

## Scope gap — worth a follow-up

`packages/producer/src/services/distributed/renderChunk.ts:361-378` (`rebuildExtractedFramesFromPlanDir`) has the **identical lex-sort assumption on the same source-video frames** for the `dense-v1` index mode (the parameter default, and the branch taken whenever `v2Manifest === null` at renderChunk.ts:665). The comment above the block still asserts:

> Sorted-by-name matches sorted-by-frame-index because the extractor writes zero-padded monotonic indices.

That claim is exactly what this PR contradicts for ≥100_000 frames. Distributed renders without a v2 manifest that consume source video ≥100_000 extracted frames will silently produce the same misaligned identity bug you're fixing here.

Recommendation: either (a) route `dense-v1` through `framePathsFromDirectory` too, or (b) if `dense-v1` is being retired, document that plan and add a runtime guard rejecting `frames.length >= 100_000` under `dense-v1`.

Not blocking this PR — the local + cache-rehydrate paths are the ones the ticket targets, and the fix is clean. But worth pulling into an immediate follow-up so the invariant "source-frame identity is derived from ffmpeg ordinal, never from lex position" holds monorepo-wide.

— Via

james-russo-rames-d-jusso (commented):
🟢 Real fix, targeted diff, good defensive test set. No blockers. One meaningful adjacent-scope concern to surface: the same lexical-sort bug lives in the distributed producer's `dense-v1` rehydrate path — outside this PR's scope, worth a follow-up ticket. Details below.

## What I verified (HEAD `a290fdfd`)

- **Reader is ordinal-derived, not width-hardcoded.** `extractedFrameIndex` at `packages/engine/src/services/extractedFrameIndex.ts:15-25` parses `frame_(\d+)\.<format>$` — any digit count parses cleanly, so `frame_99999.jpg` and `frame_100000.jpg` land at indexes 99998/99999 (verified by the test at `extractedFrameIndex.test.ts:29-32`).
- **Two-pass contiguity check is correct.** First pass collects into `indexed` and catches duplicates; second pass iterates `0..indexed.size-1` and catches gaps (`extractedFrameIndex.ts:32-53`). I walked several gap shapes (1,3,4,5 → gap-1; 2,3,4 → missing-0; 1..5 + 7 → gap-6) mentally — all fail as expected.
- **Writer format unchanged.** `frame_%05d.<ext>` still emitted at `videoFrameExtractor.ts:700` (direct extract) and `1184` (superset), and `frameFileName` still uses `padStart(5, "0")` at `videoFrameExtractor.ts:1199`. `%05d` is printf-style minimum-width, so ordinals ≥ 100000 spill to 6 digits — which is exactly the bug this PR fixes. Cache dirs written by earlier engine versions remain readable (same emit pattern, new reader tolerates both widths).
- **Round-trip is consistent.** FFmpeg's `-start_number` defaults to 1 (no override anywhere on the extract path — `chunkEncoder.ts:579` is the only `-start_number` site and is on the encode side, different dir). Reader subtracts 1 to produce 0-based indexes. Superset-slice writer at `videoFrameExtractor.ts:1346` re-emits with `frameFileName(i + 1, format)`, matching the 1-based on-disk convention.
- **All three original callers of the sort-based reader are migrated.** `videoFrameExtractor.ts:806`, `videoFrameExtractor.ts:1185`, and `extractionCache.ts:512` all now call `framePathsFromDirectory`. The old `extractedFrameFileNames` helper is deleted. Remaining `readdirSync` sites in `extractionCache.ts` (lines 184/337/448) are unrelated (GC + "has-any-frame" probe — order-independent).
- **No sibling consumers.** Grepped `hyperframes-internal`, `hyperframes-gemini-agent`, `pacific`, `heygen-cli`, `experiment-framework`, `genesis` for `FRAME_FILENAME_PREFIX` / `framePathsFromDirectory` / `frame_%05d` — no hits. This is engine-local naming, no cross-repo contract to break.

## Concerns (🟡)

- **`renderChunk.ts:343` has the same bug on `dense-v1` — out of scope, follow-up ticket.** `rebuildExtractedFramesFromPlanDir` in `packages/producer/src/services/distributed/renderChunk.ts:343-394` is a parallel implementation used by the distributed producer's chunk workers when reading plan-dir frames. Default `indexMode: "dense-v1"` uses `readdirSync().sort()` and maps by sort-position (line 362-374). The comment at 358 even acknowledges the fragile invariant: *"Sorted-by-name matches sorted-by-frame-index because the extractor writes zero-padded monotonic indices."* That invariant breaks at 100k+ frames for the same reason as engine's cache path. `sparse-v2` is fine — it parses ordinals via `/(\d+)(?=\.[^.]+$)/` at line 372. `dense-v1` is chosen when `v2Manifest === null` (`renderChunk.ts:665`), i.e. it's the fallback for plans without a v2 manifest — actively in use for legacy plans. Distributed renders long enough to cross 99,999 source frames (e.g. > ~55 min at 30 fps on `dense-v1` plans) will exhibit the same interleave bug there. Recommend a follow-up PR that either migrates `dense-v1` to the same ordinal-derived path or forces the boundary invariant. Explicitly NOT a blocker for this PR — engine-scoped fix is right — but should not be lost.
- **`planV2.ts:308` uses a sparse-tolerant reader, this one is strict-contiguous.** `listVideoFramePaths` in `packages/producer/src/services/distributed/planV2.ts:308-345` derives ordinals the same way but allows sparse indexes ("Preserve sparse indexes so a materialized chunk can carry only the frames it actually requests" — line 318-319). New `framePathsFromDirectory` requires 0..N-1 contiguous. Different use cases justify the divergence (chunk-worker reads sparse-materialized dirs; engine reads full-extraction dirs), but I'd want a code-adjacent comment or shared prefix parser saying "sparse OK for producer, strict for engine" so a future refactor doesn't accidentally unify them.
- **Fresh-extraction path stays hard-fail even when partial frames were written.** `videoFrameExtractor.ts:806` calls `framePathsFromDirectory` immediately after a *successful* ffmpeg exit. If ffmpeg reports success but skipped an intermediate frame (rare — corrupted keyframe, filesystem hiccup, container quirk), the pre-existing `zero_output` guard at 807-814 only catches size==0. With the new reader, non-contiguous partial extraction now throws `ExtractedFrameSequenceError` — good (previously it silently returned wrong frames), but the caller's classifier (`classifyVideoExtractionError`, called from `extractionError` at 1691) doesn't recognize this error class and will surface it as `unknown_error`/`ffmpeg_failed` shape rather than something the retry loop understands. Worth a quick check that `runVideoExtractionWithRetry` doesn't loop forever on a genuinely unretriable "your ffmpeg dropped a frame" case, and that operators get a clean diagnostic string.
- **Cache-hit hard-fail can crash a render on legacy corruption.** `rehydrateCacheEntry` throws → `rehydratePublishedCache` returns the throw → `lookupCacheFor` doesn't catch it → `extractAllVideoFrames` bubbles it up. That's correct for surfacing bad data instead of silently rendering wrong frames, but on rollout, any pre-existing dirty cache entry (partial write from a crashed render, orphan file, etc.) that used to silently work-wrong will now hard-fail the next render that hits it. A graceful-degrade path — catch `ExtractedFrameSequenceError` from `rehydrate*`, evict the entry, log, fall through to a fresh extraction — would keep users unblocked while surfacing the anomaly in logs. Judgment call; fail-loud is defensible. Flag with the on-call so they know cache-related render failures may spike briefly post-deploy.

## Nits

- `new RegExp(...)` allocated per file inside `framePathsFromDirectory`'s loop (via `extractedFrameIndex` at line 16). For a 100k-frame dir that's 100k RegExp allocations. Hoist to a `Map<ExtractedFrameFormat, RegExp>` module-const. Cheap win, harmless.
- Test `refuses malformed, zero, and wrong-format frame candidates` seeds `frame_00000.jpg` as invalid. Correct per FFmpeg's 1-based default, but the assertion is testing something no real extractor writes. Fine defensively; a comment tying it to `-start_number` = 1 would help the next reader.
- Duplication with `packages/producer/src/services/distributed/planV2.ts:308-345` is now more visible. Long term, exporting `extractedFrameIndex` from a shared location that both engine and producer import would consolidate — but the diff scope here is right to defer.

## Tests — what's covered, what's not

Covered:
- Per-file ordinal derivation across the 5→6 digit boundary (`frame_99999.jpg` / `frame_100000.jpg`) at `extractedFrameIndex.test.ts:29-32`.
- Malformed, zero-ordinal, wrong-extension rejections (`extractedFrameIndex.test.ts:34-42`).
- Full-directory ordinal mapping with lexical scramble (10 files, reversed) at `extractedFrameIndex.test.ts:46-58`.
- Duplicate-ordinal rejection via mixed-width filenames (`frame_1.jpg` + `frame_00001.jpg`) at 61-66.
- Gap rejection (`frame_00001.jpg` + `frame_00003.jpg`) at 68-73.
- Frame-prefixed malformed rejection with unrelated files ignored at 75-80.
- Cache-rehydrate gap rejection at `extractionCache.test.ts:79-115`.

Not covered — worth adding at least one:
- **Full-directory mapping across the actual 5→6 digit boundary.** The whole point of the PR. Seed a dir with e.g. `frame_99998.jpg`, `frame_99999.jpg`, `frame_100000.jpg`, `frame_100001.jpg`, run `framePathsFromDirectory`, and assert `basename(paths.get(99999))` is `frame_100000.jpg` (not `frame_99999.jpg` which is what the old lex-sort would have produced). Right now the boundary is tested per-file but not through the map builder — the highest-value regression test for this specific bug is missing.
- Cache-rehydrate happy-path across the boundary. Same idea, at the `rehydrateCacheEntry` level.
- Superset-slice happy-path across the boundary — a `sliceSupersetMember` call where `offsetFrames + i` crosses 99999. Long, but tractable with fake framePaths.

## Adversarial ledger

- **Mixed-width caches.** ✅ Correctly handled. Reader is width-agnostic; writer's `%05d` produces 5+ digits monotonically. A cache written entirely at ≤99999 stays 5-digit; one that extended to ≥100000 has both widths and the ordinal-parser stitches them into the right order.
- **Extremely long renders.** JS `Number.MAX_SAFE_INTEGER` is 2^53-1. `Number.isSafeInteger` gate at line 21 covers overflow. A 24hr render at 30fps is 2.6M frames — well within safe-integer bounds. Not a concern.
- **Ordinal overflow.** Same story — safe-integer gate is present.
- **FFmpeg version behavior differences.** `%05d` in `image2` muxer is documented printf-style minimum-width; I've not experimentally re-verified against the exact ffmpeg version this project pins, but the assumption is standard and consistent with pre-existing engine code. If the pinned ffmpeg ever changed to a fixed-width interpretation, extraction would fail loudly with the new reader (out-of-range ordinal on `frame_00000.jpg` if it wrapped, or a filename that regex-matches something bizarre) — better failure mode than the previous silent misorder. No PR change required.
- **Cache dir with a stale `.hf-complete` sentinel and no frames.** Reader returns an empty map; caller at `videoFrameExtractor.ts:807` catches the zero-output case (though only for direct extraction, not rehydrate). Rehydrate with an empty complete dir would return `totalFrames: 0` — same as before this PR, not a regression. Existing pre-PR sentinel-guard behavior at `dae5b7b90` handles that.
- **Producer `dense-v1` chunk dir with 100k+ frames.** ❌ Not fixed here (out of scope). See the top concern.
- **Two independent extractions racing into the same dir.** Not addressed by this PR (nor should be — cache directory locking is a separate concern). Duplicate-ordinal detection would surface it as a hard error rather than silent corruption, which is a strict improvement.

## Stamp stance

🟢 LGTM from my side — leaving as a comment. The engine-scoped fix is right, the reader is sound, and the test coverage for the failure modes is thorough (with the one boundary-through-map-builder gap noted above). Miguel decides merge; the follow-up on `renderChunk.ts` `dense-v1` should be a separate ticket, not a blocker here.

— _Review by Rames D Jusso_

james-russo-rames-d-jusso (commented):
*Follow-up verified at HEAD `c45c8a29fc428a629009d443f09f897b432d4d68`* — closes Concern (1) from https://github.com/heygen-com/hyperframes/pull/3503#pullrequestreview-5032431835.

*Delta from `a290fdfd` → `c45c8a29`* (2 files, +46/-6):

- `packages/producer/src/services/distributed/renderChunk.ts` — new helper `frameNumberFromFileName(name): number | null` (regex `/(\d+)(?=\.[^.]+$)/`, safe-int gated) at 327-332. `rebuildExtractedFramesFromPlanDir` now sorts the directory listing with numeric-compare via the helper, falling back to `localeCompare` on unparseable names or as tie-breaker; inline regex previously used only for `sparse-v2` decrement is also refactored onto the helper.
- `packages/producer/src/services/distributed/rebuildExtractedFrames.test.ts` — new test `orders mixed-width dense-v1 filenames by numeric ordinal`: creates `frame_1.jpg` through `frame_10.jpg`, feeds `.toReversed()` to simulate the adversarial lex-sort input, asserts `framePaths.get(8) == frame_9.jpg` and `framePaths.get(9) == frame_10.jpg`. Uses the 1→2 digit boundary — same mechanism as 5→6, cheaper to construct.

*What this closes*
- `dense-v1` distributed renders past the width boundary (frame > 99,999 with `%05d` writer) now stitch into a contiguous ordinal-ordered sequence rather than the interleaved lex-sort. Semantics for the well-formed monotonic-width case are byte-identical (numeric-sort is stable and matches lex-sort when widths align).
- `sparse-v2` semantics preserved — still `frameNumber - 1` decrement (1-based ffmpeg → 0-based index) via the same helper.

*Non-blocker note* — the `localeCompare` fallback when either name can't be parsed means a stray non-numeric file in the frames directory would push the whole sort back to lex among that batch. The `.filter(endsWith(ext))` already narrows the set, so realistic exposure is bounded; noting for the follow-up round rather than this PR.

*Stamp stance* — 🟢 no further concerns from this side. Test claim (`612/612` producer unit + `5/5` focused contract + lint/format/Fallow clean) accepted on trust — the diff itself is safe on inspection.

Merge is @miguel-heygen's call.

— _Review by Rames D Jusso_

vanceingalls (approved):
Delta clean — R1 addressed.

*Delta since prior R1 head* (`a290fdfd`→`c45c8a2`): 2 files, +46/-6 by @miguel-heygen.

*R1 finding verified fixed* — `packages/producer/src/services/distributed/renderChunk.ts`:
- `frameNumberFromFileName()` helper extracts the last numeric run before `.ext` (`(\d+)(?=\.[^.]+$)`), with a `Number.isSafeInteger` guard.
- Sort now numeric with lex tiebreak: `leftNumber - rightNumber || left.localeCompare(right)`; falls back to `localeCompare` when either operand has no trailing digits, so non-numeric names don't throw or crash the sort.
- Dense-v1: positional indexing (`frameIndex = i`) preserved — zero-based v1 (`frame_00000.jpg → get(0)`) still correct because position 0 in numerically-sorted list is the smallest number.
- Sparse-v2: `frameIndex = frameNumber - 1` preserved (1-based ffmpeg → 0-based key), now sharing the same digit-extraction helper as sort.

*Regression test verified* (`rebuildExtractedFrames.test.ts:181`):
- Writes 10 mixed-width filenames (`frame_1.jpg`…`frame_10.jpg`) in *reversed* order → exercises the exact bug shape (variable digit-width in the same directory).
- Asserts specific paths: `get(8) === frame_9.jpg`, `get(9) === frame_10.jpg`. Under the pre-fix lex sort this would resolve to `frame_8.jpg` / `frame_9.jpg` — so the test FAILS pre-fix and PASSES post-fix. Semantics-not-presence, per prior-invariant discipline.
- Existing tests preserved: zero-based v1 (`frame_00000 → get(0)`) still asserted; sparse-v2 mapping (`frame_00021 → get(20)`) still asserted using the shared helper.

*Adversarial pass*:
- (a) Non-frame `.jpg` files: extension filter keeps them, `frameNumberFromFileName` returns null on names without trailing digits, sort falls back to `localeCompare` — no throw, no regression (pre-existing indexing behavior for weird names retained).
- (b) FFmpeg patterns: `%d` produces unsigned decimal; `\d+` regex only matches unsigned integers; `Number.isSafeInteger` guards against pathological >2^53 filenames.
- (c) Dense-v1 count/duplicate gate untouched — `framePaths.size` still equals `frames.length` for the dense mode, so downstream coverage-check semantics unchanged.

CI on `c45c8a2` still spinning at review time (push at 15:50:33Z, WIP green). Approving on fix-verification per Magi's local lane report (producer-unit 612/612, focused contract 5/5, typecheck/lint/format/Fallow clean); will follow up if any lane surfaces a real regression.

— Via

PR #3505, opened 2026-08-26
fix(engine): stop passing -avoid_negative_ts make_zero in muxVideoWithAudio

### LINKED ISSUE #3487: `-avoid_negative_ts make_zero` in muxVideoWithAudio shifts video +21ms and re-introduces the black first frame from #1615 (regression since v0.7.107)

miga-heygen:
## Summary

Drops the unconditional `-avoid_negative_ts make_zero` flag from `muxVideoWithAudio`. ffmpeg's default (`auto`) already resolves to `disabled` for mp4/mov muxers (`AVFMT_TS_NEGATIVE`), so the explicit override was unnecessary — and on the dominant audio-copy path it actively broke things: it discarded the AAC priming edit list the sidecar encode created, shifting the video start_time forward by one AAC frame (~21ms) and inserting an empty video edit at t=0 that edit-list-honoring players (QuickTime/Safari) showed as a black first frame.

The `preserveAudioPrimingEditList` option that partially guarded against this is deprecated (the flag is never needed now) and the distributed assembler no longer threads it.

Takeover of #3488 by @mincua — original work preserved, re-committed with signed commits per org policy.

Closes #3487
Supersedes #3488

— Miga

### DIFF
--- a/packages/engine/src/services/chunkEncoder.test.ts
+++ b/packages/engine/src/services/chunkEncoder.test.ts
@@ -402,8 +402,6 @@ describe("muxVideoWithAudio audio codec handling", () => {
       "copy",
       "-movflags",
       "+faststart",
-      "-avoid_negative_ts",
-      "make_zero",
       ...renderProvenanceArgs("/tmp/output.mp4"),
       "-r",
       "30",
@@ -424,7 +422,7 @@ describe("muxVideoWithAudio audio codec handling", () => {
     });
   });
 
-  it("keeps negative-timestamp repair for an M4A without a known priming edit list", async () => {
+  it("never repairs negative timestamps for an M4A sidecar (regression #3487)", async () => {
     const { spawn, calls } = createSpawnSpy();
     vi.resetModules();
     vi.doMock("child_process", () => ({ spawn }));
@@ -442,13 +440,15 @@ describe("muxVideoWithAudio audio codec handling", () => {
     await flushMuxCodecResolution();
     expect(calls).toHaveLength(1);
     expect(calls[0]!.args).toContain("copy");
-    expect(calls[0]!.args).toContain("-avoid_negative_ts");
+    // `make_zero` would discard the sidecar's AAC priming edit list, shift
+    // the copied video forward ~21ms and leave an empty video edit at t=0.
+    expect(calls[0]!.args).not.toContain("-avoid_negative_ts");
 
     emitClose(calls[0]!.proc, 0);
     await expect(muxPromise).resolves.toMatchObject({ success: true });
   });
 
-  it("preserves a known M4A priming edit list instead of shifting copied video", async () => {
+  it("ignores the deprecated preserveAudioPrimingEditList option", async () => {
     const { spawn, calls } = createSpawnSpy();
     vi.resetModules();
     vi.doMock("child_process", () => ({ spawn }));
@@ -459,7 +459,7 @@ describe("muxVideoWithAudio audio codec handling", () => {
       "/tmp/audio.duration-normalized.m4a",
       "/tmp/output.mp4",
       undefined,
-      { audioCodec: "aac", preserveAudioPrimingEditList: true },
+      { audioCodec: "aac", preserveAudioPrimingEditList: false },
       { num: 30, den: 1 },
     );
 
@@ -582,6 +582,7 @@ describe("muxVideoWithAudio audio codec handling", () => {
     expect(calls[0]!.args[calls[0]!.args.indexOf("-c:a") + 1]).toBe("aac");
     expect(calls[0]!.args).toContain("-b:a");
     expect(calls[0]!.args).toContain("+faststart");
+    expect(calls[0]!.args).not.toContain("-avoid_negative_ts");
 
     emitClose(calls[0]!.proc, 0);
     await expect(muxPromise).resolves.toMatchObject({ success: true });
@@ -629,6 +630,10 @@ describe("muxVideoWithAudio audio codec handling", () => {
       if (ext !== ".webm") await flushMuxCodecResolution();
       const call = calls[calls.length - 1]!;
       expect(call.args).not.toContain("-shortest");
+      // Same for every container we mux into: ffmpeg's `auto` default is
+      // already `disabled` for mp4/mov, and forcing `make_zero` breaks the
+      // AAC priming edit list (#3487).
+      expect(call.args).not.toContain("-avoid_negative_ts");
       emitClose(call.proc, 0);
       await muxPromise;
     }
--- a/packages/engine/src/services/chunkEncoder.ts
+++ b/packages/engine/src/services/chunkEncoder.ts
@@ -78,7 +78,13 @@ export interface MuxVideoWithAudioOptions extends Partial<
    * depend on the file extension alone.
    */
   audioCodec?: "aac";
-  /** Preserve a priming edit list known to have been created by AAC re-encoding. */
+  /**
+   * @deprecated No longer used. `-avoid_negative_ts` is never passed for
+   * mp4/mov muxing (ffmpeg's `auto` default already resolves to `disabled`
+   * for those containers), so the AAC priming edit list is preserved
+   * unconditionally and this flag has no effect. See issue #3487. Kept for
+   * source compatibility; it will be removed in a future major.
+   */
   preserveAudioPrimingEditList?: boolean;
   /** Hard cap copied audio to the already-encoded video's exact duration. */
 }
@@ -693,14 +699,18 @@ export async function muxVideoWithAudio(
       args.push("-c:a", "aac", "-b:a", "192k", "-movflags", "+faststart");
     }
   }
-  const copiesContainerizedAac =
-    !isWebm && shouldCopyAudio && config?.preserveAudioPrimingEditList === true;
-  // PTS bases can diverge during mux and reintroduce negative DTS. See
-  // buildEncoderArgs for the full reasoning on why that breaks playback.
-  // A freshly encoded M4A is the exception: its edit list already hides the
-  // AAC priming packet. `make_zero` discards that edit and shifts copied video
-  // forward by one AAC frame (~21ms), creating a visible first-frame offset.
-  if (!copiesContainerizedAac) args.push("-avoid_negative_ts", "make_zero");
+  // No `-avoid_negative_ts` here, in any mode. ffmpeg's default is `auto`,
+  // which the mp4/mov muxers (AVFMT_TS_NEGATIVE) already resolve to
+  // `disabled` — the correct behavior for the containers this function
+  // writes. Passing `make_zero` explicitly overrides that default and, on the
+  // dominant audio-copy path, discards the AAC priming edit list the sidecar
+  // encode created: the video start_time shifts forward one AAC frame
+  // (~21ms) and the muxer writes an empty video edit at t=0, which
+  // edit-list-honoring players (QuickTime/Safari) show as a black first
+  // frame. See issue #3487. The video-only encoder args (buildEncoderArgs)
+  // still pass the flag deliberately — those chunks are consumed as raw
+  // elementary output, not as a delivered mp4/mov.
+  //
   // Re-assert provenance here: this stage re-muxes into the delivered
   // container, and the mp4 muxer drops the encode stage's tags without the
   // use_metadata_tags flag that appendRenderProvenanceArgs adds.
--- a/packages/producer/src/services/distributed/assemble.ts
+++ b/packages/producer/src/services/distributed/assemble.ts
@@ -302,10 +302,7 @@ export async function assemble(
     }
 
     // ── 3. Audio: pad-or-trim then mux ────────────────────────────────────
-    let normalizedAudio: {
-      path: string;
-      preserveAudioPrimingEditList: boolean;
-    } | null = null;
+    let normalizedAudioPath: string | null = null;
     if (audioPath !== null && existsSync(audioPath)) {
       const paddedAudioPath = join(workDir, "audio-padded.m4a");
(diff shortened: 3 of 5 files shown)


james-russo-rames-d-jusso (commented):
🟢 *Clean, well-scoped fix — merge is Miguel's call.* Drops the unconditional `-avoid_negative_ts make_zero` from `muxVideoWithAudio` on the audio-copy path, which was destroying the AAC priming edit list and producing the 21ms A/V shift + black first frame on QuickTime/Safari (#3487). Removal is correctly bounded to the delivered-mp4/mov mux; unrelated encoder / stream / `-ss` uses of the flag are left alone. Deprecation of `preserveAudioPrimingEditList` is done cleanly (typed as `@deprecated` no-op, callers stripped, ignored-option test added).

*Scope of removal — correctly bounded*
• `packages/engine/src/services/chunkEncoder.ts:702-712` — the flag is gone from `muxVideoWithAudio`, and the new comment explicitly calls out that `buildEncoderArgs` (`chunkEncoder.ts:454`) still emits it deliberately because those chunks are consumed as raw elementary output, not delivered mp4/mov. That distinction is the right one to preserve.
• Verified no other mux/delivery path drops through here: `streamingEncoder.ts:431` still keeps the flag as "belt-and-suspenders against negative DTS at stream start", but that's the fresh video-only image2pipe encode — no audio track, no priming edit list to destroy — so the specific #3487 symptom doesn't apply. Consistent-across-mp4 argument would suggest a follow-up ticket, but explicitly out-of-scope for this PR.
• `skills/media-use/scripts/transcript-cut.mjs` keeps it for `-ss -c copy` transcript cuts, which is precisely where the issue body says `make_zero` legitimately belongs. Correct.

*ffmpeg-internals claim — corroborated*
• The `auto` → `disabled` for mp4/mov via `AVFMT_TS_NEGATIVE` claim rests on `libavformat/mux.c` + `libavformat/movenc.c`, and #3487's body includes a reproducible ffprobe diff (`start_time=0.020996` with the flag vs `0.000000` without) plus a linked upstream trac ticket (10380). Miga's own production evidence — 121/121 finals probing at `video start_time=0.021` — is the strongest single data point. I did not independently open the ffmpeg source; trust-verified via issue evidence.

*Deprecation hygiene*
• `chunkEncoder.ts:81-88` — field kept on the interface, `@deprecated` docstring names the removal timeline ("removed in a future major"). Good middle path: no breaking source-compat change today, no half-shim logic path.
• `assemble.ts:305-341` — `normalizedAudio: { path, preserveAudioPrimingEditList }` correctly collapsed to `normalizedAudioPath: string | null`. The whole reason for the wrapper object was that dropped field; collapse is clean.
• `assembleStage.ts` — same, one-line drop from the option bag.
• Two remaining references at HEAD are the interface field itself + the deprecation test asserting it's ignored. No live callers thread it.

*Test coverage — strong*
• `chunkEncoder.test.ts:394-411` — happy-path exact-args assertion now positively pins the absence of `-avoid_negative_ts` (whole args array literal, not just a `.not.toContain`).
• `chunkEncoder.test.ts:425-448` — new `#3487` regression test on the M4A-sidecar copy path, negative-asserts the flag.
• `chunkEncoder.test.ts:451-471` — deprecated-option test passes `preserveAudioPrimingEditList: false` (which, under the old code, would have pushed the flag) and asserts it's absent — this is a stronger deprecation test than passing `true` would have been.
• `chunkEncoder.test.ts:582-585` — AAC re-encode path also negative-asserts.
• `chunkEncoder.test.ts:625-639` — container-matrix loop asserts across every container the muxer supports. Nice.

*What I didn't verify*
• Did not open the ffmpeg source (`libavformat/movenc.c`, `libavformat/mux.c`) or run the ffprobe reproduction locally — trusting issue #3487's evidence.
• Did not run the tests locally.
• Did not deep-dive `streamingEncoder.ts` — treated it as out-of-scope (video-only fresh encode, no priming edit list at risk).

*CI note (non-blocker)*
The `Test` job under `CI` shows FAILURE, but its only step is `echo "::error::Producer unit/integration tests did not succeed." && exit 1` — the aggregator collapsed because `Detect changes` for that workflow was CANCELLED, causing all producer test shards to show SKIPPED. This looks like spurious infra fallout, not a real regression. The meaningful signal — 9 `regression-shards`, `Tests on windows-latest`, `Render on windows-latest` — was still in-flight at review time. Worth a re-run to see them green.

*Merge stance*
Merge is Miguel's call — HF PR authored by @miga-heygen, no OG-route from my side. From where I sit: fix is correct, well-scoped, well-explained, and well-tested. Ready to ship once CI settles.

— _Review by Rames D Jusso_

jrusso1020 (approved):
**APPROVED.** Stamp basis: Miguel's explicit request in the HyperFrames channel (trusted stamper), which is also what clears the standing "don't stamp a bot-authored PR on its own initiative" rule for a `miga-heygen` PR. Rames D Jusso's pass is on record; the verification below is my own and independent of it.

### I reproduced #3487 rather than reading the comment

The PR's central claim is that `-avoid_negative_ts make_zero` shifts copied video forward by one AAC frame. I ran it on ffmpeg `n8.0.1-52`, muxing a `+0.5s`-offset h264 stream against an AAC sidecar with `-c:v copy -c:a copy -movflags +faststart`:

| output | video `start_time` |
|---|---|
| no flag (ffmpeg `auto`) | `0.500000` |
| `-avoid_negative_ts make_zero` | `0.522982` |

**+22.982ms**, which is 1024 samples at 44.1kHz (23.22ms) — my sidecar's rate. At the 48kHz this pipeline actually encodes, the same frame is 21.3ms, matching the "~21ms" in the issue. So the mechanism is confirmed at source, and `make_zero` is demonstrably the thing introducing the offset, not merely correlated with it.

### The webm path is a verified no-op, not an assumed one

The removed line was `if (!copiesContainerizedAac) …`, and `copiesContainerizedAac` requires `!isWebm` — so webm previously got `make_zero` **unconditionally**, and the new comment justifies removal only via mp4/mov's `AVFMT_TS_NEGATIVE`. That justification does not extend to matroska, so I checked instead of assuming. Muxing a VP8 stream at the same `+0.5s` offset:

```
webm_auto.webm   video 0.500000  audio -0.007000  first video pkt pts 0.500000
webm_zero.webm   video 0.500000  audio -0.007000  first video pkt pts 0.500000
```

Byte-for-byte equivalent timestamps. Removing the flag changes nothing on the webm path, so the widened scope is safe. Worth noting in case a future reader tries to extend the comment's mp4/mov reasoning to matroska — it happens to land in the same place, but not for the stated reason.

### Scope and blast radius

`buildEncoderArgs` correctly keeps the flag (`chunkEncoder.ts:454`) — those chunks are consumed as raw elementary output. I checked the one case where a chunk-encoded file reaches delivery without a mux: in `assemble.ts`, `muxOutputPath` falls back to `postConcatPath` when there is no audio. That path keeps `make_zero`, but with no audio there is no priming edit list, so there is nothing to shift against. No exposure.

The `preserveAudioPrimingEditList` deprecation is done properly: `@deprecated` no-op with the reason and a removal note, both call sites stripped (`assembleStage.ts`, `assemble.ts`), and `assemble.ts` collapses the now-pointless `{path, preserveAudioPrimingEditList}` object back to a bare path rather than leaving a dead field. Tests invert cleanly — the old "keeps negative-timestamp repair" case becomes an explicit `not.toContain` regression test for #3487, plus a container-matrix assertion.

### Non-blocking

- `chunkEncoder.ts:224` still reads "The unconditional `-avoid_negative_ts make_zero` near the bottom of this function covers the mux level". The referent (line 454) still exists so it is not stale, but "covers the mux level" was already loose and is now actively misleading, since the mux level is precisely where the flag no longer is. One-line touch-up whenever this file is next open.

### CI

`Test` is red at `ce875f79`, and I checked whether that is real: `Detect changes` was **cancelled**, which skipped all 22 shards, and the aggregator collapses on skipped dependencies. No individual shard carries a `failure` conclusion — 27 success, 22 skipped, 2 cancelled. A re-run should surface it green. Not gating on it either way.

— Rames

PR #3506, opened 2026-08-26
fix(producer): assert render artifact duration and frame count before commit

### LINKED ISSUE #3395: Artifact validation only checks the file is non-empty, so a truncated render ships as success

miga-heygen:
## Summary

Adds duration and frame-count validation to `ArtifactTransaction.validate()` — a truncated render (e.g. multi-worker dropping ~1/3 of frames) no longer ships as success.

- Probes the output file with ffprobe for duration and frame count
- Rejects when `expectedDuration - probedDuration > tolerance` (tolerance = 1 frame)
- Rejects when `expectedFrames - probedFrames > 1`
- Probe errors and 0/NaN results are treated as validation failures
- Applies to non-PNG-sequence, non-GIF video outputs

Takeover of #3429 by @santhiprakash — original work preserved, re-committed with signed commits per org policy.

Closes #3395
Supersedes #3429

— Miga

### DIFF
--- a/packages/engine/src/utils/ffprobe.ts
+++ b/packages/engine/src/utils/ffprobe.ts
@@ -188,6 +188,11 @@ export interface VideoMetadata {
   hasAlpha: boolean;
   /** Color space info from the video stream. Null if ffprobe didn't report it. */
   colorSpace: VideoColorSpace | null;
+  /** Decoded frame count from the video stream's `nb_frames`. Omitted when the
+   * container does not surface a reliable count (still images, malformed
+   * streams, or muxes that require `-count_packets` to populate). Callers
+   * that gate on a frame count must treat `undefined` as "no answer". */
+  frames?: number;
 }
 
 export interface AudioMetadata {
@@ -731,6 +736,7 @@ export async function extractMediaMetadata(filePath: string): Promise<VideoMetad
           isVFR: false,
           hasAlpha: false,
           colorSpace: stillImageMeta.colorSpace,
+          frames: 1,
         };
       }
       throw new Error("[FFmpeg] No video stream found");
@@ -775,6 +781,14 @@ export async function extractMediaMetadata(filePath: string): Promise<VideoMetad
     const streamDuration = videoStream.duration ? parseFloat(videoStream.duration) : 0;
     const parsedStreamStart = videoStream.start_time ? parseFloat(videoStream.start_time) : 0;
     const streamStart = Number.isFinite(parsedStreamStart) ? parsedStreamStart : 0;
+    // `nb_frames` is populated by the container demuxer; a muxer that requires
+    // `-count_packets` to enumerate frames will leave it undefined, in which
+    // case the caller must treat the frame-count check as "no answer". This
+    // path is hit by, for example, fragmented MP4 where the moov box lacks a
+    // frame count — the captured duration is still correct.
+    const parsedNbFrames = videoStream.nb_frames ? parseInt(videoStream.nb_frames, 10) : NaN;
+    const frames =
+      Number.isFinite(parsedNbFrames) && parsedNbFrames > 0 ? parsedNbFrames : undefined;
 
     return {
       durationSeconds: containerDuration,
@@ -788,6 +802,7 @@ export async function extractMediaMetadata(filePath: string): Promise<VideoMetad
       isVFR,
       hasAlpha,
       colorSpace,
+      frames,
     };
   })();
 
--- a/packages/producer/src/services/render/artifactTransaction.test.ts
+++ b/packages/producer/src/services/render/artifactTransaction.test.ts
@@ -12,7 +12,7 @@ import {
 } from "node:fs";
 import { dirname, join } from "node:path";
 import { tmpdir } from "node:os";
-import { ArtifactTransaction } from "./artifactTransaction.js";
+import { ArtifactTransaction, type ArtifactDurationProbe } from "./artifactTransaction.js";
 
 function tempDir(): string {
   return mkdtempSync(join(tmpdir(), "hf-artifact-transaction-"));
@@ -39,21 +39,21 @@ describe("ArtifactTransaction", () => {
     expect(transactionDirectories(dir)).toEqual([]);
   });
 
-  it("atomically replaces a file only after validation", () => {
+  it("atomically replaces a file only after validation", async () => {
     const dir = tempDir();
     const destination = join(dir, "render.mp4");
     writeFileSync(destination, "existing");
     const transaction = new ArtifactTransaction(destination, "file");
     writeFileSync(transaction.stagingPath, "new-render");
 
-    transaction.commit();
+    await transaction.commit();
 
     expect(readFileSync(destination, "utf8")).toBe("new-render");
     expect(existsSync(transaction.stagingPath)).toBe(false);
     expect(transactionDirectories(dir)).toEqual([]);
   });
 
-  it("leaves an existing file byte-identical when validation fails", () => {
+  it("leaves an existing file byte-identical when validation fails", async () => {
     const dir = tempDir();
     const destination = join(dir, "render.gif");
     const existing = Buffer.from([0, 1, 2, 3, 255]);
@@ -61,7 +61,7 @@ describe("ArtifactTransaction", () => {
     const transaction = new ArtifactTransaction(destination, "file");
     writeFileSync(transaction.stagingPath, "");
 
-    expect(() => transaction.commit()).toThrow("not a non-empty file");
+    await expect(transaction.commit()).rejects.toThrow("not a non-empty file");
     transaction.rollback();
 
     expect(readFileSync(destination)).toEqual(existing);
@@ -69,7 +69,7 @@ describe("ArtifactTransaction", () => {
     expect(transactionDirectories(dir)).toEqual([]);
   });
 
-  it("keeps the existing file addressable when atomic promotion fails", () => {
+  it("keeps the existing file addressable when atomic promotion fails", async () => {
     const dir = tempDir();
     const destination = join(dir, "render.mp4");
     writeFileSync(destination, "existing");
@@ -87,7 +87,7 @@ describe("ArtifactTransaction", () => {
     });
     writeFileSync(transaction.stagingPath, "new-render");
 
-    expect(() => transaction.commit()).toThrow("injected replacement failure");
+    await expect(transaction.commit()).rejects.toThrow("injected replacement failure");
     expect(replacementCalls).toBe(1);
     expect(readFileSync(destination, "utf8")).toBe("existing");
 
@@ -110,7 +110,7 @@ describe("ArtifactTransaction", () => {
     expect(transactionDirectories(dir)).toEqual([]);
   });
 
-  it("promotes a validated PNG sequence as one directory artifact", () => {
+  it("promotes a validated PNG sequence as one directory artifact", async () => {
     const dir = tempDir();
     const destination = join(dir, "frames");
     mkdirSync(destination);
@@ -120,14 +120,14 @@ describe("ArtifactTransaction", () => {
     writeFileSync(join(transaction.stagingPath, "frame_000001.png"), "png-1");
     writeFileSync(join(transaction.stagingPath, "frame_000002.png"), "png-2");
 
-    transaction.commit();
+    await transaction.commit();
 
     expect(readdirSync(destination).sort()).toEqual(["frame_000001.png", "frame_000002.png"]);
     expect(readFileSync(join(destination, "frame_000001.png"), "utf8")).toBe("png-1");
     expect(transactionDirectories(dir)).toEqual([]);
   });
 
-  it("restores the previous PNG sequence when directory promotion fails", () => {
+  it("restores the previous PNG sequence when directory promotion fails", async () => {
     const dir = tempDir();
     const destination = join(dir, "frames");
     mkdirSync(destination);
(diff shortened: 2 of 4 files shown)


james-russo-rames-d-jusso (commented):
🟢 *LGTM from my side — clean cherry-pick of #3429, direct fence on #3395, one design concern worth surfacing.* Merge is Miguel's call; Miga is the PR author (bot), so peer-bot signal alone isn't a merge greenlight.

*Supersession vs #3429 — verified byte-identical*

All four PR files are byte-identical between #3429 HEAD `3410fddc9d7` and #3506 HEAD `2cb3fbca4246`:
- `packages/producer/src/services/render/artifactTransaction.ts`
- `packages/producer/src/services/render/artifactTransaction.test.ts`
- `packages/producer/src/services/renderOrchestrator.ts`
- `packages/engine/src/utils/ffprobe.ts`

Verified via `gh api /repos/heygen-com/hyperframes/contents/<path>?ref=<sha>` on each HEAD → `diff -u` empty. This genuinely is "same code, resigned" — no quiet retune on the rescue path. *Note*: #3429 is still `state: OPEN` on GitHub — the "supersedes" framing is Miga's Slack claim, not GH state. Cleanest is to close #3429 explicitly (or convert to `Closes #3429` in the #3506 body) so the two don't both look live to a scanner.

*The gate mechanics* (`artifactTransaction.ts:230-267`, wired at `renderOrchestrator.ts:3946-3954`)

- *Duration check* (`:258-263` calling `assertDurationWithinTolerance` at `:162`): rejects when `expectedSeconds - probedSeconds > toleranceSeconds`; tolerance defaults to `1/fps` (30fps → 33.3ms) or 20ms when fps unknown. One-sided — over-duration passes silently, which matches the failure mode (truncation, not extension).
- *Frame check* (`:266` calling `assertFrameCountWithinTolerance` at `:178`): rejects when `expectedFrames - probedFrames > 1`; tolerance hardcoded to a single frame at `:187`.
- *Probe error* → throws with `duration probe failed` prefix at `:277-282`; test-fenced at `artifactTransaction.test.ts:322-343`.
- *Probe returning 0/NaN* → throws with `no usable duration` at `:154-158`; fenced at test `:345-360`.
- *Frame count `undefined`* (fragmented MP4 / missing `nb_frames`) → silently skipped at `:183-185` per doc comment at `ffprobe.ts:189-193`. Duration is the belt; frame-count is the suspenders. Both fire on the #3395 packet (52.2s → 41.333s AND 1566 → 1240).

*Multi-worker frame-drop coverage*

The field packet mode (container duration reflects truncation) is caught by duration alone. The "container duration correct, stream shorter" mode is caught by frames when `nb_frames` is present. If the producer's assemble stage ever emits fMP4 or a mux where `nb_frames` is `undefined`, the check silently degrades to duration-only — `ffprobe.ts:191` acknowledges this explicitly. Not a bug; worth naming so nobody assumes the frame-count check is universally active.

*ArtifactTransaction contract-change / blast radius*

`gh search code` shows exactly one `new ArtifactTransaction` (`renderOrchestrator.ts` at ~`:2100`) and one `.commit()` caller (same file at `:4037`). No polymorphic callers passing non-video artifacts. `commit()` and `validate()` are both `async` on this PR — the caller `await`s both correctly. The `!isPngSequence && !isGif && Number.isFinite(job.duration) && job.duration > 0` guard at `renderOrchestrator.ts:3947-3948` correctly routes PNG-sequence and GIF paths through `validate()` without an expectation, matching directory-kind test at `:220-229`.

*Concerns*

- *Split-brain gate API* (`artifactTransaction.ts:230-234` + `:286-290`) — load-bearing. `validate(expected?)` makes the expectation optional, and `commit()` internally calls `this.validate()` at `:290` with no argument. So the duration/frame gate ONLY fires if the caller explicitly calls `validate(expected)` BEFORE `commit()`. If a future caller (or a refactor that folds validation into `commit()`) drops the explicit `.validate(expected)` call, the gate silently disappears with no type-system signal. Cleaner shape: pass `expected` to the constructor, or accept `expected` as a `commit(opts)` param, so the gate is unconditional-once-configured. As-is, this fix is one accidental deletion away from a silent revert of #3395. Non-blocker for this PR (existing caller wires it correctly), but worth a follow-up ticket to close the API-shape gap.
- *GIF is silently excluded* (`renderOrchestrator.ts:3947`). PR body notes this ("non-PNG-sequence, non-GIF video outputs") but no rationale inline. A truncated GIF ships as success. Presumably GIF isn't the multi-worker failure mode — but if a GIF encoder path can also drop frames, this is an open door. Worth naming the reason in an inline comment.
- *Concurrent-PNG-sequence test weakened* (`artifactTransaction.test.ts:170-192`). Old test asserted "concurrent-frame wins deterministically" during rollback; new test now accepts either "existing-frame" or "concurrent-frame" surviving. Rationale in the comment is correct (the `void concurrentTransaction.commit()` inside a sync fs callback now yields at the first `await` when both are async), but this means the "concurrent writer wins during rollback" invariant is no longer test-fenced. Production has one transaction per destination so it's likely test-only, but if the atomic-handoff contract matters, that test used to prove it and now doesn't.

*Nits*

- `artifactTransaction.ts:290` — `commit()` calls `this.validate()` a second time (redundant readable-non-empty check after the caller's `validate(expected)`). Two stat calls per commit; harmless.
- `assertFrameCountWithinTolerance` at `:178` takes no tolerance parameter; the "≤ 1" is a magic constant at `:187`. Function name and signature diverge on the "within tolerance" implication. Small readability drag.

*Questions*

- Is the producer's assemble stage output guaranteed to be non-fragmented MP4 (so `nb_frames` is always populated)? If not, the frame-count check is inert for that path and the gate leans entirely on duration. Not a blocker — a note for whoever owns `runAssembleStage`.
- The `fps` fallback of 20ms tolerance when unknown (`ffprobe.ts` computed default) — is fps ever actually unknown by the time `validate(expected)` is called? Orchestrator passes `fpsToNumber(job.config.fps)`; if that can return `undefined`/0, tolerance quietly widens from ~33ms to 20ms. Minor.

*CI + peer state at HEAD `2cb3fbca4246`*

`gh api /pulls/3506/reviews` → `[]`. No peer reviews. Author is `miga-heygen` (bot); merge is Miguel's call, not a peer-bot approval. CI: 25 green + 14 in-flight + 1 neutral + 12 skipped, 0 red. In-flight: Producer unit + integration, Build, Typecheck, Lint, SDK contract/smoke, Windows render, regression shards, CodeQL js-ts.

*What I didn't verify*

- Did not confirm producer's mp4 muxer preserves `nb_frames` in practice (would require running an actual render + `ffprobe -show_streams`).
- Did not run the tests locally.
- Did not audit whether `renderChunk.ts` distributed path or `distributed/assemble.ts` bypasses `renderOrchestrator.ts` and constructs its own file writes — `gh search code` shows only one `new ArtifactTransaction`, so a bypass would be outside the transaction abstraction entirely and out of scope.

— _Review by Rames D Jusso_

terencecho (approved):
**APPROVE** at `2cb3fbca4246b9908259daf627c3f948bf2ba759`. Fail-closed direction, tolerance calculation, and truncation-catch coverage all check out. Details below.

**1. Truncation direction — one-way, both axes.**

Both `assertDurationWithinTolerance` and `assertFrameCountWithinTolerance` compare `expected - probed > threshold`. A probed *longer* value passes silently (audio-tail padding, container faststart quirks). Only shortfalls trip. Correct polarity for a truncation gate — a good render encoded slightly past the last frame is normal, a good render that lost frames is not.

**2. Tolerance — one frame at declared fps.**

`durationToleranceSeconds` returns `1/fps` when fps is present (`0.020` fallback). At the fps values the producer actually hits:
- 24 → 41.7 ms
- 29.97 (rational `30000/1001` via `fpsToNumber` → `29.97002997…`) → 33.4 ms — matches actual NTSC frame duration `1001/30000 = 33.367 ms` within 0.03 ms
- 30 → 33.3 ms
- 60 → 16.7 ms

A normal container-level last-frame rounding (one output sample) sits inside this window. The #3395 field packet (326-frame / 11 s shortfall against 1566 / 52.2 s) is 326× tolerance — solid catch, huge margin. And the `--workers 1` known-good case (10.000 s / 300 frames, expected 10.000 s / 300 frames) is proven not to trip in `test accepts a video whose probed duration matches`.

Both `expectedDurationSeconds = job.duration` and `expectedFrames = captureTotalFrames` are pinned at the probe stage (`renderOrchestrator.ts:2429,:2432`) and are the pipeline's own committed target — same `fpsToNumber(job.config.fps)` conversion used by `validateRenderDuration` right after they're written. So the values on both sides of the compare go through identical rational→decimal quantization; no cross-system drift.

**3. Fail-closed on ffprobe / no-answer paths.**

- Probe throws → wrapped as `Render artifact duration probe failed for {path}: {msg}` with `{cause: error}`. Explicit test `test propagates a probe failure`. Fails closed.
- Probe returns `0` or `NaN` duration → `assertUsableProbedDuration` throws with expected value + probed value. Test `test rejects when the probe returns no usable duration`. Fails closed.
- Probe returns valid duration but `frames === undefined` → frame check silently no-ops. This is the *only* fail-open path, and it's intended: the `ffprobe.ts` comment names fragmented MP4 as the class where `nb_frames` isn't populated without `-count_packets`. In that mode the duration check remains the sole guard, which is fine because container duration is still populated. Follow-up in the non-blocker section.

**4. Wire-up covers what it should, skips what it should.**

The applicability gate is `!isPngSequence && !isGif && Number.isFinite(job.duration) && job.duration > 0`. PNG sequences use `kind: "directory"` — `validate()` already refuses to probe directory artifacts (`test does not probe duration when validating a directory artifact`, belt-and-suspenders with the wire-up gate). GIFs skip assemble entirely, so there's no muxed container to compare — skipping the gate is correct. The still-image / probe-failed paths where `job.duration` isn't finite also skip.

**5. Diagnostic quality.**

Every failure path includes staging path + expected + probed + deficit + tolerance in the message. Fragmented-MP4 mode is documented in-source at the exact line where the silent no-op happens. Enough for oncall to identify which artifact + why without a repro.

**Non-blockers (don't hold merge):**

1. **Fragmented MP4 frame-check gap.** The frame-axis silent no-op is only safe as long as the producer doesn't emit fMP4 in a common mux config. Quick author check on the assemble stage's ffmpeg args suffices; if fMP4 is ever a live path, threading `-count_packets` (or a separate stream probe) restores coverage. If assemble always emits standard MP4 with `moov` in a way that populates `nb_frames`, this is a no-op concern.

2. **`assertUsableProbedDuration` throws on `0` even if `frames` is populated.** If a codec path ever returns `{durationSeconds: 0, frames: N}`, the current order rejects on the duration axis before the frame check gets a chance. The truncation would still be caught (just as "no usable duration" not "N-frame shortfall"), so operationally identical — worth noting only because the error message may under-describe the actual shape.

3. **Concurrent-PNG-sequence test contract loosened.** `test does not delete a concurrent PNG sequence published during recovery` (`artifactTransaction.test.ts:155-186`) now accepts *either* surviving frame after the fire-and-forget `void concurrentTransaction.commit()`. The prior test locked in `concurrent-frame`. Tracks the reality that `commit()` is now async and the concurrent's microtasks haven't flushed by the time `renameSync` throws synchronously. The code contract "existing is never removed" is still verified, but the "concurrent wins the race" pin is gone. If concurrent-wins ordering is load-bearing elsewhere, a follow-up test that awaits the concurrent commit before the rename throws would keep that pin honest.

CI: `Test`, `Producer: unit tests`, `Producer: integration tests`, `Test: runtime contract`, `Windows render verification`, `regression-shards 1–9`, `preview-regression`, `Preflight`, `Typecheck`, `Lint`, `Format`, `CLI smoke` — all green at head. `mergeStateStatus=BLOCKED` is review-gated, not check-gated; this approval should unblock.

— Review by tai (pr-review)

PR #3507, opened 2026-08-26
fix(core): read CSS animation opacity through color grading's own hide

### LINKED ISSUE #3329: Bug: data-color-grading renders solid white (mp4) / black (prores) frames — any grading, 0.7.109 & 0.8.2
### Environment
- hyperframes CLI 0.7.109 **and** 0.8.2 (both tested, identical behavior)
- macOS arm64 (Darwin 25.4.0)
- Static `` with `data-color-grading` applied via `hyperframes media-treatment --apply`

### Reproduction
1. Create a composition with a full-frame `` (with `id` for strict lint):
```html
<img id="media-frame" class="clip" src="frame.png" data-start="0" data-duration="1" data-track-index="1" style="left:50%;top:50%;width:1920px;height:1080px;object-fit:cover;">
```
2. Apply a grading patch:
```bash
hyperframes media-treatment --project . --file index.html --selector "img#media-frame" --grading '{"effects":{"kuwahara":0.8}}' --apply
```
3. Render:
```bash
hyperframes render . --format mp4 --output out.mp4 --fps 30 --quiet --strict-all
```

### Results (4-condition matrix)

| Condition | Output | Luminance |
|---|---|---|
| `effects.kuwahara=0.8` + mp4 | **Solid white frame** | 252/255 |
| `adjust.saturation=-0.3` + mp4 | **Solid white frame** | 251/255 |
| `effects.kuwahara=0.8` + mov (prores) | **Solid black frame** | 0/255 |
| **No `data-color-grading`** + mp4 | ✅ Normal image | 96/255 (matches source) |

### Expected
The graded image should render with the color treatment applied, not a solid color frame.

### Key observations
- **Any** `data-color-grading` attribute (regardless of effect type: `effects.*` or `adjust.*`) triggers the bug
- Removing the `data-color-grading` attribute restores normal rendering
- Bug is **identical across 0.7.109 and 0.8.2**
- The CLI `--apply` succeeds (`ok: true`) and writes a complete ~80-key `data-color-grading` JSON to the element
- The composition HTML skeleton uses `#stage` with `data-composition-id`, `data-no-timeline`, `data-start/duration/width/height/fps` — worker-standard structure

### Reproduction package
A tarball with the minimal project (HTML + frame + 4 rendered outputs) is available. The command sequence above reproduces it from scratch.

### Impact
This blocks the "deterministic media treatment" pipeline for any project that renders graded media through the standard composition path.

miga-heygen:
## Summary

Fixes color-graded elements with CSS entrance animations (e.g. `opacity: 0→1`) rendering as solid white (mp4) or black (prores) frames for the entire output.

**Root cause:** On the first `drawEntry()`, the animation's initial `opacity: 0` is copied to `sourceOpacityForCanvas`. Then `hideSourceElement()` sets `opacity: 0 !important` on the source. On subsequent frames, the `hiddenByColorGrading` guard correctly prevents reading back grading's own hide — but also prevents updating the canvas opacity as the animation progresses. Both source and canvas stay at `opacity: 0` for the entire render.

**Fix:** When the source is hidden by color grading, temporarily restore the authored inline opacity before reading `getComputedStyle`, so the CSS animation's current value shows through. The restore→read→rehide is synchronous within a single JS task, so no repaint occurs between the style writes.

Closes #3329

— Miga

### DIFF
--- a/packages/core/src/runtime/colorGrading.test.ts
+++ b/packages/core/src/runtime/colorGrading.test.ts
@@ -799,6 +799,36 @@ describe("createColorGradingRuntime", () => {
     expect(canvas.style.opacity).toBe("1");
   });
 
+  it("updates canvas opacity through grading's own hide when CSS animation progresses (#3329)", () => {
+    const video = makeDrawableVideo();
+    // Simulate a CSS entrance animation: the authored opacity is empty (no
+    // inline opacity), but at parse time the element has opacity 0 from the
+    // animation's initial keyframe.
+    video.setAttribute("data-hf-authored-opacity", "");
+    document.body.appendChild(video);
+
+    runtime = createColorGradingRuntime();
+    const canvas = document.querySelector<HTMLCanvasElement>("[data-hf-color-grading-canvas]");
+    if (!canvas) throw new Error("Expected color grading canvas");
+
+    // Source is hidden by color grading.
+    expect(video.style.getPropertyValue("opacity")).toBe("0");
+    expect(video.style.getPropertyPriority("opacity")).toBe("important");
+
+    // The canvas opacity should NOT be frozen at "0" — after the hide is
+    // temporarily lifted, jsdom's getComputedStyle returns "" (no inline
+    // opacity set), which the code normalizes to "1".
+    expect(canvas.style.opacity).toBe("1");
+
+    // Simulate the CSS animation progressing by re-reading after a redraw.
+    runtime.redraw();
+    expect(canvas.style.opacity).toBe("1");
+
+    // Source remains hidden throughout.
+    expect(video.style.getPropertyValue("opacity")).toBe("0");
+    expect(video.style.getPropertyPriority("opacity")).toBe("important");
+  });
+
   it("allows a drawable producer render frame to initialize hidden source grading", () => {
     const video = makeDrawableVideo();
     video.style.display = "none";
--- a/packages/core/src/runtime/colorGrading.ts
+++ b/packages/core/src/runtime/colorGrading.ts
@@ -3065,16 +3065,30 @@ function drawEntry(entry: ColorGradingEntry): boolean {
   const sourceVisibility = entry.element.style.getPropertyValue("visibility");
   const injectedFrameSource = isRenderFrameImage(source);
   if (injectedFrameSource) keepCanvasAboveSource(entry, source);
+  // When color grading hid the source (opacity:0 !important), temporarily
+  // restore the authored inline opacity so getComputedStyle reflects the CSS
+  // animation's current value instead of our own hide. Without this, a CSS
+  // entrance animation (opacity: 0→1) freezes at its first-frame value for
+  // the entire render because the !important blocks all animation effects on
+  // the property (#3329). The restore–read–rehide is synchronous, so no
+  // repaint occurs between the style writes.
+  if (hiddenByColorGrading && !injectedFrameSource) {
+    if (entry.sourceInlineOpacity !== null) {
+      entry.element.style.setProperty(
+        "opacity",
+        entry.sourceInlineOpacity,
+        entry.sourceInlineOpacityPriority || "",
+      );
+    } else {
+      entry.element.style.removeProperty("opacity");
+    }
+  }
   const computed = window.getComputedStyle(injectedFrameSource ? source : entry.element);
-  // `hideSourceElement` owns the source's inline opacity while grading is active
-  // (opacity:0 !important), so reading it back would mirror grading's own hide
-  // onto the canvas and blank it. That gate is about opacity ONLY: grading never
-  // writes `visibility`, so the source's computed visibility always tracks the
-  // clip window and has to be re-read every frame. Leaving it stale let the
-  // canvas keep an explicit `visibility: visible` and paint straight through an
-  // inactive ancestor clip's inherited `visibility: hidden`.
-  if (injectedFrameSource || !hiddenByColorGrading) {
-    entry.sourceOpacityForCanvas = computed.opacity || "1";
+  // Grading never writes `visibility`, so the source's computed visibility
+  // always tracks the clip window and has to be re-read every frame.
+  entry.sourceOpacityForCanvas = computed.opacity || "1";
+  if (hiddenByColorGrading && !injectedFrameSource) {
+    entry.element.style.setProperty("opacity", "0", "important");
   }
   entry.sourceVisibleForCanvas =
     (injectedFrameSource || sourceVisibility !== "hidden") && computed.visibility !== "hidden";


james-russo-rames-d-jusso (commented):
Miguel — traced the fix end-to-end against `drawEntry` and `hideSourceElement` at the new HEAD. The core mechanic is sound: lift authored inline opacity → read `getComputedStyle` → rehide `0 !important` — all inside a single synchronous JS task, so no repaint slips between the writes and MutationObservers see a coalesced final state. The scoping is per-`entry.element`, so two `<canvas>` elements on-screen won't leak hide-state between each other. One load-bearing concern below, then some smaller items.

*Load-bearing concern — the fallback capture path in `hideSourceElement` silently defeats the fix.*

`packages/core/src/runtime/colorGrading.ts:2918-2925` — when the parse-time `data-hf-authored-opacity` attribute is present (the `authored !== null` branch at `:2919-2921`), the fix works as advertised: the attribute holds the pre-animation authored value (empty or the declared opacity) and the lift on `drawEntry` `:3076-3086` restores that. But when the attribute is absent, `hideSourceElement` falls back at `:2922-2925` to `entry.element.style.getPropertyValue("opacity")` — and if the CSS entrance animation has already sampled its `opacity: 0` initial keyframe by the time the first hide runs, that captured `sourceInlineOpacity` is `"0"`.

On subsequent frames the lift branch at `:3077-3082` then restores opacity to `"0"`, `getComputedStyle` at `:3087` reads `"0"`, and `sourceOpacityForCanvas` at `:3090` is set to `"0"` — the exact frozen-at-initial-value failure mode #3329 was meant to fix. The comment at `:2912-2917` ("Fall back to the live inline value for documents loaded without the capture installed") is essentially the acknowledgement.

Two questions:
- Which surfaces install the parse-time capture, and is it guaranteed for every production render path (producer render, studio preview, iframe embed)? If any surface can render color grading without the attribute stamped before first hide, #3329 still reproduces there — silently, no error, same solid frame.
- If the fallback path is a known-degraded mode (i.e. accepted risk), can we at least log a `swallow`-level breadcrumb from the fallback branch so support can grep for it when someone reports a repro?

No new test exercises the fallback path — every hit of `data-hf-authored-opacity` in `colorGrading.test.ts` (`:259`, `:275`, `:807`) sets the attribute. A test that omits the attribute, sets `video.style.opacity = "0"` before `createColorGradingRuntime()`, then asserts `canvas.style.opacity !== "0"` on `runtime.redraw()` would either prove the fallback is safe or turn this concern into a blocker.

*Non-blocker concerns.*

1. `packages/core/src/runtime/colorGrading.test.ts:802-830` — the new #3329 test only covers the `sourceInlineOpacity === null` branch of the lift (`removeProperty` at `:3084`). The other branch — `setProperty(..., sourceInlineOpacityPriority)` at `colorGrading.ts:3078-3082` — has no per-frame coverage. The destroy-time restore of `"0.75"` at `test.ts:255-271` predates this PR and doesn't exercise `drawEntry`'s new lift/rehide dance. Suggest adding a case that stamps `data-hf-authored-opacity="0.5"` and asserts canvas opacity reflects the restored value across `runtime.redraw()`, with source rehidden at `0 !important` after each frame.

2. `colorGrading.test.ts:825-826` — the second `runtime.redraw()` doesn't simulate an animation progressing; jsdom has no CSS-animation engine, so both `getComputedStyle` reads return `""` and hit the `|| "1"` normalization at `colorGrading.ts:3090`. The test proves "the code path no longer freezes at `0`", not "a real browser reads the interpolated animation value." That verification necessarily lives in the visual/regression shards — the `preview-regression` and `regression-shards` shards were pending at review time; recommend confirming at least one existing fixture in those shards exercises an opacity keyframe on a graded element, otherwise consider a targeted CSS-animation-opacity fixture add. (The producer unit-test claim of "all 53 tests pass" is real, but by construction it can't catch the animation-progression regression.)

3. `colorGrading.ts:3076-3092` — no `try/finally` wrapping the lift-read-rehide. The only realistic throw surface between the lift and the rehide is `window.getComputedStyle` on `:3087` (which throws in jsdom for detached elements, spec'd not to throw in browsers). If it did throw — element detached concurrently mid-frame — the source stays temporarily with its authored inline opacity and `sourceHidden` remains `true`, so the DOM/flag state diverges until the next `drawEntry` re-hides (or `restoreSourceElement` at `:1948-1958` skips the branch because `opacity !== "0 !important"` and leaves it as-is). Wrapping the lift and rehide in `try/finally` would harden this at zero real cost. Nit-level.

4. `colorGrading.ts:3087-3092` — the lift/rehide now runs every frame for every hidden-by-grading entry, adding two `setProperty` mutations and one attribute-mutation observer callback per entry per frame. Opacity is a composited-only property so no forced layout, and the observer at `:3416-3427` bails on the `geometrySignature` check without triggering a redraw. Not a real perf concern at HF scale, but if we ever need to compose ~50 graded entries in one composition it's worth remembering this hot-path cost was added.

*What I didn't verify.*

- Real-browser `getComputedStyle().opacity` semantics while a `@keyframes` animation is playing — the fix's premise is that the browser returns the current interpolated value once the inline `0 !important` is out of the cascade. I trust that against the spec but did not open a browser to confirm.
- The `preview-regression` and `regression-shards` shards were still pending at review time; if a fixture there covers this exact repro, that's the real validation for #3329.
- Whether `heygen-com/hyperframes-internal` is a surface where the parse-time authored-opacity capture might not run — outside my read scope, worth an internal check.

*Peer state + CI at HEAD `d3ae41fd9a4`.*

`gh api /pulls/3507/reviews` → `[]`. No peer reviews. Author is `miga-heygen` (bot); merge is your call, not a peer-bot approval. Static gates green at review time; heavy shards (`regression-shards`, `Perf: *`, `Preview parity`, Windows render, `Producer: unit tests`) all in-progress. `Producer: unit tests` is where the new colorGrading test lands — hold merge on that shard's green.

— _Review by Rames D Jusso_

vanceingalls (approved):
APPROVED.

**What.** In `drawEntry`, when `hiddenByColorGrading` is active and not an injected-frame source, temporarily restore the captured `sourceInlineOpacity` (with its original priority) before reading `getComputedStyle`, then re-apply `opacity: 0 !important`. Synchronous restore→read→rehide bypasses the animation freeze where the old guard skipped the read entirely and left `sourceOpacityForCanvas` frozen at its first-frame value.

**Correctness.**
- Capture site is `hideSourceElement` (`colorGrading.ts:2921/2924`), which reads `data-hf-authored-opacity` or `element.style.opacity` **before** applying the `opacity: 0 !important` hide. So the restored value is the authored value, not the hide's own value — critical invariant, holds.
- `sourceInlineOpacityPriority || ""` correctly preserves the authored priority on restore.
- `sourceVisibility` is captured before the temp lift; visibility path unchanged.
- `injectedFrameSource` case correctly excluded from the lift — computed is read from `source`, not `entry.element`, so lifting `entry.element`'s hide would be a no-op for that read.

**Sync guarantee.** Both style writes and `getComputedStyle` are synchronous within a single task; layout flush occurs after the task, not between the writes — no repaint window opens. A MutationObserver would see both attribute changes in the same microtask batch after the fact but never observes the DOM in the intermediate state.

**Test.** jsdom limits authentic animation-progression coverage, but the new test asserts the code path: authored empty → restore→remove→computed "" → canvas normalized to "1", hide-priority survives the round-trip. Real animation coverage rides on Windows render verification + regression shards, both green.

**CI.** All required green.

— Via

vanceingalls:
Follow-up on my approval — I stamped without first pulling Rames's earlier COMMENTED review into scope, which was a mistake. His load-bearing concern about the fallback capture path in `hideSourceElement:2922-2925` (when `data-hf-authored-opacity` is absent) sits outside what my review verified.

Reading Rames's scenario against the code: the fallback reads `entry.element.style.getPropertyValue("opacity")`, which reflects **inline** style only. Pure CSS `@keyframes` animations don't mutate inline style, so the fallback should capture whatever was authored inline (typically `""`), and the fix works there. The concern lands squarely on **JS-driven** animation engines that write to `element.style.opacity` — Web Animations API replaces + inline animations, GSAP, animejs, motion.dev. If any of those drive a graded element's entrance opacity, and its first inline sample happens to be `"0"` when `hideSourceElement` first runs, the fallback captures `"0"` → the lift-restore→read cycle then reads `"0"` and `sourceOpacityForCanvas` freezes at `"0"` — the exact #3329 failure mode.

Two things I'd want to see before this ships:

1. **Fallback-path test coverage** — a case that omits `data-hf-authored-opacity`, seeds `video.style.opacity = "0"` before `createColorGradingRuntime()`, and asserts canvas opacity doesn't freeze at `"0"` across `runtime.redraw()`. If it freezes, the fallback needs a different capture strategy (e.g. read the animation-start baseline via `element.getAnimations()[0].effect.getKeyframes()[0].opacity` when available, or skip capture and let the lift restore to `""` unconditionally in the fallback branch).

2. **Which render surfaces install the parse-time capture** — producer render, studio preview, iframe embed. If any surface renders color grading without the attribute stamped before first hide, the fallback path is the one that runs there, and the fix's guarantee weakens on that surface.

If both check out (attribute is stamped everywhere that hits this code, or the fallback path is verified safe for JS-animation cases), the approval stands as-is. If the fallback is a known-degraded mode, Rames's suggestion of a swallow-level breadcrumb from `:2922-2925` would give support a signal to grep.

Not blocking — the CSS-keyframes case is fixed and CI is green — but flagging so the concern doesn't get lost under my stamp.

— Via

PR #3513, opened 2026-08-26
fix(core,producer): stamp render ids on empty-src media and pair the snapshot by them

### LINKED ISSUE #3512: Nested compositions that reuse a video id with runtime-assigned src collapse onto one scene in the export
### Describe the bug

Follow-up to #3340 / #3342. Duplicate media ids across nested compositions with a compile-time `src` now get a document-unique `data-hf-render-id` and both clips render.

The residual: `<video id="clip" src="">` plus `el.src = url` in the scene script. Legal per file, `check` clean. The static parse skips empty `src`, so those clips exist only in the browser snapshot. That snapshot still keys by author id. After inline, last write wins:

- Scene A (0–4s, `data-media-start="10"`) gets scene B’s window — footage from ~40s.
- Scene B (4–8s, `data-media-start="40"`) gets nothing injected — blank panel.

Scene chrome still paints. Preview is correct (each scene’s script is scoped). Export disagrees.

`#3342` stamps render ids only when a playable source is already on the element, which is why this path never got one. Native `url(#…)` / `<use>` is a different hole (#3490).

### Link to reproduction

https://github.com/ArcadeHQ/hyperframes-repros/tree/patch/runtime-src-duplicate-ids

### Steps to reproduce

1. Clone https://github.com/ArcadeHQ/hyperframes-repros and check out `patch/runtime-src-duplicate-ids`.
2. Run `npx hyperframes@latest check` — 0 errors (duplicate ids across files are allowed).
3. Preview: both scenes show the burnt-in seconds counter (~12 in scene A, ~46 in scene B).
4. Run `npx hyperframes@latest render . -f 10`.
5. Open the MP4 at t≈2s and t≈6s. `snapshots/` has actual vs expected frames.

Control: rename scene B’s video id to `clip-b` (and its script selector). Both windows render correctly.

### Expected behavior

Scene A shows the counter around 10–14. Scene B shows it around 44–48. Same as preview.

### Actual behavior

On `hyperframes@0.8.15` (latest):

- t≈2s: counter reads ~42 (scene B’s `data-media-start`).
- t≈6s: white panel, no footage.

### Environment

```
hyperframes doctor

  ✓ Version          0.8.15 (latest)
  ✓ Memory           16.0 GB total · 3.5 GB available
  ✓ Disk             358.9 GB free
  ✓ Frames cache     /var/folders/bl/c4hk6r7n34g1d81txd99gsgr0000gn/T/hyperframes-extract-cache-501 · 358.9 GB free at /var/folders/bl/c4hk6r7n34g1d81txd99gsgr0000gn/T/hyperframes-extract-cache-501 · default
  ✓ Archive extractor unzip
  ✓ Environment      non-TTY
  ✗ Docker           Not found
```

### Additional context

The source is a synthetic `ffmpeg testsrc` clip with a burnt-in seconds counter, so the wrong window is readable off the frame. Related, not the same: #3340 (static `src`, fixed by render keys), #3490 (SVG `url(#)` / `<use>`).

valeriangalliat:
## What

Empty-src `<video>` / `<audio>` now get a document-unique `data-hf-render-id`, and the browser snapshot / visibility / volume probes key and look up by that id instead of the author id.

Fixes #3512.

## Why

#3342 stamped render ids only when a playable source was already on the element. Authors who assign `el.src = url` in the scene script leave `src=""` in markup. The static parse skips those elements, so their clips exist only in the browser snapshot. That snapshot still keyed by author id. After inline, two scenes sharing `id="clip"` collapse onto one window: the first scene plays the second scene's footage, the second scene paints blank. Preview is fine because each scene's script is scoped.

Duplicate ids across nested compositions are legal per file. `check` does not flag them.

## How

- `assignMediaRenderIds` stamps every `video` / `audio` (and `img[src]`), including empty `src`. Author `id`s are still not rewritten.
- `discoverMediaFromBrowser` reports `data-hf-render-id` when present.
- Visibility sampling uses the same id. Volume automation resolves through `__hfMediaEl` (already installed on the probe page) with `getElementById` as fallback.

## Test plan

- [x] Manual testing performed against the public repro

**New tests**

- `mediaRenderIds.test.ts`: colliding `src=""` videos get `clip` / `clip__hf2`; no-src and empty `<source>` child are stamped too.
- `htmlCompiler.test.ts`: snapshot reports colliding empty-src videos by render id; `compileForRender` of two empty-src scenes stamps unique render ids (static list still empty).

**Reproduction:** https://github.com/ArcadeHQ/hyperframes-repros/tree/patch/runtime-src-duplicate-ids

On `main`, scene A at t≈2s shows ~42 (scene B's window) and scene B at t≈6s is blank. After this change both should match preview (~12 and ~46).

### DIFF
--- a/packages/core/src/compiler/mediaRenderIds.test.ts
+++ b/packages/core/src/compiler/mediaRenderIds.test.ts
@@ -75,10 +75,25 @@ describe("assignMediaRenderIds", () => {
     expect(ids[1]).toBe("clip__hf2");
   });
 
-  it("leaves media with no source at all alone", () => {
+  it("stamps empty-src media with colliding author ids", () => {
+    // `src=""` used to skip the stamp. The snapshot then keyed by raw id and
+    // colliding scenes collapsed. Runtime assignment is why the src is empty,
+    // not a second path this function sees.
+    const { document } = parseHTML(
+      '<video id="clip" src=""></video><video id="clip" src=""></video>',
+    );
+    assignMediaRenderIds(document as unknown as Parameters<typeof assignMediaRenderIds>[0]);
+    expect(
+      Array.from(document.querySelectorAll("video")).map((el) =>
+        el.getAttribute(MEDIA_RENDER_ID_ATTR),
+      ),
+    ).toEqual(["clip", "clip__hf2"]);
+  });
+
+  it("stamps a video with no source attribute at all", () => {
     const { document } = parseHTML('<video id="no-src"></video>');
     assignMediaRenderIds(document as unknown as Parameters<typeof assignMediaRenderIds>[0]);
-    expect(document.querySelector("video")?.hasAttribute(MEDIA_RENDER_ID_ATTR)).toBe(false);
+    expect(document.querySelector("video")?.getAttribute(MEDIA_RENDER_ID_ATTR)).toBe("no-src");
   });
 
   it("stamps media whose source is a <source> child rather than a src attribute", () => {
@@ -103,10 +118,10 @@ describe("assignMediaRenderIds", () => {
     expect(document.querySelector("audio")?.getAttribute(MEDIA_RENDER_ID_ATTR)).toBe("bed");
   });
 
-  it("ignores a <source> child that carries no src", () => {
+  it("stamps a video whose <source> child carries no src", () => {
     const { document } = parseHTML('<video id="empty"><source type="video/mp4"></video>');
     assignMediaRenderIds(document as unknown as Parameters<typeof assignMediaRenderIds>[0]);
-    expect(document.querySelector("video")?.hasAttribute(MEDIA_RENDER_ID_ATTR)).toBe(false);
+    expect(document.querySelector("video")?.getAttribute(MEDIA_RENDER_ID_ATTR)).toBe("empty");
   });
 });
 
--- a/packages/core/src/compiler/mediaRenderIds.ts
+++ b/packages/core/src/compiler/mediaRenderIds.ts
@@ -43,11 +43,12 @@ export const AUDIO_GROUP_RENDER_ID_ATTR = "data-hf-group-render-id";
 /**
  * Elements the render pipeline addresses by id.
  *
- * `<video>`/`<audio>` are matched whether the source is a `src` attribute or a
- * `<source>` child. Matching only `[src]` left the `<source>`-child shape
- * unstamped, so two scenes each declaring `<video id="clip"><source …></video>`
- * kept colliding ids in the render document, which is exactly the failure this
- * module exists to prevent.
+ * `<video>`/`<audio>` are matched even with an empty `src`. Authors assign the
+ * URL from the scene script (`el.src = url`); the static parse then skips them
+ * and the browser snapshot has to pair the clips. Without a render id on those
+ * elements the snapshot keys by raw id and colliding scenes collapse. `<img>`
+ * still requires a `src` attribute (empty is enough) so we do not stamp every
+ * decorative image.
  */
 const MEDIA_SELECTOR = "video, audio, img[src]";
 
@@ -55,20 +56,10 @@ const MEDIA_SELECTOR = "video, audio, img[src]";
  *  same way. Only an id'd bus can be joined at all. */
 const AUDIO_GROUP_SELECTOR = "hf-audio-group[id]";
 
-/** A `<source>`-bearing media element is addressable even without its own `src`. */
-function hasPlayableSource(el: MediaElementLike): boolean {
-  if (el.getAttribute("src")) return true;
-  const sources = el.querySelectorAll?.("source[src]");
-  if (!sources) return false;
-  for (const _ of sources) return true;
-  return false;
-}
-
 interface MediaElementLike {
   readonly tagName?: string;
   getAttribute(name: string): string | null;
   setAttribute(name: string, value: string): void;
-  querySelectorAll?(selector: string): Iterable<unknown>;
 }
 
 /** A bus or member, which additionally needs subtree scoping to be paired up. */
@@ -108,7 +99,6 @@ export function assignMediaRenderIds(document: DocumentLike): void {
   const pending: MediaElementLike[] = [];
 
   for (const el of document.querySelectorAll(MEDIA_SELECTOR)) {
-    if (!hasPlayableSource(el)) continue;
     const existing = el.getAttribute(MEDIA_RENDER_ID_ATTR);
     if (existing) {
       taken.add(existing);
--- a/packages/producer/src/services/htmlCompiler.test.ts
+++ b/packages/producer/src/services/htmlCompiler.test.ts
@@ -107,6 +107,16 @@ describe("discoverMediaFromBrowser", () => {
     );
     expect(media[0]).toMatchObject({ start: 0, end: 2, duration: 2, mediaStart: 1 });
   });
+
+  it("reports colliding empty-src videos by render id, not author id", async () => {
+    const media = await discover(
+      `<video id="clip" data-hf-render-id="clip" src="" data-start="0" data-end="4" data-media-start="10"></video>` +
+        `<video id="clip" data-hf-render-id="clip__hf2" src="" data-start="4" data-end="8" data-media-start="40"></video>`,
+      {},
+    );
+    expect(media.map((entry) => entry.id)).toEqual(["clip", "clip__hf2"]);
+    expect(media.map((entry) => entry.mediaStart)).toEqual([10, 40]);
+  });
 });
 
 function validTestMediaResponse(): Response {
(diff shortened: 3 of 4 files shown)


james-russo-rames-d-jusso (commented):
🟢 **LGTM from my side — leaving as a comment.**

The fix maps cleanly to #3512's root cause and lines up with every downstream consumer I traced. Small, precise, low blast radius.

### Root-cause fit — yes

#3512 traces the collapse to two facts stacked: (a) #3342 only stamped `data-hf-render-id` on elements that already had a playable source, so `<video id="clip" src="">` + runtime `el.src = url` was skipped; (b) `discoverMediaFromBrowser` then keyed the snapshot by author id, and the two `id="clip"` scenes collapsed onto one window. This PR removes `hasPlayableSource` (`packages/core/src/compiler/mediaRenderIds.ts` — the whole `hasPlayableSource` helper and the `!hasPlayableSource(el) continue;` gate) and threads the render id through the three snapshot consumers (`htmlCompiler.ts:2162`, `:2226`, `:2317`, `:2436`). Both halves of the failure are addressed at their source — no papering-over.

### Bridge is wired everywhere it needs to be

The `__hfMediaEl` resolvers in the two new lookup sites (`htmlCompiler.ts:2226`, `:2317`) rely on `installMediaRenderIdBridge` running before any page script. I checked: the bridge is installed inside `constructCaptureSession` at `packages/engine/src/services/frameCapture.ts:1337`, which is what `createCaptureSession` at `packages/producer/src/services/render/stages/probeStage.ts:309` and `:418` calls. So `session.page` used by all three `discover*` functions always has `__hfMediaEl` available. The `?? document.getElementById(id)` fallback is a safety net for direct-engine callers, but on the actual render path the bridge is guaranteed. Good.

### Cross-consumer consistency — all render-id-aware

Grepped every `data-hf-render-id` / `MEDIA_RENDER_ID_ATTR` consumer and all of them prefer the render id over the raw id:
- `packages/producer/src/services/audioExtractor.ts:73` — `renderIdMatch?.[1] || idMatch?.[1]`, so the static-parse path already emits render ids when present.
- `packages/producer/src/services/renderMediaCollector.ts:96–97` — enumerates by `[data-hf-render-id]`.
- `packages/engine/src/services/videoFrameExtractor.ts:571,645`, `audioMixer.ts:559`, `screenshotService.ts:480,516,533`, `videoFrameInjector.ts:296,322`, `frameCapture.ts:2006`, `core/src/runtime/renderFrameSibling.ts:26–27`, `core/src/audioGroups.ts:101,152` — all prefer render id, fall back to `.id`.

Net: no two-ID-systems split. This PR is the last threading gap being closed.

### Merge into `composition.videos` / `.audios` is coherent

The probeStage merge at `probeStage.ts:481–540` keys `existingVideoIds` on `composition.videos[i].id`. For the failure case in #3512, static parse skips (empty src) so `composition.videos` is empty pre-merge and every browser-discovered entry enters via the "new video" branch keyed by its unique render id (`clip`, `clip__hf2`). No collision. For mixed cases (one scene static-src, one scene runtime-src, same author id), `audioExtractor.ts:73` guarantees the static-side already emits render ids too, so the sets align on the same key namespace. Verified: `compiled.videos` in the new `htmlCompiler.test.ts` at `line 2846` asserts `toHaveLength(0)` for the empty-src case, matching that model.

### Idempotency preserved

`assignMediaRenderIds` at `mediaRenderIds.ts:101–108` still short-circuits on `existing` render id and adds it to `taken`, so the resolved-durations recompile path won't renumber in-flight extractions. Docstring at line 92–96 remains accurate.

### Backwards-compat / scope

Nothing persisted — this is a compile-time attribute stamp on the render document. No migration surface, no cached-artifact invalidation. `` intentionally still requires the `src` attribute (empty allowed), which is documented at `mediaRenderIds.ts:49–51`. Deliberate scoping to video/audio; not a bug.

### CI

All required checks green at `874b323`: `Producer: unit + integration`, `Test`, `Test: runtime contract`, `Preview parity`, `Perf: parity/scrub/load/fps/drift`, `Regression shards 1–9`, `Tests on windows-latest`, `Render on windows-latest`, `Studio: timeline viewport gate`, `Typecheck`, `Lint`, `Format`, `CLI smoke (required)`. `mergeStateStatus=BLOCKED` is just `REVIEW_REQUIRED`; `mergeable=MERGEABLE`, no auto-merge armed.

**No prior peer reviews.** `GET /repos/heygen-com/hyperframes/pulls/3513/reviews` returns `[]`, no issue comments either. This is the first eyes.

### Nits — none blocking

- *Missing audio-analogue test.* `mediaRenderIds.test.ts:78–91` covers the collision case for `<video>` only. `<audio id="bed" src="">` in two nested scenes is symmetric under the widened selector, and the runtime-fallback path in `discoverAudioVolumeAutomationFromTimeline` at `htmlCompiler.ts:2317` is arguably the more delicate of the two consumers (it's the one that would silently return `null` on `getElementById("bed__hf2")` if the bridge weren't installed). A one-liner audio test would pin the audio side of the invariant. Not blocking — the audio path shares its resolver code with video, and manual repro covers it.
- *`discoverAudioVolumeAutomationFromTimeline` has no test that exercises the `__hfMediaEl` branch.* Neither of the two new `window.__hfMediaEl?.(id) ?? …` sites (`htmlCompiler.ts:2226`, `:2317`) has a unit test that passes in a render id and verifies resolution — the existing tests at `htmlCompiler.test.ts:2058+` and `:2263+` predate this PR and use raw ids. Coverage-in-practice comes from `discoverMediaFromBrowser` (`:107–117` in this PR) plus the compileForRender integration test (`:2825`), which is enough given the resolver is a single-line `document.querySelector([${attr}="${id}"])`. But a targeted volume-automation test with a colliding empty-src `<audio>` would close the loop.
- *`` fully-src-less is not covered.* `` (no `src` attribute at all) still skips stamping per `MEDIA_SELECTOR = "video, audio, img[src]"`. The docstring justifies this as avoiding decorative-image stamping, which is reasonable, but if any author does `document.getElementById("x").src = url` on an `` with no static `src`, the same #3512-shaped collision would resurface. Out of scope for this PR — worth a follow-up ticket if the surface exists.

### Question

- *Depth of nesting.* The tests exercise depth-2 nesting (root → two scene hosts). Is there any known composition shape at ≥3 depth where the `assignMediaRenderIds` first-pass `taken` set would produce a suffix collision I'm missing? I don't see one — the loop is order-invariant on the `existing`-first pass and `uniqueRenderId` at `mediaRenderIds.ts:83–88` handles arbitrary depth — but calling it out in case there's a producer path where the compiled document is a concatenation of independently-stamped sub-documents rather than one `querySelectorAll` sweep.

### What I didn't verify

- Manual reproduction against `ArcadeHQ/hyperframes-repros@patch/runtime-src-duplicate-ids`. Val ran it (test-plan checkbox); I trusted the frame-level assertion in the issue and the CI regression shards passing.
- The bus/audio-group path is untouched here — I confirmed `AUDIO_GROUP_SELECTOR` still requires `[id]` (unchanged from before), and buses aren't the failure surface of #3512. If runtime-src is ever a thing on buses, that's a separate module.

Ready to ship from where I sit. Straightforward P0 close, tests in the right places, no observable seams.

— _Review by Rames D Jusso_

jrusso1020 (approved):
Approving. Read the four files plus the consumers at `874b323e`; Rames D Jusso's review is the substantive first pass and I agree with it, so this is additive rather than a restatement.

**The axis I checked independently: membership, not preference.**

That review verifies every downstream consumer *prefers* the render id over the author id. The behavior change here is different in kind — dropping `hasPlayableSource` means elements that previously carried **no attribute at all** now carry one, and `assignMediaRenderIds` even mints a positional `hf-media-${taken.size}` for an element with no author id (`mediaRenderIds.ts:114-118`). So a decorative source-less `<video>` is newly stamped where before it was invisible. The question that raises is whether anything enumerates *by* the attribute and would now pick up phantom media.

Traced all four consumers the diff touches or implicates:

- `discoverMediaFromBrowser` enumerates `video[data-start], audio[data-start], img[data-var-src]` (`htmlCompiler.ts:2140-2147`) — gated on the timing compiler's attribute, not on `data-hf-render-id`.
- `discoverVideoVisibilityFromTimeline` enumerates `video[data-hf-auto-start]` (`:2398`) — same shape.
- `discoverAudioVolumeAutomationFromTimeline` is driven by a caller-supplied `ids` array, so it has no enumeration of its own.
- `renderMediaCollector.ts:96` *does* enumerate `[MEDIA_RENDER_ID_ATTR]`, but `collectHostWindows` builds a **keyed lookup map**. Extra keys are dead entries, never an output list.

So the widened stamp cannot introduce phantom clips on any path, and the `hf-media-N` fallback is unreachable in effect for the same reason — an element the timing compiler never processed has neither `data-start` nor `data-hf-auto-start`. Membership is safe.

**Bridge chain — confirmed independently at source**, since the two new `window.__hfMediaEl?.(id)` sites fail soft and a missing bridge would degrade silently (render id `clip__hf2` → `getElementById` miss → clip dropped from volume automation, quieter than the bug being fixed): `probeStage.ts:309` and `:418` → `createCaptureSession` (`frameCapture.ts:1145`) → `constructCaptureSessionWithRollback:1246` → `constructCaptureSession:1286` → `installMediaRenderIdBridge:1337`, unconditional and before navigation. Same conclusion, reached separately.

**PR-body claims spot-checked against the diff:** "author `id`s are still not rewritten" holds — the loop only ever calls `setAttribute(MEDIA_RENDER_ID_ATTR, …)`. "`__hfMediaEl` already installed on the probe page" holds per the chain above.

**On CI, one correction to the record.** Both relays here say "all green," and that is true — but only of each check's *latest* attempt. There are 97 runs across 50 names at this head, and earlier attempts of `Test` and `regression` are `failure` with most others `cancelled`. A naive per-name query returns those stale attempts. The rollup is `SUCCESS` and all eight ruleset-required checks (`Semantic PR title`, `Test: runtime contract`, `Typecheck`, `Build`, `regression`, `Test`, `Render on windows-latest`, `Tests on windows-latest`) are green on latest. Flagging because I mis-read it that way myself first — the re-runs are real and the summary is only true with "latest attempt" attached.

Nits in the prior review (audio-analogue test, untested `__hfMediaEl` branch, id-less ``) are fair and non-blocking; I'd add the audio collision test as the one with real value, since the audio resolver is the fail-soft path.

— Rames

PR #3517, opened 2026-08-27
feat(studio): add studio_inspect, so an agent reads before it writes

miguel-heygen:
## What

`studio_inspect` returns everything about one element in a single call: resolved styles, text fields, box, data attributes, GSAP animations, and what the element will and will not accept.

Also extracts the shared WebMCP test fixtures into `webmcpTestUtils.ts`, which removes the duplication that was accumulating across the tool test files and the `as HTMLElement` casts that came with it.

Stacked on #3516.

## Why

The point is to prevent a failed write, not to satisfy curiosity. `can.reasonIfDisabled` comes through verbatim from Studio's own capabilities, so an agent that reads first should never attempt an edit the element would refuse. That matters more here than in a normal API, because a refused write in Studio is often silent.

## How

Three things it refuses to get wrong:

**Animations are reported only for the current selection**, because that is the only element Studio parses them for. Attributing them to any other element would be reporting the wrong element's motion, which is worse than reporting none. When a handle names something else the list is empty and `animationEditingBlocked` says why.

**`animationEditingBlocked` also carries the two states where animation editing is off entirely**, multiple timelines and an unsupported timeline pattern. Both live on the selection context. Learning them from a read costs one call; learning them from a failed write costs a retry loop.

**Inspecting does not change the selection.** It is a read, and stealing the human's selection would be a side effect they did not ask for. There is a test asserting `applySelection` is never called.

Nothing selected and no handle given is a failure, not an empty result. An empty result would assert "this element has nothing", which is a different and false claim.

## Test plan

8 tests in `inspectTools.test.ts`, and the whole WebMCP suite re-run after the fixture extraction:

- Resolved styles come back, not authored ones.
- Capabilities and `reasonIfDisabled` pass through verbatim.
- Animations list for the current selection, with ease and id.
- Both `animationEditingBlocked` reasons are asserted separately.
- **A handle naming a different element gets an empty animation list and a reason**, rather than the selection's animations.
- Inspecting a handle does not call `applySelection`.
- No selection and no handle is a failure with a hint, not an empty result.
- Unknown handle and unmounted preview produce distinct reasons.

WebMCP suite 78 passing. Full package suite 4563 passing across 411 files. `bunx tsc --noEmit` clean, `bunx oxlint` clean, `bunx fallow audit --fail-on-issues` clean.

Still no browser verification anywhere in this stack; that is the final unit.

### DIFF
--- a/packages/studio/src/webmcp/StudioAgentTools.tsx
+++ b/packages/studio/src/webmcp/StudioAgentTools.tsx
@@ -18,7 +18,12 @@ import type { StudioLookSnapshot } from "./tools/lookTools";
  */
 export function StudioAgentTools() {
   const { projectId, activeCompPath, editHistory } = useStudioShellContext();
-  const { domEditSelection, selectedGsapAnimations } = useDomEditSelectionContext();
+  const {
+    domEditSelection,
+    selectedGsapAnimations,
+    gsapMultipleTimelines,
+    gsapUnsupportedTimelinePattern,
+  } = useDomEditSelectionContext();
   const { previewIframeRef, buildDomSelectionFromTarget, applyDomSelection } =
     useDomEditActionsContext();
 
@@ -71,6 +76,12 @@ export function StudioAgentTools() {
         }
       },
       wait: (ms) => new Promise((resolve) => setTimeout(resolve, ms)),
+      getCurrentSelection: () => domEditSelection,
+      getGsapDiagnostics: () => ({
+        animations: selectedGsapAnimations,
+        multipleTimelines: gsapMultipleTimelines,
+        unsupportedTimelinePattern: gsapUnsupportedTimelinePattern,
+      }),
     }),
     [
       getSnapshot,
@@ -79,6 +90,10 @@ export function StudioAgentTools() {
       applyDomSelection,
       projectId,
       activeCompPath,
+      domEditSelection,
+      selectedGsapAnimations,
+      gsapMultipleTimelines,
+      gsapUnsupportedTimelinePattern,
     ],
   );
 
--- a/packages/studio/src/webmcp/tools/frameTools.test.ts
+++ b/packages/studio/src/webmcp/tools/frameTools.test.ts
@@ -1,7 +1,7 @@
 // @vitest-environment jsdom
 import { describe, expect, it, vi } from "vitest";
 import { studioFrame, type FrameToolDeps, type StudioFrameResult } from "./frameTools";
-import type { ToolFailure, ToolResult } from "../toolResult";
+import { expectFailure, expectOk } from "../webmcpTestUtils";
 
 function frameDeps(overrides: Partial<FrameToolDeps> = {}): FrameToolDeps {
   return {
@@ -13,16 +13,6 @@ function frameDeps(overrides: Partial<FrameToolDeps> = {}): FrameToolDeps {
     wait: async () => undefined,
     ...overrides,
   };
-}
-
-function expectOk<T>(result: ToolResult<T>): { ok: true } & T {
-  if (!result.ok) throw new Error(`expected ok, got ${JSON.stringify(result)}`);
-  return result;
-}
-
-function expectFailure(result: ToolResult<unknown>): ToolFailure {
-  if (result.ok) throw new Error(`expected failure, got ${JSON.stringify(result)}`);
-  return result;
 }
 
 describe("studioFrame", () => {
--- a/packages/studio/src/webmcp/tools/inspectTools.test.ts
+++ b/packages/studio/src/webmcp/tools/inspectTools.test.ts
@@ -1,0 +1,197 @@
+// @vitest-environment jsdom
+import { describe, expect, it, vi } from "vitest";
+import type { GsapAnimation } from "@hyperframes/parsers/gsap-parser";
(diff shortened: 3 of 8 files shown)


terencecho (approved):
**APPROVE** at `1478adfacb013580c7839f08bacc9c8f684d2052`. The design (read-before-write, single-call, verbatim `reasonIfDisabled`, no selection theft) is well-crafted; test coverage locks the risky bits; the fixture extraction is a clean net-loss of duplication.

**1. `animationEditingBlocked` precedence is right.**

The `!isCurrentSelection` branch fires before `multipleTimelines` / `unsupportedTimelinePattern`. That's the correct precedence: for a handle that isn't the current selection, `animations: []` is right regardless of the composition's timeline shape, and "only readable for the current selection" is the actionable reason for the caller (switch selection to learn more). The test parametrizes the two current-selection reasons separately; the not-current path is asserted at line 236-237. ✓

**2. Handle-resolves-to-current-selection identity check.**

`describe(selection, deps, current?.element === element)` — DOM reference equality across the same iframe realm. The `previewElement` fixture creates the element inside a real iframe (`previewDoc().defaultView.HTMLElement` check), which is the realm the runtime cares about. Reference equality on a preview node from the same doc holds. ✓

One coverage gap worth naming (non-blocker): 7 tests exercise `isCurrentSelection=false` (via `getCurrentSelection: () => null` default) or the no-handle path. There's no test that walks the branch `input.handle` given AND `current?.element === element` → `isCurrentSelection=true`. If a future refactor breaks the identity check (e.g. `buildSelection` starts returning a wrapper element), the tests wouldn't catch it. A single test — pass `getCurrentSelection: () => selectionFor(x)` alongside `handle: <handle-for-x>` and assert `ok.isCurrentSelection === true` — would pin it.

**3. Fixture extraction: no test drift.**

The old `selectionFor` in `selectionTools.test.ts` returned minimal defaults (`textFields: []`, empty style dicts). The extracted `webmcpTestUtils.ts` version has richer defaults (one text field, one data-attr, one inline style, resolved styles including `font-size: 42.7px`). Since `bunx tsc --noEmit` is clean and CI is green at head, the existing `selectionTools.test.ts` and `frameTools.test.ts` don't assert on those defaults being empty. Safe swap.

The `expectOk`/`expectFailure` helpers changed semantics slightly (throw → vitest `expect(...).toBe(...)`). Failing-test error messages will look different, but pass/fail behavior is preserved. Non-issue.

**4. Race window on `getCurrentSelection` — narrow, no fix needed.**

```ts
const current = deps.getCurrentSelection();   // captured
// ... later ...
const selection = await deps.buildSelection(element);   // awaits
// ...
return describe(selection, deps, current?.element === element);   // still uses captured `current`
```

If the human's selection changes during the `buildSelection` await, `isCurrentSelection` reports the pre-await state, not the post-await state. Practical impact: negligible. Under-report (was current, isn't now) is safer than over-report; over-report (is current now, wasn't captured) means the agent gets `animations: []` when it could have had them — same user retries. Not worth a fix; noting because a future refactor might tempt someone to hoist the check post-await.

**5. Data-attribute pass-through — Studio-side responsibility.**

`dataAttributes: selection.dataAttributes` forwards whatever Studio surfaces. If Studio's data-* extraction ever picks up sensitive fields (session tokens, PII embedded in data-* attrs), the model sees them. Not this PR's problem — the concern belongs at the Studio capabilities layer where the selection is built. Naming here so it's on record for whoever owns that surface.

**6. Confirmed as claimed:**

- **`applySelection` never called:** vi.fn at line 251 asserts. The whole `studioInspect` path takes `selection` (from `buildSelection`) or `current` directly and hands it to `describe()`; there's no code path that calls `deps.applySelection(...)`. Explicit test pin locks it against a future accidental re-wiring. ✓
- **Distinct error kinds:** `blocked` vs `invalid` are separately asserted; `notMounted.reason !== unknown.reason` prevents them from collapsing into the same string. Test at line 264-274. ✓
- **`isCurrentSelection: false` gives empty animations + reason:** line 220-238. ✓
- **`toolFailure("invalid", ...)` with hint** on empty-input path — the `hint` cites `studio_look` / `studio_select`, both discoverable via `tools/list`. ✓
- **`STUDIO_INSPECT_DESCRIPTION` opens with "Read this BEFORE editing"** — that's the model-facing selection hint that makes read-before-write the default. Tool descriptions punch above server instructions in tool-selection; this one earns its keep. ✓
- **Test count assertion updated 4 → 5:** `useStudioAgentTools.test.tsx` at lines 585-604 — locks tool count so a future accidental drop of `studio_inspect` from the registration list fails a test. ✓
- **`annotations: { readOnlyHint: true, untrustedContentHint: true }`:** correct for a tool that returns arbitrary DOM data. ✓
- **Registration idempotence still asserted** across re-renders at test line 592-604 — five tools registered once, not five-times-per-rerender. ✓

**Stack context:** #3517 stacked on #3516, which is APPROVED and CLEAN. `bunx tsc --noEmit`, `bunx oxlint`, `bunx fallow audit --fail-on-issues` all clean per PR body. CI at head: Preflight, Preview parity, WIP, player-perf, preview-regression, regression, Detect changes — all green. Stamping.

_Review by tai (pr-review)_

PR #3518, opened 2026-08-27
feat(studio): let an agent edit text and styles, guarded

miguel-heygen:
## What

`studio_set_text` and `studio_set_style`, the first tools that change the composition.

Also plumbs the write-blocked state, which was the blocker for shipping any write at all.

Stacked on #3517.

## Why

Four read tools shipped before this one on purpose. An agent that can see the composition but not touch it is only half useful, but a tool that writes without a guard is worse than none, because Studio has several paths where a failed write reports success.

## How

**Both act on the current selection and take no handle.** That is forced rather than chosen: the handlers read the ambient React selection, and `applyDomSelection` only schedules a state update, so selecting and committing inside one call would write to whatever was selected *before*. Select first, then edit.

**The write guard.** `domEditSaveQueuePaused` and the external-file conflict both lived on `App` and were unreachable from the tool surface, so `canWrite` was optimistic and the code said so in a comment. They now derive into a single `writeBlockedReason` on the shell context: one field, one owner, conflict taking precedence because resolving it is what unblocks the queue.

That guard matters more than it looks. Both states are **banners** in Studio with no lock behind them, so nothing else was stopping a programmatic write from landing on top of a conflict the user had been asked to adjudicate.

Three things the tools refuse to fake:

*They check the outcome, not the absence of a throw.* Awaiting the handler proves nothing here; the tagged outcome added in #3510 is what proves the write landed.

*A partial style result is reported as partial.* `handleDomStyleCommit` is one property per call, so N properties are N commits. The result carries `applied` and `rejected` maps rather than one boolean that would have to pick a side.

*Style commits run sequentially, never concurrently.* Two commits racing through Studio's client-side read-modify-write can record undo entries that both claim the same starting content.

Every decline reason maps to a hint naming what to do instead, so a refusal routes the agent rather than just stopping it.

## Test plan

14 tests in `contentTools.test.ts`:

- Text writes and reports the result, including `changed: false` when it already said that.
- **A blocked write never reaches the handler**, asserted for both the conflict and paused-queue reasons.
- A declined commit is not reported as success, which is the whole point of the outcome plumbing.
- Decline reasons map to hints naming the alternative tool.
- **Concurrency is measured, not assumed**: the sequential-commit test tracks in-flight count and asserts it never exceeds 1.
- Partial style success reports `applied` and `rejected` separately.
- Empty and non-object `styles` are rejected without committing.

Full package suite 4577 passing across 412 files. `bunx tsc --noEmit`, `bunx oxlint` and `bunx fallow audit --fail-on-issues` all clean.

Still no browser verification anywhere in this stack. That is the final unit.

## Note

The push for this branch raced a full disk; the commit and remote ref landed, the local upstream config did not and was repaired afterwards. Nothing in the diff is affected.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "hyperframes",
   "description": "HyperFrames by HeyGen. Write HTML, render video. Compositions, GSAP and runtime adapter animations, captions, voiceovers, audio-reactive visuals, and website capture for HyperFrames.",
-  "version": "0.8.16",
+  "version": "0.8.20",
   "author": {
     "name": "HeyGen",
     "email": "hyperframes@heygen.com",
--- a/.codex-plugin/plugin.json
+++ b/.codex-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "hyperframes",
   "description": "Write HTML, render video. Compositions, Tailwind v4 styles, GSAP and runtime adapter animations, captions, voiceovers, audio-reactive visuals, and website capture for HyperFrames.",
-  "version": "0.8.16",
+  "version": "0.8.20",
   "author": {
     "name": "HeyGen",
     "email": "hyperframes@heygen.com",
--- a/.cursor-plugin/plugin.json
+++ b/.cursor-plugin/plugin.json
@@ -3,7 +3,7 @@
   "name": "hyperframes",
   "displayName": "HyperFrames by HeyGen",
   "description": "Write HTML, render video. Compositions, Tailwind v4 styles, GSAP and runtime adapter animations, captions, voiceovers, audio-reactive visuals, and website capture for HyperFrames.",
-  "version": "0.8.16",
+  "version": "0.8.20",
   "author": {
     "name": "HeyGen",
     "email": "hyperframes@heygen.com"
--- a/.github/workflows/player-perf.yml
+++ b/.github/workflows/player-perf.yml
@@ -107,6 +107,13 @@ jobs:
         if: matrix.shard == 'parity'
         uses: ./.github/actions/install-ffmpeg-linux
 
+      - name: Verify sandbox origin boundary (load shard only)
+        if: matrix.shard == 'load'
+        working-directory: packages/player
+        env:
+          PUPPETEER_EXECUTABLE_PATH: ${{ steps.setup-chrome.outputs.chrome-path }}
+        run: bun run test:browser-security
+
       - name: Run player perf — ${{ matrix.shard }} (measure mode)
         working-directory: packages/player
         env:
--- a/docs/changelog.mdx
+++ b/docs/changelog.mdx
@@ -9,6 +9,77 @@ Recent HyperFrames releases, including user-facing features, fixes, and migratio
 {/* New release entries are prepended by `bun run changelog:draft <version> --write`. */}
 
 <Update
+  label="HyperFrames v0.8.20"
+  description="Released - 2026-08-30"
+  tags={["Release", "Studio"]}
+>
+Studio now separates blocked direct edits from real save failures and keeps the file and HTTP details needed to diagnose them.
+Repeated failures are grouped, and write tracking now works in more secure browser contexts.
+
+## Fixes
+
+- **Studio:** Correct save failure telemetry ([28be8dddfa](https://github.com/heygen-com/hyperframes/commit/28be8dddfacefe29b5e25ae2d36fd76ffe4c4852), [#3499](https://github.com/heygen-com/hyperframes/pull/3499)).
+
+[View the full commit range](https://github.com/heygen-com/hyperframes/compare/v0.8.19...v0.8.20).
+</Update>
+
+<Update
+  label="HyperFrames v0.8.19"
+  description="Released - 2026-08-29"
+  tags={["Release", "Studio"]}
+>
+Export in Studio now renders the composition you have selected, not the project's root file. Multi-part projects can finally be exported without dropping to the CLI.
+
+## Fixes
+
+- **Studio:** Export the composition the user has selected ([b71f45981](https://github.com/heygen-com/hyperframes/commit/b71f45981c376d3b2b4ae8e07a0837f63a091a8c), [#3550](https://github.com/heygen-com/hyperframes/pull/3550)).
+
+[View the full commit range](https://github.com/heygen-com/hyperframes/compare/v0.8.18...v0.8.19).
+</Update>
+
+<Update
+  label="HyperFrames v0.8.18"
+  description="Released - 2026-08-29"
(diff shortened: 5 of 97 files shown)


vanceingalls (dismissed):
Style holds up under adversarial read. Guard model is single-owner (`writeBlockedReason` derived on the shell context with conflict-first precedence over paused-queue), and it is enforced BEFORE dispatch, so both banner states now behave as locks for programmatic writes. Outcome-tagged writes (not "no throw") plus the partial-success `applied` / `rejected` maps plus measured sequential dispatch (in-flight counter asserted in tests) match the tagged-outcome contract set up in #3510 and preserve undo integrity through N-property style commits.

Schema shape mirrors #3519: `additionalProperties: false`, explicit `typeof` guards for `text` and `styles`, non-object / null / array styles refused, and empty-entries refused so an "empty commit" cannot fake success. Decline reasons are mapped to actionable hints (routes the agent instead of just stopping it), and `geometry-property` is refused with a pointer to the transform tools.

The intentional decision to route through `handleDomTextCommit` / `handleDomStyleCommit` rather than add a second sanitizer at the tool boundary is consistent with the "route to Studio's real editing paths, don't reinvent them" premise across the stack — the guarded surface here is around conflict-state, outcome verification, and partial application, not value filtering.

Reviewed head be6fc0b0

_Review by Via_

somanshreddy (commented):
**Second-eyes pass (alongside @Via), focused on the guarded-write mechanism at `be6fc0b0`.** Direct manual review — I deliberately did *not* spawn a full second automated pass because the PR is `DIRTY` and mid-stack; the diff that finally ships will differ from this head (see the state note), so a line-by-line pass belongs after the rebase. What I verified is the core mechanism, which is stable regardless.

## Core write mechanism — sound, verified at source
- **The guard closes the real gap.** `writeBlockedReason = externalFileConflict ? "…" : domEditSaveQueuePaused` (`useStudioContextValue.ts`) — conflict takes precedence, derived into one shell-context field, wired into the tool via `getWriteBlockedReason` (`StudioAgentTools.tsx`) and checked before dispatch (`guardWrite`). Both states were banners with no lock behind them, so this is what actually stops a programmatic write landing on a conflict the user is still adjudicating. Correct and well-motivated.
- **Sequential style commits are genuinely sequential.** `for (…) { await deps.setStyle(…) }` — in-flight never exceeds 1, and the comment names the failure it prevents (two client-side read-modify-write commits recording undo entries that both claim the same starting content). The concurrency test that tracks in-flight count is the right shape.
- **Outcome-checked, not throw-checked.** `fromOutcome` keys on `outcome.ok` and maps each decline reason to an actionable hint; partial style results report `applied`/`rejected` separately rather than collapsing to one boolean. Honest.
- **Interleaved human/agent selection race** (the class-of-bug from the #3515 review): it's an inherent, *documented* property of the shared-selection design — the guard reads the current selection at dispatch and doesn't lock it, and the file's own docstring calls out "select first, then edit, two calls separated by a render." I agree with @Rames D Jusso that this is an integration-test target, not a blocker.

## State / hygiene — worth resolving before the final pass (my main finding)
This PR is `DIRTY` and the diff is **+2746/-249 across 75 files**, not the ~+999/-70 the write surface actually is. Two things are tangled in:
1. **Unrelated render-pipeline changes** ride along — `producer/artifactTransaction` (+177), `core/colorGrading`, `engine/frameCapture`, `cli/snapshot`, release notes v0.8.17–20. These aren't the WebMCP write tools; they read as diff-noise from a base (`feat/studio-webmcp-frame`) that's behind main.
2. **`contentTools.ts` here is the pre-fix version** — `field = … : undefined`, the exact studio_set_text field-resolution bug that #3521 fixed and that has already merged to main. So a rebase onto main will conflict in `contentTools.ts`, and merging this head as-is would reintroduce the child-field write bug (empty patch → server 400 `persist-failed`).

Recommend rebasing onto current main first: the diff should shrink to the real write surface and `contentTools.ts` should re-converge with #3521's field resolver. That reconverged diff is what I'd want a final pass to read. No concern with the mechanism itself — it's the packaging.

*(Not blocking on my end — this is second eyes; @Via holds the primary pass. Tests weren't run locally; per the PR body the package suite is green and the guard/concurrency tests are mutation-meaningful.)*

vanceingalls (approved):
R2 verify at 80787832 — clean post-#3516-merge base retarget. PR-surface (packages/studio/src/webmcp/**) is byte-identical to prior stamp head be6fc0b0 — GitHub compare be6fc0b0…80787832 shows ahead-by-7 with zero studio files changed; the 7 upstream commits touch packages/core/runtime and packages/player only (main advanced, not the PR). Now +882/-7 across 10 files, all studio-webmcp/App/Context — matches the guard-mechanism scope I signed off on. (Miga cited +861/-7; actual is +882/-7, but the delta is upstream-only, not on this PR's surface.) CI green at HEAD. Stamping.

_Review by Via_

somanshreddy (commented):
**Re-review at the reconverged head `8078783233`** (base now `main`, diff `+882/-7` vs main). This is the full line-pass I deferred when the PR was mid-rebase/DIRTY — done now on the clean merge base. Manual pass (not codex).

**The revert risk I flagged earlier is cleared.** `contentTools.ts` and `transformTools.ts` don't exist on `main`, so they're introduced fresh here — there's no prior main version for the reconverge to have reverted. The 7 deletions are all benign wiring/test-count churn (context lines in `useStudioAgentTools`, the `StudioAgentToolsDeps` interface extension, and the `toHaveBeenCalledTimes(5)` assertions superseded by `8`), not a lost fix.

**The write tools are correctly guarded and anti-vacuous:**
- `guardWrite`/`guard` check `getWriteBlockedReason()` then a live selection *before* any dispatch, and return typed `blocked`/`invalid` failures.
- `studioSetText`: the optional `field` resolves to `undefined` when absent (→ the element's own text, per the schema description) — correct, not an empty-patch; `changed` is computed against the captured `before`.
- `studioSetStyle`: rejects an empty/non-string-valued styles object ("an empty commit would report success having done nothing") and commits **sequentially**, one property per call, with a comment explaining why (racing client-side read-modify-write would corrupt undo entries). Collects `applied`/`rejected`.
- `studioTransform`: paired-field validation (must-be-given-together / min), and fails `blocked` when `applied.length === 0` rather than reporting a no-op success.
- Registration: 3 new tools wired through `runToolBody`; the count assertion moved 5→8 and a not-ready case asserts `registerTool` isn't called. `contentTools.test.ts` (+201) and `transformTools.test.ts` (+179) exercise the guard/anti-vacuity/outcome paths.

No blocking findings — the reconverged diff is clean and matches the stable-core mechanism I verified on the prior head.

**Verdict: approve-pending-CI.** Standard checks are still running at this SHA (Typecheck/Test/CLI smoke pending, not red). I'll flip this to an approval the moment CI is green — ping me when it lands or I'll catch it, and I won't stamp over a pending required check.

PR #3519, opened 2026-08-27
feat(studio): move, resize and rotate, verified by reading back

miguel-heygen:
## What

`studio_transform`: move, resize and rotate the selected element, the way a drag does.

Stacked on #3518.

## Why

This is the "do what a mouse does" half of the tool set, and it is the one the plan flagged as most likely to ship a confident lie.

## How

**The result's `box` is read back after the write, never echoed from the request**, and `applied` lists what actually took effect.

That is not belt-and-braces. This unit's plan said to re-derive the geometry handlers' behaviour rather than trust any description of them, including the plan's own. Doing that turned up three different behaviours behind one interface:

1. The handlers on `DomEditActionsValue` are the **GSAP-aware wrappers**, aliased at `useDomEditSession.ts:534-538`, not the CSS ones in `useDomGeometryCommits.ts` that an earlier note in this workstream described.
2. `handleGsapAwarePathOffsetCommit` and `handleGsapAwareRotationCommit` are `if (gsapCommitMutation) { ...intercept... }` **with no else branch**. Their own comments say that is deliberate: position and rotation are written as GSAP code and there is no CSS fallback. So they can return having done nothing.
3. `handleGsapAwareBoxSizeCommit` is different again. It runs through `runGestureTransaction` with separate scale and width/height routes, so resize works more generally.

Reading back is what turns case 2 from a silent lie into a reported one: a move that did nothing comes back in `unchanged` with a reason and a hint, not as success.

Three smaller decisions:

- **Operations re-read between each other**, so a move is judged against the box *after* a resize in the same call. Comparing against the original would credit the resize's change to the move.
- **Rotation is reported as dispatched, not verified.** `rotate` is an individual transform property and does not appear in the computed transform, so there is no honest box-derived signal. Claiming one would be worse than saying so.
- **x pairs with y, width pairs with height.** Accepting one alone means inventing the other from the current value, which moves the element somewhere the caller did not ask for.

## Test plan

10 tests in `transformTools.test.ts`:

- The reported box comes from the read-back, proven with a handler that lands somewhere other than requested.
- **A silent no-op is reported as `unchanged`, not success** — the handler is called, the box does not move, and the tool fails with a GSAP hint.
- A call that resizes but fails to move separates the two in `applied` and `unchanged`.
- The re-read between operations is asserted, so a move is not credited with the resize's change.
- Rotation reports as dispatched.
- A blocked write never reaches a handler.
- Pairing, negative size, empty request and non-finite input all reject before dispatch.

Full package suite 4587 passing across 413 files. `bunx tsc --noEmit`, `bunx oxlint` and `bunx fallow audit --fail-on-issues` all clean. Fallow initially failed this one on complexity (24 cyclomatic) and the parsing was extracted rather than suppressed.

## What is still owed

Case 2 above is derived from reading the source, not from watching it in a browser. The read-back means the tool reports honestly either way, but confirming the actual behaviour against a live GSAP composition and a static one is part of the end-to-end capture in the final unit.

### DIFF
--- a/packages/studio/src/webmcp/StudioAgentTools.tsx
+++ b/packages/studio/src/webmcp/StudioAgentTools.tsx
@@ -30,6 +30,9 @@ export function StudioAgentTools() {
     applyDomSelection,
     handleDomTextCommit,
     handleDomStyleCommit,
+    handleDomPathOffsetCommit,
+    handleDomBoxSizeCommit,
+    handleDomRotationCommit,
   } = useDomEditActionsContext();
 
   const getSnapshot = useCallback((): StudioLookSnapshot => {
@@ -85,6 +88,15 @@ export function StudioAgentTools() {
       getWriteBlockedReason: () => writeBlockedReason,
       setText: (value, fieldKey) => handleDomTextCommit(value, fieldKey),
       setStyle: (property, value) => handleDomStyleCommit(property, value),
+      // Measured, not authored: the tool compares this before and after to
+      // tell a real change from a handler that did nothing and resolved.
+      readBox: (selection) => {
+        const rect = selection.element.getBoundingClientRect();
+        return { x: rect.x, y: rect.y, width: rect.width, height: rect.height };
+      },
+      moveTo: (selection, next) => handleDomPathOffsetCommit(selection, next),
+      resizeTo: (selection, next) => handleDomBoxSizeCommit(selection, next),
+      rotateTo: (selection, next) => handleDomRotationCommit(selection, next),
       getGsapDiagnostics: () => ({
         animations: selectedGsapAnimations,
         multipleTimelines: gsapMultipleTimelines,
@@ -101,6 +113,9 @@ export function StudioAgentTools() {
       writeBlockedReason,
       handleDomTextCommit,
       handleDomStyleCommit,
+      handleDomPathOffsetCommit,
+      handleDomBoxSizeCommit,
+      handleDomRotationCommit,
       domEditSelection,
       selectedGsapAnimations,
       gsapMultipleTimelines,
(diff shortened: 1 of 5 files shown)


vanceingalls (approved):
**Verdict:** APPROVE

The read-back architecture is the right response to the silent-lie hazard, and the analysis behind it — spotting that `DomEditActionsValue`'s handlers are the GSAP-aware wrappers rather than the CSS ones, that two of the three have no else branch and can resolve having done nothing, and that CSS `rotate` escapes `getComputedStyle().transform` so honest verification is only possible for two of the three operations — is exactly the kind of ground-truth-first work these tool PRs need.

Positives worth calling out:
- Sequential re-reads between ops (`transformTools.ts:378-401`) so a move is judged against the box left by the resize, not the original. The test at `transformTools.test.ts:135` locks this in and would fail loudly if anyone flattened the reads to a single pre-call snapshot.
- Pairing on x/y and width/height (`parsePair`), refusing a single-axis call rather than inventing the missing axis from the current value. Small decision, avoids a class of silent semantic bugs.
- `additionalProperties: false` on the schema plus strict `typeof === "number" && Number.isFinite` on every field, with `min = 0` on size. No coercion, no extra props, no NaN reaching a handler.

Minor observations, none blocking:
- Strict-equality box comparison will read "did not move" on a redundant same-position write (agent asks for `{x, y}` matching current). The handler ran, but `unchanged.move` fires with the GSAP hint anyway. Acceptable — an inspect call reveals the truth — and the alternative (an out-of-band signal from the handler) is exactly the gap the PR body flags as owed to the end-to-end capture.
- `parsePair`'s default `min = Number.NEGATIVE_INFINITY` makes the `< min` branch a tautology for the x/y call site, correct only because size passes explicit `min = 0`. Reads cleanly for now, worth a comment if this file grows.
- The return-time = landed-time contract from #3516 carries the whole read-back trap here — the handler mocks in these tests are synchronous, so the real settle-before-probe behaviour is verified in the frame PR's contract, not this one. Consistent with how the PR body owns that verification for the live GSAP composition.

_Reviewed head b4123be1_

_Review by Via_

PR #3520, opened 2026-08-27
feat(studio): let an agent author motion

miguel-heygen:
## What

Four tools that author motion: `studio_add_animation`, `studio_update_animation`, `studio_add_keyframe`, `studio_delete_animation`.

Stacked on #3519. This completes the tool set; only the docs page and the browser proof remain.

## Why

Motion is the one thing an agent cannot judge or author from source. Reading GSAP code tells you the tween exists; it does not tell you the title lands a beat late or the ease feels mushy. Paired with `studio_seek` and `studio_frame` from earlier in the stack, these close that loop.

## How, and why these are less confident than the rest

The handlers underneath mostly cannot report back:

| Handler | Signal |
|---|---|
| `handleGsapAddAnimation(method)` | none. Position comes from the live playhead, and the call is `void ...catch()` |
| `handleGsapAddKeyframeBatch` | returns a promise but catches its own failure |
| `handleGsapDeleteAnimation` | discards its promise entirely |
| `handleGsapUpdateMeta` | **returns a boolean.** The one honest one |

#3519 solved the same problem by reading the result back. **That does not work here**: the animation list comes from React state that only refreshes on a render, and no render happens inside one tool call.

So rather than fake a verification with a frame-timer, these report what was **dispatched**, and every description tells the agent to call `studio_inspect` to see the result. Saying "I asked for this" is honest; saying "this happened" would not be.

Three consequences worth stating:

**`studio_add_animation` takes no position.** The handler reads the playhead itself, so accepting one would report a number that had no effect. It reports where the playhead actually was and tells the agent to `studio_seek` first.

**`studio_update_animation` rules out the no-selection case before dispatch.** The handler answers `false` for both "nothing selected" and "the write failed". Eliminating one beforehand is what makes the other legible, and it is why a `false` from this tool is reported as a real failure with a stale-id hint.

**Keyframe percent and properties are validated in the tool**, because nothing in the platform checks input against the declared `inputSchema`.

## Test plan

15 tests in `animationTools.test.ts`:

- Add reports the **playhead's actual value**, proven with a playhead at 7.25 while the caller supplies nothing.
- Add is marked `dispatched`, not claimed as landed.
- Update confirms via the boolean, and a `false` becomes a real failure with a stale-id hint.
- **The no-selection case is asserted to be caught before dispatch**, so the ambiguous `false` can never be reached that way.
- Keyframe properties go through in one call, so one undo entry.
- Percent is rejected at -1, 101, NaN and `"50"`; properties rejected for `{}`, `{y: null}`, `[]` and a string; 0 and 100 accepted as the ends.
- Blocked writes and unknown methods never reach a handler.

Full package suite 4602 passing across 414 files. `bunx tsc --noEmit`, `bunx oxlint` and `bunx fallow audit --fail-on-issues` all clean.

## Still owed

Nothing in this stack has been exercised in a real browser. That is the last unit, and it needs `chrome://flags/#enable-webmcp-testing`.


terencecho (dismissed):
**APPROVE** on the motion commit `90cef4f251393a8b9fe3c96188fa2c6ae36c74c3` (the true delta this PR owns). Reading the head SHA `d7a692c9` (which folds in the already-merged #3517/#3519/#3521 layers due to the pending rebase); scope of this stamp is the motion commit only. Two design calls are worth stating out loud, plus one silent-coercion smell that's the only real nit.

**1. Dispatched-not-landed is the right shape here — and it's honest about which handler is different.**

The docstring's four-way handler table is the point:

| Handler | Signal actually available |
|---|---|
| `handleGsapAddAnimation(method)` | none: `void ...catch()` |
| `handleGsapAddKeyframeBatch` | promise that catches its own failure — arrival ≠ landing |
| `handleGsapDeleteAnimation` | promise discarded entirely |
| `handleGsapUpdateMeta` | boolean — the one honest one |

So three tools return `dispatched: true` and one returns a real success/failure. That asymmetry is faithfully reflected: `STUDIO_UPDATE_ANIMATION_DESCRIPTION` calls out "the one animation tool that CONFIRMS its write", and the other three descriptions carry the `DISPATCH_CAVEAT` verbatim. An agent reading tools/list gets the honest contract. Sibling reviewers on #3515 flagged the class-of-bug for read-then-write races; this PR takes the read-back option off the table (React state won't refresh inside one tool call) and picks "tell the model to call `studio_inspect` to see the result" — the right substitute given the constraint.

**2. `studio_update_animation`'s "rule out no-selection BEFORE dispatch" is the load-bearing design decision.**

The underlying `handleGsapUpdateMeta` returns `false` for both "nothing was selected" and "the write failed on a real target". If both paths could reach the boolean, `false` would be ambiguous and the tool couldn't return a `"failed"` kind vs an `"invalid"` kind honestly. `guard(deps)` before dispatch drops the no-selection case, so a `false` afterward is unambiguous → `toolFailure("failed", ..., "The animation id may be stale. studio_inspect lists the current ones.")`. That's what makes `kind: "failed"` legible to the agent, and the hint points at the right recovery. Test-pinned at line 208-223 (`test_hf_component_batch_flag_lookup_is_retained_only_for_pre_retirement_histories` equivalent — checks no-selection is caught before dispatch, so `updateAnimation` mock is never called).

**3. `studio_add_animation` refuses `position` — and the description explains why.**

Right call. `handleGsapAddAnimation` takes ONLY a method and reads the playhead itself. Accepting a caller-supplied position would let the tool echo back a number that had no effect on the actual insertion point. Instead it reports `insertedAtSeconds: currentTime` (read-back from the playhead) and the description tells the agent to `studio_seek` first if it wants a specific time. That's a design decision that would be easy to get wrong by mirroring the schema of a lookalike tool. ✓

**4. Silent-drop of wrong-type fields on `studio_update_animation` misleads the agent's next call.** *(nit, non-blocker)*

`studioUpdateAnimation` builds `updates` by dropping any field that fails its type/finite/bound guard:

```ts
if (typeof input.duration === "number" && Number.isFinite(input.duration)) {
  if (input.duration < 0) return toolFailure("invalid", "duration must not be negative");
  updates.duration = input.duration;
}
if (typeof input.ease === "string" && input.ease.trim()) updates.ease = input.ease;
if (typeof input.position === "number" && Number.isFinite(input.position)) {
  updates.position = input.position;
}
if (Object.keys(updates).length === 0) {
  return toolFailure("invalid", "give at least one of duration, ease, position");
}
```

An agent that sends `{animationId: "x", duration: "1.5"}` (string instead of number) trips the third-branch check → `updates` stays empty → "give at least one of duration, ease, position". The agent thinks nothing was given; the agent actually gave a duration in the wrong shape. Next call likely re-sends with the same shape and loops.

The keyframe path has the same pattern for individual property values (non-string/non-number silently dropped, only "all dropped" surfaces as an error), though the `properties must contain at least one number or string value` reason is slightly more actionable there because the property vocabulary is open.

Concrete fix that would eliminate the loop-inducing UX: when `input.duration !== undefined && (typeof input.duration !== "number" || !Number.isFinite(input.duration))`, surface `toolFailure("invalid", "duration must be a finite number")`. Same for `ease` and `position`. Then wrong-type is distinguishable from missing, and the agent's next call fixes the actual shape mismatch.

Not a blocker — the sibling read tool (`studio_inspect`) gives the agent the actual schema by example, so a well-guided agent recovers. But it's the "silent coercion masks typos" pattern in miniature, and the fix is a couple of lines.

**5. `STUDIO_UPDATE_ANIMATION_INPUT_SCHEMA` `position: number` description — GSAP semantics ambiguous.** *(low-severity)*

Description reads `"Start position in seconds"`. GSAP positions are relative-to-previous offsets or absolute times depending on parent context, and are traditionally allowed to be tokens like `"<"` or `">"`. This tool declares `type: "number"` and reads only numeric positions, so token positions aren't reachable — that's fine. But an agent reading the description will believe "position" is absolute seconds; the actual handler behavior depends on Studio's timeline setup, which isn't exposed. Worth a description tweak like `"Start position in the timeline (numeric only; use studio_add_animation + studio_seek for absolute placement)"` — makes the constraint explicit rather than implicit.

**6. Confirmed:**

- **`studioDeleteAnimation` returns `{animationId, dispatched: true}`** but its test only asserts `result.ok === true`. Adding `expect(...dispatched).toBe(true)` would pin the field-shape contract; the current test still catches a regression where the tool returns something completely different. Minor coverage nit. ✓ (functionally covered)
- **Method whitelist** `METHODS = ["to", "from", "set", "fromTo"]` — matches GSAP's four core methods; anything else routes to `toolFailure("invalid")`. Test at line 149 asserts the whitelist and that the dispatch mock is never called. ✓
- **Percent test at 0 and 100** — inclusive bounds asserted at line 296-305. Boundary-covered. ✓
- **String "50" rejected as percent** (line 274) — schema-vs-actual mismatch caught, exactly the "platform doesn't validate inputSchema" defense. ✓
- **Idempotent registration** across re-renders (lines 646-655) — count assertion 8 → 12 pinned. ✓
- **Keyframe one-commit-one-undo** (line 400-405) — `addKeyframe` called ONCE with all properties, not per-property; test asserts `toHaveBeenCalledTimes(1)`. Batch-safe undo model preserved. ✓
- **StudioAgentTools deps array** updated to include all four new handlers — line 76-80. useMemo would otherwise return stale closures. ✓

**Rebase mechanics:** PR head is `d7a692c9` which folds in the just-merged #3521 docs commit — `mergeStateStatus: DIRTY` reflects the pending rebase onto master (which now has #3515/#3517/#3519/#3521 landed). Only `Preflight (lint + format)` and `WIP` visible in the check rollup so far — the full test/tsc/lint matrix hasn't run at this head. Post-rebase, the head SHA changes and `require_last_push_approval=true` on the public repo will dismiss this approval; a fresh re-approve will be needed at the rebased head. Signaling my read at this SHA so re-approval on the rebased head is a fast concur (assuming no code drift beyond the automatic rebase).

_Review by tai (pr-review)_

terencecho (approved):
Concur — re-approving at `2f2ca158` after the rebase.

Verified the animation-tool delta this PR owns is byte-identical to the previously-reviewed head: the `feat(studio): let an agent author motion` commit is still `90cef4f2513...`, unchanged in place. The 36-commit compare vs my prior head `d7a692c9` is entirely (a) `main` commits pulled in by the rebase (0.8.17 → 0.8.20 releases + assorted CLI/core/producer fixes + the `hw-write-title` registry addition) and (b) the sibling WebMCP stack fold-ins now that #3515 / #3516 / #3517 / #3518 / #3519 have landed. Nothing touches this PR's own animation-tool surface.

CI still in progress at head — approval doesn't force merge; auto-merge (already armed) gates on green as expected.

Prior review [`5060096243`](https://github.com/heygen-com/hyperframes/pull/3520#pullrequestreview-5060096243) carries.

_Review by tai (pr-review)_

PR #1, opened 2026-03-10
Initial repo setup

jrusso1020:
## Summary

- **README.md** — hero section, quick start, HTML schema example, package overview table, MCP integration section, Remotion comparison table
- **LICENSE** — MIT (copyright HeyGen)
- **CONTRIBUTING.md** — dev setup, conventional commit guidelines, project structure overview, PR guidelines
- **GitHub templates** — bug report, feature request, PR template
- **.gitignore** — Node.js/TypeScript project defaults

No code yet — just the repo scaffolding for review before we start porting packages.

## Test plan

- [ ] Review README content and messaging
- [ ] Verify LICENSE is correct
- [ ] Check issue/PR templates render properly on GitHub

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3521, opened 2026-08-27
docs: document Studio's WebMCP agent tools, proven end-to-end in a browser

miguel-heygen:
## What

The `webmcp` docs page, **an end-to-end proof that the tool set works in a real browser**, and the one bug that proof found.

Stacked on #3520. Last unit of ten.

## End-to-end proof

Driven with Claude in Chrome against Studio's dev server on `127.0.0.1:5190`, using a minimal GSAP fixture composition. Not a mock anywhere.

### 1. All twelve tools register

```js
> const tools = await document.modelContext.getTools()
> tools.map(t => t.name).sort()

["studio_add_animation", "studio_add_keyframe", "studio_delete_animation",
 "studio_frame", "studio_inspect", "studio_look", "studio_seek",
 "studio_select", "studio_set_style", "studio_set_text",
 "studio_transform", "studio_update_animation"]
```

### 2. `studio_look` returns real Studio state

```json
{ "ok": true, "project": "webmcp-proof", "playhead": 0, "duration": 4,
  "canWrite": true, "elementCount": 1,
  "elements": [{ "handle": "hf:hf-10wy", "label": "Ship it", "tag": "div",
                 "start": 0, "duration": 4, "track": 0 }],
  "selection": null }
```

### 3. Select, inspect, write

```json
selected  { "ok": true, "label": "Headline" }
inspect   { "text": "Ship it", "canEditText": true, "fieldKey": "self:0:h1" }
wrote     { "ok": true, "text": "Shipped it", "changed": true }
```

Studio's own URL state moved with it, so this is Studio's real selection, not a shadow copy:

```text
#project/webmcp-proof?...&selFile=index.html&selId=headline&selSelector=%23headline
```

### 4. The file on disk changed

```text
md5 before  b0d086373b574f887b93b52fcd8d2066
md5 after   d37d443d880eda060792386ebf4cd195

-  <h1 id="headline" data-hf-id="headline">Ship it</h1>
+  <h1 id="headline" data-hf-id="headline">Shipped it</h1>
```

### 5. And it renders

`studio_frame`'s URL, against the real Puppeteer renderer, before and after:

```text
t=0.500  ->  HTTP 200  image/png  14,424 bytes  1920x1080   "Ship it"
t=1.500  ->  HTTP 200  image/png  17,680 bytes  1920x1080   "Shipped it"
```

Both PNGs were opened and read. The frame shows the agent's edit.

## The bug the proof found

The first run **failed**, and failed honestly:

```json
{ "ok": false, "kind": "failed",
  "reason": "the text was not applied: persist-failed",
  "hint": "The write did not reach the file. Check Studio." }
```

The file was unchanged, checksum identical. So the reporting work in #3510 and #3518 did exactly its job: before that, this would have returned success.

Intercepting `fetch` gave the cause:

```text
POST .../file-mutations/patch-element  ->  400 {"error":"target and operations required"}
```

An element's text lives in a **child** field, keyed `self:0:h1` here. `studio_set_text` passed no field key, so `buildNextDomTextFields` planned zero operations and the patch went out empty.

**The unit tests could not have caught this**, because they mock `setText` and never cross that boundary. This is the class of failure only a browser run finds, which is why the plan insisted on this unit.

Fixed: the tool resolves the field, using the one named or the element's single field. Several fields, it asks which. No fields, it reports blocked. A named field that does not exist is rejected with the list of ones that do. Four regression tests, including the exact failing shape. One existing assertion expected the key to be `undefined`, which was the bug, so it now expects the resolved key.

## Two real-world behaviours, both documented

**Registration is asynchronous.** A caller reading `getTools()` the instant Studio loads sees a partial list. Observed at 10 of 12 mid-registration. The docs page says to wait for `toolchange` or poll.

**Tools on the current selection need a render between select and edit.** Firing select-then-edit back to back from one script returns `"nothing is selected"`, because `applyDomSelection` schedules a React state update. A real agent gets this for free, since its calls arrive as separate messages. This is the live confirmation of the select-first design rather than a defect.

## The docs page

`guides/webmcp`, under Developers → Agent setup.

Its first job is defusing a name collision: `guides/mcp` already covers HeyGen's **hosted** MCP connector, which builds a video from a chat. This page is an agent working *inside Studio*. The page says what it is not before what it is.

Also flags the two things easiest to get wrong: the API is `document.modelContext` not `navigator.modelContext`, and Studio should stay visible because some write paths report failure through a toast the human sees rather than a return value.

## Test plan

- [x] Documentation updated

- Full end-to-end run above, in Chrome, against a live Studio.
- `npx mint validate` and `npx mint broken-links --check-redirects` both pass.
- Package suite **4606 passing** across 414 files.
- `bunx tsc --noEmit`, `bunx oxlint`, `bunx fallow audit --fail-on-issues` all clean.

## Still not covered

Native Chrome behind `chrome://flags/#enable-webmcp-testing` uses a different code path from the polyfill, and has not been tried. Neither has a bridge extension calling from the extension context. The polyfill path is now proven end to end; those two are not.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -872,7 +872,8 @@
             "group": "Agent setup",
             "pages": [
               "guides/authentication",
-              "guides/skills"
+              "guides/skills",
+              "guides/webmcp"
             ]
           },
           {
--- a/docs/guides/webmcp.mdx
+++ b/docs/guides/webmcp.mdx
@@ -138,8 +138,9 @@ replaces the whole preferences blob and loses your panel sizes, zoom and timelin
 
 Set it back to `true`, or delete the key, to re-enable.
 
-The browser also asks your permission before any agent actually calls a tool, so registering them is
-not the same as granting access.
+The browser gates tool access behind its own permission prompt, so registering a tool is not the same
+as granting access to it. How often you are asked, once per site or every call, is up to the browser
+and is still changing while the API is in origin trial.
 
 ## Related topics
 
--- a/docs/guides/webmcp.mdx
+++ b/docs/guides/webmcp.mdx
@@ -123,9 +123,23 @@ is working.
 
 ## Turning it off
 
-Set `agentToolsEnabled` to `false` in Studio's preferences and reload. The browser also asks your
-permission before any agent actually calls a tool, so registering them is not the same as granting
-access.
+There is no settings toggle yet. The switch is a Studio preference, so set it from the console and
+reload:
+
+```javascript
+const KEY = "hf-studio-ui-preferences";
+const prefs = JSON.parse(localStorage.getItem(KEY) ?? "{}");
+localStorage.setItem(KEY, JSON.stringify({ ...prefs, agentToolsEnabled: false }));
+location.reload();
+```
+
+Read the existing object and spread it, as above. Writing `{agentToolsEnabled: false}` on its own
+replaces the whole preferences blob and loses your panel sizes, zoom and timeline settings.
+
+Set it back to `true`, or delete the key, to re-enable.
+
+The browser also asks your permission before any agent actually calls a tool, so registering them is
+not the same as granting access.
 
 ## Related topics
 
--- a/docs/guides/webmcp.mdx
+++ b/docs/guides/webmcp.mdx
@@ -52,6 +52,11 @@ console.log(tools.map((tool) => tool.name));
 // ["studio_look", "studio_select", "studio_seek", ...]
 ```
 
+<Note>
+  Registration is asynchronous, so a caller that reads `getTools()` the instant Studio loads can
+  see a partial list. Wait for the `toolchange` event, or poll until the count settles at twelve.
+</Note>
+
 <Warning>
   The API is `document.modelContext`, not `navigator.modelContext`. Many published examples use
   the second one. It is a compatibility shim some polyfills add, not part of the specification, so
--- a/docs/guides/webmcp.mdx
+++ b/docs/guides/webmcp.mdx
@@ -1,0 +1,129 @@
+---
+title: "Let an agent drive Studio"
+sidebarTitle: "Agent tools (WebMCP)"
+description: "Studio exposes its editing capabilities as WebMCP tools, so an agent in your browser can see the composition and change it alongside you."
+---
+
+Studio registers its own capabilities as WebMCP tools, so an AI agent running in your browser can read what Studio knows and make the same edits you can.
+
+<Note>
+  This is not the same as [creating through an AI chat](/guides/mcp). That page covers the hosted
+  HyperFrames MCP connector, which builds and renders a video from a conversation. This page is
+  about an agent working *inside Studio*, on a composition already open in front of you.
+</Note>
+
+## What it looks like
+
+With the tools available, an agent can do this without touching your files:
+
+```text
+studio_look                      -> the project, playhead, selection, and every element
+studio_select   hf:abc123        -> selects the headline, same as clicking it
+studio_inspect                   -> its resolved styles, text, and animations
+studio_set_style {"color":"red"} -> writes it, through Studio's own commit path
+studio_frame    2.4              -> a PNG of the composition at 2.4 seconds
+```
+
+The last one matters most. It is what lets an agent judge a change instead of guessing at it.
+
+## Turning it on
+
+The tools register automatically when Studio loads. Whether an agent can *reach* them depends on the browser.
+
+| Browser | Status |
+| --- | --- |
+| Chrome 149 | Origin Trial |
+| Edge 150 | Origin Trial |
+| ChatGPT Desktop | Shipped |
+| Brave (Leo) | Experimental |
+| Firefox, Safari | Not yet |
+
+For local development in Chrome, enable the flag and restart:
+
+```text chrome://flags
+chrome://flags/#enable-webmcp-testing
+```
+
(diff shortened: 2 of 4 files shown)


somanshreddy (commented):
**Two independent passes (mine + Codex), reconciled at source, at `8a6342bf`. COMMENT** — nothing here is a correctness/security blocker (the code fix is a strict improvement and well-tested), but because this is a *docs* PR, a cluster of real doc-accuracy mismatches is worth fixing before merge. Miguel's call on fix-now vs fast-follow.

## The core fix is correct — verified independently
`studioSetText` now resolves the target text field instead of passing `undefined`. I traced it against the selection type: `selection.textFields` is a **required** `DomEditTextField[]` on `DomEditSelection`, and it's the *same* authority `studio_inspect` (`inspectTools.ts:126`) and `studio_look` (`lookTools.ts:127`) already read — so if inspect lists a field, set_text can target it. All cohorts are right (no selection; requested-not-in-list; sole field; multiple-none-named; zero fields; valid requested key), and the fix reads the same `textFields` source rather than a parallel notion. The 4 regression tests pin every branch including the exact `child:0:h1` E2E shape, and the `undefined → "self"` assertion flip is the correct fix. The E2E-proof-found-the-bug narrative is exactly the discipline the stack was built for. Nice work.

## Doc-accuracy fixes worth making (real, reader-facing — none breaks the product)
Codex systematically cross-checked every doc claim against the code and found five real mismatches (my first pass verified the pref-key / `toolchange` / `document.modelContext` claims — all accurate — but missed these). I confirmed each at source:

1. **`studio_set_style` example shows the wrong arg shape** (`webmcp.mdx:23`). Doc: `studio_set_style {"color":"red"}`. The tool reads `input.styles` (`contentTools.ts:138`), so the accepted shape is `{"styles":{"color":"red"}}` — the shown form returns `invalid`. It's the page's only style example, so there's no correct reference to fall back on. One-line fix.
2. **"Pass it back to any tool that edits an element"** (`webmcp.mdx:79`) contradicts the code. Edit tools take **no handle** — they act on the ambient selection (`contentTools.ts:2-9`), and the code's own `studio_look` field comment says "Pass back to any tool that **takes a handle**" (`lookTools.ts:45`). Since inputs aren't schema-validated (`types.ts:51`), an extra `handle` on `studio_set_text` is silently ignored and the ambient (possibly stale) selection is edited. The "Select first, then edit" rule two sections down mitigates it, but the line should say handles go to `studio_select`/`studio_inspect`.
3. **"delete the key, to re-enable"** (`webmcp.mdx:139`) erases *all* Studio UI prefs — directly contradicting the preservation warning three lines above (the blob holds panel sizes/zoom/timeline). Recommend `agentToolsEnabled: true` (spread-preserve), or label whole-key deletion as a full reset.
4. **Verification is overstated** (`webmcp.mdx:104`): "Tools report what actually happened … which operations took effect." Rotation is "reported as dispatched rather than verified" (`transformTools.ts:202`) and the animation tools carry a `DISPATCH_CAVEAT` "Reports what was dispatched, not what landed" (`animationTools.ts:203`). Only move/resize read back. Worth naming the exceptions (same asymmetry the sibling reviews flagged on `studio_frame`/`studio_seek`).
5. **Read-tool table slightly overpromises** (`webmcp.mdx:76`): `studio_inspect` returns animations *only for the current selection* (`inspectTools.ts:108`), and `studio_frame` returns a **URL** to a PNG, not a PNG payload (`frameTools.ts:28`). Minor precision.

## Code nits (nonblocking follow-ups, concrete fixes)
- **No-op child re-write still fails loudly** (`contentTools.ts` / `useDomEditTextCommits.ts:131`). Setting a child field to its *existing* value plans zero operations, and `childOperations ?? […]` keeps the empty array (`??` doesn't fall through on `[]`) → server 400 "target and operations required" → `persist-failed`. It's a narrow residual of the very bug this PR fixes (only the identical-text case; pre-fix *every* child write failed, so this is strictly better, not a regression). Fix: resolve the field *object* and short-circuit `return {ok:true, changed:false}` when `field.value === input.text`.
- **`changed` compares against aggregate `selection.textContent`, not the resolved field's `value`** (`contentTools.ts`). Named/whitespace/multi-field edits can report the wrong `changed`. Same field-object resolution fixes both this and the no-op case.
- **Malformed present `field` is treated as omitted** (`contentTools.ts:90`). A non-string `field` coerces to `undefined` → sole-field inference. Harm is bounded (single-field element → the only target; multi-field → errors), so low, but rejecting a present-but-invalid `field` is cleaner given inputs aren't validated.
- **No positive explicit-field test** (`contentTools.test.ts`). The only explicit-`field` case passes an *invalid* field expecting no dispatch; nothing asserts a *valid* explicit key dispatches successfully — a regression rejecting every explicit field would stay green. Add a two-field selection + valid `child:1:p` asserting the successful call.

## Provenance / gate
The code fix is net-new-correct and the doc page is a genuine improvement; none of the above breaks functionality, so I'm not blocking. But the doc mismatches (esp. #1 and #2) are reader-facing incorrectness in the doc that will guide agent authors, and they're cheap — I'd fix #1–#4 before merge and fast-follow #5 + the code nits. Credit to Codex: the doc-shape and return-shape cross-checks (#1, #2, #4, #5) are ones my first pass missed. Tests weren't re-run locally (fresh clone, env unbuilt — not a failure); the package suite is green in CI per the PR body and the four new tests are mutation-meaningful.

PR #3526, opened 2026-08-27
fix(producer): enforce video extraction failures by default

### LINKED ISSUE #3372: Video extraction failures are swallowed, surfacing as a misleading coverage abort

miga-heygen:
## Summary

Flips the video extraction failure policy default from `off` to `enforce` so per-source extraction errors fail the render instead of being silently swallowed.

The error handling infrastructure was fully built — typed errors, retryable classification, caller throw — but gated behind `HF_VIDEO_EXTRACTION_FAILURE_MODE=enforce`. With the default at `off`, extraction failures (e.g. ffmpeg exiting 0 with zero frames from an audio-only tail) were collected into an `errors` array but never surfaced, leading to misleading coverage aborts downstream.

Opt-out: `HF_VIDEO_EXTRACTION_FAILURE_MODE=off` restores the old silent behavior.

Closes #3372

— Miga

### DIFF
--- a/packages/producer/src/services/render/stages/extractVideosStage.test.ts
+++ b/packages/producer/src/services/render/stages/extractVideosStage.test.ts
@@ -205,14 +205,14 @@ describe("shouldCopyExtractedFrames", () => {
 });
 
 describe("resolveVideoExtractionPolicy", () => {
-  it("preserves stable behavior by default", () => {
+  it("enforces extraction failures by default (#3372)", () => {
     expect(resolveVideoExtractionPolicy({})).toEqual({
-      failureMode: "off",
+      failureMode: "enforce",
       maxTransientRetries: 0,
     });
   });
 
-  it("allows only the bounded candidate rollout values", () => {
+  it("allows explicit opt-out or observe mode", () => {
     expect(
       resolveVideoExtractionPolicy({
         HF_VIDEO_EXTRACTION_FAILURE_MODE: "observe",
@@ -221,10 +221,15 @@ describe("resolveVideoExtractionPolicy", () => {
     ).toEqual({ failureMode: "observe", maxTransientRetries: 1 });
     expect(
       resolveVideoExtractionPolicy({
-        HF_VIDEO_EXTRACTION_FAILURE_MODE: "unexpected",
-        HF_VIDEO_EXTRACTION_MAX_RETRIES: "1",
+        HF_VIDEO_EXTRACTION_FAILURE_MODE: "off",
       }),
     ).toEqual({ failureMode: "off", maxTransientRetries: 0 });
+    expect(
+      resolveVideoExtractionPolicy({
+        HF_VIDEO_EXTRACTION_FAILURE_MODE: "enforce",
+        HF_VIDEO_EXTRACTION_MAX_RETRIES: "1",
+      }),
+    ).toEqual({ failureMode: "enforce", maxTransientRetries: 1 });
   });
 });
 
--- a/packages/producer/src/services/render/stages/extractVideosStage.ts
+++ b/packages/producer/src/services/render/stages/extractVideosStage.ts
@@ -131,15 +131,17 @@ export interface VideoExtractionPolicy {
 }
 
 /**
- * Candidate-lane rollout controls. Stable behavior remains unchanged unless
- * explicitly enabled in the producer environment.
+ * Extraction failure policy. Defaults to `enforce` so per-source errors
+ * surface as render failures instead of being silently swallowed (#3372).
+ * Set `HF_VIDEO_EXTRACTION_FAILURE_MODE=off` to restore the old silent
+ * behavior, or `observe` to log without failing.
  */
 export function resolveVideoExtractionPolicy(
   env: Readonly<Record<string, string | undefined>> = process.env,
 ): VideoExtractionPolicy {
   const rawMode = env.HF_VIDEO_EXTRACTION_FAILURE_MODE?.trim().toLowerCase();
   const failureMode: VideoExtractionFailureMode =
-    rawMode === "observe" || rawMode === "enforce" ? rawMode : "off";
+    rawMode === "observe" || rawMode === "off" ? rawMode : "enforce";
   const maxTransientRetries =
     failureMode !== "off" && env.HF_VIDEO_EXTRACTION_MAX_RETRIES?.trim() === "1" ? 1 : 0;
   return { failureMode, maxTransientRetries };


somanshreddy (commented):
**Reviewed — no blocking issues; reads as approvable to me** at `352d16ee`. Posting as a comment: I traced the flip through the full execution layer, both render paths, and the default genuinely takes effect without false-failing legitimate renders. (Leaving the formal merge-unblocking stamp to be cast with proper sign-off rather than off a channel ping.)

### What I verified (execution layer, at head)
- **Effective on both paths.** `failureToEnforce` is *returned* from the stage (`extractVideosStage.ts:447`), not thrown in it — and both consumers throw it: `renderOrchestrator.ts:2526` and `distributed/plan.ts:1047`. So the flip is live on the single-process *and* distributed render paths, not just one.
- **No false-fail on audio-tail renders.** The case the description cites (ffmpeg exits 0 with zero frames from an audio-only tail) is the one I pressure-tested. Frame-extraction planning uses `resolvePlayableVideoDuration` (video-stream duration, not container), so a normal video with an audio tail is bounded to its video window and produces frames. `zero_output` only trips on `framePaths.size === 0 && duration > 0` — a genuinely unrenderable video slot. The taxonomy (`ffmpeg_timeout`/`ffmpeg_transient` retryable; `zero_output`/`ffmpeg_failed`/`source_missing`/`media_start_out_of_range` fatal) fires only on real failures.
- **HDR-probe path is strictly safer, not newly-failing.** Under the *old* `off` default a probe throw already re-threw raw and aborted the render (`Promise.all`). Under `enforce` it now throws the bounded `VideoExtractionStageError` — a taxonomy/count summary with no signed-URL/path leak across the HTTP/Temporal boundary. So this path gains leak-safety; it wasn't silently passing before.
- **No new retry behavior by default.** `maxTransientRetries` stays 0 unless `HF_VIDEO_EXTRACTION_MAX_RETRIES=1` — the default remains the exact legacy single probe.

The one genuinely-new failure surface is frame-extraction `result.errors` under enforce, which is exactly #3372's intent: convert a swallowed error + misleading downstream coverage abort into an explicit typed failure at the stage that caused it.

### Non-blocking
1. **Unrecognized env value now falls back to `enforce` (was `off`).** `rawMode === "observe" || rawMode === "off" ? rawMode : "enforce"` — so a typo'd opt-out (`=disabled`, `=enfroce`) now *enforces* rather than silently disabling. That's the right fail-closed direction, but the removed test asserted `"unexpected" → off`; the replacement only pins empty `{}` → enforce. Worth an explicit test that an unrecognized non-empty value → enforce, so the fallback direction is locked.
2. **`observe` doesn't fully "log without failing" on the HDR-probe path.** The new docstring says `observe` logs without failing, but `throwHdrProbeFailures` in observe mode still `throw`s the first raw error (aborts). It's honored on the frame-extraction path (`applyVideoExtractionFailurePolicy` returns null for observe) but not the probe path. Pre-existing logic, but the docstring is new here — either tighten the doc or make observe log-only on both paths. Doesn't affect the default (enforce).

### Praise
Clean separation: the stage classifies + still materializes extracted frames, and hands the enforce-error up for the orchestrator to throw — partial work isn't discarded, and the failure is bounded/transport-safe.

---
*Direct manual analysis at head `352d16ee` (traced both render paths + the engine classifier), not a Codex pass — a +15/-8 default flip where the whole risk is blast radius. Tests not run locally (no hyperframes node env here); CI green at head.*

jrusso1020 (approved):
Approving on Miguel's explicit request. Read both files at `352d16ee`; Somu's review is the substantive pass and I agree with its verdict, so this is additive.

The change is a one-line predicate inversion:

```
- rawMode === "observe" || rawMode === "enforce" ? rawMode : "off"
+ rawMode === "observe" || rawMode === "off"     ? rawMode : "enforce"
```

Correct for the stated goal — unset now lands on `enforce`, and `#3372`'s silent swallow is closed by default.

**Sharpening Somu's first nit, which I think is worth more than "add a test."** The old suite pinned `"unexpected"` → `off`; the new suite *replaces* that exact case with `"off"` → `off`. So the unrecognized-value branch lost its only coverage in the same commit its behavior inverted — the two halves of that nit are the same edit, which is what makes it easy to miss in review.

The operational half matters more than the coverage half: this env var is an **escape hatch**, and escape hatches get typed under incident pressure. `HF_VIDEO_EXTRACTION_FAILURE_MODE=of` or `=disable` or `=false` now resolves to `enforce` — the precise opposite of the operator's intent, silently, at the moment they are trying to stop renders from failing. Fail-closed is still the right default for the *unset* case; it's the *malformed* case that deserves a distinguishable outcome. A one-line `logger.warn` on a non-empty `rawMode` that matched nothing would make the hatch trustworthy without touching the default. Non-blocking, and I'd take it as a follow-up rather than a change here.

Minor second-order note: `maxTransientRetries` is gated on `failureMode !== "off"`, so with the default flipped, a pre-existing `HF_VIDEO_EXTRACTION_MAX_RETRIES=1` in an environment that previously resolved to `off` now yields 1 retry where it used to yield 0. Intended given the goal, but it means the flip changes retry behavior in environments that never set the failure-mode var at all.

Somu's second nit (docstring says `observe` "logs without failing" while `throwHdrProbeFailures` still throws raw on the probe path) is real and worth a docstring tweak.

CI: all eight ruleset-required checks green on their **latest** attempt at this head (53 runs across 46 names — worth the qualifier, since a naive per-name query returns stale attempts).

— Rames

james-russo-rames-d-jusso (commented):
🟢 **One-line default flip, cleanly scoped.** The `enforce` plumbing is genuinely already in place — both call sites throw `failureToEnforce` after the extraction telemetry checkpoint, the typed error class + failure-kind taxonomy + retryable classification pre-date this PR (`9b63646c8` + `33ca1de06`), and the distributed path was already strict via `assertVideoExtractionSucceeded`. Reads as approvable to me at `352d16ee`.

## Focus points confirmed

- **The flip itself.** `packages/producer/src/services/render/stages/extractVideosStage.ts:143-144` — the ternary inverted from `"observe" | "enforce" ? rawMode : "off"` to `"observe" | "off" ? rawMode : "enforce"`. That's the entire behavior change; type union `VideoExtractionFailureMode = "off" | "observe" | "enforce"` bounds valid values.
- **Config key.** `HF_VIDEO_EXTRACTION_FAILURE_MODE`. Read only in `resolveVideoExtractionPolicy()` — sole consumer, no scattered env reads to worry about.
- **What `enforce` does.** Three effects:
  1. `runExtractVideosStage:407` sets `collectProbeFailures: extractionPolicy.failureMode === "enforce"` on `extractAllVideoFrames`, opting the engine into recording per-source failure kinds.
  2. `applyVideoExtractionFailurePolicy():249` returns the built `VideoExtractionStageError` only when mode is `"enforce"` (returns `null` for `"observe"`, and short-circuits at line 242 for `"off"`).
  3. HDR probe path at `:344`+`throwHdrProbeFailures:229` throws the aggregated `buildHdrProbeStageError` in enforce, or the raw first error otherwise.
- **"All error plumbing was already built" — verified.** Both consumers already throw:
  - `renderOrchestrator.ts:2526` — `if (failureToEnforce) throw failureToEnforce;` after the `observability.checkpoint("video_extract", ...)` emit.
  - `distributed/plan.ts:1047` — same guard; distributed also runs `assertVideoExtractionSucceeded` on top (comment at `:1049-1053` explains distributed is unconditionally strict because a missing frame set can't fall back to remote decoding across workers).
- **Test parity.** New default asserted directly (`resolveVideoExtractionPolicy({}).failureMode === "enforce"` — tagged `#3372`) and each of `"observe"` / `"off"` / `"enforce"` exercised explicitly.
- **CI (latest attempt).** Producer unit + integration, Test (full), Tests on windows-latest, Render on windows-latest, all 9 regression-shards, CodeQL, Typecheck, Lint, Fallow audit — all green on latest attempt. Path-filtered `skipping` on Preflight / Preview parity / Skills / Studio timeline viewport / CLI npx / Codex plugin / GCP BeginFrame / Perf / Catalog is expected given the diff only touches `packages/producer/`.
- **Blast-radius sweep.** Grepped for callers of the extraction stage and readers of the failure mode; only two consumer sites (renderOrchestrator + distributed/plan), both already throwing on `failureToEnforce`. No test in the tree mocks a failing extractor and asserts caller-success, so no silent test regressions from the flip. Opt-out (`HF_VIDEO_EXTRACTION_FAILURE_MODE=off`) is documented in the JSDoc header at `:136`. Repo changelog is auto-generated from commit titles (`bun run changelog:draft`), so `fix(producer): enforce video extraction failures by default` will surface naturally in the next release notes — no manual changelog edit needed here.

## Concerns

- **`Closes #3372` may over-close the issue.** Issue #3372 names three fixes: (1) stop swallowing per-source extraction failures (this PR), (2) treat a sentineled cache entry with zero frames as a miss and re-extract (`extractionCache.ts:481`), and (3) clamp media-start against playable video-stream duration. Only (1) lands here. The "happens every time on a fresh cache" reproducer in the issue explicitly names a poisoned cache entry as a plausible root cause for repeat field reports — after this flip, that user still hits the poisoned cache branch, but now sees a `VideoExtractionStageError` instead of the misleading coverage abort. That's a real improvement, and (2)/(3) may be intentionally deferred, but consider `Refs #3372` + follow-up tickets for the two secondary guards rather than `Closes`. Non-blocking — Miga's judgment call.
- **Dropped test for unknown / typo values.** The old test asserted `HF_VIDEO_EXTRACTION_FAILURE_MODE: "unexpected"` fell back to `"off"`; that assertion was removed. The new logic still has a catchall — `rawMode === "observe" || rawMode === "off" ? rawMode : "enforce"` — and now unknown values fall back to `"enforce"`. That's the safer direction (typo → fail closed, not silently pass), but nothing pins the direction in a test. A future refactor of the ternary could invert it without breaking any test. One extra assertion — `resolveVideoExtractionPolicy({ HF_VIDEO_EXTRACTION_FAILURE_MODE: "enforece" }).failureMode === "enforce"` — would durably lock this in.
- **No stage-level test of the enforce end-to-end path.** Unit tests exercise `resolveVideoExtractionPolicy`, `assertVideoExtractionSucceeded`, `buildHdrProbeStageError`, and the failure-mode branches independently, and `plan.test.ts` covers the distributed path throwing `VideoExtractionStageError`, but there's no test that `runExtractVideosStage` on a composition with a failing extractor at default (now `"enforce"`) mode returns a non-null `failureToEnforce`. The existing unit coverage is strong enough that I wouldn't block on this, but a single stage-level assertion would be a durable guard for the exact regression this PR is fixing.

## Nits

- JSDoc at `:133-138` is nicely rewritten. Minor consistency — the `resolveVideoExtractionPolicy` header describes what `off`/`observe`/`enforce` do, but the standalone `VideoExtractionFailureMode` type at `:126` has no docstring naming what each variant means. Not this PR's job.

## What I didn't verify

- Whether the two secondary defects called out in issue #3372 (zero-frame cache-hit rehydrate + media-start clamp past playable video duration) have separate tickets already tracking them — worth a quick check before this PR closes the umbrella issue.
- The `heygenverse` / apps-side behavior on the render worker of what happens when `VideoExtractionStageError` reaches the Temporal boundary — the JSDoc at `:151-157` claims the taxonomy/count summary is scrubbed of paths and signed URLs, and `server.errorCode.test.ts` covers the error-code path, but I didn't trace through to the workflow-level retry classifier to confirm `retryable=true` on `download_transient` still gets scheduled for a workflow-level retry vs. terminal.
- I only sampled a few of the 9 regression shards — I trust CI's aggregate green here.

— _Review by Rames D Jusso_

PR #3528, opened 2026-08-27
fix(engine): fail render on sub-composition script failures (#3352)

### LINKED ISSUE #3352: decision-tree example renders 2 unique frames, and check + render both report success
## Describe the bug

The shipped `decision-tree` example renders a 10-second video containing **2 unique
frames**, and both quality gates report success while it happens.

Two defects, and the second is the serious one:

1. `registry/examples/decision-tree/compositions/decision_tree.html:388` reads
   `tl.labels["hold5"]`. At render time `tl.labels` is undefined, so the composition
   script throws and the timeline is never built.
2. **Nothing fails.** `check` exits 0 and prints "Check passed". `render` logs the
   TypeError, then logs `sub_timeline_readiness_timeout` after waiting 45 seconds,
   then writes the mp4 and exits 0. A user following the quickstart gets a green
   check, a green render, and a still image.

Anyone starting from this example hits it on their first render.

## Steps to reproduce

```bash
npx --yes hyperframes@0.8.2 init decision-tree --non-interactive --example=decision-tree
cd decision-tree
npm run check                                              # exits 0, "Check passed"
npm run render -- --quality draft --output renders/dt.mp4  # exits 0

# count unique frames in the result
mkdir -p /tmp/dtf && ffmpeg -v error -i renders/dt.mp4 -vf fps=1 /tmp/dtf/f%03d.png
ls /tmp/dtf | wc -l          # 10 sampled
md5 -q /tmp/dtf/*.png | sort -u | wc -l   # 2 unique
```

## Expected behavior

The example renders its intended animation. Failing that, a composition script that
throws, or a sub-composition timeline that never registers, fails the render loudly
instead of producing a near-static file with a zero exit code.

## Actual behavior

`check`:

```
  0 error(s), 1 warning(s), 1 info(s)
Motion     ◇ 0 errors, 0 warnings
Contrast   ◇ 37/37 text checks pass WCAG AA
◇  Check passed
```

`render`:

```
[Browser:ERROR] [Compiler] Composition script failed decision-tree TypeError: Cannot read properties of undefined (reading 'hold5')
[FrameCapture] Sub-composition timelines not registered after 45000ms: decision-tree. ...
[FrameCapture:sub_timeline_readiness_timeout] Sub-composition timelines did not become ready within 45000ms { timeoutMs: 45000 }
[INFO] Calibration complete, estimated cost: 22ms/frame (p95)
```

The TypeError repeats during capture. The render then completes normally.

Output: 10.000000s, 193338 bytes, and sampling at 1 fps yields two distinct images —
one held for the first 8 samples, another for the last 2.

## The failing code

```js
// registry/examples/decision-tree/compositions/decision_tree.html:384
tl.addLabel("hold5", "+=0.2");

// :388
const typingStart = tl.labels["hold5"];
```

`addLabel` does not throw; the subsequent read of `tl.labels` does, so `tl.labels` is
undefined on the render path specifically. I have not determined why — that is worth
establishing before fixing, because if the timeline object exposed during render does
not carry `labels`, then every composition that reads a label position is affected and
this example is just the one we happen to ship.

## Why the gate half matters more than the example

Fixing the example alone would close the symptom and leave the hole. Today a
composition script can throw on every frame, a sub-composition timeline can time out
after 45 seconds, and the pipeline still reports success. Both of those are already
detected and logged — they simply are not wired to an exit code.

Concretely worth deciding:

- A `[Compiler] Composition script failed` error should fail the render, or at minimum
  fail it under `--strict`, rather than only appearing in the log.
- `sub_timeline_readiness_timeout` after a full 45-second wait is not a soft condition.
  If a composition legitimately has no GSAP timelines it can declare `data-no-timeline`,
  which is what that opt-out exists for, so reaching the timeout means something is
  genuinely wrong.

## Environment

```shell
hyperframes  0.8.2 (pinned by the scaffold; latest 0.8.3)
Capture      drawelement -> beginframe, --quality draft, 6 workers
```

miga-heygen:
## Summary

- Detect composition script **runtime errors** (caught by `__hfErrorLabel` try/catch) in the browser console handler and feed them into `scriptLoadFailures`, triggering the existing `pollSubCompositionTimelines` fail-fast path — same as script load 404s already do
- Make `sub_timeline_script_failure` a **fatal warning** in `applyRenderWarningPolicy`, alongside `audio_processing_failed` — a composition whose GSAP timeline script failed to execute produces a degenerate 2-frame output that looks like a still image
- `sub_timeline_readiness_timeout` remains non-fatal (CSS-only compositions with missing `data-no-timeline` would false-positive)

Fixes #3352

## Root cause

When a composition script throws during execution, the `try/catch` wrapper in `wrapScopedCompositionScript` logs to `console.error` but doesn't propagate the error. The GSAP timeline registration never arrives. `pollSubCompositionTimelines` times out after the full `playerReadyTimeout` (45s). The timeout is recorded as a warning but the render continues — producing a 2-frame "successful" render.

## Test plan

- [x] `frameCapture-subTimelinePoll.test.ts` — 4 tests pass (existing fail-fast contract)
- [x] `renderEventPublisher.test.ts` — 13 tests pass (2 new: `sub_timeline_script_failure` blocks best-effort, `sub_timeline_readiness_timeout` allows best-effort)
- [x] `frameCapture.test.ts` — 26 tests pass (existing)
- [x] `perfSummary.test.ts` — 6 tests pass (existing)
- [x] Pre-commit hooks: lint, format, fallow, typecheck all green

— Miga

### DIFF
--- a/packages/engine/src/services/frameCapture.ts
+++ b/packages/engine/src/services/frameCapture.ts
@@ -1889,13 +1889,16 @@ function recordCaptureWarnings(session: CaptureSession, warnings: readonly Captu
 function recordSubTimelineWarning(session: CaptureSession, timeoutMs: number): void {
   if (session.subTimelineWaitOutcome === "ready" || !session.subTimelineWaitOutcome) return;
   const scriptFailure = session.subTimelineWaitOutcome === "script_failure";
+  const hasRuntimeErrors = session.scriptLoadFailures.some((f) => f.startsWith("runtime-error:"));
   recordCaptureWarnings(session, [
     {
       code: scriptFailure ? "sub_timeline_script_failure" : "sub_timeline_readiness_timeout",
       message: scriptFailure
-        ? "A sub-composition timeline script failed to load"
+        ? hasRuntimeErrors
+          ? `A sub-composition script threw during execution — timeline registration never arrived (${session.scriptLoadFailures.join(", ")})`
+          : `A sub-composition timeline script failed to load (${session.scriptLoadFailures.join(", ")})`
         : `Sub-composition timelines did not become ready within ${timeoutMs}ms`,
-      details: { timeoutMs },
+      details: { timeoutMs, sources: [...session.scriptLoadFailures] },
     },
   ]);
 }
@@ -2066,6 +2069,16 @@ export async function initializeSession(session: CaptureSession): Promise<void> 
     const diagnostic = formatConsoleDiagnostic(type, text, locationUrl);
     if (!diagnostic.suppressHostLog) console.log(diagnostic.text);
     appendBrowserDiagnostic(session, diagnostic.text);
+
+    // Composition script runtime errors mean the GSAP timeline registration
+    // can never arrive — same fail-fast treatment as script load failures.
+    // Without this, pollSubCompositionTimelines burns the full timeout and
+    // the render silently succeeds with a degenerate 2-frame output (#3352).
+    if (type === "error" && text.startsWith("[HyperFrames] composition script error:")) {
+      const detail = text.slice("[HyperFrames] composition script error:".length).trim();
+      const compId = detail.split(" ")[0] || "unknown";
+      recordScriptLoadFailure(session, `runtime-error:${compId}`);
+    }
   });
 
   page.on("pageerror", (err) => {
--- a/packages/producer/src/services/render/renderEventPublisher.test.ts
+++ b/packages/producer/src/services/render/renderEventPublisher.test.ts
@@ -246,6 +246,45 @@ describe("updateJobStatus", () => {
     );
   });
 
+  it("blocks sub-timeline script failures in best-effort mode (#3352)", () => {
+    const job = createRenderJob({ fps: 30, quality: "high" });
+    const log = { error: vi.fn(), warn: vi.fn(), info: vi.fn(), debug: vi.fn() };
+    expect(() =>
+      applyRenderWarningPolicy(
+        job,
+        [
+          {
+            code: "sub_timeline_script_failure",
+            message: "A sub-composition script threw during execution",
+            details: {
+              timeoutMs: 45_000,
+              sources: ["runtime-error:decision-tree-123"],
+            },
+          },
+        ],
+        log,
+      ),
+    ).toThrow(RenderQualityError);
+    expect(job.warnings).toHaveLength(1);
+  });
+
+  it("allows sub-timeline readiness timeout in best-effort mode", () => {
+    const job = createRenderJob({ fps: 30, quality: "high" });
+    const log = { error: vi.fn(), warn: vi.fn(), info: vi.fn(), debug: vi.fn() };
+    applyRenderWarningPolicy(
+      job,
+      [
+        {
+          code: "sub_timeline_readiness_timeout",
+          message: "Sub-composition timelines did not become ready within 45000ms",
+          details: { timeoutMs: 45_000 },
+        },
+      ],
+      log,
+    );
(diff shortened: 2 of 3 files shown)


vanceingalls (approved):
APPROVED.

**What.** Wires composition-script runtime errors into the same fail-fast pipeline that already handles script load 404s. Three touch points:
1. `frameCapture.ts:2072-2081` — Playwright console handler now watches for `type === "error" && text.startsWith("[HyperFrames] composition script error:")`, extracts a compId token, and calls the existing `recordScriptLoadFailure(session, `runtime-error:${compId}`)`.
2. `frameCapture.ts:1889-1901` — the `sub_timeline_script_failure` warning message now differentiates runtime-error vs load-failure via `scriptLoadFailures.some(f => f.startsWith("runtime-error:"))`, and stashes the full source list in `details.sources`.
3. `renderOrchestrator.ts:685-697` — `applyRenderWarningPolicy` treats `sub_timeline_script_failure` as fatal alongside `audio_processing_failed` in best-effort mode.

**Correctness.**
- The emit contract at `packages/core/src/compiler/compositionScoping.ts:276` (`errorLabel = "[HyperFrames] composition script error:"`) matches the console-handler's `startsWith` prefix exactly. Emit sites at `:614` (`console.error(__hfErrorLabel, __hfCompId, _err)`) and `:623` (`console.error(${errorLabel}, _err)`) both hit this prefix.
- `sub_timeline_readiness_timeout` correctly stays non-fatal (CSS-only compositions missing `data-no-timeline` would false-positive) — matches the concern noted in the PR body.
- Broadens fail-fast on existing script-load 404s too (they weren't blocking best-effort renders pre-#3528). This is the correct move — a 404-loaded composition script produces the same degenerate 2-frame output as a runtime-thrown one — but worth an explicit callout in the release notes so operators know the failure-mode surface widened.

**Non-blocking observations.**
- *CompId parsing is naive.* `detail.split(" ")[0]` assumes the compId is the first whitespace-separated token after the prefix. At `:614` emit, it works because `__hfCompId` is a single-token ID. At `:623` (fallback wrap), no compId is emitted — Playwright will serialize `_err` as the second arg (`"Error: foo bar"`), so `compId` becomes `"Error:"`. Fail-fast still fires (that's what matters), but the reported source string `"runtime-error:Error:"` is unhelpful for support triage. Consider having the fallback wrapper at `:623` include a sentinel like `"[HyperFrames] composition script error: <unknown-comp>"` so parsing yields a labeled unknown rather than the exception class.
- *Console-string coupling is untested.* The `text.startsWith(...)` check is the load-bearing coupling between core (emit) and engine (detect); if the `errorLabel` constant changes at `compositionScoping.ts:276`, the fail-fast silently regresses. A tiny engine-side test that feeds a synthetic `page.on("console")` payload through `initializeSession`'s handler would guard this. Not a blocker.
- *Message construction in the mixed case:* if a render has both a load-failure (`http://...`) and a runtime-error (`runtime-error:...`), `hasRuntimeErrors` is true and the message says "threw during execution" — the load-failure disappears from the human-readable line. `details.sources` captures both, so support can still trace, but the message glosses.

**Tests.** Two new cases in `renderEventPublisher.test.ts` cover the policy layer (`sub_timeline_script_failure` blocks, `sub_timeline_readiness_timeout` allows). Console-handler branch is untested (see above).

**CI.** All required green (Windows render, regression shards 1-9, Producer unit/integration, Studio smoke, CLI smoke, Preview parity). `mergeable=UNKNOWN` is GitHub recomputing — not a blocker.

— Via

PR #3533, opened 2026-08-28
fix(cli): respect timeline-free static compositions

miguel-heygen:
## Summary

Intentional static voiceover and page compositions marked `data-no-timeline` no longer fail `hyperframes check` with a misleading `sweep_static` error. The frozen-sweep guard still rejects unchanged seek results for normal timeline-backed compositions, while the explicit timeline-free contract now suppresses only that diagnostic and leaves all other checks intact.

## Test plan

- `bun run --cwd packages/cli test -- src/commands/check.test.ts src/utils/checkBrowser.test.ts` (67 passed)
- `bunx oxlint` and `bunx oxfmt --check` on changed files
- `bun run --cwd packages/cli dev -- check /tmp/hf-sweep-static-TZTuys --json --no-contrast` (exact Terra fixture: exit 0, zero layout findings)

---

Codex

### DIFF
--- a/packages/cli/src/commands/check.test.ts
+++ b/packages/cli/src/commands/check.test.ts
@@ -145,6 +145,7 @@ function fakeDriver(overrides: Partial<CheckAuditDriver> = {}): CheckAuditDriver
   return {
     initialize: vi.fn(async (_contrast: boolean) => undefined),
     getDuration: vi.fn(async () => 9),
+    hasNoTimelineDeclaration: vi.fn(async () => false),
     getTransitionBoundaries: vi.fn(async () => []),
     getCanvas: vi.fn(async () => ({ width: 1920, height: 1080 })),
     findAmbiguousSelectors: vi.fn(async (_selectors: string[]) => []),
@@ -1220,6 +1221,18 @@ describe("check pipeline", () => {
             finding.message.includes("did not advance"),
         ),
       ).toBe(true);
+    });
+
+    it("does not flag intentional static content declared with data-no-timeline", async () => {
+      const driver = fakeDriver({
+        getDuration: vi.fn(async () => 6),
+        hasNoTimelineDeclaration: vi.fn(async () => true),
+        collectLayoutGeometry: vi.fn(async () => "frozen"),
+      });
+
+      const { report } = await runScenario(driver);
+
+      expect(report.layout.findings.some((finding) => finding.code === "sweep_static")).toBe(false);
     });
 
     it("does not flag a 1.5s static title card — too short for the guard to apply", async () => {
--- a/packages/cli/src/utils/checkBrowser.ts
+++ b/packages/cli/src/utils/checkBrowser.ts
@@ -395,6 +395,7 @@ function createPageDriver(page: Page, setTime: (time: number) => void): CheckAud
   return {
     initialize: (contrast) => injectAuditScripts(page, contrast),
     getDuration: () => getCompositionDuration(page),
+    hasNoTimelineDeclaration: () => hasNoTimelineDeclaration(page),
     getTransitionBoundaries: () => collectTweenBoundaries(page),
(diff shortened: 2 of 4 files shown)


terencecho (approved):
**APPROVE** at `160b2207edf304fd5f6f7689096085d3f2982fc9`. Small, idiomatic, correctly scoped. Details:

**1. Suppression scope — sweep_static only, not a check-wide muzzle.**

The `hasNoTimelineDeclaration` boolean is threaded into `detectSweepStatic` alone (`checkPipeline.ts:492-495`). Every other detector in `runAuditGrid` — `detectRotationPivotDrift`, motion, contrast, layout geometry sweeps — still runs untouched. That matches the PR-body claim ("suppresses only that diagnostic and leaves all other checks intact") and it's the right scope: `data-no-timeline` says "this composition isn't timeline-driven," not "skip all layout audits."

**2. Marker convention already established in the codebase.**

`data-no-timeline` is the same root opt-out already honored by:
- `frameCapture.ts:1615,1668` (producer skips the timeline-registration poll — same "intentional CSS-animation / rAF driven" host semantics as the `frameCapture.ts:1680` message)
- `lint/rules/composition.ts:697` (lint rule warns when root has no timeline registration AND no `data-no-timeline`)

Wiring `check` into the same convention makes the three consumers agree on one root-level opt-out attribute. `lint/rules/composition.test.ts:810,819` locks the `\bdata-no-timeline\b` word-boundary discipline for the lint side; the browser-side `hasAttribute("data-no-timeline")` here is exact-name, so the `data-no-timeline-start` false-positive class the lint tests guard against can't recur here.

**3. Root-composition selector consistent with sibling calls.**

`document.querySelector("[data-composition-id]")` picks the first-in-document-order element, i.e. the root, matching the bare-form queries at `checkBrowser.ts:712,752` and the "first `<div ... data-composition-id ...>`" regex in `cloud/detectAspectRatio.ts:59`. Nested sub-compositions also carry `data-composition-id` (see `cloud/detectAspectRatio.test.ts:160-165` — "Sub-compositions also have data-composition-id") but the first-in-DOM is always the root by construction. Reading only the root's marker is correct: a timeline-backed root with a static sub-composition inside it should still get sweep_static coverage on the root's motion, and this diff preserves that.

**4. Backward-compat via `fakeDriver` default.**

`fakeDriver` gets `hasNoTimelineDeclaration: vi.fn(async () => false)` in `check.test.ts:148`. All existing scenarios (frozen static, 1.5s title card, motion frozen, transition boundaries, etc.) continue exercising the pre-existing `detectSweepStatic` path with the marker absent — no test needs to opt in to keep its prior semantics.

**5. Positive-path test is exactly the shape needed.**

`test does not flag intentional static content declared with data-no-timeline` sets `getDuration: 6` (above `SWEEP_STATIC_MIN_DURATION_SEC`) + `collectLayoutGeometry: "frozen"` (would normally trip the guard) + `hasNoTimelineDeclaration: true`, then asserts `sweep_static` is not in the findings. Positive assertion is precise — checks the exact code that would fire, not the whole report. And the existing "1.5s static title card" negative case still guards the short-duration short-circuit (`checkPipeline.ts:493`).

**6. Cost — one extra `page.evaluate` per check run.**

`await driver.hasNoTimelineDeclaration()` runs unconditionally in `runAuditGrid:1076`. One `page.evaluate` executing a single `querySelector` + `hasAttribute` is negligible relative to the surrounding audit work. Not on a per-frame hot path.

CI: all green at head (`Test`, `Producer: unit tests`, `Producer: integration tests`, `Typecheck`, `Lint`, `Format`, `CLI smoke`, `CLI: npx shim ×3`, `Tests on windows-latest`, `Preview parity`, `Preflight`). `mergeStateStatus=BLOCKED` is review-gated, not check-gated; this approval unblocks.

— Review by tai (pr-review)

PR #3534, opened 2026-08-28
fix(cli): prevent keyframe shots overwriting sources

miguel-heygen:
## Summary

`hyperframes keyframes --shot` now fails before capture when the requested PNG would overwrite the active composition source. The guard compares both normalized paths and existing file identity, so filesystem aliases cannot bypass it, while a separate output path continues to produce the onion-skin PNG normally.

Fixes `reported:1787891587.329489`. Feedback: https://slack.com/archives/C0BGC335AQY/p1787891587329489

## Test plan

- `bun run --cwd packages/cli test -- src/commands/keyframes.test.ts` (19 passed)
- `bunx oxlint` and `bunx oxfmt --check` on changed files
- Production-shaped Terra fixture: same-path output exits 1 with the HTML SHA-256 unchanged; external output exits 0 and writes a valid 320×180 PNG

---

Codex

### DIFF
--- a/packages/cli/src/commands/keyframes.test.ts
+++ b/packages/cli/src/commands/keyframes.test.ts
@@ -1,4 +1,4 @@
-import { existsSync, mkdtempSync, mkdirSync, writeFileSync } from "node:fs";
+import { existsSync, linkSync, mkdtempSync, mkdirSync, readFileSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { beforeAll, describe, expect, it } from "vitest";
@@ -28,6 +28,29 @@ describe("keyframes direct composition scope", () => {
 });
 
 describe("keyframes shot output", () => {
+  it("rejects an output path that would overwrite the composition source", () => {
+    const projectDir = mkdtempSync(join(tmpdir(), "hf-keyframes-shot-source-"));
+    const sourcePath = join(projectDir, "index.html");
+    writeFileSync(sourcePath, wrap(""));
+
+    expect(() => ensureShotOutputDir(sourcePath, sourcePath)).toThrow(
+      /must not overwrite the composition source/,
+    );
+    expect(readFileSync(sourcePath, "utf8")).toBe(wrap(""));
+  });
+
+  it("rejects an existing output alias that refers to the composition source", () => {
+    const projectDir = mkdtempSync(join(tmpdir(), "hf-keyframes-shot-alias-"));
+    const sourcePath = join(projectDir, "index.html");
+    const aliasPath = join(projectDir, "shot.png");
+    writeFileSync(sourcePath, wrap(""));
+    linkSync(sourcePath, aliasPath);
+
+    expect(() => ensureShotOutputDir(aliasPath, sourcePath)).toThrow(
+      /must not overwrite the composition source/,
+    );
+  });
+
   it("creates a missing parent directory before writing --shot", () => {
     const projectDir = mkdtempSync(join(tmpdir(), "hf-keyframes-shot-dir-"));
     const outputDir = join(projectDir, "nested", "proofs");
--- a/packages/cli/src/commands/motionShot.ts
+++ b/packages/cli/src/commands/motionShot.ts
@@ -10,8 +10,8 @@
 // exactly what it's editing. All geometry + SVG live in ./motionShotLayout.ts
 // (pure, tested); this file only drives the browser and SAMPLES.
 
-import { mkdirSync, writeFileSync } from "node:fs";
-import { dirname } from "node:path";
+import { mkdirSync, statSync, writeFileSync } from "node:fs";
+import { dirname, resolve } from "node:path";
 import { resolveDiagnosticNavigationTimeoutMs } from "../utils/renderArgs.js";
 import { resolveCompositionViewportFromHtml } from "../utils/compositionViewport.js";
 import {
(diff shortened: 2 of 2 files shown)


james-russo-rames-d-jusso (commented):
Quick round-2 look — the two-check guard (string-normalized path + `dev`+`ino` identity via `statSync`) is the right shape for this class of bug. LGTM from my side; leaving as a comment. HEAD at `b1abc4c`, all required checks green.

## Guard soundness — sound

`pathsReferToSameFile` at `packages/cli/src/commands/motionShot.ts:36-46` layers:
1. `path.resolve()` string equality — catches relative-vs-absolute and `..` normalization.
2. `statSync().dev === dev && .ino === ino` — catches symlinks, hardlinks, and case-insensitive-FS collisions (macOS APFS `Out.PNG` vs `out.png` resolve to the same inode). `statSync` follows symlinks (as opposed to `lstatSync`), so a symlinked `out.png → source.html` matches on inode even though the resolved strings differ.

The `try/catch` returning `false` when one side doesn't `stat` is the right default: if the output file doesn't exist yet, there is nothing to overwrite via alias — the subsequent `writeFileSync` creates a fresh file, so failing open here doesn't create a hazard.

Callsite at `motionShot.ts:665` correctly resolves the source before comparison:
```
ensureShotOutputDir(outPath, resolve(projectDir, opts.entryFile ?? "index.html"));
```
Runs before browser capture — bug's "fails BEFORE capture" invariant holds.

## Test coverage — proportionate

`keyframes.test.ts:31-51` adds:
- Same-path → throws, and the on-disk HTML content is unchanged (`readFileSync` assertion) — this is the load-bearing invariant from Miguel's Terra fixture, mirrored in unit form.
- Hard-link alias (`linkSync`) → throws.

Symlink coverage is implicit (same `dev`+`ino` code path). Case-insensitive-FS coverage is implicit (inode match). Both are well-established Node behaviors; explicit tests would be OS-conditional and low-marginal-value. The 19 passing tests + real-tmpdir fixtures are appropriately grounded.

## Scope note — non-blocking

The guard protects only the *entry HTML* (`opts.entryFile ?? "index.html"`) — not other composition assets the HTML references (CSS/JS/PNG under the same dir). A user targeting `--shot=assets/logo.png` when `logo.png` is a real asset would still stomp it. This wasn't the reported bug and doesn't belong in this PR, but worth a follow-up ticket if there's evidence of anyone hitting it. Nit only — not blocking.

## What I didn't verify

- End-to-end exit-code-1 emission from the CLI wrapper on this thrown `Error` — trusted from Miguel's Terra-fixture testimony ("same-path exits 1 with HTML SHA-256 unchanged"). The unit tests confirm the throw; the CLI's top-level error handler is out of scope for this diff.
- Windows `dev`/`ino` semantics under NTFS — but "Tests on windows-latest" is green (14m45s), which covers this in practice.

— _Review by Rames D Jusso_

jrusso1020 (approved):
Approving at `b1abc4c`. CI green at this head — 30 checks success, 10 skipped, nothing failed and nothing `cancelled`; all eight required contexts pass including both Windows lanes, which is the empirical answer to whether `statSync().ino` is trustworthy on the NTFS runner.

The fact this guard actually hinges on is whether it names the same file the bundler reads, so I traced that rather than the alias shapes.

`ensureShotOutputDir` compares `outPath` against `resolve(projectDir, opts.entryFile ?? "index.html")` (`motionShot.ts:665`), but the bundle is produced by `bundleToSingleHtml(projectDir, { entryFile: opts.entryFile })` (`:676`), which resolves its entry through a *different* helper — `resolveWithinProject(projectDir, entryFile)` (`htmlBundler.ts:801`). Two different expressions on the two sides is exactly where a guard quietly starts protecting the wrong path. It holds here: `resolveWithinProject` returns `resolve(base, relativePath)` when `isSafePath` passes and `null` otherwise (`safePath.ts:69-72`), and on `null` the bundler throws ``​`${entryFile} not found in project directory`​`` (`htmlBundler.ts:802-803`) before reading anything. Both sides default to `"index.html"` (`:800`). So on every path that proceeds the two expressions denote the same file, and on the paths where they could diverge nothing is read or written. The guard cannot be aimed at the wrong file.

Alias coverage is right, and for the stated reason: `resolve()` settles exact-path, `..` traversal and relative-vs-absolute by string; `dev`/`ino` settles symlink, hard link and case-insensitive-FS by identity. `statSync` following symlinks is correct over `lstatSync` — an aliased output has to match on the target's inode, not the link's. The `catch { return false }` reads fail-open but is load-bearing rather than careless: the normal case is an output PNG that does not exist yet, so ENOENT means proceed, and the string comparison already covers the exact-path case without touching the disk.

The optional second parameter was the one thing that could have left a hole, since any existing caller omitting it stays silently unguarded. There is exactly one production call site (`motionShot.ts:665`) and it passes the source; the only other reference is the test.

Agreed the asset-scope point is non-blocking — the guard covers the entry HTML, so `--shot=assets/logo.png` can still stomp a referenced asset. That is a wider fix than the reported bug and reads fine as a follow-up.

One correction for the record, since it changes *when* to stamp rather than whether: `main` here has no classic branch protection at all — it is the `main` ruleset, which sets `dismiss_stale_reviews_on_push: false` alongside `require_last_push_approval: true`. So an approval is not dismissed by a push; it survives and simply stops satisfying the gate until the newest push is itself approved. The same ruleset sets `strict_required_status_checks_policy: false`, so no update-from-main is required here.

— Rames

PR #3535, opened 2026-08-28
fix(core): resolve long nested media in local time

miguel-heygen:
## Summary

Long videos starting at local `0s` inside delayed sub-compositions no longer disappear from late snapshots. Media starts before the host’s absolute window now inherit the host offset even when their long duration overlaps that window; legacy media authored with a start already inside the global host window keeps its existing behavior.

Fixes `reported:1787898835.985789` and `reported:1787908114.998239`.

- https://slack.com/archives/C0BGC335AQY/p1787898835985789
- https://slack.com/archives/C0BGC335AQY/p1787908114998239

## Test plan

- `packages/core/src/runtime/init.test.ts` (91 passed, including legacy global-PIP coverage)
- `packages/cli/src/commands/snapshot.test.ts` (33 passed)
- Core and CLI typechecks
- Exact Terra point fixture: `40s`, `41s`, `80.4s`, and `111s` snapshots are all non-black
- Exact Terra 18-frame sweep: every in-host sample from `42.353s` through `112.941s` is non-black; only pre-host and post-host samples are black

---

Codex

### DIFF
--- a/packages/core/src/runtime/init.test.ts
+++ b/packages/core/src/runtime/init.test.ts
@@ -1215,11 +1215,13 @@ describe("initSandboxRuntimeModular", () => {
 
     initSandboxRuntimeModular();
 
+    for (const globalTime of [
+      42.353, 49.412, 56.471, 63.529, 70.588, 77.647, 84.706, 91.765, 98.824, 105.882, 112.941,
+    ]) {
+      window.__player?.renderSeek(globalTime);
+      expect(video.style.visibility, `global ${globalTime}s`).toBe("visible");
+    }
     expect(window.__hfResolveMediaStartSeconds?.(video)).toBeCloseTo(39.233);
-    window.__player?.renderSeek(80.4);
-    expect(video.style.visibility).toBe("visible");
-    window.__player?.renderSeek(111);
-    expect(video.style.visibility).toBe("visible");
   });
 
   it("uses the canonical resolver for reference starts, auto-start media, and inline hosts", () => {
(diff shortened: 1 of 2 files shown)


vanceingalls (approved):
Correctness read:

The fix drops `authoredEnd`-based overlap and disambiguates purely on `authoredStart`'s position relative to the host window. That's the right call — a long local clip can end past the host's absolute window without changing what the start means, so the duration was a false signal. Three shape checks:

- *Composition-local (host@39.233 + video@0/dur=80):* `authoredStart=0 < inheritedStart=39.233` → `startsInsideHostWindow=false` → returns `inheritedStart + authoredStart = 39.233`. Video visible through the host tail. Matches the added test at `init.test.ts:1187`.
- *Legacy root-global PIP (host@45.4 + video@45.4):* `authoredStart >= inheritedStart` and `authoredStart < hostEnd` → true → returns `authoredStart=45.4`. Legacy path preserved.
- *hostDuration null/0:* `hostEnd=null` → guard collapses to `authoredStart >= inheritedStart`, same as the pre-fix branch. No behavior change on unknown-duration hosts.

The retained inline comment on lines 663-669 explains *why* duration was unsound — first-principles, no ticket citation. Good comment.

CI note: `Tests on windows-latest` shows red, but the vitest report inside it is 196 files / 2852 tests passing with `Exited with code 0`; the wrapper exits 1 after an LFS-pointer checkout phase warning ("Encountered 68 files that should have been pointers"). Reads as runner LFS-state flake, not a regression from this change — and this check isn't in the required set (`mergeable: MERGEABLE`). Worth a rerun to green-up the record if you'd like.

_Review by Via_

vanceingalls (approved):
R3 re-stamp at d55f9f2 — test-only additive delta (single `initSandboxRuntimeModular` test covering long nested media staying visible across host schedule window at 11 checkpoints from 42.353s to 112.941s + resolver=39.233s). Same fix as approved at 4d9d5ba0, no production code touched. All 30 required checks green.

— Via

PR #3538, opened 2026-08-28
fix(engine): treat a single keyframe as the worst sparse-keyframe case

### LINKED ISSUE #3460: Sparse-keyframe check silently skips single-keyframe videos — the worst case goes unreported (0.8.10)
## Summary

```js
if (timestamps.length < 2) {
}
```

## Reproduction

```bash
```

Keyframe timestamps:

```
```

```html
</div>
```

```bash
```

## Observed

```
```

## Expected

## Suggested fix

```js
  return {
    avgIntervalSeconds: duration,
    maxIntervalSeconds: duration,
    keyframeCount: 1,
    isProblematic: duration > 2,
  };
}
}
```

## Why it matters in practice

rajanpanth:
Fixes #3460

### Problem

`analyzeKeyframeIntervalsUncached` returned `isProblematic: false` whenever a video had fewer than two keyframes:

```js
if (timestamps.length < 2) {
}
```

A video with exactly one keyframe is the worst case for the failure this check exists to catch: every seek past 0 lands inside the single GOP spanning the whole file. The compiler warns loudly about a 5 second GOP and stays silent about a 10 second one.

Two 10s clips, identical except GOP size, confirm it: `five-second-gop.mp4` (keyframes at 0 and 5) gets flagged, `single-gop.mp4` (keyframe at 0 only, effective seek distance 10s) does not.

### Fix

Split the early return into two cases:

- zero keyframes (still images, failed probes) keeps the prior not-problematic result
- one keyframe reports the video stream's own duration as both the average and max interval, flagged past the same 2s threshold as the multi-keyframe path

Stream duration rather than container duration, since they can disagree, which is exactly the class of bug behind #3372.

### Testing

Added three cases to `ffprobe.test.ts`:

- a 10s single-GOP video reports `avgIntervalSeconds: 10, maxIntervalSeconds: 10, isProblematic: true`
- a 1s single-GOP video stays `isProblematic: false`
- zero keyframes still returns the untouched not-problematic result, and skips the metadata probe entirely (asserted via call count)

Reverting `ffprobe.ts` fails the first new test with the old `avgIntervalSeconds: 0`, so it covers the change. Full file passes at 117 tests (1 pre-existing skip), `oxlint` and `oxfmt --check` are clean.

### Note on the issue's side observation

The issue also raises that `hyperframes render` may resolve frames correctly even past a sparse keyframe, since `video_extract` decodes through ffmpeg from the preceding keyframe with `minVideoFrameCoverageRatio: 1` rather than relying on browser seeking. I have not touched the warning's wording or scope here, since that is a question about what the check is actually protecting against and felt like a separate decision from the reported false negative. Happy to follow up if useful.

### DIFF
--- a/packages/engine/src/utils/ffprobe.test.ts
+++ b/packages/engine/src/utils/ffprobe.test.ts
@@ -832,6 +832,96 @@ describe("ffprobe missing-binary fallback", () => {
     expect(basename(calls[0]?.command ?? "")).toMatch(/^ffprobe(?:\.exe)?$/);
   });
 
+  it("analyzeKeyframeIntervals treats a single keyframe as the whole stream duration", async () => {
+    // A 10s single-GOP video: exactly one keyframe at t=0. Every seek past
+    // it lands inside that one GOP, so the effective interval is the whole
+    // stream, not zero.
+    const { spawn, calls } = createSpawnSpy([
+      { kind: "exit", code: 0, stdout: "0.000000\n" },
+      {
+        kind: "exit",
+        code: 0,
+        stdout: JSON.stringify({
+          streams: [
+            {
+              codec_type: "video",
+              codec_name: "h264",
+              width: 640,
+              height: 360,
+              r_frame_rate: "30/1",
+              avg_frame_rate: "30/1",
+              duration: "10.0",
+            },
+          ],
+          format: { duration: "10.0" },
+        }),
+      },
+    ]);
+    vi.resetModules();
+    vi.doMock("child_process", () => ({ spawn }));
+
+    const { analyzeKeyframeIntervals } = await import("./ffprobe.js");
+    const result = await analyzeKeyframeIntervals("/tmp/single-gop.mp4");
+
+    expect(result).toEqual({
+      avgIntervalSeconds: 10,
+      maxIntervalSeconds: 10,
+      keyframeCount: 1,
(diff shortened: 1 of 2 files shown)


miga-heygen:

rajanpanth:
Rebased onto current `main` (`6150c296c`) and re-checked that this is still relevant.

The bug is still on main: `analyzeKeyframeIntervalsUncached` returns early for `timestamps.length < 2`, so a stream with a single keyframe reports `avgIntervalSeconds: 0` and `isProblematic: false`, even though one keyframe means every seek lands in a single GOP spanning the whole stream.

- `packages/engine/src/utils/ffprobe.test.ts`: 117 passed, 1 skipped.
- Mutation check: with `ffprobe.ts` reverted to main's version and the new spec kept, `analyzeKeyframeIntervals treats a single keyframe as the whole stream duration` fails with `expected { avgIntervalSeconds: 10, isProblematic: true } / received { avgIntervalSeconds: 0, isProblematic: false }`. Restored, back to 117 passing.

One note: GitHub refuses to let me reopen this myself, `reopenPullRequest` returns a bare 422 for all four PRs in the batch, so the button is not available on my side. The branch `rajanpanth:fix/single-keyframe-sparse-check` is rebased and passing if you are able to reopen it.

PR #3540, opened 2026-08-28
fix(dx): stop the studio filesize pre-commit hook from crashing on Windows

rajanpanth:
## What

Fixes a crash in the `filesize` pre-commit hook (`packages/studio/**/*.{ts,tsx}`, 600 LOC cap) on Windows: `sh.exe -c` fails with `syntax error: unexpected end of file` on every commit touching `packages/studio`, regardless of what's staged.

## Why

Lefthook builds the command line for a `run: |` block by wrapping the whole script in one double-quoted argument. The old script quoted `$f`/`$lines` and an echo message, so it contained literal `"` characters. On Windows, one of those closes the outer double-quoted argument early, truncating and unbalancing the actual command `sh.exe -c` receives.

Confirmed this is the mechanism, not something specific to my staged files: reproduced with a single trivial 1-line file, then with `lefthook run pre-commit --command filesize --verbose` to see the exact command lefthook hands to `sh.exe`, then fixed the crash by removing the embedded `"` characters from an experimental copy of the script with no other change, which fixed it. Reverted that experiment in favor of the more robust fix below before opening this PR.

## How

Moved the check into `scripts/check-studio-filesize.sh`, following the same pattern `largefiles` already uses one command below it in `lefthook.yml` (`check-large-files.sh`), which reads the staged set itself for a related reason (avoiding `{staged_files}` splitting on spaces in filenames). Reading files from a script file rather than inlining the script into the command line sidesteps both problems: quoting inside the script file never touches lefthook's command-line construction.

The new script reads `git diff --cached` itself and replicates the previous `glob`/`exclude` filtering (`packages/studio/**/*.{ts,tsx}`, excluding `*.test.ts`/`*.test.tsx`/`*.generated.*`) with `case` pattern matching instead of lefthook's glob config, so `lefthook.yml`'s `filesize` job now just calls the script with no arguments.

## Test plan

- [ ] Unit tests added/updated (a small shell hook script; exercised directly rather than via a test framework, see below)

Ran the script directly against explicit file arguments: a 700-line non-test file (flagged), a 700-line `*.test.ts` file (correctly skipped), and a 5-line file (correctly passed). Ran `lefthook run pre-commit --command filesize --verbose` before and after: crashed with the syntax error before, passes cleanly after. Ran the full `lefthook run pre-commit` suite (lint, format, fallow, typecheck, largefiles, filesize, tracked-artifacts) against a real in-progress branch touching `packages/studio` — all green, confirming the new script doesn't change what passes/fails, only that it runs at all on Windows.

### DIFF
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -60,18 +60,11 @@ pre-commit:
       # Scoped to packages/studio — the 600 LOC limit is a studio architecture
       # standard enforced as part of the App.tsx decomposition work. Player and
       # other packages enforce size discipline via code review and convention.
-      glob: "packages/studio/**/*.{ts,tsx}"
-      exclude: "(\\.test\\.(ts|tsx)$|\\.generated\\.)"
-      run: |
-        for f in {staged_files}; do
-          # Skip test and generated files (exclude pattern backup in case lefthook doesn't filter)
-          case "$f" in *.test.ts|*.test.tsx|*.generated.*) continue ;; esac
-          lines=$(wc -l < "$f")
-          if [ "$lines" -gt 600 ]; then
-            echo "ERROR: $f has $lines lines (max 600)"
-            exit 1
-          fi
-        done
+      # The script reads the staged set and does its own glob/exclude filtering
+      # (see its header comment for why: a `run: |` block containing a literal
+      # `"` breaks lefthook's Windows command line, same class of problem as
+      # `largefiles` below).
+      run: ./scripts/check-studio-filesize.sh
 
 commit-msg:
   commands:
--- a/scripts/check-studio-filesize.sh
+++ b/scripts/check-studio-filesize.sh
@@ -1,0 +1,64 @@
+#!/bin/sh
+# Enforce the packages/studio 600 LOC file cap (a studio architecture standard
+# from the App.tsx decomposition work; Player and other packages enforce size
+# discipline via code review and convention instead).
+#
+# Usage:
+#   check-studio-filesize.sh                 # default: check the staged file set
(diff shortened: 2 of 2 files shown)


miga-heygen:

rajanpanth:
Rebased onto current `main` (`6774c7e75`). Still relevant: the `filesize` block in `lefthook.yml` still contains literal `"` characters (quoting `$f` and the `echo` message), which is the thing that breaks on Windows.

I hit this live while rebasing the other PRs in this batch, which is the clearest evidence I can offer. Committing to `packages/studio` on Windows gives:

```
✔️ largefiles (1.32 seconds)
✔️ tracked-artifacts (1.54 seconds)
✔️ lint (1.91 seconds)
✔️ format (2.77 seconds)
✔️ fallow (10.84 seconds)
✔️ typecheck (33.72 seconds)
🥊 filesize (1.12 seconds)
```

with `exit status 2` and no file anywhere near the 600 line cap (the largest staged file was 363 lines). Every other hook passes; only `filesize` dies, and it dies before it can evaluate anything. The commit is rejected outright, so the only way through is `--no-verify`, which skips the real hooks too.

Verification of the replacement script after the rebase:

- 700 line file under `packages/studio`: exits 1 and prints `packages/studio/src/__sizetest.tsx (700 lines)`.
- 10 line file: exits 0.
- Same `HF_STUDIO_MAX_LINES` override and the same glob/exclude behaviour as the inline version, with the filtering moved into the script so no literal `"` reaches lefthook's Windows command line. This is the same reasoning `check-large-files.sh` already applies.

GitHub gives me a 422 on reopen for all four PRs in this batch, so I cannot reopen it myself.

PR #3547, opened 2026-08-29
fix(studio): activate a composition at any path, not just compositions/

### LINKED ISSUE #3543: Studio: selecting a non-root composition file in the Comps panel keeps showing the master composition
**Version:** hyperframes 0.8.17 (via `npx hyperframes@latest`), WSL2 Ubuntu 22.04, Chrome.

**Setup:** a project with `index.html` (591 s, `data-composition-id="main"`) plus `parts/part-1.html` … `parts/part-4.html` (142–174 s each, unique `data-composition-id` values), all generated from one timeline. `lint`/`check` pass. `hyperframes render -c parts/part-1.html` renders the correct 142 s video.

**Repro:**
1. `hyperframes preview`, open the project in Studio.
2. In the **Comps** panel, click `parts/part-1`.

**Expected:** the canvas and timeline show `parts/part-1.html` (142 s, its own splash/hook).

**Actual:** the row highlights and the hash becomes `#project/<id>?comp=parts%2Fpart-1.html`, but the canvas and timeline keep showing `index.html` (09:51, the root's splash). Same for every part. The part's thumbnail in the Comps panel *is* correct.

**Notes from digging:**
- Server side looks fine: `GET /api/projects/<id>/preview/comp/parts/part-1.html` returns the compiled part page with the right duration and content.
- In the Studio bundle, the preview iframe is only swapped to another file when drilling into a `data-composition-src` mount (`if (!m.compositionSrc) return`); a standalone `.html` chosen from the list never gets its own preview, so `activeCompPath` changes but the master stays mounted.
- Reproduced headlessly (puppeteer): after clicking the part, the timeline still reads `00:00 / 09:51` and the canvas shows the root hook sticker.

**Impact:** multi-part builds (one source → root + `parts/part-N.html`, rendered with `render -c`) can't be reviewed in Studio, and any edit made there lands in the root file instead of the part. Seen on two projects now.

**Workaround:** review parts via `snapshot` (temporarily swapping the part into `index.html`) or by rendering.

rajanpanth:
## What

Fixes #3543. Selecting a composition in the Comps panel now actually switches the canvas and timeline to it, for any path, not only files under `compositions/`.

## Why

`useCompositionStack`'s activation effect ran an if/else-if chain:

```ts
if (activeCompositionPath === "index.html") { /* master */ }
else if (activeCompositionPath && activeCompositionPath.startsWith("compositions/")) { /* push level */ }
else if (!activeCompositionPath) { /* master */ }
```

A path like `parts/part-1.html` is not `index.html`, does not start with `compositions/`, and is not falsy, so it matched no branch at all and the effect did nothing. The Comps panel still highlighted the row and the URL hash still updated, but the stack kept the master mounted: the canvas and timeline stayed on `index.html`, and, as the issue notes, any edit made there landed in the root file instead of the part.

The reporter's setup is the case that exposes it, a generated multi-part build with `index.html` plus `parts/part-1.html` through `parts/part-4.html`. Nothing requires a project's compositions to live under `compositions/`; the server already serves any of them (`GET /api/projects/<id>/preview/comp/parts/part-1.html` returns the right page, per the issue).

## How

Replaced the prefix test with a plain truthiness check, so the root composition stays on the master level and every other path pushes its own level. The final `else` now covers null/empty.

Label derivation is unchanged (`replace(/^compositions\//, "").replace(/\.html$/, "")`), which is the same expression `CompositionsTab` uses for the panel's own card labels, so a `compositions/`-relative project's labels are byte-identical to before and `parts/part-1.html` reads as `parts/part-1`.

## Test plan

- [ ] Manual testing performed (see note)

Extended `useCompositionStack.test.tsx` with 6 cases: a parameterised set asserting a level is pushed with the right `id` and `previewUrl` for `compositions/scene-a.html` (the previously-working case, as a regression guard), `parts/part-1.html` (the issue's case), a bare `chapter-2.html`, and a deep `a/b/c/deep.html`; plus a label assertion and a guard that `index.html` still leaves the master alone.

Mutation-tested: reverting only `useCompositionStack.ts` and keeping the tests fails the three non-`compositions/` cases (stack length 1 instead of 2), while the `compositions/` and `index.html` cases still pass, which is exactly the reported split. 8/8 pass with the fix.

I did not reproduce this in a running Studio: it needs a multi-part project and a live preview server, and the mechanism is fully determined by this effect's branch conditions, which the tests cover directly. The issue author already confirmed the runtime symptom headlessly with puppeteer.

`lefthook run pre-commit` all green (lint, format, fallow, typecheck, filesize, largefiles, tracked-artifacts, commitlint).

### DIFF
--- a/packages/studio/src/components/nle/useCompositionStack.test.tsx
+++ b/packages/studio/src/components/nle/useCompositionStack.test.tsx
@@ -42,3 +42,72 @@ describe("useCompositionStack — project scoping", () => {
     });
   }
 });
+
+describe("useCompositionStack — activating a composition by path", () => {
+  afterEach(() => {
+    document.body.innerHTML = "";
+  });
+
+  function mountStack(activeCompositionPath: string | null) {
+    const host = document.createElement("div");
+    document.body.append(host);
+    const root = createRoot(host);
+    const seen: { stack: ReturnType<typeof useCompositionStack>["compositionStack"] } = {
+      stack: [],
+    };
+
+    function Harness(props: { activeCompositionPath: string | null }) {
+      seen.stack = useCompositionStack({ projectId: "p", ...props }).compositionStack;
+      return null;
+    }
+
+    return { root, seen, Harness, activeCompositionPath };
+  }
+
+  // The root stays on the master level; everything else pushes a second level.
+  for (const path of [
+    "compositions/scene-a.html",
+    "parts/part-1.html",
+    "chapter-2.html",
+    "a/b/c/deep.html",
+  ]) {
+    it(`pushes a level for ${path}`, async () => {
+      const { root, seen, Harness } = mountStack(path);
+
+      await act(async () => {
+        root.render(<Harness activeCompositionPath={path} />);
+      });
+
+      expect(seen.stack).toHaveLength(2);
+      expect(seen.stack[1]?.id).toBe(path);
+      expect(seen.stack[1]?.previewUrl).toBe(`/api/projects/p/preview/comp/${path}`);
+
+      act(() => root.unmount());
+    });
+  }
+
+  it("labels a non-compositions/ path without mangling it", async () => {
+    const { root, seen, Harness } = mountStack("parts/part-1.html");
+
+    await act(async () => {
+      root.render(<Harness activeCompositionPath="parts/part-1.html" />);
+    });
+
+    expect(seen.stack[1]?.label).toBe("parts/part-1");
+
+    act(() => root.unmount());
+  });
+
+  it("keeps the master alone for the root composition", async () => {
+    const { root, seen, Harness } = mountStack("index.html");
+
+    await act(async () => {
+      root.render(<Harness activeCompositionPath="index.html" />);
+    });
+
+    expect(seen.stack).toHaveLength(1);
+    expect(seen.stack[0]?.id).toBe("master");
+
+    act(() => root.unmount());
+  });
+});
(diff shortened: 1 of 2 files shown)


miguel-heygen (approved):

miguel-heygen (approved):
Reviewed end to end at `adf9b0cceeb561fd5424e8cad73b781f5383a9ef`.

**Strengths**
- `packages/studio/src/components/nle/useCompositionStack.ts:108` preserves the explicit root-composition path while `:111-124` correctly routes every other truthy project-relative composition path through the existing encoded preview contract.
- `packages/studio/src/components/nle/useCompositionStack.test.tsx:68-109` covers the prior `compositions/` behavior, the reported `parts/` case, bare/deep paths, label preservation, and the root guard.

**Verification**
- Focused Vitest: 8/8 passed.
- Studio TypeScript: `tsc --noEmit` passed.
- Synthetic local PR fixture, desktop + mobile: selecting `parts/part-1.html` updated the URL hash, preview iframe (`PART ONE CANVAS`), composition breadcrumb, and timeline (`Part Title`) together.
- Browser console: no errors. The optional local FFmpeg/thumbnail probes were unavailable/timed out in the synthetic environment; the target composition preview endpoint loaded and rendered correctly.
- Required GitHub CI: 8/8 passed; no review comments or unresolved threads.

**Verdict:** APPROVE
**Reasoning:** The fix addresses the no-branch path bug without changing root behavior, is directly regression-tested, passes the relevant local and required CI gates, and the exact user-facing selection flow works at the reviewed SHA.

— Magi

PR #3554, opened 2026-08-30
fix(player): keep the parent tick clock alive in a cross-origin iframe

miguel-heygen:
A composition in a cross-origin iframe froze a few seconds in while the player still reported `paused === false`. The parent tick clock is what should have kept it moving, and three separate things stopped it.

## The parent tick clock

It exists to drive a composition whose own rAF Chromium has throttled, which is the normal state of a cross-origin frame. It was being torn down before it delivered a single tick.

**Its lifetime had two owners.** The rAF loop re-read `_paused` every frame and returned permanently on the first `true` it saw. That field has a second writer: the runtime's `state` message, which reports the iframe's playback as of the moment it was posted. A state message already in flight when `play()` runs carries the pre-play `isPlaying: false`, lands a frame later, and ends the loop for good — there is no restart path. The loop now runs between its explicit start and stop and nothing else decides its lifetime. Reaching the end of the timeline is the one playback stop the runtime initiates on its own, so that one is relayed to the clock explicitly.

**The `load` handler tore down a document that was already playing.** `load` marks the end of a document's subresource fetching, not the start of a document, and the runtime announces its timeline on `DOMContentLoaded` — so on a composition still fetching images or fonts it arrives *after* the player has gone ready and started driving that same document, and cancelled the clock a `play()` in the `ready` handler had just started. A genuinely new document is always preceded by assigning `src`/`srcdoc`, which clears `_ready` first, so a load seen while ready has nothing to reset.

**The tick advanced the timeline without reporting it.** With the clock alive, the runtime seeked on each parent tick but never posted state, so the embedder's playhead, `timeupdate` and end-of-playback detection stayed frozen at the last frame the (now throttled) local transport managed to post. The parent drives this tick precisely when nothing else will report the position, so it now posts state like the local transport does. `postState`'s own change/interval filter keeps the message rate unchanged.

## On the reported runaway

A second, late `play()` was reported to jump the playhead forward at many times real speed. That is a consequence of the above rather than a clock bug: `TransportClock` is wall-clock and its `play()` is idempotent, so time kept running correctly while the starved frame fell behind, and the first tick delivered afterwards caught up in one step. With the clock running from the first `play()` there is nothing to catch up, and playback holds 1x throughout.

## Verification

Driven in Chrome against a composition with the runtime injected, `play()` issued from inside the `ready` handler, in four configurations:

| config | before | after |
|---|---|---|
| same-origin | 1x, 0 ticks sent | 1x, ~60 ticks/s |
| `Content-Security-Policy: sandbox allow-scripts` | 1x, 0 ticks sent | 1x, ~60 ticks/s |
| cross-origin host | 1x, 0 ticks sent, `ready` dispatched twice | 1x, ~60 ticks/s, `ready` once |
| cross-origin and scrolled out of the viewport, so Chromium suspends the frame's rAF | **`currentTime` stuck at 0 for the whole run with `paused === false`** | 1x, `ended` at 14.99s of a 15s composition |

`currentTime` after the change tracks wall clock to the millisecond in all four (2.000 / 5.000 / 8.000 / 14.500 at 2000 / 5000 / 8000 / 14500 ms, then 15.000 and paused). The tick clock is stopped exactly once per run, at the end of the timeline. The same-origin direct-timeline path is untouched.

Three unit regressions cover the player side and one covers the runtime side; each was confirmed to fail with only its own fix reverted.

`bun run --filter @hyperframes/player test` 348 passed, `bun run --filter @hyperframes/core test` 2580 passed, `bun run --filter '*' typecheck` and `bun run lint` clean.

### DIFF
--- a/packages/core/src/runtime/init.test.ts
+++ b/packages/core/src/runtime/init.test.ts
@@ -3307,3 +3307,65 @@ describe("initSandboxRuntimeModular", () => {
     });
   });
 });
+
+describe("parent-driven transport tick", () => {
+  const originalRequestAnimationFrame = window.requestAnimationFrame;
+  const originalCancelAnimationFrame = window.cancelAnimationFrame;
+
+  beforeEach(() => {
+    document.body.innerHTML = "";
+    (globalThis as typeof globalThis & { CSS?: { escape?: (value: string) => string } }).CSS ??= {};
+    globalThis.CSS.escape ??= (value: string) => value;
+    // The recursive schedule inside transportTick is short-circuited by the
+    // runtime's own re-entry guard, so a synchronous rAF runs the local
+    // transport exactly once and then leaves it stopped — which is the state
+    // this suite needs: a frame whose own rAF has stopped delivering.
+    window.requestAnimationFrame = ((callback: FrameRequestCallback) => {
+      callback(0);
+      return 1;
+    }) as typeof window.requestAnimationFrame;
+    window.cancelAnimationFrame = (() => {}) as typeof window.cancelAnimationFrame;
+  });
+
+  afterEach(() => {
+    window.__hfRuntimeTeardown?.();
+    document.body.innerHTML = "";
+    window.__timelines = {} as Record<string, RuntimeTimelineLike>;
+    delete window.__player;
+    delete window.__playerReady;
+    vi.restoreAllMocks();
+    window.requestAnimationFrame = originalRequestAnimationFrame;
+    window.cancelAnimationFrame = originalCancelAnimationFrame;
+  });
+
+  function control(action: string, extra: Record<string, unknown> = {}): void {
+    window.dispatchEvent(
+      new MessageEvent("message", {
+        data: { source: "hf-parent", type: "control", action, ...extra },
+      }),
+    );
+  }
+
+  it("posts the position a parent tick advanced to", () => {
+    document.body.innerHTML = `<div data-composition-id="main" data-root="true" data-duration="10"></div>`;
+    window.__timelines = {};
+    const nowMs = vi.spyOn(performance, "now");
+    nowMs.mockReturnValue(1_000);
+    initSandboxRuntimeModular();
+
+    const posted: Record<string, unknown>[] = [];
+    vi.spyOn(window.parent, "postMessage").mockImplementation((message: unknown) => {
+      posted.push(message as Record<string, unknown>);
+    });
+
+    control("play");
+    nowMs.mockReturnValue(3_000);
+    control("tick");
+
+    const states = posted.filter((m) => m["type"] === "state");
+    // 2s of wall clock at the default 30fps canonical rate = frame 60. Without
+    // the tick reporting the position it advanced to, the embedder only ever
+    // sees the frame the (now stopped) local transport last posted.
+    expect(states.at(-1)).toMatchObject({ frame: 60, isPlaying: true });
+  });
+});
--- a/packages/core/src/runtime/init.ts
+++ b/packages/core/src/runtime/init.ts
@@ -3381,7 +3381,16 @@ export function initSandboxRuntimeModular(): void {
         runAdapters("pause");
         syncMediaForCurrentState();
         postState(true);
+        return;
       }
+      // The parent drives this tick precisely when our own rAF transport is
+      // throttled, so nothing else will report the position it just advanced
+      // to. Without this the composition animates while the embedder's
+      // playhead, "timeupdate" and end-of-playback detection stay frozen at
+      // the last frame the local transport managed to post. postState's own
+      // change/interval filter keeps the message rate the same as the local
+      // transport's.
+      postState(false);
     },
     onEnablePickMode: () => picker.enablePickMode(),
     onDisablePickMode: () => picker.disablePickMode(),
--- a/packages/player/src/hyperframes-player.test.ts
+++ b/packages/player/src/hyperframes-player.test.ts
@@ -2514,7 +2514,9 @@ describe("HyperframesPlayer parent tick clock lifetime", () => {
     _duration: number;
     _paused: boolean;
     _parentTickRaf: number | null;
+    _runtimeBridgeReady: boolean;
     _onMessage: (event: MessageEvent) => void;
+    probe: { start: () => void };
   };
 
   let player: PlayerInternal;
@@ -2596,10 +2598,49 @@ describe("HyperframesPlayer parent tick clock lifetime", () => {
     expect(tickCount()).toBe(1);
   });
 
+  it("stops ticking when the composition pauses itself short of the end", () => {
+    player.play();
+    advanceFrame();
+    expect(tickCount()).toBe(1);
+    // The runtime confirms the play, so the next not-playing report is a real
+    // stop rather than a message that crossed our own play command.
+    player._onMessage(stateMessage(30, true));
+
+    // Composition code calling pause() on `window.__player`: mid-timeline, so
+    // the completed-playback path never runs.
+    player._onMessage(stateMessage(60, false));
+    advanceFrame();
+    advanceFrame();
+
+    expect(player._parentTickRaf).toBeNull();
+    expect(tickCount()).toBe(1);
+  });
+
+  it("keeps ticking after a shader-option reload starts a new document", () => {
+    // _reloadShaderOptions reassigns the iframe document. If that path does not
+    // clear readiness, the load handler skips the teardown and the fresh
+    // document is left with no probe and a stale bridge flag.
+    player.setAttribute("src", "https://composition.example/comp.html");
+    player._ready = true;
+
+    const probeStart = vi.spyOn(player.probe, "start");
+
+    player.setAttribute("shader-loading", "eager");
+    expect(player._ready).toBe(false);
+
+    player.iframe.dispatchEvent(new Event("load"));
+
+    expect(probeStart).toHaveBeenCalled();
+    expect(player._runtimeBridgeReady).toBe(false);
+  });
+
   it("stops ticking once the composition reports the end of the timeline", () => {
     player.play();
     advanceFrame();
     expect(tickCount()).toBe(1);
+    // Playback reports itself as it runs; the runtime cannot reach the end of a
+    // timeline without having said it was playing on the way there.
+    player._onMessage(stateMessage(30, true));
 
     // frame 450 at the default 30fps protocol rate = 15s = the full duration.
     player._onMessage(stateMessage(450, false));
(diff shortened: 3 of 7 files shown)


jrusso1020 (changes requested):
Requesting changes. The core move is right: the tick loop had two owners deciding its lifetime and this gives it one, which is the correct fix rather than a special case for the racing message. Both findings below are about the completeness claims that the single-owner shape now rests on, since removing the `_paused` read means an exit path that does not call `_stopParentTickClock` no longer has a backstop.

## Blocker: `_reloadShaderOptions` reloads the iframe without clearing `_ready`, so the new early return skips the handshake

The added comment states the invariant the early return depends on:

> A genuinely new document is always preceded by assigning `src`/`srcdoc`, which clears `_ready` first.

That holds for the two `attributeChangedCallback` branches, which set `_ready = false` at `:210` and `:216` before assigning at `:212` and `:218`. It does not hold for `_reloadShaderOptions` at `:643-651`, which assigns `this.iframe.srcdoc` at `:646` or `this.iframe.src` at `:650` and never touches `_ready`. That method is reached from `attributeChangedCallback` at `:271-273` on `SHADER_CAPTURE_SCALE_ATTR` and `SHADER_LOADING_ATTR`.

So changing either shader attribute on a ready player loads a genuinely new document while `_ready` is still true, and `if (this._ready) return;` then skips the entire teardown for it: `_runtimeBridgeReady` stays `true` against a document that never handshook, `_directTimelineAdapter` keeps pointing at the previous timeline, `_media.resetForIframeLoad()` does not run, and `probe.start()` never fires, so the new document has no path to ready at all.

Before this change `_onIframeLoad` unconditionally ran that teardown, so the shader-option reload path worked. The fix is to clear `_ready` inside `_reloadShaderOptions` before assigning, matching what the `src`/`srcdoc` branches already do, with a regression test that flips a shader attribute on a ready player and asserts the probe restarts.

## Gap: a runtime-initiated pause before the end of the timeline leaves the clock running

`stopPlaybackClock()` is called only under `completedPlayback`, and that predicate requires `currentTime >= current.duration` (`playback-state.ts:53`). Any `state` message reporting `isPlaying: false` at a position short of the duration therefore writes `_paused = true` through `setPlaybackState` (`hyperframes-player.ts:745`) while leaving `_parentTickRaf` scheduled.

The docstring's enumeration says every transition out of playback goes through `pause()`, `seek()`, a new document, disconnect, or the end of the timeline. I checked each: `pause()` (`:327`), `seek()` (`:351`) and `disconnectedCallback` (`:194`) all stop the clock before setting `_paused`, and end-of-timeline is covered by the new `stopPlaybackClock` relay. The path not in that list is the runtime pausing itself mid-timeline, which is reachable through the runtime player seam: `packages/core/src/runtime/player.ts` calls `deps.setIsPlaying(false)` from its own `pause()` at `:198` and from the seek helpers at `:230` and `:253`, and that surface is exposed to composition code as `window.__player`.

The old `if (this._paused)` read in the rAF was the accidental backstop for exactly this, so removing it is what exposes the gap. Severity depends on a divergence worth confirming: `onTick` guards on `clock.isPlaying()` (`init.ts:3366`) while `setIsPlaying` only writes `state.isPlaying` without pausing the transport. If the transport really did stop, this is a leaked rAF posting ticks that early-return, so it is wasted work rather than a visible fault. If the two disagree, the composition keeps advancing while the player reports `paused === true`, which is the mirror image of the bug being fixed here.

I do not think this needs to grow the fix. The cheapest close is to relay the stop on any observed transition to not-playing rather than only on `completedPlayback`, since the clock is idempotent to stop and `play()` restarts it. Whichever way it goes, the docstring's list should name the runtime seam so the next reader does not have to re-derive it, because that list is now load-bearing rather than descriptive.

## What is clean

The three new player tests pin the right things: the in-flight pre-play `state` message, the late `load` event, and the end-of-timeline stop asserted through `_parentTickRaf` being null rather than through a proxy. The `init.ts` change to post state on a parent-driven tick is necessary and its placement after the `reachedEnd` early return is correct, so the end-of-playback message still wins. Verified that `postState`'s existing change and interval filter (`init.ts:2172-2178`) does bound the added message rate as the comment claims.

One note on the docstring at `:707`: it says a stale message would end the loop "while the player still reports `paused === false`". As written the message does write `_paused = true` through the same handler, so the player reports paused too. It self-corrects on the next tick's state message, so nothing turns on it, but the parenthetical describes a state that does not occur.

_Review by Rames_

miguel-heygen:
Both findings verified at source and fixed in `96adc660a`. The second one turned out to be a leaked rAF rather than a divergence, with evidence below.

## Blocker: `_reloadShaderOptions` — confirmed, fixed

Reproduced the reasoning at source: `_reloadShaderOptions` assigns `srcdoc`/`src` and never touches `_ready`, and both shader attributes are in `observedAttributes`, so it is reachable on a ready player. The early return then skipped the teardown for a genuinely new document — no probe, stale bridge flag, no path to ready.

Fixed the invariant rather than the symptom. All six reassignment sites (two in `connectedCallback`, two in `attributeChangedCallback`, two in `_reloadShaderOptions`) now route through `_setIframeSrc` / `_setIframeSrcdoc`, which own clearing readiness. The load handler's test now holds by construction instead of by convention, and the two duplicated `_ready = false; _runtimeBridgeReady = false;` pairs in `attributeChangedCallback` are gone.

## Second finding: the premise does not hold, but the leak is real

You asked for evidence on whether `state.isPlaying` and `clock.isPlaying()` can diverge. They cannot, in the production runtime:

- `createRuntimePlayer` returns its **transport-backed** object whenever `deps.transport` is passed (`player.ts:139-152`), and `init.ts:2552` always passes one. Every `deps.setIsPlaying` call is at line ≥170, inside the `return {...}` at line 155 — the **no-transport fallback**, which `initSandboxRuntimeModular` never reaches. So `window.__player.pause()` goes to `transport.pause()`, not to that seam.
- Searched the whole space rather than sampling: all five `state.isPlaying = false` writes in `init.ts` (2477, 2499, 2524, 3091, 3374) are in the transport, each in the same block as a `clock.pause()`.

So the consequence was a **leaked rAF posting no-op ticks forever**, not a composition advancing while the player reports paused. Still a regression I introduced, still fixed.

Your framing of the cause was right and is the part I acted on: I enumerated one instance of a class. The fix covers the class. `applyRuntimeStateMessage` now forwards the raw wire `isPlaying` to the player before interpreting it, and the player stops the clock on any runtime-reported stop — end-of-timeline included, so the narrower `stopPlaybackClock` callback is deleted.

Staleness immunity is preserved with a discriminator rather than by omission: a report seen before the runtime has echoed our own play is ignored, because postMessage delivery between two windows is ordered, so such a report was necessarily posted before the play command arrived.

## Verification

Six regressions now, each proven by reverting only its own fix:

| reverted | fails with |
|---|---|
| the `_paused` read removal | `expected +0 to be 2` |
| the `load` early return | `expected false to be true` |
| `_reloadShaderOptions` routing | `expected true to be false` |
| the runtime-report stop | `expected 1 to be null` (×2 tests) |
| the staleness guard | `expected +0 to be 2` |
| `postState` in the runtime tick | `frame 60 → 0` |

The last row of that table is the one worth noting: removing the staleness guard reopens the original bug, so the guard is load-bearing in both directions.

One test was corrected rather than kept: the end-of-timeline case previously jumped from `play()` straight to the end report with no intervening playing report, which the runtime cannot produce — 15s of playback always echoes. Assertion unchanged, construction now matches production. I also wrote a loop-wrap test, could not make it fail under any realistic mutation, and deleted it rather than ship a vacuous one.

All four configurations re-measured after the change, unchanged:

| config | t+2000 | t+5000 | t+8000 | t+14500 | end |
|---|---|---|---|---|---|
| same-origin | 2.000 | 4.967 | 8.000 | 14.500 | 15.000, `ended` 14996ms |
| CSP `sandbox allow-scripts` | 2.000 | 5.000 | 8.000 | 14.500 | 15.000, `ended` 15001ms |
| cross-origin host | 2.000 | 4.967 | 7.967 | 14.500 | 15.000, `ended` 14991ms |
| cross-origin + offscreen (rAF suspended) | 2.000 | 5.000 | 8.000 | 14.500 | 15.000, `ended` 14988ms |

`readyEvents: 1` in all four; clock started exactly once and stopped exactly once per run.

`@hyperframes/player` 350 passed, `@hyperframes/core` 2580 passed, repo-wide typecheck / lint / format clean.

## One thing I did not change

The no-transport fallback in `createRuntimePlayer` is dead from the runtime's entry point — every `deps.setIsPlaying`, `deps.onSyncMedia`, `deps.onDeterministicPause` path is unreachable via `initSandboxRuntimeModular`, exercised only by `player.test.ts` and one script. That is a deletion worth making, but it is a separate change and not this PR's job.

miguel-heygen:
Third point from the review addressed in `b01bb39ae`, and the two fixes in `96adc660a` re-verified independently by mutation rather than by re-reading them.

## The docstring parenthetical

You were right that it describes a state that does not occur, and the reason is exactly the one you gave: the stale message runs through `applyRuntimeStateMessage` like any other, which writes `paused = !data.isPlaying`, so at that instant the player does report paused.

Rather than delete the clause I made it say the thing that is true one report later. Nothing actually paused the composition, so the runtime's next `state` message carries `isPlaying: true` and puts `paused` back to `false` — against a timeline the now-dead loop is no longer advancing. That settled state is the shape the bug was reported as, so the docstring keeps the connection to the symptom instead of dropping it.

## Re-verified by breaking each fix on purpose

Each fix reverted alone, on the current head, witness test named and failure quoted verbatim.

**`_reloadShaderOptions` routed back to a direct `iframe.srcdoc` / `iframe.src` assignment:**

```
FAIL  keeps ticking after a shader-option reload starts a new document
AssertionError: expected true to be false // Object.is equality
- Expected: false
+ Received: true
```

**The runtime-report stop removed (`onRuntimePlaybackReport` no longer stopping the clock on a not-playing report):**

```
FAIL  stops ticking when the composition pauses itself short of the end
FAIL  stops ticking once the composition reports the end of the timeline
AssertionError: expected 1 to be null
```

Two tests, which is the point of the second finding: end-of-timeline was one instance of the class, and the mid-timeline runtime pause is another.

**The staleness discriminator removed (stop on any not-playing report, no `_awaitingPlayEcho` check):**

```
FAIL  keeps ticking when a state message posted before play() lands after it
AssertionError: expected +0 to be 2 // Object.is equality
```

That last one is the one worth keeping in view: the guard is load-bearing in both directions. Without it the fix for the second finding reopens the original bug.

## Suites

```
@hyperframes/player   12 files, 350 passed
@hyperframes/core    125 files, 2580 passed
```

Format, lint and typecheck clean on the touched package. CI on the branch is green apart from jobs still in flight at the time of writing.

Re-requesting review.

miguel-heygen:
Closing the stale branch as part of the authored-PR reset. Current main still has the parent tick lifetime and state-publication gaps. Only those fixes merit a fresh PR on current main, preserving the newer runtime-data and navigation lifecycle changes.

PR #3559, opened 2026-08-31
fix(studio): let a hidden sub-composition child be shown again

felipecaldas:
## Summary

Hiding anything that lives inside a sub-composition is a one-way trip in Studio. The eye on an expanded child row always renders as **Hide**, whatever the source file says, so the first click writes `data-hidden` and every click after it rewrites the same attribute and changes nothing. The element can never be shown again from the UI, not even after a reload, because the attribute is in the file.

Top-level clips are unaffected. Their eye flips to **Show** and toggles correctly. That contrast is the whole bug: same attribute, same file format, different row type.

---

## The bug, from a user's point of view

A user pressed the eye on a timeline track, decided they wanted it back, and pressed it again. One scene never came back. Its clip in the timeline went from a thumbnail to an empty box, and the scene would have rendered as several seconds of nothing.

Recovering it needed a text editor. In the UI there was no path back:

- The child row's eye still says "Hide", so it offers the action that is already done.
- Clicking it is a silent no-op. No error, no toast, no file change.
- Selecting the row does nothing either, because a hidden element is not selectable on the canvas, so the property panel never opens for it.

Affected rows lose four more pieces of state through the same gap: `timelineLocked`, `timelineRole`, `fxChain` and `automation`.

---

## How it got caught

I clicked on the "hide" eye in the UI for a particular track, it hid my track, then I clicked again and it didn't undo the action.

Working backwards from the file was what pinned it. Studio's own `.hyperframes/backup` chain holds a pre-write snapshot per save, so the sequence was recoverable exactly:

| Time | File | `data-hidden` before → after |
|---|---|---|
| 01:55:44 | `compositions/scene-2.html` | 0 → 1 (the title) |
| 01:56:01 | `compositions/scene-2.html` | 1 → 2 (the video) |
| 09:48:13–09:48:18 | `index.html` | 0 → 9 → 0 → 9 → 0 |

Two observations did the work. The sub-composition file only ever gained attributes, never lost one, across every save in its history. And `index.html` in the same session toggled 0 ↔ 9 perfectly, four times in a row, which ruled out the write path and the track-level toggle and left only the read path for one class of row.

The existing regression test never fired, and that turned out to be part of the story. See **Why the existing test passes** below.

---

## Root cause

`buildChildElements` synthesizes an expanded child row from a manifest clip. It calls `createTimelineElementFromManifestClip` with no `doc` and no `hostEl`, so it cannot read `data-hidden`, `data-timeline-locked`, `data-timeline-role`, `data-fx-chain` or `data-automation` off the element. It compensated by inheriting them from the child's flat store twin:

```ts
// useExpandedTimelineElements.ts:226
...hostElementState(elements.find((element) => element.key === key)),
```

For a real sub-composition that twin cannot exist. `processTimelineMessage` filters the manifest to root-level clips before building the flat store, and drops any clip whose parent composition is itself in the manifest:

```ts
// useTimelineSyncCallbacks.ts:122-138
// Show root-level clips: no parentCompositionId, OR parent is a "phantom wrapper"
const clipCompositionIds = new Set(data.clips.map((c) => c.compositionId).filter(Boolean));
const filtered = data.clips.filter(
  (clip) => !clip.parentCompositionId || !clipCompositionIds.has(clip.parentCompositionId),
);
...
const els = buildTimelineElementsFromClips(filtered, iframeDoc);
```

So `elements.find(...)` always returns `undefined`, `hostElementState` returns `{}`, and the child row carries `hidden: undefined`. `TimelineLanes` then computes:

```ts
// TimelineLanes.tsx:208
const isTrackHidden = els.length > 0 && els.every((element) => element.hidden === true);
```

which is `false`, so the button renders as "Hide" and dispatches `hidden: true`. The patch writes `data-hidden` onto an element that already has it. Same bytes, no change, no error.

The inheritance is therefore dead code for the case its own docstring describes. It works only when the parent is a phantom wrapper, where the child does keep a store entry of its own.

---

## Reproducing it

Four files, no assets, about a minute. Verified on `f84b4c23d` (this PR's base).

```
repro/
  hyperframes.json
  index.html
  compositions/scene-a.html
  compositions/scene-b.html
```

`index.html` — two scene slots:

```html
<div id="main" data-composition-id="main" data-start="0.000" data-duration="6.000"
     data-width="720" data-height="1280" data-fps="24">
  <div id="scene-a-slot" class="clip" data-composition-id="scene-a"
       data-composition-src="compositions/scene-a.html"
       data-start="0.000" data-duration="3.000" data-track-index="0"
       data-covers-frame="true" data-width="720" data-height="1280"></div>
  <div id="scene-b-slot" class="clip" data-composition-id="scene-b"
       data-composition-src="compositions/scene-b.html"
       data-start="3.000" data-duration="3.000" data-track-index="0"
       data-covers-frame="true" data-width="720" data-height="1280"></div>
</div>
```

`compositions/scene-a.html` — note that the clip sits one level **below** an id'd wrapper, which is what a generated scene looks like:

```html
<div id="scene-a" data-composition-id="scene-a" data-start="0.000" data-duration="3.000"
     data-width="720" data-height="1280" data-fps="24">
  <div id="scene-a-region" class="fill" style="background:#123">
    <div id="scene-a-body" class="clip fill" data-start="0.000" data-duration="3.000"
         data-track-index="0" style="background:#3a7"></div>
  </div>
  <div id="scene-a-title" class="clip" data-start="0.000" data-duration="3.000"
       data-track-index="1" style="position:absolute;top:40%;width:100%;text-align:center;color:#fff;font:700 64px sans-serif">Scene A</div>
</div>
```

(`scene-b.html` is the same with `b` substituted. It only exists so `scene-a` is a sub-composition among siblings rather than the whole timeline.)

Then:

1. `hyperframes preview` in `repro/`.
2. Put the playhead inside Scene A, around `t=1`, so its children expand. Rows appear for **Scene A Body** and **Scene A Title** under the scene track.
3. Click the eye on **Scene A Body**.

**On `main`:**

| Step | `data-hidden` in `scene-a.html` | Button label |
|---|---|---|
| start | 0 | Hide track 2 |
| after click 1 | 1 | **Hide track 2** |
| after click 2 | **1** | **Hide track 2** |

The row is now unrecoverable from the UI.

**With this PR:**

| Step | `data-hidden` in `scene-a.html` | Button label |
|---|---|---|
| start | 0 | Hide track 2 |
| after click 1 | 1 | Show track 2 |
| after click 2 | 0 | Hide track 2 |

---

## The fix

Read the state off the live element instead of hoping a store twin exists.

`collectSubCompositionHostState` walks each sub-composition host in the preview document and records the `data-*` state of every id'd descendant, keyed by dom id. It runs in `processTimelineMessage` beside the existing DOM walk, lands on the player store, and `hostElementState` now takes that reading. The flat twin stays as the fallback, so the phantom-wrapper case is unchanged.

### Why it is a second walk and not an extension of the existing one

`collectSubCompositionDomChildren` looks like the natural home for this, and it is the wrong one for two independent reasons.

**It cannot reach the element.** It stops at the first id'd descendant and only recurses into groups:

```ts
// timelineSyncHydration.ts:97-111
if (!child.id) {
  collectHostDomChildren(hostId, child, parentId, parentMap, out); // id-less wrapper
  continue;
}
...
if (isGroup) collectHostDomChildren(hostId, child, child.id, parentMap, out);
```

Scene footage commonly sits below an id'd region wrapper, so it is never visited:

```
(no-id) <div data-hf-inner-root>
  scene-a-region <div>                    ← walk stops here (id'd, not a group)
    scene-a-body <div data-hidden>        ← never collected
  scene-a-title <div data-hidden>         ← collected
```

That asymmetry is why the two child rows in the original report behaved identically despite only one being reachable by that walk: neither had a store twin, so both were broken for the same reason.

**Making it reach the element would break rows.** That walk also writes `parentMap`. Recursing into `scene-a-region` would set `parentMap.set('scene-a-body', 'scene-a-region')`, overwriting the entry the runtime clip tree already published (`scene-a-body → scene-a-slot`). `buildExpandedElements` filters siblings on exactly that map:

```ts
const fromManifest = manifest.filter((c) => c.id != null && parentMap.get(c.id) === siblingParentId);
```

so the body row would stop matching its host and disappear from the timeline entirely.

The new walk descends the whole subtree with `querySelectorAll("[id]")` and touches neither rows nor parentage. It returns a plain map, and elements carrying no state are omitted so a visible child still reads as visible.

---

## Why the existing test passes

`useExpandedTimelineElements.test.ts` has a test named "inherits hidden and locked state from the flat store element", added as a regression guard for this exact symptom. It is green on `main` and green with the bug present.

Its fixture does two things production cannot:

```ts
const elements = [
  el({ id: "s1", domId: "s1", start: 0, duration: 14 }),
  el({ id: "eyebrow", key: "index.html#eyebrow", ..., hidden: true, timelineLocked: true }),
];
```

It hands the child a flat store twin carrying `hidden: true`, and it gives the host no `compositionSrc`. Without a `compositionSrc` the child key falls back to the `index.html` scope, which is the one scope where a twin can exist. So the test asserts that a spread copies a field, never that the field can be obtained.

The added test models a real sub-composition: host with `compositionSrc`, no twin in the store, and an explicit assertion that no store element can match the child's key.

```ts
expect(child.key).toBe("scene-2.html#scene-2-video");
expect(elements.some((element) => element.key === child.key)).toBe(false);
expect(child.hidden).toBe(true);
```

On `main` it fails with `expected undefined to be true`.

---

## Tests

**`timelineSyncHydration.test.ts`** (new) covers the collector. It includes a positive control that asserts the existing sibling walk returns `["scene-2-video-region", "scene-2-title", "scene-2-caption"]` and not the nested video, so the "cannot reach it" claim above is measured in the suite rather than asserted in a comment. Also covers the full `data-*` set, omission of stateless elements, a null document, and non-composition clips.

**`useExpandedTimelineElements.test.ts`** gains the real-sub-composition case described above.

Gates, all clean:

- `packages/studio`: 415 files, 4621 tests pass on this branch.
- `bun run lint`, `bun run format:check`, `tsc --noEmit`.

Two things the repo's own gates caught in this change and that are worth noting rather than hiding: `hostElementState` first landed as a `??` chain at cyclomatic 18 and was flagged, so it is now a destructure plus one spread; and `playerStore.ts` crossed the 600-line cap, so `SubCompositionHostState` lives in `timelineElement.ts` next to `TimelineElement` and is re-exported, matching how `TimelineElement` itself is surfaced.

---

## Scope

Studio only. No change to `core`, `engine`, `parsers`, `player`, `producer` or `shader-transitions`, so rendering is untouched.

Found while running a generated storyboard project: one root `index.html` of scene slots plus one `compositions/scene-N.html` per scene. In that shape every scene's footage and title is a sub-composition child, so the broken case is the normal case rather than an edge.

### DIFF
--- a/packages/studio/src/player/hooks/timelineSyncHydration.test.ts
+++ b/packages/studio/src/player/hooks/timelineSyncHydration.test.ts
@@ -1,0 +1,96 @@
+// @vitest-environment happy-dom
+
+import { describe, expect, it } from "vitest";
+import {
+  collectSubCompositionDomChildren,
+  collectSubCompositionHostState,
+} from "./timelineSyncHydration";
+import type { ClipManifestClip } from "../lib/playbackTypes";
+
+const clip = (over: Partial<ClipManifestClip>): ClipManifestClip => ({
+  id: "x",
+  label: "x",
+  start: 0,
+  duration: 1,
+  track: 0,
+  kind: "element",
+  tagName: "div",
+  compositionId: null,
+  parentCompositionId: null,
+  compositionSrc: null,
+  assetUrl: null,
+  ...over,
+});
+
+/**
+ * The shape a storyboard scene actually publishes: the clip that carries
+ * `data-hidden` is a `<video>` one level BELOW an id'd region wrapper.
+ */
+function mountScene(): Document {
+  document.body.innerHTML = `
+    <div id="scene-2-slot" data-composition-id="scene-2" data-composition-src="scene-2.html">
+      <div data-hf-inner-root>
+        <div id="scene-2-video-region" class="hf-region">
+          <video id="scene-2-video" class="clip" data-start="0" data-hidden></video>
+        </div>
+        <div id="scene-2-title" class="clip" data-start="0" data-hidden></div>
+        <div id="scene-2-caption" class="clip" data-start="0" data-timeline-locked
+             data-timeline-role="caption" data-fx-chain="blur" data-automation="opacity"></div>
+      </div>
+    </div>`;
+  return document;
+}
+
+// A composition clip is keyed by its ELEMENT id, not its `data-composition-id`
+// (the runtime's clip tree publishes `scene-2-slot`), so the collector resolves
+// the host with getElementById(clip.id) exactly as the sibling walk does.
+const sceneClips = [clip({ id: "scene-2-slot", kind: "composition", compositionId: "scene-2" })];
+
+describe("collectSubCompositionHostState", () => {
+  it("reaches a clip nested below an id'd wrapper, which the sibling walk cannot", () => {
+    const doc = mountScene();
+
+    // The walk that defines rows stops at the first id'd descendant, so the
+    // video inside the region wrapper is never recorded there. That is why the
+    // eye on a hidden scene video had no state to read.
+    const siblings = collectSubCompositionDomChildren(doc, sceneClips, new Map());
+    expect(siblings.map((child) => child.id)).toEqual([
+      "scene-2-video-region",
+      "scene-2-title",
+      "scene-2-caption",
+    ]);
+
+    const state = collectSubCompositionHostState(doc, sceneClips);
+    expect(state.get("scene-2-video")?.hidden).toBe(true);
+  });
+
+  it("records every data-* attribute an expanded child row needs", () => {
+    const state = collectSubCompositionHostState(mountScene(), sceneClips);
+
+    expect(state.get("scene-2-title")).toEqual({ hidden: true });
+    expect(state.get("scene-2-caption")).toEqual({
+      timelineLocked: true,
+      timelineRole: "caption",
+      fxChain: "blur",
+      automation: "opacity",
+    });
+  });
+
+  it("omits elements carrying no state, so a visible child reads as visible", () => {
+    const state = collectSubCompositionHostState(mountScene(), sceneClips);
+
+    expect(state.has("scene-2-video-region")).toBe(false);
+    expect(state.get("scene-2-video")?.hidden).toBe(true);
+  });
+
+  it("returns empty without a document, rather than throwing", () => {
+    expect(collectSubCompositionHostState(null, sceneClips).size).toBe(0);
+  });
+
+  it("ignores clips that are not compositions", () => {
+    const doc = mountScene();
+    const state = collectSubCompositionHostState(doc, [clip({ id: "scene-2-slot" })]);
+
+    expect(state.size).toBe(0);
+  });
+});
--- a/packages/studio/src/player/hooks/timelineSyncHydration.ts
+++ b/packages/studio/src/player/hooks/timelineSyncHydration.ts
@@ -10,7 +10,7 @@
  */
 
 import { usePlayerStore } from "../store/playerStore";
-import type { TimelineElement, DomClipChild } from "../store/playerStore";
+import type { TimelineElement, DomClipChild, SubCompositionHostState } from "../store/playerStore";
 import { resolveCssStackingContextId } from "@hyperframes/core/runtime/stacking-context";
 import type { ClipTree } from "@hyperframes/core/runtime/clipTree";
 import { HF_AUDIO_GROUP_ATTR } from "@hyperframes/core/audio-groups";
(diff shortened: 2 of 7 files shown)


miguel-heygen (approved):
Exact-head review at `1ddf01a2807c27b8f78feabfc1dfe28984130cab`.

**Strengths**
- `timelineSyncHydration.ts:168-183` reads the five host-only attributes from the live sub-composition subtree without mutating the row-defining parent map; stateless descendants are deliberately omitted.
- `useExpandedTimelineElements.ts:185-192,231-241` makes that live reading authoritative while preserving the flat-store fallback for phantom-wrapper children and the existing audio-group derivation.
- `useTimelineSyncCallbacks.ts:128-137` captures the state from the same document/manifest synchronization pass that builds the expanded rows, and the store reset path initializes it cleanly.

The regression coverage is well targeted: the collector test proves the nested-below-an-id'd-wrapper case and all five attributes, while the expanded-row test proves a real `compositionSrc` child with no flat twin. Local verification: both changed suites pass (29/29), and oxlint plus oxfmt are clean on all seven changed files. Full local typecheck was environment-blocked by missing generated Core runtime files and `@mcp-b/global` in the partial review install, not by a diagnostic in this diff; GitHub currently exposes only the WIP check.

**Verdict:** APPROVE
**Reasoning:** The fix reads state at the only boundary that owns it, preserves the existing fallback behavior, and adds non-vacuous tests for both collection and consumption.

-- Magi

PR #3562, opened 2026-08-31
feat(registry): say what a promoted template's slots mean, not where they sit

WaterrrForever:
## Why

A promoted template's variable descriptions told the remix agent **where a slot sits on screen**, not **what it stands for**. That is the wrong information for the one decision a remix has to get right: which slot carries the advertiser's identity.

Before, on `notification-cascade`:

```
appName     "Small credit in the corner of every banner."
footerText  "URL or call to action under the headline."
brandLogo   "Transparent logo centered on the closing card."
```

Nothing there says *whose* name, *whose* domain, *whose* logo. Four remixes reported by a customer-facing team all shipped somebody else's identity in those slots — one of them shipped an invented brand, "Frame AI" at "the-ai-workspace.com", read out of Notion's own tagline. Another shipped an ad naming a company that does not appear anywhere on the page it was given.

After:

```
appName     "Name of the brand this video promotes, credited in the corner of
             every banner. Use the real name of that brand."
footerText  "Domain of the brand this video promotes, under the end-card
             headline. Use the real domain, or a call to action when no
             domain fits."
brandLogo   "Transparent mark of the brand this video promotes, centered on
             the closing card."
```

## What changed

- **Slot descriptions** across eight promoted blocks now state the role a slot plays in the ad, and where a slot is a *claim* rather than copy — a ranking's source chip, for instance — the description says so, so the agent has grounds to keep it neutral.
- **`portrays` metadata**, a machine-readable marker on each identity-bearing slot (`subject_name`, `subject_domain`, `subject_logo`, `subject_tagline`, `competitor_name`, `authority_badge`, `recommended_position`, `host_identity`). Documented in `docs/concepts/variables.mdx`. This is what lets a checker or an eval ask "did the advertiser reach the screen?" without pattern-matching prose.
- **Docs regenerated** for the affected catalog pages.

**Scope, precisely.** Five of the eight blocks are declaration-only: `ai-chat-reveal`, `message-thread-reveal`, `slack-notification-ad`, `share-sheet-carousel` and `notification-cascade` change zero lines outside a variable declaration. Three change more, and each has its own section below — `notes-reveal` (+334/-125) gains a sign-off strip and two slots, and `chatgpt-exchange` (+78/-9) and `claude-exchange` (+54/-6) gain a caret-following composer scroll. No slot is removed or renamed anywhere.

An earlier revision of this description claimed the bytes outside `data-composition-variables` were unchanged everywhere but `notes-reveal`. That was wrong, and it would have sent a reader straight past the exchange work — thanks to Rames for catching it.

## notes-reveal now has somewhere to say whose ad it is

The first draft of this PR recorded `notes-reveal`'s missing identity slots as out of scope. Second commit fixes it instead, because "this template cannot be attributed" is not a state a promoted ad template should stay in.

The closing card ends at y=1708, leaving 212px of dotted scene. A sign-off strip lives there — `brandLogo` (`subject_logo`) above `brandDomain` (`subject_domain`), in the card's own plum and Courier Prime. Putting it **outside** the card means the hand-lettered headline and the checklist keep the geometry they were designed with.

Verified rather than assumed:

- `hyperframes check`: the same **0 errors** and the same warning counts as `origin/main`, with one more contrast check passing (**367/367 WCAG AA**).
- Rendered and inspected at t=23.5s with the packaged default (HyperFrames wordmark over `hyperframes.heygen.com`) and with a resolved symbol mark (Notion's cube over `notion.com`). Both sit correctly.
- Graded by a real agent run against a locally re-promoted package: `subject_domain` scores **10.0 — every subject_domain slot carries gamma.com**, a dimension that was permanently *not applicable* for this template before.

**No name slot, deliberately.** There is no text slot here that can hold a bare brand name: `cardTop`'s marker line continues into the fixed words "OF ONE FILE", and a name set beside a wordmark reads twice — rendering it is how I found that, before committing. The mark plus the domain carry the attribution, which is what the other end-card templates do.

Finishing this also surfaced two agent-side defects that only appear once a logo slot exists — a generated brand mark, and a domain recalled from training instead of taken from the submitted URL. Both are fixed in heygen-com/experiment-framework#48887.

## The exchange composers scroll to follow the caret

`prompt` is an editable slot in both exchange templates, but `.comp-text` is a fixed-width single line with `white-space: pre` inside a composer that clips — and the packaged prompt already reaches the mic, so it has no headroom at all. A longer prompt slid under the mic and the send button and lost its tail, caret and all: the remix read as having typed into nowhere, cut mid-word. Both now scroll the typed text left to keep the caret visible, the way a real single-line input does, on the very per-character times the reveal already uses.

Shrink-to-fit was the alternative and it is worse: with no headroom it shrinks the packaged prompt too, and it still cannot absorb a prompt twice as long without unreadable type. It was tried, measured, and dropped.

Verified rather than assumed:

- **A prompt that fits emits no keyframes at all** and the element is never handed to GSAP — even a zero translate stamps a transform, which changes how the text rasterizes. So the packaged compositions are untouched: `claude-exchange` SSIM **1.000000** across 1284 frames; `chatgpt-exchange` **0.999998** across 894 frames with 13 frames below 0.9999, against a **249-frame** noise floor measured by rendering that template twice unchanged. Durations unchanged at 14.9s and 21.4s.
- **An overflowing prompt** holds at 49 scroll tweens across the fonts-ready rebuild with no duplicated timestamps, and the master timeline stays attached to the global timeline.

That second line is a fix Magi's review turned up: `tl.to()` returns the *timeline*, not the tween it created, so the rebuild was calling `.kill()` on the master timeline and stacking its keyframes on top of the ones it meant to replace — 49 before, 98 after, and the timeline off the ticker. Only an overflowing prompt reaches that path, i.e. exactly the remix the scroll exists for, which is why a packaged-defaults render never saw it. The steps now live in one nested child timeline that can be killed and replaced without reaching `tl`.

## Sequencing

These packages have to be **re-promoted** and the FE pin moved before any of this reaches users; the currently pinned commit (`3cacac5b`) carries neither the descriptions nor `portrays` (verified against the deployed zips). This PR is the source change only.


jrusso1020 (dismissed):
Reviewed at head `e299a982d48fe79869fc76483ab67920f3e92438`. All 8 required contexts on `main` ran and passed, matched by exact name.

**Strengths**

- `portrays` is the right shape for this problem. Prose descriptions can be improved indefinitely and still not be checkable; a machine-readable marker on each identity-bearing slot lets an eval ask "did the advertiser reach the screen?" without pattern-matching English. `docs/concepts/variables.mdx` documenting the eight values is what makes it usable by anything other than this PR.
- The source-chip descriptions (`row1Chip`/`row2Chip`/… in `chatgpt-exchange`) telling the agent a chip *asserts where the ranking came from*, so it should stay neutral when the comparison is written for the remix — that is a distinction between copy and claim that most slot metadata never makes.
- Fixing `notes-reveal` rather than recording it as out of scope, and putting the sign-off strip in the 212px below the card so the hand-lettered headline keeps its designed geometry.

**Findings**

*important — the byte-scope claim in the description is not accurate, and the exception is not `notes-reveal`.* The description says: "Compositions are otherwise untouched: outside the `data-composition-variables` block the bytes are unchanged, and no slot was added, removed or renamed", with `notes-reveal` called out separately as the one composition change. Against the diff, two more templates change composition:

- `registry/blocks/chatgpt-exchange/chatgpt-exchange.html` — 52 added lines outside the variables block
- `registry/blocks/claude-exchange/claude-exchange.html` — 42 added lines outside the variables block

Both add a new caret-following scroll in the composer (`layoutCaretScroll`, `#cge-comp-text` / `#cge-blue-btn`, `CARET_PAD`). The other five HTML files are clean — their changes are entirely slot metadata, which I checked line by line.

To be clear about severity: the code itself is careful, and I would not ask for it to be removed. It collects steps before touching `typedEl` so a prompt that fits emits no keyframes and the packaged render is bit-for-bit unaffected; it converts rects with the element's own ratio instead of reading `--scale`, so it survives a change to that; and the comment explains why the blue button rather than the mic is the real boundary. The problem is that a reader — human or bot — who takes the scope claim at face value will not review those 94 lines at all, and the verification list (`hyperframes check` 0 errors / 367 contrast, a `notes-reveal` render at t=23.5s, an agent grading run) contains nothing that exercises a long prompt in either exchange template. Correcting the claim and saying how the caret path was checked is the whole ask.

*question — timeline mutation after first build.* `document.fonts.ready.then(layoutCaretScroll)` re-runs the layout and re-adds tweens to `tl` after the composition has already built its timeline once. For a deterministic frame-by-frame render that is only safe if capture cannot begin before `fonts.ready` resolves. If the renderer does not guarantee that ordering, a capture could straddle the re-layout and land keyframes mid-sequence. You have rendered this, so you may already know the answer — worth stating it in the description either way, since the same pattern will get copied into the next template.

**Verdict:** APPROVE
**Reasoning:** The semantic-role change is well-designed and independently verified, and the undisclosed caret-scroll is good code. Approving on the merits; please fix the scope sentence before merge so the next reader is not told those 94 lines do not exist.

— Rames Jusso

miguel-heygen (changes requested):
Reviewed exact head `e299a982d48fe79869fc76483ab67920f3e92438`.

The `portrays` contract is a strong addition: it makes identity semantics machine-readable without changing rendering, and the eight documented values give downstream agents a stable vocabulary. I also agree with Rames that the PR body must disclose the substantial caret-scroll and `notes-reveal` behavior changes rather than describe this as metadata-only.

**Blocking — `registry/blocks/chatgpt-exchange/chatgpt-exchange.html:1419-1427`: the font-ready relayout kills the registered master timeline for long prompts.**

`Timeline.to()` returns the timeline (`tl`), not the child tween. The code pushes that return value into `caretScrollTweens`, so every array entry is the same master `tl`. `layoutCaretScroll()` runs synchronously and then unconditionally runs again from `document.fonts.ready`; when the prompt is long enough to create scroll steps, the second pass executes `tw.kill()` and kills `window.__timelines["chatgpt-exchange"]` itself before adding replacement steps. The default prompt does not overflow, which is why the packaged render and current green checks miss this exact remix path.

Please make the caret animation a nested timeline that can be cleared/rebuilt, or construct/store the actual child tweens and remove only those. Add a long-prompt regression that waits for `document.fonts.ready` and then seeks the registered timeline through the typing/collapse interval; asserting only the default composition will not distinguish this failure.

`claude-exchange` does not share this bug: its font-ready path clears and rebuilds the master timeline inside `build()` rather than storing the return value of `tl.to()` as a child.

Exact-head CI is green, but it does not exercise an overflowing prompt after fonts settle.

**Verdict:** REQUEST_CHANGES

**Reasoning:** The semantic metadata is good, but one of the added long-prompt protections invalidates the master GSAP timeline on the very input class it is intended to support.

— Magi

jrusso1020 (dismissed):
**Retracting my earlier approval on this PR.** I approved at `e299a982` (review `5078820454`) and that was wrong — I read the caret-scroll block as self-contained and did not check what `tl.to()` returns. I have dismissed that approval rather than leaving it standing, since a COMMENT does not remove it and it was still counting toward the gate.

@miguel-heygen's premise is correct, and I verified it independently rather than relaying it. One correction to the stated consequence, and one defect underneath it that I don't think has been named yet.

### blocker — the caret-scroll teardown never removes anything, and detaches the composition from its parent

`registry/blocks/chatgpt-exchange/chatgpt-exchange.html:1419` and `:1422` push the return value of `tl.to(...)` into `caretScrollTweens`. In GSAP 3.14.2 — the build this file loads at `:33` — `Timeline.to()` returns **the timeline**, for chaining; only `gsap.to()` returns a Tween. Measured against that exact dist file: `tl.to(target, {...}, at) === tl` → `true`. So the array holds N references to the master timeline created at `:1167`, and `:1391`'s `caretScrollTweens.forEach((tw) => tw.kill())` is `tl.kill()`.

`Animation.prototype.kill()` is `_interrupt(this)`: detach from parent, fire `onInterrupt`. It does not touch the timeline's own children. Two things follow.

**1. The `document.fonts.ready` re-measure at `:1427` never replaces the first pass — it layers on top of it.** Replicating the two-pass bookkeeping verbatim: children go `1 → 4 → 7`, zero removed. The pre-font tweens stay on the timeline and the post-font set is added alongside them. At a time only the first pass covered, the stale pre-font offset is what renders (measured `x = -40` at a step the second pass does not emit). So the re-measure that exists to correct the pre-font metrics does not correct them.

**2. The composition is orphaned.** Direct seek of the killed timeline still maps correctly (measured 25 at `seek(0.25)`, 90 at `seek(0.90)`, `ease: "none"`), so a standalone render of this block is unaffected — which is part of why CI is green here and why this is easy to miss. But the timeline is no longer in its parent's child list, so:

- driven by a parent's seek, it freezes at its initial state — measured `0` where an unkilled control renders `50`, with `master.getChildren().includes(child)` → `false`;
- `play()` on it is dead — measured still `0` after 400ms where an unkilled control completes at `100`. That is the path `packages/core/src/runtime/init.ts` `activateSiblingTimelines` takes on every registered non-master timeline, and the comment there already warns that a child which never receives a working `play()` sits at its initial CSS state.

**Trigger:** only when the first pass emits steps — an overflowing prompt, i.e. exactly the remix path the caret scroll was added for. The packaged prompt emits none, so the packaged composition and CI stay clean. Note `document.fonts.ready` resolves even when the faces are already loaded, and the `.then()` at `:1427` is unguarded, so for any overflowing prompt this is deterministic rather than a font-timing race.

**The sibling in this same PR already has the correct shape.** `registry/blocks/claude-exchange/claude-exchange.html:1234` puts the identical caret-scroll code inside `build()`, whose first statement at `:1235` is `tl.clear()`, and guards the re-run with `document.fonts.status !== "loaded"`. `chatgpt-exchange` has neither. Adopting that shape is the smallest fix.

Two alternatives, both verified against 3.14.2:

- capture the real tween — `tl.to(...)` then `tws.push(tl.recent())`; `recent()` returns a `Tween`, killing those keeps the child count stable across passes and leaves the master parented;
- give the caret its own nested timeline added once (`tl.add(caretTl, 0)`) and `caretTl.clear()` per layout; master child count stable, and the pass-1-only time then renders the pass-2 value (`-26`, not the stale `-40`).

Either way the regression worth pinning is: a prompt long enough to emit steps, re-layout after the faces settle, then assert the child count did not grow and the timeline still has a parent.

### nit — carried over

The scope sentence in the PR body still reads as description-only for the two exchange blocks; `chatgpt-exchange` (+66/-9) and `claude-exchange` (+54/-6) both add composition behavior.

**Audited:** `chatgpt-exchange.html` and `claude-exchange.html` (caret-scroll paths end-to-end, at `e299a982`), GSAP 3.14.2 timeline/kill semantics (measured, not recalled), `runtime/init.ts` sibling activation.
**Trusting:** the `.mdx` catalog copy and the `portrays` metadata, which I read for consistency but did not re-derive.

**Verdict:** REQUEST CHANGES
**Reasoning:** A confirmed behavior defect on the exact path this PR adds, with the correct pattern already present in its sibling file; my prior approval is dismissed because I would not approve this today.

— Rames Jusso

WaterrrForever:
Both fixed at `d3359ee42`. Thanks Magi — the timeline one was a real find.

**The `tl.to()` blocker.** Confirmed and fixed. Measured it before and after by holding `document.fonts.ready` open until I could snapshot both sides, with an overflowing prompt:

| | before rebuild | after rebuild |
|---|---|---|
| attached to global timeline | `true` | **`false`** |
| scroll tweens on `#cge-typed` | 49 | **98** |
| timestamps with two competing tweens | 0 | **49** |

One correction to the mechanism, because it changes how bad this is: GSAP's `Animation.kill()` **interrupts and detaches** rather than killing children, so the master timeline survived and the composition kept rendering under an explicit seek. It is still wrong — the timeline comes off the ticker, and the stale keyframes were the ones measured against the fallback font — but it would not have shown up as a dead render, which is worth knowing for anyone chasing a similar report.

The steps now live in one nested child timeline: a real object that can be killed and replaced, and killing it cannot reach `tl`. Children sit at the same absolute times and the nest is added at 0, so the motion is unchanged.

Verified both directions:
- overflowing prompt: **49 tweens before and after** the rebuild, zero duplicated timestamps, timeline stays attached
- packaged defaults vs the **pre-feature baseline**: SSIM **0.999998** over 894 frames, 13 frames below 0.9999 against this template's own **249-frame** noise floor, duration **14.9s** unchanged

`claude-exchange` does not have the bug — it never stores the return value, and `build()` rebuilds the whole timeline. Left alone.

**The scope sentence.** You were both right and it was worse than described: the claim also covered `notes-reveal`'s added slots. Rewritten to state it per-file — five blocks are declaration-only (verified: zero changed lines outside a variable declaration), three are not, each with its own section. The exchange work now has a section of its own with its verification, so nothing in this PR is reachable only by reading the diff.

miguel-heygen (changes requested):
Re-reviewed exact head `d3359ee42b5e71531b87e7df6879338f574391f4`.

**The implementation defect from my prior review is fixed.** The caret steps now live in one nested `caretTl`; the font-ready pass kills only that child, clears the transform only when a prior scroll existed, rebuilds the steps at the same absolute times, and re-adds the child at position 0. The master `tl` remains registered and parent-driven seeking is preserved. The PR body also now states the real three-template behavioral scope accurately.

**Remaining important gate — the requested regression is not committed.** The delta from the blocked head changes only `chatgpt-exchange.html`; no test or regression fixture exercises an overflowing prompt through `document.fonts.ready` and parent-driven seek. The current required suite renders the packaged prompt, which emits zero caret-scroll keyframes and therefore cannot distinguish the original failure from this fix. That is exactly how the broken implementation stayed fully green.

Please commit a browser/runtime regression with a prompt long enough to emit steps, wait for the font-ready rebuild, then assert at minimum that the registered master timeline remains parented and the caret child count/timestamps do not double before seeking through the typing and collapse interval. The manual measurement in the PR body is good evidence for this head, but it is not a guard against the next edit; this repo requires changed logic branches to have a test that can fail when production behavior regresses.

All required exact-head checks are green. This hold is only for the missing discriminating regression; the code fix itself is accepted.

**Verdict:** REQUEST_CHANGES

**Reasoning:** The timeline bug is repaired, but the only path capable of detecting it remains outside CI, despite the prior review explicitly requesting that regression and the default-render suite being structurally unable to cover it.

— Magi

jrusso1020 (commented):
Re-review at `d3359ee4`. Posting this as a comment rather than an approval: @miguel-heygen's `CHANGES_REQUESTED` is pinned to this exact head, so it is a live block and not mine to stamp over. On the code I am satisfied — details below, including a measured reason the specific test being asked for would not do what it looks like it does.

### The fix is correct, verified at the mechanism

I checked the new shape against the same harness that found the defect rather than against the diff's own reasoning. Replicating `layoutCaretScroll` verbatim on `gsap@3.14.2`:

- master child count stable across two passes with different metrics (`5 → 5`) — the stale nest is genuinely discarded, where the old shape went `4 → 7` with nothing removed;
- the master stays parented, so the sub-composition and `play()` paths the old `tl.kill()` severed are intact — under a parent-driven seek the value now propagates (`-52`) where the old shape froze at its initial state;
- a time only the first pass covered now renders the *second* pass's value (`-26`, not the stale `-40`), which is the bug the `document.fonts.ready` re-measure exists to prevent and previously did not;
- `tl.add(caretTl, 0)` preserves absolute times — a step authored at `1.0` reads `-13.00` at `t=1.04` under `ease: "none"`;
- master `totalDuration()` unchanged, so nesting does not extend the composition.

This is also stronger than the `tl.recent()` alternative I offered. That one stays correct only while every future edit remembers to capture the return value; killing a nested child *cannot* reach its parent, so the property holds structurally. The comment at `:1388-1394` records the trap for the next person, which is the part that outlives us both.

### On the requested regression — the shape matters, and I measured it

A long-prompt + font-ready + parent-seek test committed against this repo's unit-test DOM would pass vacuously, on the broken code and the fixed code alike. Measured in `happy-dom@20`, the environment `packages/player` runs:

- `getBoundingClientRect()` on the composer text returns `{left: 0, width: 0}`, and `offsetWidth` / `offsetLeft` are `0`;
- so `limit` computes to `-8`, and `if (!(limit > 0)) return;` fires before any step is collected — zero caret keyframes regardless of prompt length;
- and `document.fonts` does not exist at all, so the second pass never runs.

Both preconditions for the defect are unreachable there. A green test in that environment would read as protection while asserting nothing, which is worse than the absence of one.

Where it *could* live is the render regression corpus — `regression.yml` renders named fixtures for real, in shards, which is the only place layout and `document.fonts` both exist. That is a new fixture plus shard budget rather than a unit test, so it is a reasonable follow-up ask but a heavy gate on a fix that is already structurally correct. Calibrating it as *important*, not a blocker.

**CI at this head:** 57 distinct checks, 0 not green.

**Verdict:** COMMENT — I would approve this on the merits; withholding the stamp only because a peer block is live at this head.
**Reasoning:** The defect is closed at the mechanism and verified by measurement; my one substantive difference is that the requested test shape cannot observe the bug it is meant to pin.

— Rames Jusso

WaterrrForever:
Regression added at `458a6101d` — but not the shape you asked for, because I measured that shape and it would not have failed. Both alternatives first, since the reasoning is the point:

**A unit test cannot see the bug.** Rames's happy-dom measurement is correct; I reproduced it:

```
getBoundingClientRect: {"left":0,"width":0}
offsetWidth: 0  offsetLeft: 0
document.fonts exists: undefined
limit computed by the template's guard: -8  →  guard returns early: true
```

`limit = -8`, so `if (!(limit > 0)) return;` fires before a single keyframe is collected, at any prompt length — and with no `document.fonts` the second pass never runs. Both preconditions for the bug are unreachable. The test would pass on the broken code and the fixed code alike.

**A render fixture would not have failed either**, which I did not expect and had to check before spending a golden mp4 on it. Long prompt, real pipeline, broken template vs fixed:

```
SSIM All: 0.999994 over 894 frames
frames below 0.999: 0 of 894      worst frame: 0.999871
duration: 14.900 both
```

The renderer drives the composition by explicit seek, and the duplicated keyframes are appended last so they win the tick. That is precisely *why* `.kill()` on the master timeline stayed invisible for so long: the two observable effects — detachment from the parent, and 49 stale keyframes — are both things an explicit seek papers over. A fixture plus shard budget would have guarded nothing.

**So the guard is the root cause, which is static and needs no DOM.** `tl.to()` returns the timeline, `gsap.to()` returns a tween, they read identically. New lint rule `gsap_timeline_return_used_as_tween` in `packages/lint/src/rules/gsap.ts`, error severity. It keys on the timeline handles `collectTimelineVarNames` already collects — that is what keeps `gsap.to()` and `Array.from()` out. Pushing a timeline return into an array is flagged outright; a bound return only once something tween-scoped is aimed at it.

Verified against the real files rather than synthetic ones:

- pre-fix `chatgpt-exchange`: **1 error**, snippet points at `caretScrollTweens.push(tl.to(typedEl, ...))`
- fixed `chatgpt-exchange`: **0**
- all **642** HTML files under `registry/` and `examples/`: **0** findings
- the 8 promoted blocks vs `origin/main`: identical error-for-error and warning-for-warning, so error severity costs no existing template
- 540 lint tests pass, 6 new — including the nested-timeline shape as an accepted case, so the fix hint is covered by a test rather than just asserted

This covers every composition in the registry instead of one template, and it fails in CI on the source, where the bug actually lives.

If you still want the render-corpus fixture on top of this I'll add it, but I'd rather not check in a golden that cannot go red — say the word and I'll open it as a follow-up instead.

miguel-heygen (approved):
Re-reviewed exact head `458a6101d8c7a38ea9b34f28ee06efb8c76aef40`.

The new guard satisfies the durability concern from my prior review, and it is the right kind of test for this failure.

I agree with the measured constraint: happy-dom has no usable layout metrics or `document.fonts`, so a nominal long-prompt unit test would return before producing a caret keyframe on both the broken and fixed implementations. A golden render is also the wrong oracle because explicit seek lets the duplicated last-added keyframes win, leaving near-identical pixels while live parent-driven playback is broken.

The lint rule instead pins the semantic root cause directly. It derives actual timeline handles through `collectTimelineVarNames`, rejects collection of `tl.to()` / `from()` / `set()` returns, and rejects a bound timeline return only when tween-scoped lifecycle methods are later aimed at it. It does not flag `gsap.to()`, ordinary chaining, `Array.from()`, or the nested-timeline fix. The six tests cover the original collect-and-kill shape, the bound-return shape, both valid alternatives, and two false-positive boundaries. Error severity is justified because the failure detaches a registered composition timeline.

I also verified the production fix remains the nested child shape, so the new rule and the repaired template agree. No blocker remains from my pass.

Required exact-head checks are still running with no completed failures; runtime contract and semantic title are already green. This approval is on code quality, not a merge instruction—wait for the remaining required contexts.

**Verdict:** APPROVE

**Reasoning:** The master-timeline defect is fixed structurally, and the committed lint rule is a non-vacuous, repo-wide regression guard that catches the precise API misuse which caused it.

— Magi

jrusso1020 (approved):
Third-round review, at exact head `458a6101`. Additive over @miguel-heygen's approval, which already covered the rule's adequacy, the six tests, why the runtime-test alternatives are vacuous here, and that the repaired template still uses the nested child. I re-derived the rule independently rather than checking its test names, and I'm adding a ground-truth control, a registry-wide sweep, and one coverage gap.

**The rule is validated against ground truth — this is the part that matters.** I ported the rule's logic and helpers verbatim (`stripJsComments`, `escapeRegExp`, `collectTimelineVarNames`) and first confirmed the port reproduces all six of this PR's own asserted outcomes, 6/6. Then I ran it against the *actual* pre-fix source at `e299a982`: it fires **twice**, at both `caretScrollTweens.push(tl.to(...))` sites. Against the fix at this head: **zero**. A guard that fires on the specific code that shipped the defect, and not on its repair, is a different thing from a guard that fires on a paraphrase of it — last round the concern was precisely that a plausible-looking test here would pass identically on broken and fixed code. This one doesn't.

**Strengths**
- `gsap.ts:2168-2176` keys the pattern on `collectTimelineVarNames` rather than on `.to(`. That's the design decision the rule lives or dies on: it's what keeps `gsap.to()` — which really does return a Tween — and `Array.from()` outside the match, instead of flagging every `.to(` in the file.
- The bound-name branch (`:2180-2186`) only reports once something tween-scoped is aimed at the alias, so ordinary chaining doesn't generate noise. Test 6 pins exactly that, and it's the test I'd have asked for.
- The `fixHint` prescribes the nested-timeline shape *and* says children keep their absolute times when the nest is added at 0. That second clause is the non-obvious half, and it's what makes the hint actionable instead of merely correct.
- Blast radius is clean: I swept the rule over all **162** registry block HTML files at this head — 162/162 have a detected timeline (so the zero isn't a detector that saw nothing), **0 findings**. No false positives across the registry, and no other latent instance of the pattern.

**important — the detector cannot see the shape the fixHint prescribes**

`collectTimelineVarNames` (`gsap.ts:826`) requires a declarator adjacent to the constructor:

```
/(?:const|let|var)\s+([A-Za-z_$][\w$]*)\s*=\s*gsap\.timeline\b/
```

The landed fix builds its nested timeline as `let caretTl = null;` (`chatgpt-exchange.html:1395`) and then `caretTl = gsap.timeline();` (`:1428`) — a plain reassignment. So `caretTl` never enters `timelineVars`, and every rule keyed on that helper is blind to it.

Measured, not inferred: reintroducing the exact original collect-and-discard bug on the nested timeline — `bad.push(caretTl.to(typedEl, {...}, at))` — yields **0 findings**. The same mutation against a `const`-declared nested timeline yields **1**. So the miss is specifically the declaration form, and it's the form this PR's own fix and fixHint use.

Why I'm not blocking on it: nothing is live (all 162 blocks are clean above), and widening the helper is not free — `gsap_callback_dom_measurement` (`:2070`, severity `warning`) shares it, so a wider regex newly registers reassigned timeline handles for that rule too and can surface new warnings across the registry. That deserves a deliberate change, not a one-line drive-by inside this PR. The cheap move here, if you want one, is a line in the rule comment noting that a reassigned timeline handle isn't covered — so the next author doesn't read a clean lint as proof of absence.

**note — what the guard does and doesn't buy**

The six tests run under the required `Test` context (`bun run --filter '!@hyperframes/producer' test` includes this package), so the rule's *logic* is pinned by CI. Its application to authored blocks is not: `bun run lint` is `oxlint` plus the `check:*` scripts and never calls the linter (and `Lint` isn't a required context anyway); `registryBlocks.test.ts` doesn't lint; `registryComponents.test.ts` covers `registry/components` and filters to `media_missing_src`; and the compile-time path (`staticGuard.ts:19-21` consumed at `htmlBundler.ts:814`) only `console.warn`s on an error-severity finding rather than failing the bundle. So this reliably catches the pattern for anyone running `check`, and it will not by itself turn a future PR red. That's pre-existing wiring shared by every rule in this file — not something this PR changed. I flag it only so "durable guard" isn't over-read.

**CI at this head:** 6 of the 8 required contexts are green. `Test` is still in progress — that's the one running these six tests — and `regression` / `Tests on windows-latest` are summary contexts that haven't been created yet while their matrices finish. Nothing is failing. So this is a code approval, not a merge signal.

**Verdict:** APPROVE
**Reasoning:** The fix I verified last round is intact, and the new rule is a real guard rather than a vacuous one — proven by firing twice on the exact pre-fix source, staying silent on the repair, and producing zero false positives across all 162 registry blocks. The detector gap is a coverage limit inherited from a shared helper, with nothing live behind it, so it belongs in a follow-up rather than gating this.

— Rames Jusso

PR #3571, opened 2026-08-31
fix(cli): guard PowerShell process queries against exited PIDs

miga-heygen:
## Summary

`processIdentity` and `processParentPid` call `Get-CimInstance Win32_Process` to look up process metadata on Windows. When the target process has already exited, `Get-CimInstance` returns null and calling `.CreationDate.ToFileTimeUtc()` or `.ParentProcessId` on it throws `InvokeMethodOnNull`. The try/catch handles it, but PowerShell writes the error to stderr, which pollutes the test runner's output and causes spurious exit code 1 on Windows CI.

Two fixes per call site:
- Null-check the CimInstance before accessing properties (`$p = ...; if ($p) { $p.Property }`)
- `-ErrorAction SilentlyContinue` + `stdio: ["pipe", "pipe", "ignore"]` to suppress any residual stderr

Fixes the recurring `Tests on windows-latest` flake on main.

## Test plan

- [x] All 196 CLI test files pass locally
- [ ] Windows CI should no longer exit 1 from PowerShell stderr noise

— Miga

### DIFF
--- a/packages/cli/src/utils/orphanCleanup.ts
+++ b/packages/cli/src/utils/orphanCleanup.ts
@@ -101,9 +101,9 @@ export function processIdentity(pid: number): string | null {
           "-NoProfile",
           "-NonInteractive",
           "-Command",
-          `(Get-CimInstance Win32_Process -Filter 'ProcessId = ${pid}').CreationDate.ToFileTimeUtc()`,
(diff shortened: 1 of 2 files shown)


miguel-heygen (approved):

miguel-heygen:
### Merge activity

* **Aug 31, 9:15 PM UTC**: @miguel-heygen merged this pull request with [Graphite](https://app.graphite.com/github/pr/heygen-com/hyperframes/3571).

PR #3572, opened 2026-08-31
fix(producer): degrade gracefully when font cache directory is unwritable

### LINKED ISSUE #3412: Unwritable font cache directory aborts the render instead of degrading

miga-heygen:
## Summary

When the font cache directory can't be created (`EPERM`/`EACCES` on `mkdirSync`), the render aborted with a raw filesystem error. The cache is an optimization — it stores Google Fonts downloads between runs — so an unwritable cache should degrade, not crash.

The fix wraps `mkdirSync` in `fontCacheDir` with a try/catch. On failure, it falls back to a temp directory (`mkdtempSync`) for the process lifetime, mirroring the existing Lambda fallback pattern. A warning is logged so users know fonts are cached to a temp location.

## Root cause

`fontCacheDir` in `deterministicFonts.ts` called `mkdirSync` unguarded. When `~/.cache/hyperframes/fonts/` was unwritable (permissions, read-only filesystem, corporate policy), the throw propagated through compile and failed the render.

Fixes #3412.

## Test plan

- [x] All 7 `deterministicFonts.test.ts` tests pass
- [x] No type errors in `deterministicFonts.ts`
- [ ] CI green

— Miga

### DIFF
--- a/packages/producer/src/services/deterministicFonts.ts
+++ b/packages/producer/src/services/deterministicFonts.ts
@@ -746,14 +746,22 @@ function fontSlug(familyName: string): string {
     .replace(/^-|-$/g, "");
 }
 
+let ephemeralFontCacheRoot: string | undefined;
+
 function fontCacheDir(slug: string): string {
   const dir = join(resolveFontCacheRoot(), slug);
   if (!existsSync(dir)) {
     try {
       mkdirSync(dir, { recursive: true });
     } catch {
-      const fallback = join(tmpdir(), `hyperframes-fonts-fallback`, slug);
+      ephemeralFontCacheRoot ??= mkdtempSync(join(tmpdir(), "hyperframes-fonts-"));
+      const fallback = join(ephemeralFontCacheRoot, slug);
       mkdirSync(fallback, { recursive: true });
+      defaultLogger.warn(
+        `Font cache directory is unwritable (${dir}). ` +
+          `Using temporary fallback — fonts will re-download each run. ` +
+          `Fix with: chmod 755 ${resolveFontCacheRoot()}`,
+      );
       return fallback;
     }
   }
--- a/packages/producer/src/services/deterministicFonts.ts
+++ b/packages/producer/src/services/deterministicFonts.ts
@@ -749,7 +749,13 @@ function fontSlug(familyName: string): string {
 function fontCacheDir(slug: string): string {
   const dir = join(resolveFontCacheRoot(), slug);
   if (!existsSync(dir)) {
-    mkdirSync(dir, { recursive: true });
+    try {
+      mkdirSync(dir, { recursive: true });
+    } catch {
+      const fallback = join(tmpdir(), `hyperframes-fonts-fallback`, slug);
+      mkdirSync(fallback, { recursive: true });
+      return fallback;
+    }
   }
   return dir;
 }
--- a/packages/producer/src/services/deterministicFonts.ts
+++ b/packages/producer/src/services/deterministicFonts.ts
@@ -754,14 +754,17 @@ function fontCacheDir(slug: string): string {
     try {
       mkdirSync(dir, { recursive: true });
     } catch {
+      const firstFallback = ephemeralFontCacheRoot === undefined;
       ephemeralFontCacheRoot ??= mkdtempSync(join(tmpdir(), "hyperframes-fonts-"));
       const fallback = join(ephemeralFontCacheRoot, slug);
       mkdirSync(fallback, { recursive: true });
-      defaultLogger.warn(
-        `Font cache directory is unwritable (${dir}). ` +
-          `Using temporary fallback — fonts will re-download each run. ` +
-          `Fix with: chmod 755 ${resolveFontCacheRoot()}`,
-      );
+      if (firstFallback) {
+        defaultLogger.warn(
+          `Font cache directory is unwritable (${dir}). ` +
+            `Using temporary fallback — fonts will re-download each run. ` +
+            `Fix with: chmod 755 ${resolveFontCacheRoot()}`,
+        );
+      }
       return fallback;
     }
   }


jrusso1020 (commented):
Reviewed at `80357092c375fe84f95e8964b06466c3bae508fc`.

**This branch moved backwards.** I first read it at `a0242e0b1d1502ebcaf4ac93ea68de2643ae9bee` (`+16/-2`) and by the time I went to post, the head was `80357092c3` (`+7/-1`, pushed 22:25:47Z). The earlier version was the better one and I would restore it.

What the earlier revision did:

- `mkdtempSync(join(tmpdir(), "hyperframes-fonts-"))`, so the fallback got a unique directory created with private permissions.
- Logged a warning naming the unwritable root and the fallback, so the degradation was visible in render output rather than silent.
- Cached the fallback in `fallbackCacheRoot`, so every later slug resolved under the same directory instead of re-deciding per font.
- Threw a contextual error if the fallback itself failed, with the recursion bounded to one level.

That also matched the convention already in this file: `resolveFontCacheRoot` uses exactly this `mkdtempSync` + module-level-cache shape for `lambdaFontCacheRoot`.

What the current revision does instead:

- Builds a fixed, guessable path at `join(tmpdir(), "hyperframes-fonts-fallback", slug)`. On a shared host that is a world-writable directory with a predictable name, which invites collisions between concurrent renders and is the classic setup for a symlink swap.
- Drops the warning entirely, so an unwritable font cache now degrades silently.
- Does not guard the fallback `mkdirSync`, so if that also fails the raw error propagates with no context.
- Re-derives the fallback per slug rather than remembering it.

**It is also now a duplicate.** This revision is byte-identical to the font-cache half of #3576: both go from base blob `7b99aad6e1` to `8c41c886fe`. Two open PRs carrying the same change to the same file will conflict on whichever merges second.

One gap that applies to either revision: `fontCacheDir` only enters the try/catch when the directory does not exist. A directory that exists but is not writable, such as a root-owned `~/.cache/hyperframes/fonts` left by an earlier privileged run, returns unchanged and fails later at write time, past the fallback. Given the PR title, that is arguably the main case.

Recommendation: restore `a0242e0b1d`, and close either this or #3576 so only one carries the change.

Not approving. This PR is bot-authored, and my standing rule is that a bot's autonomously created PR needs the owner's explicit go before I stamp it.

_Review by Rames_

jrusso1020 (commented):
Re-reviewed at `db15463d7cd8fec1e2d5cdb40ec6a96c7326dae6`. This resolves what I raised, and it is better than the `a0242e0b1d` revision I asked you to restore.

Confirmed at source:

- `mkdtempSync(join(tmpdir(), "hyperframes-fonts-"))` is back, so the fallback directory is unguessable and created with private permissions. That closes the collision and symlink concern with the fixed `hyperframes-fonts-fallback` path.
- `ephemeralFontCacheRoot ??=` caches the root, so every slug in a run resolves under one directory instead of re-deciding.
- The warning is back and now names the unwritable directory and gives a `chmod 755` remedy, which the earlier revision did not.
- Computing `fallback` directly and returning it is simpler than the previous recursion through `fontCacheDir`, and it removes the need to reason about re-entry at all.
- No longer duplicates #3576, which is now closed.

One nit, worth fixing while you are in here.

**The warning fires once per font, not once per run.** The try/catch is per-slug and `??=` only suppresses the repeat `mkdtempSync`, not the repeat `defaultLogger.warn`. A composition using a dozen families will emit a dozen identical warnings naming a dozen different `dir` paths. The revision I originally pointed at guarded the warn behind the first fallback so it fired once. Gating on whether `ephemeralFontCacheRoot` was already set before this call would restore that.

Two smaller notes, neither blocking:

- The inner `mkdirSync(fallback, { recursive: true })` is unguarded, so if the temp directory is also unwritable the raw error propagates without context. Unlikely to happen, but the earlier revision threw a contextual error there.
- The `existsSync(dir)` short-circuit still means a directory that exists but is not writable never reaches this fallback and fails later at write time. That is the one gap from my first review that is still open. Fine to leave for a follow-up, but it is arguably the more common form of "unwritable font cache".

Still not approving, for the same reason as before rather than anything about the code: this PR is bot-authored, and a bot's autonomously created PR needs the owner's explicit go before I stamp it. On the substance I have no objection left beyond the warn-per-font nit.

_Review by Rames_

jrusso1020 (approved):
Reviewed at `6ea0e7dee56aa71485e1adc1ac3964d8f9e6342c`. Clean, and the nit from my earlier pass is fixed correctly. Approving.

## The warn-once is right, and it is the part that is easy to get wrong

`firstFallback` is captured **before** the `??=`, not after. That ordering is the whole mechanism: read it after and `ephemeralFontCacheRoot` is already assigned, so the flag is false on the very first fallback and the warning never fires at all. As written the operator gets exactly one line the first time the cache degrades, which is the intended behaviour.

## The fallback fits the existing design rather than duplicating it

`resolveFontCacheRoot` already handles one unwritable-root case proactively, the Lambda `/home/sbx_*` tree, by minting a per-process `mkdtempSync` root. This adds the reactive counterpart for every other environment: an unwritable `~/.cache/hyperframes/fonts`, or a `HYPERFRAMES_FONT_CACHE_DIR` pointing somewhere the process cannot write. Two module-level roots for two genuinely different triggers, not the same mechanism written twice.

Imports resolve (`mkdtempSync` from `node:fs`, `tmpdir` from `node:os`), and the `return fallback` inside the `catch` correctly bypasses the trailing `return dir`.

## Non-blocking: one shape of "unwritable" still gets through

`if (!existsSync(dir))` guards the whole `try`, so the fallback is only reachable when the directory does **not** exist. A slug directory that exists but cannot be written to returns `dir` unchanged and the caller fails on the write, with the original unhandled error the PR title is about.

The case is narrower than it first looks, because a font that was never cached has no slug directory, so `mkdirSync` throws and the fallback works. It bites specifically on a partially populated cache that has become read-only: the slug directory is there, some subsets are present, and a newly needed subset has to be written. A container that bakes the cache at build time as root and then runs as a non-root user is the shape I would expect to hit it.

Not blocking, because this is strictly better than before and does not regress anything. Worth knowing the gate is `existsSync`, not writability.

_Review by Rames_

PR #968, opened 2026-05-19
docs(deploy): templates-on-lambda guide for personalised video at scale

jrusso1020:
## What

Phase 9 PR 9.5 of the distributed rendering plan — the user-facing `templates-on-lambda` guide. Documents the personalised-video-at-scale flow now shippable end-to-end after PRs 9.1–9.4.

## Why

PRs 9.1–9.4 plumb variables through every layer (producer plan/renderChunk, aws-lambda SDK, single-render CLI, batch CLI). Without a guide, users have to reverse-engineer the flow from `--help` output and source comments. The docs ARE the deliverable that makes "easy template pipelines" real — that's why this PR exists separate from the code changes.

## How

`docs/deploy/templates-on-lambda.mdx` walks all 11 sections required by the Phase 9 plan:

- (a) What's a template (composition + `data-composition-variables`)
- (b) Declaring variables (syntax, types, defaults, `getVariables()`)
- (c) Local iteration loop (`hyperframes render --variables` / `--variables-file` / `--strict-variables`)
- (d) Deploying to Lambda (pointer to deploy guide + `sites create`)
- (e) Single personalised render (`lambda render --variables`)
- (f) Batch pipeline — the headline (`lambda render-batch --batch users.jsonl`, with a worked 5-row example, manifest output, progress polling, `--dry-run`)
- (g) Programmatic via SDK (TypeScript example with `deploySite` + `Promise.all(renderToLambda)`)
- (h) Working with large variables (the 256 KiB Step Functions Standard ceiling, URL-your-assets convention, one-line escape note for genuine > 256 KiB cases)
- (i) Cost + scale considerations (Lambda concurrency, `--max-parallel-chunks` vs `--max-concurrent`, in-process vs distributed crossover)
- (j) Migrating from `@remotion/lambda` `inputProps` (side-by-side table; same 256 KiB cap + same URL-your-assets convention, so migration is mechanical)

Plus a Mermaid architecture diagram for the site-upload-once + N-execution fan-out flow at the top.

`docs/docs.json` adds the page to the Deploy navigation group between the existing `aws-lambda` and `migrating-to-hyperframes-lambda` pages.

## Test plan

- [x] Guide structure matches the 10-section spec from the Phase 9 plan (verified via `grep "^## "`).
- [x] Mermaid diagram renders (Mintlify's preview will catch syntax errors).
[...]

PR #3578, opened 2026-09-01
fix(encoder): signal host interruptions for retry

jrusso1020:
## What

Classify verified external FFmpeg SIGTERM interruptions and expose a sanitized `ENCODER_INTERRUPTED` producer error across streaming, GIF, disk encode, audio normalization, mux, and faststart paths.

## Why

Spot-node drains terminate active FFmpeg processes. Those failures were flattened into generic exit-code 255 errors, so Temporal treated an infrastructure interruption as a terminal render failure instead of retrying on a fresh producer. Context: [incident thread](https://heygen.slack.com/archives/C09JCPXF3HS/p1788219792750419?thread_ts=1788216100.950949&cid=C09JCPXF3HS).

## How

- Detect direct SIGTERM or FFmpeg's exact handled-signal-15 exit signature while excluding app aborts, deadlines, inactivity timeouts, SIGKILL, and generic code 255 failures.
- Propagate a stable `external_interruption` reason through every render-time FFmpeg stage.
- Convert it to a fixed, retryable, system-owned producer error while retaining bounded raw diagnostics only in local logs.
- Bypass same-host capture fallback and close the stdin/child-exit race that could erase the signal.

Activation dependency: merge and publish this OSS package before deploying the internal producer consumer PR.

## Test plan

426 focused tests pass across engine and producer paths. Engine and producer TypeScript checks, Oxlint/Oxfmt, the repository pre-commit checks, and a cross-repository internal producer build against this checkout pass.

Not exercised in dev: the OSS package must first be published and consumed by the internal producer image before a real Spot/SIGTERM retry can be injected.


miguel-heygen (changes requested):
The core classifier and same-host fallback boundary are well designed:

- `packages/engine/src/utils/runFfmpeg.ts:34-50` requires either a direct SIGTERM or FFmpeg's exact handled-signal-15/exit-255 signature, while managed abort/deadline/inactivity reasons take precedence.
- `packages/producer/src/services/render/encoderInterruption.ts:1-29` keeps the wire-facing message fixed and the raw diagnostic local.
- `packages/producer/src/services/renderOrchestrator.ts:1843-1851,3603-3610` correctly prevents a host interruption from being hidden behind the same-host pinned capture fallback.

**Blocker — the new `runFfmpeg().failureReason` contract is only consumed by selected encode/assemble-stage call sites, so the same Spot/SIGTERM failure remains terminal elsewhere in the same render pipeline.**

The clearest in-process example is audio: `packages/engine/src/services/audioMixer.ts:376-418` ignores `result.failureReason`, defaults an external SIGTERM to `reason="ffmpeg_failed", retryable=false`, and `packages/producer/src/services/render/stages/audioStage.ts:133-155` returns that into the fatal audio-warning path. A Spot drain while FFmpeg is extracting/preparing/mixing audio therefore becomes `RenderQualityError`, not `EncoderInterruptedError`, so the EF consumer never receives `ENCODER_INTERRUPTED` and does not retry on a fresh host.

The decision also drifts in sibling encode/mux paths: `packages/producer/src/services/distributed/assemble.ts:184-225,289-315,334-363` drops `failureReason` for remux, concat, CFR re-encode, pad/trim, mux, and faststart, even though this PR explicitly adds the reason to those shared engine results. `packages/producer/src/services/render/stages/captureHdrResources.ts:491-502` does the same for HDR FFmpeg pre-extraction.

Please make external-interruption propagation a single producer-owned boundary (or explicitly handle the reason at every `runFfmpeg`/mux/faststart result consumer) and add mutation-sensitive coverage for at least the standard audio stage plus distributed assembly. Otherwise the retry works only when the host happens to drain during the subset of FFmpeg stages changed here.

**Verdict:** REQUEST CHANGES
**Reasoning:** False-positive classification and redaction are sound, but the same verified host interruption still terminates through uncovered FFmpeg consumers, so the cross-stack retry contract is incomplete.

— Magi

somanshreddy:
**Review at head `79afdaf9` — focused on the classification boundary + same-host fallback (per the scope split), two passes (Codex + mine) reconciled at source.** I read `runFfmpeg.ts`/`managedChildProcess.ts`/`processTracker.ts`/`streamingEncoder.ts`, the producer render stages (encode/assemble/captureHdr/captureStreaming/audioPadTrim), `encoderInterruption.ts`, `renderOrchestrator.ts`, `audioExtractor.ts`, `audioMixer.ts`, `distributed/assemble.ts`+`renderChunk.ts`, `server.ts`, and the classification tests, all at this head.

## The classification boundary is sound — this part is right

The false-positive defense holds precisely, and I verified each anchor:
- **Managed kills can't be misclassified.** `isExternalFfmpegInterruption` returns false unless `terminationReason === "exit"`, and `ManagedChildProcess.onClose` sets `reason: this.requestedReason ?? "exit"` — so every abort/deadline/inactivity termination (which *does* internally `kill("SIGTERM")`) is tagged with its request reason, and the `signal === "SIGTERM"` branch is never reached for our own kills. The test matrix pins exactly this (all three managed reasons stay non-retryable even with `signal:"SIGTERM"` **and** the stderr line present).
- **The two positive branches are well-chosen.** `signal === "SIGTERM"` is Node-reported and unspoofable; `exitCode === 255 && /^Exiting normally, received signal 15\.?\r?$/m` is ffmpeg's own SIGTERM-handler signature. SIGKILL, ordinary exit codes, and generic 255-without-the-line are all correctly excluded, each with a test.
- **Same-host fallback bypass is precise.** `shouldRetryViaPinnedFallback` returns false only for `EncoderInterruptedError` (alongside cancellation), so a drained host isn't retried in-place, while *non-interruption* failures keep their normal pinned fallback. Good.
- **Redaction is at the write site.** `EncoderInterruptedError.message` is a fixed string; the raw ffmpeg diagnostics live only in `diagnosticMessage`, which the orchestrator logs bounded to 2 KB locally and (per the comment + PR #968) never puts on the wire. This is the correct shape.

## Blocking — the interruption signal is not emitted on every FFmpeg path (confirmed at source; concurs with Magi's + Codex's CHANGES_REQUESTED)

The retry contract only holds for the paths that thread `failureReason` → `encoderFailureError`. A Spot drain during any of the following still terminalizes as a generic/non-retryable failure, so EF never sees `ENCODER_INTERRUPTED` and never retries on a fresh host — a false-*negative* against the fix's own goal. I verified each:

- **`distributed/assemble.ts:184` (+ concat/mux/faststart) — highest priority.** The distributed assembler runs its own `runFfmpeg` remux/concat and throws `new Error(...)` on `!success`, ignoring `remuxResult.failureReason`. Distributed is the scaled production render path and its per-chunk encode *is* covered (`renderChunk` → `runEncodeStage`), which makes the uncovered final-assembly stage the most consequential hole.
- **`captureHdrResources.ts:491` (HDR pre-extraction).** `runFfmpegImpl` returns `failureReason`, but the caller throws a bare `Error("HDR frame extraction failed …")` and drops it.
- **`audioMixer.ts` render-time mixing** and **`audioExtractor.ts:92` extraction** funnel into `AudioProcessingFailure`/`RenderQualityError` (the warnings path), not `encoderFailureError` — so a drain mid-audio is a non-retryable audio failure.
- **`videoFrameExtractor.ts:787`** uses `runFfmpeg` but its result type doesn't carry `failureReason` onward.

Net: the covered set is the in-process encode/mux/HDR-encode/normalize/gif stages + distributed chunk encode; the gap is distributed assembly, HDR pre-extraction, and the audio/frame paths. Either extend `failureReason` through those (distributed/assemble first), or make it an explicit, documented v1 scope with the rest tracked — but as written, drains on the scaled path's assembly stage won't retry.

**Lower / with the same fix:** `captureStreamingStage` (~:855) — if the capture loop throws a *non-interruption* error while the encoder separately died from SIGTERM, `wrapCaptureStageError` attributes the browser error and the encoder's `getExitFailureReason()` is not consulted, so a genuine drain can be masked by the co-occurring capture error (both usually die together on a drain). Narrow, but it's in an otherwise-covered path. (Codex flagged this; plausible, medium.)

## Dropped from the Codex pass (verified not real)
- **`processTracker.ts:21` "application cleanup SIGTERM misclassified"** — not reachable in the producer. `killTrackedProcesses()` is only called from the CLI (`preview.ts`); `server.ts`'s SIGTERM `shutdown()` drains the browser pool and closes the server but never calls it. Producer ffmpeg children receive SIGTERM only from the OS on an actual drain — the correct case. (If `killTrackedProcesses` is ever wired into producer shutdown, revisit — though a drain-time SIGTERM would still be legitimately retryable.)

## One boundary nuance (non-blocking, low)
The `exitCode === 255 && stderr-line` branch trusts ffmpeg stderr content. The signal branch is unspoofable, but stderr can echo customer-influenced text (metadata tags, filenames). A crafted input that makes ffmpeg both exit 255 *and* emit a line exactly matching `Exiting normally, received signal 15.` would be misclassified → a bounded spurious retry (EF caps it, no child multiplication). Contrived and low-impact, but it's the one soft spot in an otherwise tight classifier — worth a comment acknowledging the trust boundary.

## Praise
The `isExternalFfmpegInterruption` test suite is exactly right — SIGKILL excluded, 255-without-line excluded, line-without-255 excluded, and managed abort/deadline/inactivity kept non-retryable even under a SIGTERM+line combination. That's a genuinely mutation-sensitive boundary suite.

---
**Verdict (COMMENTED, concurring with Magi's CHANGES_REQUESTED).** The classification core, redaction, and same-host fallback bypass — the parts in my scope — are correct and well-tested. The blocker is completeness: the interruption signal doesn't reach EF from distributed assembly, HDR pre-extraction, or the audio/frame paths, so drains there won't retry. Extend coverage (distributed/assemble first) or scope it explicitly. Merge gate is a `text-editor-squad` code-owner regardless.

somanshreddy:
**Re-review at `dc32b8791` — completeness blocker RESOLVED. Verified at source.**

I re-traced every path I flagged, at the new head:
- **Distributed assembly (`distributed/assemble.ts`)** — all six FFmpeg passes now route through `encoderFailureError` with `failureReason`: single-chunk remux, concat-copy, CFR re-encode, audio pad/trim, audio mux, faststart. This was my highest-priority gap (the scaled path); closed.
- **Standard audio** — `audioMixer.ffmpegFailure` maps `failureReason==="external_interruption"` → retryable `external_interruption`, **and** suppresses the same-host automation retry for it (also closes the "another attempt on the same host" concern); `audioStage` surfaces that failure as `encoderFailureError`.
- **Video frame extraction + HDR probe** (`videoFrameExtractor` → `external_interruption` typed error; `extractVideosStage.throwIfEncoderInterrupted` / `throwHdrProbeFailures`).
- **HDR raw pre-extraction** (`captureHdrResources.extractHdrVideoFrames:497`) — the exact path I cited, fixed in the `dc32b8791` follow-up.
- **Animated GIF prep** — threads it.

**Full sweep confirms coverage is now complete for the live render surface.** The only FFmpeg sites still not threading `failureReason` are benign: `ffprobe`/`gpuEncoder` (fast capability/metadata probes), `distributed/shared` (`ffmpeg -version`), the regression harness (tests), and `audioExtractor.processAudio` (no non-test importers — the live audio path is `processCompositionAudio`→`audioMixer`, which is covered).

The classification core (managed-reason precedence, unspoofable signal vs 255+signal-15 signature, same-host fallback bypass, write-site redaction) was already sound and is unchanged.

**Verdict: no blockers remain from my scope (OSS classification + same-host fallback + completeness). LGTM at `dc32b8791`.** Two non-blocking residuals I'll leave for a future pass, neither gating: the `255 + "Exiting normally, received signal 15."` branch trusts FFmpeg stderr (bounded spurious-retry risk from crafted metadata), and the narrow `captureStreamingStage` case where a co-occurring non-interruption capture error can mask an encoder interruption. Merge gate is the `text-editor-squad`/code-owner; required-head checks still finishing (no failures).

miguel-heygen (approved):
Final re-review at `dc32b879`: the completeness blocker is fully resolved.

- The `b92f3b40` delta propagates the structured interruption through standard audio, normal video/HDR probe extraction, GIF prep, and every distributed remux/concat/CFR/pad/mux/faststart result while suppressing same-host fallback/degradation retries.
- `packages/producer/src/services/render/stages/captureHdrResources.ts:491-503` now routes the last raw HDR pre-extraction subprocess through `encoderFailureError`, and `captureHdrResources.test.ts` positively pins `EncoderInterruptedError` for that result.
- The original false-positive boundary, fixed diagnostic message, JSON/SSE transport contract, and same-host capture-fallback bypass remain intact.

The render-time FFmpeg inventory from my blocker is now closed, and all exact-head required checks (including Windows and every regression shard) are green.

**Verdict:** APPROVE
**Reasoning:** Every production FFmpeg stage in scope now preserves the same single structured interruption signal, with the final HDR gap covered by a targeted regression and no broadening of the classifier.

— Magi

PR #3580, opened 2026-09-01
fix(cli): flag caption-zone by DOM box overlap

xuanruli:
## What

`--caption-zone` now flags a text element when its DOM box (`getBoundingClientRect`) overlaps the keepout, instead of when the box center sits inside the band.

## Why

A landscape card centered at `y=.860` can pass a 87.5–92.5% center gate and still cover the painted V2A/Pacific caption. The keepout size stays the 5% one-line pill; the predicate is what changes.

## How

AABB intersect of `elementRect` against the fractional band. Full-frame wrappers and sub-4px boxes stay excluded via `candidateIsSized`. Finding copy says "overlaps" instead of "is centred in".

## Test plan

- [x] Unit tests added/updated — overlap-only fixture now fails; 1920×1080 `y=.860` / 80px card fails the 5% keepout; same center with a 20px box that stays above 87.5% still passes
- [x] `bun run --cwd packages/cli test src/commands/check.test.ts` (52 passed)
- [x] Documentation updated (`docs/packages/cli.mdx`, CLI skill)

### DIFF
--- a/docs/packages/cli.mdx
+++ b/docs/packages/cli.mdx
@@ -624,7 +624,7 @@ drift), `*.motion.json` assertions, and WCAG AA contrast.
 | `--timeout`                                  | Render-ready budget in ms; also raises page navigation above its 10s floor (default 3000)                                   |
 | `--no-contrast`                              | Skip the WCAG pass while iterating                                                                                         |
 | `--strict`                                   | Exit non-zero on warnings too (default: errors only)                                                                        |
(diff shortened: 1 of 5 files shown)


xuanruli:
Paired EF prompt sync (keepout size unchanged): https://github.com/heygen-com/experiment-framework/pull/49010

miguel-heygen (approved):

PR #3608, opened 2026-09-03
chore: release v0.8.27

miguel-heygen:


jrusso1020 (approved):
Approving at `e42a1d31a0f450def3709674a0937e26bc645be0`.

*A release PR's one real risk is claiming to carry something it does not, so that is what I checked first.* Both commits the notes cite are ancestors of this head - `compare/<commit>...<head>` reports `ahead` for `5a5e841c` (#3606) and for `ae7e530f` (#3598) - so the tag cut from this branch will contain them rather than merely mention them. #3606 merged at 03:53:45Z and #3598 at 03:25:00Z, both before this branch was cut.

*The bump is uniform and the diff carries no code.* All thirteen `packages/*/package.json` read `0.8.27`, and so do the three plugin manifests. The remaining files are `docs/changelog.mdx` and the new `releases/v0.8.27.md`. Nothing else is in the diff, which is what a release PR should look like - a version-only change is the one shape where "the tests passed on main" is still evidence about what is shipping.

*The notes describe the user-visible effect rather than the mechanism*, which is right for a changelog, and both sentences are true of the merged code. On the favicon half I reviewed the implementation itself at `c6f4b57` and the ranking behaves as the note says: sharpest declared candidate first, with `object-fit: contain` and no invented background, so a transparent mark stays transparent.

*`Perf: scrub` and `player-perf` are RED at this head and I am approving over them deliberately, so here is the reasoning rather than an override with no note.*

The failure is not a perf result. `Perf: scrub` never reached its measure step - it died in `bun run build` with `File 'packages/core/src/generated/runtime-inline.ts' is not a module`, and `Run player perf - scrub (measure mode)` is marked skipped underneath it. `Perf: drift`, `fps` and `parity` are `cancelled`, which is the matrix failing fast around it rather than three more findings.

Three things say that is infrastructure and not this cut:

- `Build` is SUCCESS on this exact commit, so the repo does build.
- `Perf: load` is SUCCESS on this exact commit, and it runs the same `bun run build` in the same matrix - one shard read a generated file another had not finished writing.
- The diff is thirteen version strings, three plugin manifests and two documents. There is no path from it to the codegen of `runtime-inline.ts`.

None of `Perf: *` or `player-perf` is in this branch's required set, so they will not gate the merge either way. Worth a re-run before you merge, so the release does not land with a red lane on its commit that somebody has to re-derive later. The rest of CI is still in flight - regression shards, the windows lanes, `CLI smoke (required)` - so I am not calling the required set green.

The merge is yours and it is gated on those finishing; I am not merging or queueing this, and this repo requires an approval on the last push, so any further push needs a fresh one.

_Review by Rames_

jerrai-bot-heygen (approved):
Reviewed release v0.8.27 at exact head `e42a1d31a0f450def3709674a0937e26bc645be0`.

The release diff is limited to a uniform 0.8.27 version bump, release notes, and changelog. It contains the already-merged #3598 missing-asset accounting and #3606 favicon-ranking behavior; package/plugin versions are aligned and no lockfile update is required for workspace dependencies. Earlier transient perf shard failures were rerun; current completed checks are green and remaining jobs are pending.

— Jerrai

jerrai-bot-heygen (approved):
Re-approved at exact head `00b3fad70e0927442197828f25a847264063fab4` after the main-only merge. The release bump content remains unchanged and verified; CI has restarted at this head with no failures.

PR #3609, opened 2026-09-03
feat(lint): error when a video or img src is the other media kind

xuanruli:
## What

Add a `media_src_kind_mismatch` lint error when a `<video>` element's own `src` is an image, or an `` element's own `src` is a video.

## Why

The existing media-src check only asks whether `src` is a resolvable URL, so `<video src="still.jpg">` and `` lint green. Agents then ship that HTML; the producer later fail-closes because the tag and file kind disagree.

Companion bake-time rewrite (does not live here): https://github.com/heygen-com/experiment-framework/pull/49536

## How

Classify the element's own `src` from `data:` MIME or file extension. This rule is an **error**, so it only fires on cases that cannot be a false positive:

- `<video src>` whose path/MIME is an image
- `` whose path/MIME is a video

Skipped on purpose (extension cannot prove a producer failure): `blob:`, extensionless URLs, `<audio>` (including `<audio src="*.mp4">` SFX), and `<source>`.

## Test plan

- [x] Unit tests added/updated (`packages/lint/src/rules/media.test.ts`: video+jpg, img+mov, data:image, matching kinds, extensionless + audio)

### DIFF
--- a/packages/lint/src/rules/media.test.ts
+++ b/packages/lint/src/rules/media.test.ts
@@ -796,3 +796,73 @@ describe("audio_carve_ungrouped_sources", () => {
     expect(finding?.severity).toBe("warning");
   });
 });
+
+describe("media_src_kind_mismatch", () => {
+  it("errors when <video> src is an image", async () => {
+    const html = `
+<html><body>
+  <div id="root" data-composition-id="c1" data-width="1920" data-height="1080">
+    <video id="bg" src="still.jpg" data-start="0" data-duration="5" muted></video>
+  </div>
+  <script>window.__timelines = {};</script>
+</body></html>`;
+    const result = await lintHyperframeHtml(html);
+    const finding = result.findings.find((f) => f.code === "media_src_kind_mismatch");
+    expect(finding?.severity).toBe("error");
+    expect(finding?.elementId).toBe("bg");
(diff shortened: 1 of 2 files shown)


miguel-heygen (approved):
Reviewed at exact head `692580de744f2199426bedeba3ebe2f00d481d44`.

- `packages/lint/src/rules/media.ts:44-130` keeps the rule on the provable cases only: element-owned `<video>`/`` src, closed image/video suffix sets, data MIME support, case-folding, query/hash stripping, and explicit fail-open for blob/extensionless input. The suffix vocabulary matches EF's shared `get_resource_type_from_extension` image/video sets today.
- The finding reconstructs a bounded diagnostic without echoing the full src, is severity `error`, and leaves `<audio>`/`<source>` to their separate policies.
- `media.test.ts:799-868` discriminates both mismatch directions, data-image input, healthy matches, and the extensionless/audio carve-outs. Exact-head CI has zero non-passing checks.

Non-blocking: the suffix vocabulary is still manually duplicated across OSS and EF, so a parity fixture would be useful if either set grows.

**Verdict:** APPROVE
**Reasoning:** The rule fails only on extension/MIME evidence that proves a tag-kind mismatch, covers both directions, and preserves the intended ambiguous-source carve-outs.

— Magi

PR #3615, opened 2026-09-03
fix(producer): skip HDR layers outside their sub-composition window

### LINKED ISSUE #3591: [feedback] Render ignores scene sub-composition time windows when an HDR-tagged asset switches the render to HLG mode
**Rating:** 4/10 (also submitted via `hyperframes feedback`)

## Summary
`hyperframes render` ignores scene sub-composition time windows when an HDR-tagged asset switches the render into HLG mode. Studio preview plays the same project correctly, so the defect only appears in the rendered file.

## Reproduction packet
```
REPRO COMMAND: npx hyperframes render --quality high --output repro.mp4   # same defect with --quality draft
EXPECTED / ACTUAL: at t=30s (scene B window, 20-40s) the frame should show scene B / it shows scene A full-frame —
  the first scene stays composited for the rest of the video ("stuck"), and scene-root backgrounds are lost so later
  scenes' elements layer over the stale scene. Trigger isolated: one <video> whose 8-bit H.264 source carries HDR
  color tags (bt2020nc + arib-std-b67, from an iPhone HLG .mov cut) -> log shows
  "HDR auto-detected from source(s) - output: HLG" + phase capture_hdr_layered. Re-encoding the SAME clip with
  bt709 tags (no other change) renders perfectly on the normal capture_disk path.
EXACT ERROR: none; render exits 0. Visual defect from the first scene boundary onward (frames ~600+ at 30fps).
  hdr_detection trace: {"effectiveHdr":"hlg","nativeHdrVideoCount":1}
OUTCOME: output corrupt (silently — exit 0, lint/check pass, preview correct)
WORKAROUND: tonemap/re-tag the source to bt709 (zscale=t=linear,tonemap=hable,zscale=p=bt709:t=bt709:m=bt709) so
  hasHdrContent stays false; also added opaque full-bleed child bg divs per the layered-path guidance, but the
  re-tag alone fixes it.
COMPOSITION_STRUCTURE:
  elements: video=1 audio=0 img=1 svg=0 canvas=0 subComps=3
  attributes: transform, overflow:hidden, z-index, data-duration, data-start, data-composition-src
  timeline: nested (3 sub-comps); driver=gsap
  delta: only difference between broken and correct render is the video's color metadata (bt2020nc/arib-std-b67 vs
    bt709); quality level does not matter. A single-scene 8s version did NOT reproduce; 3 sequential scenes / 60s
    with the HDR video in scene 3 reproduces every time. First seen on a 7-scene 238s production comp (2 failures),
    reproduced here minimally.
  defect: full-frame / all frames after the first scene's window ends (frames 600-1800 of 1800)
```

## Minimal repro
Published minimal repro: https://hyperframes.dev/p/031d1650-6870-4463-bd26-77ceb1fa20ce

Notes for bisecting: the synthetic HDR trigger clip is `testsrc2` re-tagged `-color_primaries bt2020 -color_trc arib-std-b67 -colorspace bt2020nc` (8-bit yuv420p H.264) — no true HDR pixels needed, tags alone flip the path.

## Environment
```
os=linux/x64 node=v22.17.0 cpu=28cores mem=16GB ffmpeg=yes wsl
cli=0.8.23
```

---
_Filed via `hyperframes feedback --file-issue` (draft) and submitted with user consent._

miga-heygen:
## Summary

- The HDR blit loop only checked `opacity <= 0` when deciding whether to skip a layer — it ignored the sub-composition time window entirely, so when a scene ended, its HDR video/image elements kept painting the last captured frame over all subsequent scenes
- Extracted `isHdrLayerBlittable()` predicate that gates on `(visible || renderFrameVisible)`, matching the existing DOM layer rule in `selectDomLayerShowIds`
- Native videos are deliberately hidden during capture (`visibility: hidden`), so `el.visible` is always false for them — `renderFrameVisible` tracks the replacement `` whose paintability follows the scene's GSAP timeline state
- The same gate protects transition-scene capture, and the HDR regression golden now records the corrected visibility windows

This PR does not change documented auto-HDR codec/transfer selection, Windows temp placement, or black-level behavior.

## Test plan

- [x] 5 unit tests cover HDR video/image in and out of window, native replacement visibility, and zero opacity
- [x] 7/7 HDR compositor tests pass
- [x] 15/15 transition/shared HDR-frame tests pass
- [x] Producer typecheck plus changed-file lint and format pass
- [x] `hdr-regression` re-render passes all 100 visual checkpoints with zero failed frames against the refreshed LFS golden
- [ ] CI required checks rerunning

Closes #3591

— Miga

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/packages/producer/src/services/hdrCompositor.test.ts
+++ b/packages/producer/src/services/hdrCompositor.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from "vitest";
 import type { ElementStackingInfo } from "@hyperframes/engine";
-import { selectDomLayerShowIds } from "./hdrCompositor.js";
+import { isHdrLayerBlittable, selectDomLayerShowIds } from "./hdrCompositor.js";
 
 function makeEl(id: string, overrides?: Partial<ElementStackingInfo>): ElementStackingInfo {
   return {
@@ -24,6 +24,41 @@ function makeEl(id: string, overrides?: Partial<ElementStackingInfo>): ElementSt
     ...overrides,
   };
 }
+
+describe("isHdrLayerBlittable", () => {
+  it("skips HDR video elements outside their sub-composition window", () => {
+    // Native videos are hidden during capture (visible=false), so
+    // renderFrameVisible tracks the replacement <img>'s paintability.
+    // When the scene's time window has ended, both are false.
+    expect(isHdrLayerBlittable(makeEl("v1", { visible: false, renderFrameVisible: false }))).toBe(
+      false,
+    );
+  });
+
+  it("keeps HDR video elements within their sub-composition window", () => {
+    // During capture the video itself is hidden, but the replacement frame is visible.
+    expect(isHdrLayerBlittable(makeEl("v1", { visible: false, renderFrameVisible: true }))).toBe(
+      true,
+    );
+  });
+
+  it("keeps visible HDR image elements", () => {
+    // Images are not hidden by the frame injector — visible tracks the time window.
+    expect(isHdrLayerBlittable(makeEl("img1", { visible: true, renderFrameVisible: false }))).toBe(
+      true,
+    );
+  });
+
+  it("skips HDR image elements outside their time window", () => {
+    expect(isHdrLayerBlittable(makeEl("img1", { visible: false, renderFrameVisible: false }))).toBe(
+      false,
+    );
+  });
+
+  it("skips zero-opacity elements regardless of visibility", () => {
+    expect(isHdrLayerBlittable(makeEl("v1", { opacity: 0, renderFrameVisible: true }))).toBe(false);
+  });
+});
 
 describe("selectDomLayerShowIds", () => {
   it("does not re-show hidden DOM elements while preserving visible injected video frames", () => {
--- a/packages/producer/src/services/hdrCompositor.test.ts
+++ b/packages/producer/src/services/hdrCompositor.test.ts
@@ -96,4 +96,17 @@ describe("selectDomLayerShowIds", () => {
       ),
     ).toEqual(["active-overlay"]);
   });
+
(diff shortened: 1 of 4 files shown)


miguel-heygen (changes requested):
Exact-head review at `8bcf70fb394`

The normal-frame fix is correct, but the same defect remains in the transition capture path. `captureSceneIntoBuffer` is the second production HDR blit loop used for shader/CSS transition frames; at `packages/producer/src/services/render/stages/captureHdrFrameShared.ts:206-236` it blits every HDR scene member without applying `isHdrLayerBlittable`. An HDR child whose own sub-composition window has ended can therefore still paint its last frame into a transition scene buffer.

Please apply the shared paintability predicate at that HDR entrance too and add a transition-scene regression proving a hidden HDR video/image is not blitted while a visible replacement is. It would also be safer to have `selectDomLayerShowIds` call the same shared predicate instead of retaining a byte-equivalent second condition, since parity between these paths is the contract this fix depends on.

The reported non-transition reproduction is fixed by the current change, and the focused helper cases plus completed CI checks are sound.

miga-heygen:
Pushed a second commit at `5d73c8e62` — applies the same `isHdrLayerBlittable` gate to `captureSceneIntoBuffer` in the transition capture path. Genesis correctly identified this gap.

**On the CI golden-frame failure:** the `hdr-regression` suite's golden was captured with the buggy behavior — window A's `wa-video` (opacity=1.0 from its GSAP tween, but visibility=hidden from clip management at t=1.5s) was leaking its last frame into windows B–D. The new output correctly skips that blit, so the golden needs to be regenerated with `bun run test:regression:update hdr-regression` from a machine with Chrome headless + ffmpeg.

— Miga

PR #3618, opened 2026-09-04
feat(studio): upgrade Studio to Tailwind CSS v4

miguel-heygen:
Lands unit U14 (Tailwind v4) of the Studio design-system foundation. No ratchet yet; no color literal added. One PR over the usual size guideline: a half-migrated Tailwind tree does not build, so there is no seam to split at.

# Lands U14 of the Studio design-system plan. Ratchet delta: none (the hex ratchet lands in a later unit).

## What

Studio moves from Tailwind CSS 3.4 to 4.3.3, loaded through `@tailwindcss/vite` instead of PostCSS.

- `packages/studio/tailwind.config.js` and `packages/studio/postcss.config.js` are deleted. The JS config's `studio.*` and `panel.*` palettes become an `@theme` block in `src/styles/studio.css`, byte-for-byte the same hex values.
- `@import "tailwindcss" source(none)` plus explicit `@source` globs for `src/**/*.{ts,tsx}` and `index.html`. Automatic source detection would walk up to the repository root and scan `data/projects`, which is gitignored user content rather than Studio markup.
- `autoprefixer` and `postcss` are dropped from Studio's dev dependencies; `@tailwindcss/vite` is added.
- Renamed utilities migrated across Studio's components.

The published `./tailwind-preset` subpath is untouched: `src/styles/tailwind-preset.shared.js` and `tailwind-preset.ts` still exist and still build to `dist/styles/tailwind-preset.js`. Turning it into a deprecated shim is a later unit's job.

## Why

Tailwind v4 is a prerequisite for the token work in this plan. It makes the token file CSS-first, emits every token as a real CSS custom property, and handles alpha on any color format, which removes most of the hand-built plumbing the token unit would otherwise need.

## How

The official `@tailwindcss/upgrade` tool did the utility renames. Its diff was then reviewed by hand, which mattered more than expected: **the tool rewrites bare words inside any string literal, not only inside class attributes.** Left as-is it would have shipped these defects, all of which were reverted:

- `removeEventListener("blur", ...)` became `removeEventListener("blur-sm", ...)` in five files, so focus-loss cleanup would never have fired.
- `element.style.setProperty("flex-grow"/"flex-shrink", ...)` became `setProperty("grow"/"shrink", ...)` in the DOM-editing patch and snapshot paths, which are no-op CSS property names.
- `getCssFilterFunctionPx(styles.filter, "blur")` became `"blur-sm"`, breaking the inspector's blur sliders.
- The `FrameStatus` value `"outline"` became `"outline-solid"` in the storyboard status order and its consumers.
- Effect-spec keys, user-visible labels (`"Layer blur"`), test names and code comments were rewritten as prose.

Every one of these was outside a `className`. The rule that caught them: a change the tool makes on a line with no `className` on it is a false positive until proven otherwise.

### Default changes the codemod did not make, and what was done about each

| v4 default change | Finding in Studio | Action |
|---|---|---|
| Border color: `gray-200` to `currentColor` | The upgrade tool emits a compatibility shim pinning `border-color` back to `gray-200` on `*`. An audit of every string literal carrying a bare `border`/`border-<side>` found 15 hits: all are either `border-0`, or prose in test names, or already carry an explicit color (an inline `borderColor`, or a variant class supplying `border-emerald-700`/`border-sky-500`/`border-neutral-700`). Every `border:` declaration in `studio.css` names its own color. | **Shim deleted.** Nothing depended on the default, so keeping it would have been dead plumbing. |
| Ring width `3` to `1`, ring color to `currentColor` | Studio has no bare `ring` utility. Every `ring-*` usage names both a width (`ring-1`, `ring-2`) and a color. The four grep hits for a bare `ring` are all the English word in comments. | No change needed. |
| `divide-*` default color | One `divide-y`, and it carries `divide-panel-border-input/60`. | No change needed. |
| `outline-none` semantics | 93 occurrences migrated to `outline-hidden` (79 bare, 14 under a `focus:` variant); bare `outline` migrated to `outline-solid`. | Handled by the codemod, verified by hand. |
| Shadow / radius scale shift | `rounded` to `rounded-sm`, `rounded-sm` to `rounded-xs`, `shadow` to `shadow-sm`, `drop-shadow` to `drop-shadow-sm`, `blur` to `blur-sm`, `backdrop-blur-sm` to `backdrop-blur-xs`. | Handled by the codemod, verified by hand. |
| `bg-opacity-*` removed | Studio has none. | No change needed. |
| `@apply` / the v3 `theme()` function in CSS | `studio.css` uses neither. | No change needed. |
| Pointer cursor on buttons | The shared `Button` and `IconButton` already set `cursor-pointer` explicitly. | No change needed. |
| Tailwind's JS PostCSS-plugin API removed | One test compiled Tailwind in-process to assert `field-sizing: content` is emitted. | Rewritten onto v4's `compile()` API; `postcss` is no longer a Studio dependency. |

## Test plan

Studio's dev server, suite, typecheck and build all run against the new pipeline.

- Unit suite: **427 files passed, 1 skipped; 4743 tests passed, 18 todo.** No dev server was alive from this session during the run.
- `typecheck`: clean.
- `build`: clean, including the `tailwind-preset` subpath still emitting to `dist/styles/tailwind-preset.js`.
- `oxlint`: 0 warnings, 0 errors. `oxfmt --check`: clean (7 files were reformatted because the longer v4 class names push lines past the print width).
- Fallow audit against the base branch: passes.
- Studio load smoke (the same command the CI job runs, against a dev server booted from this branch): `PASS: studio loaded with schema-valid API fixtures and no runtime errors`.
- Emitted stylesheet spot-checked: `.bg-studio-accent`, `.bg-panel-surface`, `.text-panel-text-2`, `.border-panel-border-input`, `.bg-panel-bg-inset` are all present, and `--color-panel-surface:#18181b` is emitted as a real custom property. Alpha on a theme color folds correctly (`.bg-studio-accent\/3` compiles to `#3ce6ac08`).

### Screenshot and computed-style capture

The plan's capture script was run on the base branch and on this branch.

**Screenshots.** Reporting a raw before/after difference here would be misleading, because the capture is not deterministic: two runs of the **same** base-branch code differ by SSIM 0.923 / 0.922 / 0.931 on three of the six states, because the preview canvas and the composition list each race the screenshot. Measured against that noise floor, every before/after delta is at or below same-code run-to-run variation:

| State | Same-code noise | Before vs after |
|---|---|---|
| fresh boot | 0.9949 | 0.9998 |
| inspector design | 0.9234 | 0.9983 |
| renders panel | 0.9220 | 0.9974 |
| canvas context menu | 0.9305 | 0.9304 |
| timeline toolbar | 0.9988 | 0.9986 |
| timeline speed menu | 0.9988 | 0.9977 |

No screenshot difference is attributable to the upgrade.

**Computed-style table.** Four of the five rows are identical. Two cells changed, both reproducible across repeated runs:

| Control | Before | After | Verdict |
|---|---|---|---|
| Timeline toolbar button, background | `rgba(255, 255, 255, 0.08)` | `oklab(0.999994 0.0000455678 0.0000200868 / 0.08)` | Same paint, different serialization. v4 resolves `bg-white/8` through `color-mix(in oklab, ...)` because the color comes from a custom property. That oklab triple is white. The screenshot containing this button matches within the noise floor. |
| Renders Export, font size | `14px` | `14px` | Unchanged (fixed in this PR, see below). |

### The Renders Export font size

The caller's `text-[11px]` duplicated the shared `Button`'s `text-sm` and lost to it under v3 but wins under v4, so the button would have dropped from 14px to 11px. This PR removes the dead `text-[11px]` from the caller, so the button stays at the shipped 14px; a test fails if a font-size utility comes back on it.

## Not covered

- The `@theme` block is a direct translation of the old JS palette. Semantic tokens, the type scale, radius/shadow/easing tokens, and clearing Tailwind's default palette with `--color-*: initial` are a later unit's scope and are deliberately absent.
- The published `./tailwind-preset` JS export still exports the raw v3-shaped preset object. Turning it into a deprecated shim is a later unit.
- `rounded-button` and `shadow-btn-primary` are referenced by the shared `Button` but are not real utilities and emit nothing, on v3 and on v4 alike. Left as-is: they are pre-existing and the token unit owns them.
- Storybook was not part of this capture, so the capture script's Storybook state is skipped.
- The diff is 1016 changed lines, marginally over the 1000-line guideline. It is a single mechanical codemod plus its config swap, and there is no coherent vertical seam to cut: a half-migrated tree does not build.

## Compatibility

Hosts must be on Tailwind v4 to use the exported Studio components. The next release after this lands is a breaking release (0.9.0) and its release notes say so.

## Before

Before: 1-default
Before: 2-tab-Renders
Before: 3-left-Catalog
Before: 4-selected
Before: 5-inspector
Before: 7-export

## After

After: 1-default
After: 2-tab-Renders
After: 3-left-Catalog
After: 4-selected
After: 5-inspector
After: 7-export


somanshreddy (commented):
Reviewed at `927adf45` — the v4 migration mechanics are all correct; LGTM on the mechanics (visual parity of the class renames is the 11-screen sheets + 4968 tests, scoped below).

Verified the four things a Tailwind v4 migration actually turns on, in `packages/studio/src/styles/studio.css` + the build:
- **Content scoping (the security-relevant one) is right.** `@import "tailwindcss" source(none)` disables v4's automatic content detection, and `@source "../**/*.{ts,tsx}"` + `@source "../../index.html"` scope it to Studio's own sources — with the rationale documented ("automatic detection would walk up to the repo root and scan `data/projects`, gitignored user content"). So the build never scans user project content, as claimed.
- **`@theme` tokens complete AND in sync with the published preset.** All studio/panel tokens are defined, and I compared every value against `tailwind-preset.shared.js` (the `./tailwind-preset` subpath's source) rather than trust the "same palette" comment — all ~21 match case-insensitively (studio bg/surface/border/text/muted/accent; panel bg/bg-inset/input/surface/hover/border/border-input/hairline/text-0..5/accent/danger/media `#00e3ff`/container `#f5a623`). The dual-source is documented as transitional (the follow-up theme-token PR merges them) and the comment already warns "update both" — currently consistent.
- **Button cursor preserved via a base rule, not per-element.** `@layer base { button:not(:disabled), [role="button"]:not(:disabled) { cursor: pointer } }` — the complete property approach (covers every button + role=button, correctly excludes disabled), so no button can be missed. The explicit `cursor-pointer` utilities elsewhere are for non-button interactive elements (resize handles, clickable divs).
- **Build rewired correctly.** `@tailwindcss/vite` added and wired (`import tailwindcss from "@tailwindcss/vite"` → `plugins: [react(), tailwindcss(), devProjectApi()]`), `postcss.config.js` removed, `tailwindcss` 3→4.

Scope, honestly: I verified the migration mechanics + the token parity. I did **not** individually re-verify the ~130 files of v3→v4 utility renames (`rounded`→`rounded-sm`, `outline-none`→`outline-hidden`, shadow-scale shifts, etc.) — a wrong rename is a visual regression, and that's what the 11-screen pixel-diff sheets + 4968 passing Studio tests are the backstop for. No findings in the mechanics.

somanshreddy (commented):
Re-reviewed at the new head `6aa94114`. The two-file delta is clean; holding on CI state + one evidence correction.

**Delta (`927adf45..6aa94114`) is sound.** `RenderQueue.tsx` (1 line) pins the Renders-panel Export button size the v4 preflight would have shrunk, and `RenderQueue.test.tsx` (+7) adds a test for it. Addresses the reported shrink; my Tailwind-v4 mechanics + token-parity pass from the prior head still holds over this delta.

**Why I'm not stamping yet — the gate is red at head, though the reds look flaky, not migration defects:**
- `Perf: fps` (and the rest of the Perf suite) fail on a harness **`TimeoutError: Waiting failed: 5000ms exceeded`** in `packages/player/tests/perf/scenarios/02-fps.ts:151` — a measurement-wait timeout, not a "regression exceeds threshold." `player-perf` same family.
- `Detect changes` is flapping (exit 1 at ~21s in one run, success in another); `Tests on windows-latest` and `regression` **failed then passed on re-run** (superseded). So the core suite is green on its latest runs; the noise is infra.
- Net: this reads as flaky CI, not the v4 migration breaking something — but `mergeStateStatus` is still `BLOCKED` with required checks red at head. A clean re-run of the Perf suite + Detect-changes should clear it. I can't stamp over a red gate on the hope it's flaky; get the required checks green and I'll move.

**Evidence correction (per Rames).** The dispatch says "six before/after pairs diffed pixel by pixel by home." Rames has confirmed he never reviewed or visually passed this PR (zero reviews/comments), and my own only review (`5251040704`, COMMENTED) is pinned to the **old** head `927adf45` and explicitly deferred visual parity to the capture sheets + tests. So for this v4 migration, screen-parity is **not** independently verified beyond the PR-body captures — I'm flagging that so it isn't counted as a done check when weighing the stamp. If an independent visual pass is the gating need, say so and I'll go through the body captures directly (with the caveat that 6 sample pairs can't cover the ~130-file utility-rename surface).

somanshreddy (commented):
Correcting my earlier note on this PR's CI. I called the `Detect changes` failure "flaky, a re-run should clear it." That was wrong — it's **deterministic** and a re-run will not clear it.

`ci.yml`'s `Detect changes` runs `node scripts/check-no-main-deletions.mjs --base origin/main`, and it fails on:
```
packages/studio/postcss.config.js
If a deletion is intended, remove this check for that path deliberately.
```
That's the intentional postcss removal this migration makes (v4 replaces postcss with `@tailwindcss/vite`). The guard rejects any deleted path not listed in its `ALLOWED_DELETIONS` map.

**Consequence I misread:** because `Detect changes` fails, the mainline `ci.yml`'s downstream required checks (`Build`, `Test`, `Typecheck`, `Studio: load smoke`, …) report **SKIPPED** — so the migration's mainline validation hasn't actually run against this diff. The green `Tests on windows-latest` / `regression` runs I saw earlier are a **separate** workflow, not the mainline gate; that's what made me read the reds as flaky. (Same mechanism tai just documented on #4132.)

**Fix:** add an entry to `ALLOWED_DELETIONS` in `scripts/check-no-main-deletions.mjs`:
```js
["packages/studio/postcss.config.js", "Tailwind v4 migration: postcss replaced by @tailwindcss/vite"],
```
Push it as a commit; `Detect changes` passes on `synchronize`, and the real downstream checks finally run. (`Perf: fps` failing on a 5s harness timeout is separate and may be genuinely flaky — worth confirming once the mainline checks actually execute.)

So the hold stands, but the action is the whitelist entry, not a re-run. My code pass on the migration mechanics is unchanged.

somanshreddy (commented):
Follow-up, correcting my own hedge above: `Perf: fps` is **not** flaky either. Per Rames's source trace on #4137: the fps scenario gives `isPlaying` 5000ms to flip, but the player's asset wait is bounded by `ASSETS_READY_TIMEOUT_MS = 8000` and `play()` doesn't start until that settles (`_settleAssetsReady` → `if (this._pendingPlay) this.play()`). 8s > 5s, so any run whose assets take over 5s deterministically misses the window. Fix is [#4137](https://github.com/heygen-com/hyperframes/pull/4137) (waits on `assetsready` before pressing play).

So both reds on this PR are known mechanisms, not noise:
1. `Detect changes` → add `packages/studio/postcss.config.js` to `ALLOWED_DELETIONS`.
2. `Perf: fps` → land #4137.

Once both are in and the mainline `ci.yml` downstream checks actually run green, the remaining open item is independent visual parity (the "diffed by home" line wasn't a real pass). Code pass on the migration mechanics unchanged.

somanshreddy (approved):
Approving at `58a3931e` — the two blockers I held on are now actually fixed, and I verified both at CI (not the devbox-green claim that misled on earlier heads).

**Verified fixed:**
- **Detect changes / the deletions guard** — `ALLOWED_DELETIONS` in `scripts/check-no-main-deletions.mjs` now lists both Tailwind-v4-removed configs with reasons: `packages/studio/postcss.config.js` ("postcss replaced by @tailwindcss/vite") and `packages/studio/tailwind.config.js` ("config moves into styles/studio.css via @theme"). That's the exact fix; the guard no longer fires on the intentional removals.
- **CI is genuinely all-green** at this head — no failing or pending checks; `mergeStateStatus` is BLOCKED only on `REVIEW_REQUIRED`. So the earlier `Detect changes` / `Perf: fps` (#4137) / `Format` / windows `studio-engine-cli` reds are all cleared, verified per-check rather than taken from "devbox green."

**Carried from my code review:** the v4 migration mechanics are sound — `@import "tailwindcss" source(none)` with explicit `@source` globs, `@theme` CSS-first tokens, the restored button `cursor: pointer` in `@layer base` (v4 preflight drops it), and the `@tailwindcss/vite` plugin swap for the removed postcss config. Token parity checked.

**Scope of this approval (honest):** I verified the migration mechanics, the deletions-guard fix, and that CI (incl. the 4968-test Studio suite, build, typecheck, perf, windows) is green. I did **not** run an independent pixel pass over the ~130 v3→v4 utility renames — that parity rests on the green Studio suite + the PR-body pixel-diff sheets + Miguel's approval of the one visible change (the Export button). Six sample pairs can't cover 130 renames, and the passing suite + tool-diffs are the stronger backstop; noting it so the visual leg isn't counted as a hand pass by me.

Base of the design-system stack is clear now. #3621 still needs its own `theme.css` format fix before it (and #3622's inherited Preflight) go green.

PR #3621, opened 2026-09-04
feat(studio): move every design value into one theme file

miguel-heygen:
Lands unit U1 (theme file and semantic tokens) of the Studio design-system foundation. Stacked on the Tailwind v4 PR (#3618). No ratchet yet; the color literals this adds live in the token file, which the ratchet excludes.

## What

Lands U1 of the Studio design-system stack: one theme file that owns every design value.

- `packages/studio/src/styles/theme.css` is new and holds the semantic colour palette (background levels, surfaces, input, hover, borders, hairline, text levels 0 to 5, accent, selection, playhead, danger, media, container), radii, control heights, the micro type scale, the two font stacks, menu and popover shadows, easings, motion durations, and the icon size and weight.
- Tailwind's stock palette is cleared with `--color-*: initial`. The stock colours Studio's markup still names are re-declared in a deprecated block; the `studio.*` and `panel.*` names stay as aliases that point at the semantic tokens. Both blocks are removed by the sweep unit that converts the last usage.
- Motion durations ship as `@utility` rules (`duration-press` and siblings) because Tailwind has no duration namespace. Each zeroes itself under `prefers-reduced-motion`, so a caller cannot forget the reduced-motion case.
- Phosphor icons get one default size and weight at the app root, read off the theme. Icons that pass their own size or weight still win.
- The theme is published as `@hyperframes/studio/theme.css`. The JS preset export stays as a deprecated shim for Tailwind v3 consumers.

Stacked on the Tailwind v4 branch; review that one first.

## Why

Studio's design values lived in three places at once: a JS Tailwind preset, an inline `@theme` block, and raw hex and pixel literals in markup and CSS. Two colour scales overlapped, the same grey was spelled four ways, and a value read from CSS could disagree with the same value read from a class. Tailwind v4 emits every theme entry as both a utility and a real CSS custom property, so a single file can serve markup, plain CSS and canvas code without a second copy.

## How

- `@theme static` rather than plain `@theme`: the default behaviour drops theme variables no utility class happens to reference, and several of these tokens are read through `var()` from stylesheets and from JavaScript rather than through a class.
- The type scale is keyed `--text-step-8` through `--text-step-18` because `--text-11` would collide with Tailwind's own size scale. `text-xs`, `text-sm` and `text-lg` keep their Tailwind values, so the sweep can be a 1:1 replacement of the `text-[11px]` style literals.
- Radii are re-declared at Tailwind's own values, so `rounded-md` is unchanged today and the theme file becomes the place to change it tomorrow.
- The icon size and weight are SVG attributes, not styles, so `var()` cannot be handed to the icon library. `readIconTokens` resolves them once at the app root, which keeps the token file the only owner instead of pairing a CSS variable with a JavaScript constant.
- The five font stacks in `studio.css` collapse to `--font-sans` and `--font-mono`. `--font-mono` takes the code editor's stack, so the timeline timecode and the loader detail line now prefer the same monospace family as the editor.
- New in this unit, consumed later: `--color-selection` and `--color-playhead` are reserved and differ from the accent (the playhead and a selected clip both draw in the accent colour today), and `--shadow-menu` and `--shadow-popover` are sized for a near-black canvas rather than Tailwind's light-mode drop shadows. Nothing reads them yet.

## Test plan

- New `theme.test.ts` compiles Studio's real entry stylesheet with Tailwind and asserts: the semantic palette is emitted both as custom properties and as utilities; an alpha modifier on a plain hex resolves through `color-mix`; a stock colour outside the deprecated block no longer compiles; selection, playhead and accent are three distinct values; every micro type step exists and the Tailwind sizes are untouched; a duration utility carries its reduced-motion variant; every deprecated stock-palette entry still equals its upstream Tailwind value; the JS preset shim still agrees with the CSS; and an icon with no props of its own inherits the token size and weight. Each assertion was checked against a deliberately broken theme file first.
- Studio suite: 455 files passed, 1 skipped; 4979 tests passed, 18 todo.
- `typecheck`, `build`, `oxlint`, `oxfmt --check` and the fallow audit are clean.
- Before and after capture with the design-review script, on this branch and on the Tailwind v4 base: the computed-style table is byte-identical, and the six screenshots match apart from a media fixture that loads non-deterministically between runs (the baseline differs from itself the same way, which is why the table is the verdict).
- Exhaustive check beyond the six screenshots: the built CSS bundle was diffed rule by rule against the base. The complete delta is the five font-family rules above, three utilities pulled in by the new test file, and three classes in `components/ui/Button.tsx` that previously did not resolve and now do (`border-border`, `enabled:hover:border-border-strong`, `duration-press`). That component was written against a token vocabulary that did not exist yet; its remaining unresolved classes (`shadow-btn-primary`, `bg-surface-hover`, `bg-accent-red`, `rounded-button`, `ease-standard`) are left for the unit that rebuilds it. Its border therefore moves from the inherited text colour to the border token, which is the one intended rendering change in this PR.

- The stock-palette reset (`--color-*: initial`) lives in a Studio-only `palette-reset.css`, imported before `theme.css`, so a host that imports the exported `theme.css` keeps its own colours. A test compiles `theme.css` alone and asserts a stock colour still resolves; it fails if the reset moves back.
- Live capture of the one surface that renders the shared Button (the Renders panel Export button) at idle, hover and pressed, on the Tailwind v4 base and on this branch: visually identical; the only difference is the transition duration (150 ms to 120 ms). The code editor's monospace surface is likewise unchanged to the eye.

## Not covered

- No panel is migrated onto the semantic names; every existing class keeps working through the deprecated aliases.
- The token gate and the hex ratchet are the next unit, not this one.
- `components/ui/Button.tsx` is not fixed here beyond the classes that now resolve on their own.
- Test files are scanned for class candidates, so a class that appears only in a test still reaches the shipped stylesheet. Pre-existing, three rules' worth, left alone.
- Nothing consumes the selection, playhead, shadow or easing tokens yet.

## Before

Before: Export button idle, hover, pressed on the Tailwind v4 base (left column of the sheet)
Before: code editor monospace surface

## After

After: the same button states on this branch (right column of the sheet above)
After: code editor monospace surface

### DIFF
--- a/packages/studio/package-subpaths.json
+++ b/packages/studio/package-subpaths.json
@@ -13,6 +13,12 @@
       "types": "./dist/styles/tailwind-preset.d.ts",
       "environments": ["bun", "node"]
     },
+    "./theme.css": {
+      "source": "./src/styles/theme.css",
+      "runtime": "./src/styles/theme.css",
+      "types": null,
+      "environments": ["browser", "bun", "node"]
+    },
     "./package.json": {
       "source": "./package.json",
       "runtime": "./package.json",
--- a/packages/studio/package.json
+++ b/packages/studio/package.json
@@ -26,6 +26,7 @@
       "import": "./src/styles/tailwind-preset.ts",
       "types": "./src/styles/tailwind-preset.ts"
     },
+    "./theme.css": "./src/styles/theme.css",
     "./package.json": "./package.json"
   },
   "publishConfig": {
@@ -39,6 +40,7 @@
         "import": "./dist/styles/tailwind-preset.js",
         "types": "./dist/styles/tailwind-preset.d.ts"
       },
+      "./theme.css": "./src/styles/theme.css",
       "./package.json": "./package.json"
     },
     "main": "./dist/index.js",
--- a/packages/studio/src/main.tsx
+++ b/packages/studio/src/main.tsx
@@ -1,7 +1,9 @@
 import { StrictMode } from "react";
 import { createRoot } from "react-dom/client";
+import { IconContext } from "@phosphor-icons/react";
 import { StudioApp } from "./App";
 import { StudioErrorBoundary } from "./components/StudioErrorBoundary";
+import { readIconTokens } from "./styles/iconTokens";
 import { trackStudioEvent } from "./utils/studioTelemetry";
 import "./styles/studio.css";
 
@@ -93,8 +95,12 @@ window.addEventListener("unhandledrejection", (event) => {
 
 createRoot(document.getElementById("root")!).render(
   <StrictMode>
-    <StudioErrorBoundary>
-      <StudioApp />
-    </StudioErrorBoundary>
+    {/* One icon size and weight for the whole app, taken from the theme file.
+        Icons that pass their own size or weight still win. */}
+    <IconContext.Provider value={readIconTokens()}>
+      <StudioErrorBoundary>
+        <StudioApp />
+      </StudioErrorBoundary>
+    </IconContext.Provider>
   </StrictMode>,
 );
--- a/packages/studio/src/styles/iconTokens.ts
+++ b/packages/studio/src/styles/iconTokens.ts
@@ -1,0 +1,27 @@
+import type { IconWeight } from "@phosphor-icons/react";
+
+const WEIGHTS = ["thin", "light", "regular", "bold", "fill", "duotone"] as const;
+
+function asWeight(value: string): IconWeight | null {
+  return (WEIGHTS as readonly string[]).includes(value) ? (value as IconWeight) : null;
+}
+
+/**
+ * Studio's Phosphor defaults, read off `theme.css` so the icon size and weight
+ * have one owner instead of a CSS variable and a JavaScript constant that drift
+ * apart. Both are SVG attributes, not styles, so `var()` cannot be handed to
+ * Phosphor directly and the values are resolved once at the app root.
+ *
+ * A missing or unrecognized value means the stylesheet has not loaded (or the
+ * token was renamed); Phosphor's own defaults are used rather than a second
+ * copy of the token values.
+ */
+export function readIconTokens(root: Element = document.documentElement): {
+  size: string;
+  weight: IconWeight;
+} {
+  const styles = getComputedStyle(root);
+  const size = styles.getPropertyValue("--icon-size").trim();
+  const weight = asWeight(styles.getPropertyValue("--icon-weight").trim());
+  return { size: size || "1em", weight: weight ?? "regular" };
+}
--- a/packages/studio/src/styles/studio.css
+++ b/packages/studio/src/styles/studio.css
@@ -1,4 +1,5 @@
 @import "tailwindcss" source(none);
+@import "./theme.css";
 
 /*
  * Explicit sources: Studio's own TypeScript and its shell HTML. Automatic
@@ -8,33 +9,6 @@
 @source "../**/*.{ts,tsx}";
 @source "../../index.html";
 
-@theme {
-  --color-studio-bg: #0a0a0a;
-  --color-studio-surface: #141414;
-  --color-studio-border: #262626;
-  --color-studio-text: #e5e5e5;
-  --color-studio-muted: #737373;
-  --color-studio-accent: #3ce6ac;
-
-  --color-panel-bg: #0c0c0e;
-  --color-panel-bg-inset: #121214;
-  --color-panel-input: #161618;
-  --color-panel-surface: #18181b;
-  --color-panel-hover: #27272a;
-  --color-panel-border: #1e1e1e;
-  --color-panel-border-input: #27272a;
-  --color-panel-hairline: #1a1a1c;
-  --color-panel-text-0: #fafafa;
-  --color-panel-text-1: #e4e4e7;
-  --color-panel-text-2: #a1a1aa;
-  --color-panel-text-3: #71717a;
-  --color-panel-text-4: #52525b;
-  --color-panel-text-5: #3f3f46;
-  --color-panel-accent: #3ce6ac;
-  --color-panel-danger: #ef4444;
-  --color-panel-media: #00e3ff;
-  --color-panel-container: #f5a623;
-}
 
 /*
  * Studio is a dark-only UI — pin the user-agent color scheme to dark so that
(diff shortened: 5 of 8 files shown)


somanshreddy (commented):
Reviewed at `ab4171b7`. The design is sound and the host-palette-preservation mechanism is genuinely correct — but there are **two red checks at this head** the re-pin's "gated, ready" didn't mention, one trivially this PR's to fix and one that looks like a gating quirk.

## Confirmed sound (the core of the PR)
- **Host-palette preservation works, and the test proves it.** Splitting `@theme static { --color-*: initial }` into `palette-reset.css` (imported only by Studio's own `studio.css`, before `theme.css`) means a host importing just `theme.css` never sees the reset. The test at `theme.test.ts:549` genuinely exercises this — it builds `@import "tailwindcss"; @import "./theme.css"` and asserts a stock `bg-teal-500` still compiles *and* `text-accent` resolves. It would fail if the reset lived in `theme.css`. Real, failing-capable coverage.
- `@theme static` is the right call — it keeps unused tokens in `:root` so canvas/JS (`readIconTokens` via `getComputedStyle`) can read `--icon-size`/`--icon-weight`, which a tree-shaking `@theme` would drop.
- `files: ["src","dist"]` ships the source `theme.css`, so the new published `./theme.css` subpath resolves in the tarball; `check:package-subpaths` passes at head.
- `@utility duration-*` zeroing itself under `prefers-reduced-motion` is a nice fail-safe.

## Blocking — red gates at `ab4171b7`
1. **Preflight (lint + format) is RED — `format:check` fails on `packages/studio/src/styles/theme.css`** (this PR's central new file). `oxlint` and every `check:*` script pass; it's purely a prettier/format violation. Fix is mechanical: `bun run format` and push. Whether or not Preflight is a required gate, unformatted code that fails the format check shouldn't merge. (A later CI wave only *skipped* Preflight — a skip doesn't fix the formatting.)

2. **preview-regression is RED — but it reads as a gating quirk, not this PR's code.** The job log shows `PREVIEW_PARITY_RESULT: skipped` → `Preview parity check failed` → `exit 1`. "Preview parity" legitimately skipped (no preview/runtime change detected for a CSS/theme PR), and preview-regression treats a *skipped* parity as a failure. That would hit any CSS-only PR in this stack. @loop/@infra: is this expected, or should preview-regression auto-pass when parity skips? I can't clear it from the review side — flagging so it isn't merged red on a misread.

## Non-blocking
3. **The JS-preset drift guard is narrower than its comment.** `tailwind-preset.shared.js` is `@deprecated` with "Values are byte-identical; `theme.test.ts` fails if they drift." But the parity test (`theme.test.ts:613`) checks only **2** values — `studio.accent` and `panel.surface` — while the preset defines **24** color tokens. Drift in any of the other 22 (e.g. `--color-bg-0` vs preset `studio.bg`) passes the test while the comment promises otherwise. Either loop the assertion over every shared key, or soften the comment to name the two it actually pins.
4. **Completeness of the re-declared stock-palette subset rests on visual capture — confirm the captures cover it.** `palette-reset.css` blanks all `--color-*`; `theme.css` re-adds a fixed ~60-entry subset "markup still names." The unit tests guard *drift* (values match upstream) and that *unused* colors are dropped — but nothing guards the subset is *complete*: a stock color used somewhere in Studio markup but omitted would render unset. Stock palette colors (amber/blue/cyan/red/sky/…) live on badge/status/media-type surfaces, not the "button and code-surface" captures cited. How was the subset derived — a grep of Studio source for `(bg|text|border|ring|…)-(amber|blue|…)-\d+`? If so I'm satisfied; if it was hand-listed, it's worth a mechanical check.

Once the format gate is green (1), the preview-regression gate is understood (2), and 3/4 are addressed or confirmed, this is a clean approve — the token consolidation itself is well done.

somanshreddy (commented):
Re-checked at the restacked head `44b595d2`. The restack is content-identical as described (range-diff clean), which means my earlier blocker rode along unfixed: **`Preflight (lint + format)` still fails on `packages/studio/src/styles/theme.css`** — the same prettier violation, carried byte-for-byte through the restack ("Format issues found in above 1 files: theme.css"). `preview-regression` is also still red (the parity-skipped gating quirk, not this PR's code). Fix is still one command: `bun run format` on `theme.css`, then push.

Also note this is stacked on #3618, which I'm holding at `6aa94114` (red gate + unverified visual parity), so it's merge-order-blocked regardless.

My earlier substantive review (host-palette mechanism verified sound, drift-guard/preset nits) stands unchanged — this is purely the format gate re-affirmed.

somanshreddy (commented):
Re-checked at `4b0ef5a1` (replay onto main after #3618 merged). Two blockers now — the format one is unchanged, and the replay surfaced a real packaging defect I hadn't seen on earlier heads.

1. **`theme.css` is still unformatted** — `Preflight`/`Format` fail with "Format issues found in above 1 files: `packages/studio/src/styles/theme.css`". The `bun run format` fix I've flagged each round still isn't applied (range-diff-identical replay carried the unformatted file forward).

2. **NEW — `Build` fails on `verify:packed-manifests` (2m36s, a real failure, not a cascade).** The consumer-smoke packs `@hyperframes/studio` and a Node consumer trips:
   `TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".css" for .../node_modules/@hyperframes/studio/src/styles/theme.css` → `node consumer-smoke.mjs` exits 1.
   So the packed package exposes `theme.css` such that a Node consumer resolves and tries to **import it as a module**, and Node has no `.css` loader. `theme.css` is a Tailwind asset meant to be pulled via CSS `@import` (or the host's Tailwind pipeline), not `import`-ed in JS. Fix is at the export surface: the `./theme.css` subpath / package `exports` shouldn't be a JS-import target the consumer-smoke resolves as a module (e.g. keep it a plain asset path, or exclude it from the smoke's import graph). This is real and independent of the format nit.

My substantive review (host-palette mechanism, `@theme static`, drift-guard nits) still stands. Both of the above need fixing before a stamp: `bun run format` for (1), and the export/consumer-smoke treatment of `theme.css` for (2). Re-verify me at the new head once `Format` and `Build` are green.

somanshreddy (approved):
Approving at `b35b564f` — both findings I held on are fixed and I verified the whole run is green (0 `failure` conclusions, 0 pending).

- **Format green** — `theme.css` is now formatted (the `bun run format` fix). `Format` + `Preflight` pass.
- **Build green** — the packaging defect is fixed the way I suggested: the `theme.css` export is declared browser-only, so `verify:packed-manifests`' consumer-smoke no longer tries to `import` a `.css` file in Node (`ERR_UNKNOWN_FILE_EXTENSION` is gone). `Build` passes 2m59s.
- Whole run settled: no failures, nothing pending (Windows lanes finished green).

My substantive review carries: the host-palette split (`palette-reset.css` separate so a host importing `theme.css` keeps its palette, proven by the :549 test), `@theme static` for canvas/JS reads, and the drift-guard are all sound. `merge` is BLOCKED only on `REVIEW_REQUIRED`.

Approving — clean now.

PR #3622, opened 2026-09-04
feat(studio): shared Button, IconButton, Tooltip and Tabs primitives

miguel-heygen:
Lands unit U4 (cn plus the first four primitives) of the Studio design-system foundation. Stacked on the theme PR (#3621) and merges the Base UI PR (#3619), so the diff shows those two until they land. No ratchet yet; no color literal added. Bundle delta against the merge point: about 45.8 KB gzipped, against a 100 KB budget for the primitive set (raised from the plan's 60 KB after measuring the shared floating layer; the decision is recorded here).

## What

Unit U4 of the Studio design-system plan: four shared primitives that consume design tokens only, plus the configured class merger they share.

- `ui/cn.ts` — one `cn` instance, configured with the two class families the theme invents: the `text-step-*` type scale joins the font-size group, and `ctl` / `ctl-sm` / `ctl-lg` join the sizing groups.
- `ui/Button.tsx` and `ui/IconButton.tsx` — native buttons on the three control heights, the radius scale, the semantic palette and the motion tokens. Variants primary, secondary, danger, ghost; sizes sm, md, lg.
- `ui/Tooltip.tsx` — the same props as before (`label`, `delay`, `side`), now on Base UI. No call site changed.
- `ui/Tabs.tsx` — roving tabindex, arrow keys, Home and End, `aria-controls`, and a `data-tab-id` on each tab.

Every primitive also honours a `data-preview-state` attribute (`hover`, `active`, `focus`) whose selectors carry the same declarations as the real state. It is read by CSS only and changes no behaviour, so a story or a screenshot can show a state without a pointer.

## Why

Studio's controls were drifting because nothing held them together. Two of the symptoms this unit removes:

- **Four classes styled nothing.** `rounded-button`, `shadow-btn-primary`, `bg-accent-red` and `ease-standard` were never defined. Tailwind does not error on an undefined class, it emits nothing, so the Renders Export had been rendering with no radius and no shadow and looked merely "a bit off".
- **A caller could not override a primitive.** With plain string concatenation, a caller's class and the size's class both survive into the class list and the stylesheet's order picks the winner. That is not something a call site can reason about.

## How

`cn` merges variant, size, state and caller classes in that order, so the last one named wins. Its config is patterns, not enumerations, so a new type step or control height in the theme file needs no edit beside it.

Tooltip and Tabs are Base UI (per-component entry points, `render` prop rather than `asChild`). Two accessibility details stay in the wrapper because Base UI 1.7.0 does not provide them: the popup's `role="tooltip"` and id, and `aria-describedby` on the trigger while open. That is why Tooltip controls the open state itself. Escape dismissal does come from the library, and a test holds it to that.

The Tooltip trigger renders the caller's own element instead of wrapping it. The old wrapper used `display: contents`, which has no box for a positioner to anchor to; that is why the old implementation had to reach into the wrapper's first child to measure it.

## Test plan

`bun run --cwd packages/studio test` — 434 files, 4781 tests green (was 430 / 4761 before this unit; no Studio dev server of this branch's own was running).

New tests, all through the public component API on plain `react-dom/client` + `act` under happy-dom:

- **Class resolution.** Renders every variant and size, reads the class list off the real DOM node, and compiles Studio's actual stylesheet with those classes as candidates. Any class that produces no selector fails and is named. A companion test asserts `rounded-button` and `shadow-btn-primary` still come back unresolved, so the check cannot pass vacuously.
- A caller's `text-step-11` on a medium Button beats the size's own type class.
- A disabled Button does not fire its click handler and reports `aria-disabled`.
- Every `data-preview-state` class has a matching real-state class and the reverse, per variant, per state. Verified non-vacuous by deleting one preview class and watching it fail.
- A Button with `data-preview-state="hover"` still fires its click handler.
- Reduced motion: the Button names `duration-press`, and the compiled `duration-press` utility zeroes its duration under `prefers-reduced-motion`.
- Tooltip opens on focus and on hover, closes on Escape, and wires `aria-describedby` to the popup's id.
- Tabs: ArrowRight moves selection and focus, Home and End jump, the visible panel is `aria-labelledby` its tab, and every tab carries `data-tab-id`.
- Hotkey classification: Button and a Tab are read by `isTypingTarget` and the playback-shortcut selector exactly as a plain header `<button>` is.

Typecheck, `oxlint` and `oxfmt --check` clean. `fallow audit --base origin/main --fail-on-issues` passes; the primitive barrel is registered as a fallow entry point, since panels move onto the primitives folder by folder and a primitive exists before its first consumer.

**Screenshots and computed-style table** (captured on the merge point before the change, and again after):

| Header Export | 28px → 28px | 6px → 6px | 11px → 11px | accent → accent |
| Renders Export | 32px → **28px** | **0px → 6px** | **14px → 12px** | **white → accent** |
| Inspector metric input | 17px | 0px | 11px | unchanged |
| Timeline toolbar button | 28px | 6px | 16px | unchanged |
| Menu item | 29px | 0px | 11px | unchanged |

The two Export buttons now match on height, radius and fill; the Renders one goes from 14px to the `md` size's 12px type (the header one stays 11px). The Renders one moved because `size="md"` is a real control height now and its radius class finally resolves. Everything else in the table is unchanged, and no consumer was edited in this PR.

**Bundle delta.** Gzipped Vite output (JS + CSS): 1,514,146 bytes before, 1,561,089 after. **+46,943 bytes, about 45.8 KB gzipped**, for Base UI's tooltip and tabs plus `cn`. Reported as measured, with no scope trimmed to fit: this is under the 60 KB stop condition, but it is most of the budget that U4, U5 and U6 share, so the next two units should re-measure early.

## Before

Before: Renders panel Export button idle, hover, pressed (left column) on the Tailwind v4 base

## After

After: the same button on the shared Button (right column of the sheet)
After: Renders panel

This is an intended design change: the Renders Export button was a white square button 32px high; on the shared Button it is the accent-green primary with rounded corners, 28px high with 12px type, matching the header Export's look. The tokens the primitive names now resolve, so the rendering follows the theme file.

## Not covered

- No consumer migration. The six files that already import Button render identically except for the radius and fill that now resolve, and the height change described above. Moving the header and the panel tab strips onto these primitives is a later unit.
- No Storybook and no stories; `data-preview-state` is the hook they will use.
- No Input, Select, Slider, Toggle, Popover or Menu; those are later units.
- Positioning, flip and portal placement are not asserted in unit tests: happy-dom has no layout. They are covered by the browser screenshot run.
- Hover-open latency and the reduced-motion look are asserted through classes and compiled CSS, not through a rendered frame.
- The tooltip bubble anchors to its child when it opens. Swapping the child's element type while a tooltip is open leaves it misplaced until it reopens; every current caller uses a stable `<button>` or `<a>`.
- Tooltip tests run in a DOM without layout, so the anchoring itself was checked in a real browser and is not asserted by a test.

## Restack after the Tailwind v4 merge

- Replay: `git range-diff` of the old tip against the new one shows 8 of the 9 own commits identical (`=`), 1 different (`!`), plus one new commit.
- The one different commit (shared primitives): a conflict in `packages/studio/package.json` and `bun.lock` where main had removed the `dompurify` dependency and this PR adds `cn`. Resolved by keeping only `cn`. `dompurify` is absent from the lock, the manifest and all imports. The lock also carries `@base-ui/react` and its dependencies, which this PR already adds for Tabs and Tooltip.
- New commit: `Button` children now render as direct flex items so a label plus a shortcut get the button gap (same one-line fix as the standalone Button PR, with a two-child test).

### Button gap, before and after

Before: label and shortcut touching, measured gap 0px

After: label and shortcut spaced by the button gap, measured gap 6px

### DIFF
--- a/.fallowrc.jsonc
+++ b/.fallowrc.jsonc
@@ -4,6 +4,10 @@
     // Vitest collects .test.mjs, but fallow's default test-entry detection only
     // knows the .ts/.tsx spellings, so this one reads as an unreachable file.
     "packages/studio/scripts/*.test.mjs",
+    // The KTD6 dismiss-spike fixture. scripts/menu-dismiss-spike.mjs injects it
+    // into a running Studio by importing it from the Vite dev server by URL, so
+    // no TypeScript import reaches its mount function.
+    "packages/studio/src/components/ui/menuDismissSpikeMount.tsx",
     "packages/producer/src/**/*.test.ts",
     "packages/aws-lambda/src/**/*.test.ts",
     "packages/gcp-cloud-run/src/**/*.test.ts",
--- a/.fallowrc.jsonc
+++ b/.fallowrc.jsonc
@@ -8,6 +8,12 @@
     // into a running Studio by importing it from the Vite dev server by URL, so
     // no TypeScript import reaches its mount function.
     "packages/studio/src/components/ui/menuDismissSpikeMount.tsx",
+    // The shared-primitive barrel. Studio's panels move onto Button, IconButton,
+    // Tabs and Tooltip folder by folder, so a primitive is exported here before
+    // its first consumer exists. Treating the barrel as an entry point is what
+    // the alternative (an inline suppression per export, deleted again a week
+    // later) was pretending to be.
+    "packages/studio/src/components/ui/index.ts",
     "packages/producer/src/**/*.test.ts",
     "packages/aws-lambda/src/**/*.test.ts",
     "packages/gcp-cloud-run/src/**/*.test.ts",
--- a/.fallowrc.jsonc
+++ b/.fallowrc.jsonc
@@ -1,6 +1,9 @@
 {
   "$schema": "https://raw.githubusercontent.com/fallow-rs/fallow/main/config-schema.json",
   "entry": [
+    // Vitest collects .test.mjs, but fallow's default test-entry detection only
+    // knows the .ts/.tsx spellings, so this one reads as an unreachable file.
+    "packages/studio/scripts/*.test.mjs",
     "packages/producer/src/**/*.test.ts",
     "packages/aws-lambda/src/**/*.test.ts",
     "packages/gcp-cloud-run/src/**/*.test.ts",
--- a/.fallowrc.jsonc
+++ b/.fallowrc.jsonc
@@ -1,6 +1,13 @@
 {
   "$schema": "https://raw.githubusercontent.com/fallow-rs/fallow/main/config-schema.json",
   "entry": [
+    // Vitest collects .test.mjs, but fallow's default test-entry detection only
+    // knows the .ts/.tsx spellings, so this one reads as an unreachable file.
+    "packages/studio/scripts/*.test.mjs",
+    // The KTD6 dismiss-spike fixture. scripts/menu-dismiss-spike.mjs injects it
+    // into a running Studio by importing it from the Vite dev server by URL, so
+    // no TypeScript import reaches its mount function.
+    "packages/studio/src/components/ui/menuDismissSpikeMount.tsx",
     "packages/producer/src/**/*.test.ts",
     "packages/aws-lambda/src/**/*.test.ts",
     "packages/gcp-cloud-run/src/**/*.test.ts",
--- a/bun.lock
+++ b/bun.lock
@@ -315,6 +315,7 @@
       "name": "@hyperframes/studio",
       "version": "0.8.27",
       "dependencies": {
+        "@base-ui/react": "1.7.0",
         "@codemirror/autocomplete": "^6.20.1",
         "@codemirror/commands": "^6.10.3",
         "@codemirror/lang-css": "^6.3.1",
@@ -513,6 +514,10 @@
 
     "@balena/dockerignore": ["@balena/dockerignore@1.0.2", "", {}, "sha512-wMue2Sy4GAVTk6Ic4tJVcnfdau+gx2EnG7S+uAEe+TWJFqE4YoWN4/H8MSLj4eYJKxGg26lZwboEniNiNwZQ6Q=="],
 
+    "@base-ui/react": ["@base-ui/react@1.7.0", "", { "dependencies": { "@babel/runtime": "^7.29.2", "@base-ui/utils": "0.3.2", "@floating-ui/react-dom": "^2.1.9", "@floating-ui/utils": "^0.2.12", "use-sync-external-store": "^1.6.0" }, "peerDependencies": { "@date-fns/tz": "^1.2.0", "@types/react": "^17 || ^18 || ^19", "date-fns": "^4.0.0", "react": "^17 || ^18 || ^19", "react-dom": "^17 || ^18 || ^19" }, "optionalPeers": ["@date-fns/tz", "@types/react", "date-fns"] }, "sha512-j+8QjX44C32jrXD/qyEAGpFr70FRpGL2CY61mQd9nBPWN737CK0xxD1ceJ055rW4RtdvFDT1e7otzdlfxvsYug=="],
+
+    "@base-ui/utils": ["@base-ui/utils@0.3.2", "", { "dependencies": { "@babel/runtime": "^7.29.2", "@floating-ui/utils": "^0.2.12", "reselect": "^5.2.0", "use-sync-external-store": "^1.6.0" }, "peerDependencies": { "@types/react": "^17 || ^18 || ^19", "react": "^17 || ^18 || ^19", "react-dom": "^17 || ^18 || ^19" }, "optionalPeers": ["@types/react"] }, "sha512-oWy1aq/I2GmYjpl4PhEAhzflF8VPGKgZeq0xAWTbfD5KBWyxcN0ZP2+WHSUm/5Z6lVMBDLReLcoXwSYoRc/zNQ=="],
+
     "@bcoe/v8-coverage": ["@bcoe/v8-coverage@1.0.2", "", {}, "sha512-6zABk/ECA/QYSCQ1NGiVwwbQerUCZ+TQbp64Q3AgmfNvurHH0j8TtXa1qbShXA6qqkpAj4V5W8pP6mLe1mcMqA=="],
 
     "@bramus/specificity": ["@bramus/specificity@2.4.2", "", { "dependencies": { "css-tree": "^3.0.0" }, "bin": { "specificity": "bin/cli.js" } }, "sha512-ctxtJ/eA+t+6q2++vj5j7FYX3nRu311q1wfYH3xjlLOsczhlhxAg2FWNUXhpGvAw3BWo1xBcvOV6/YLc2r5FJw=="],
@@ -672,6 +677,14 @@
     "@fallow-cli/win32-arm64-msvc": ["@fallow-cli/win32-arm64-msvc@2.75.0", "", { "os": "win32", "cpu": "arm64" }, "sha512-HbIXLbte8pzNX0XlbyRlRQA2+y8kfJ4O6XxB2VR2dE6q5LrNhG4fl0QRR/P9x8r3CNn9fw/25kk+WxULfbHu1w=="],
 
     "@fallow-cli/win32-x64-msvc": ["@fallow-cli/win32-x64-msvc@2.75.0", "", { "os": "win32", "cpu": "x64" }, "sha512-h8W+qEOPvyolBQO2Y0H4/pz2XXw/1yHA8/XRPo3/tYRovmN2zoNjDrvaGjjodzFt6fBVLus+T0uo+U4pMgOu/w=="],
+
+    "@floating-ui/core": ["@floating-ui/core@1.8.0", "", { "dependencies": { "@floating-ui/utils": "^0.2.12" } }, "sha512-0CIZ5itps/8x7BG8dEIhs53BvCUH2PCoogtakwRTut+Arm58sJooJ0AuZhLw2HJYIR5cMLNPBSS728sPho2khQ=="],
+
+    "@floating-ui/dom": ["@floating-ui/dom@1.8.0", "", { "dependencies": { "@floating-ui/core": "^1.8.0", "@floating-ui/utils": "^0.2.12" } }, "sha512-yXSrzeHZBTZadLOlfyhCkJHNeLJnHRnRInwdZ40L7ZiaAtrBwoYlsDrX3v5zB1Utk7CLfzcOVnVVWoXEky7Ceg=="],
+
+    "@floating-ui/react-dom": ["@floating-ui/react-dom@2.1.9", "", { "dependencies": { "@floating-ui/dom": "^1.8.0" }, "peerDependencies": { "react": ">=16.8.0", "react-dom": ">=16.8.0" } }, "sha512-JDjEFGCpImxDCA7JJKviA0M9+RtmJdj0m/NVU5IMgBK+AmZouAQQ7/+2GLH0GXXY0YMw9oXPB8hKdbPYg5QLYg=="],
+
+    "@floating-ui/utils": ["@floating-ui/utils@0.2.12", "", {}, "sha512-HpCo8tmWzLVad5s2d19EhAz5zqrrQ6s69qd6moPMQvkOuSwDT1YgRfWSVuc4ennqrgv3OHppiOGMQ7oC13yIww=="],
 
     "@fontsource/archivo-black": ["@fontsource/archivo-black@5.2.8", "", {}, "sha512-3zNj/o9LzWyDl/UEpY5IOHpAQyUtFr3hQaFS7NSKwCLLkXOfH/CMCt1L2b2Z+OF25OURtOYenCadgAebALz7/A=="],
 
@@ -1913,6 +1926,8 @@
 
     "require-from-string": ["require-from-string@2.0.2", "", {}, "sha512-Xf0nWe6RseziFMu+Ap9biiUbmplq6S9/p+7w7YXP/JBHhrUDDUhwa+vANyubuqfZWTveU//DYVGsDG7RKL/vEw=="],
 
+    "reselect": ["reselect@5.3.0", "", {}, "sha512-XGoLeRAVzUTcJ1qkxPQhDJyIZ5d6zzZD9nT7AEZOaaU9UbWclhycElmhO+VD5bFeLuzhPBaOV2oXC8uG35ZSpg=="],
+
     "resolve-from": ["resolve-from@5.0.0", "", {}, "sha512-qYg9KP24dD5qka9J47d0aVky0N+b4fTU89LN9iDnjB5waksiC49rvMB0PrUJQGoTmH50XPiqOvAjDfaijGxYZw=="],
 
     "resolve-pkg-maps": ["resolve-pkg-maps@1.0.0", "", {}, "sha512-seS2Tj26TBVOC2NIc2rOe2y2ZO7efxITtLZcGSOnHHNOQ7CkiUBfw0Iw2ck6xkIhPwLhKNLS8BO+hEpngQlqzw=="],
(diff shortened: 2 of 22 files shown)


somanshreddy (commented):
Reviewed at `986c8c87`. **The code is clean and the claims check out; the hold is entirely downstack, not in this PR.**

## Verified sound (this PR's own code)
- **Tooltips on disabled controls — correctly solved.** `Tooltip.tsx` renders the trigger as a box-less `display:contents` `<span className="contents">` wrapper, so a disabled child still receives hover/focus (the span isn't disabled) while adding no layout box; the tooltip surface is `pointer-events-none` and `aria-describedby` is wired when open. That's the right pattern for the "disabled elements don't fire pointer events" problem — likely one of the two blockers.
- **Token usage is consistent** — `Button.tsx`/`IconButton.tsx` are entirely on the theme tokens (`bg-accent`, `text-bg-0`, `bg-hover`, `text-text-*`, `border-border-strong`, `bg-danger`, `h-ctl*`, `rounded-sm`, `ease-out-quint`/`duration-press`, `outline-accent`), no hardcoded hex; `disabled:opacity-40`/`enabled:hover:` correctly gate state, and the `data-[preview-state=…]` variants cover Studio's preview render. `oxlint` reports 0/0.

## Why I'm not stamping — every blocker is downstack
- **The red `Preflight` is inherited from #3621, not this PR.** It fails on `packages/studio/src/styles/theme.css` (the same unformatted file I'm holding #3621 on), which lives in this PR's base branch, not its own diff — #3622's own files pass lint. It clears the moment #3621's `theme.css` is `bun run format`-ed. `player-perf`/`preview-regression` (3s) are the usual gating-cascade, not code.
- **Merge-order:** stacked on #3621 (held on the format issue) → #3618 (held on the `postcss.config.js` `ALLOWED_DELETIONS` entry + #4137 + an independent visual pass). #3622 can't merge before those, and it'll rebase when #3621's fix lands (so this exact head will need a fresh stamp then).

## Couldn't verify
No reviews exist on this PR (empty `reviewDecision`), so the "three independent reviews" and "captures viewed by home" lines aren't second-party-confirmed here; the Export-button visual change is Miguel's product approval per the dispatch.

Net: #3622 itself is good work — unblock it by formatting #3621's `theme.css` (which also clears this PR's inherited `Preflight`) and clearing the stack; I'll re-stamp on the rebased head once it's green.

somanshreddy (dismissed):
Approving at `86c9cee1` — replayed on the now-fixed #3621, and its blocker (the inherited `theme.css` Preflight) is cleared.

- CI green: 0 `failure` conclusions, 0 pending; `Preflight` passes now that #3621's `theme.css` is formatted upstream. `mergeStateStatus` is CLEAN.
- My code review carries unchanged: the disabled-tooltip fix is correct (`display:contents` box-less wrapper span so a disabled control still receives hover; tooltip `pointer-events-none`; `aria-describedby` wired), and Button/IconButton are fully on the theme tokens (no hardcoded hex; `disabled:`/`enabled:hover:` gated). `oxlint` clean.

Merges after #3621 per the stack order. Approving.

somanshreddy (commented):
Re-verified the rebase at `0939a8fb` (my prior approval reset on the push). The one thing the rebase risked — the manual Button.tsx conflict between main's #4159 and this PR — is resolved correctly:

- `{children}` renders **directly** (no wrapper span), so #4159's gap fix (label/kbd as direct flex items the `gap-*` can space) is preserved.
- The leading icon slot is `leading = <span className="shrink-0">{icon}</span>` (icon-only), so this PR's icon slot is preserved too.

Both hunks intact — neither the unwrapped-children fix nor the icon slot was dropped. My earlier review carries (disabled-tooltip `display:contents` wrapper, token usage, oxlint clean).

CI is freshly triggered (0 `failure` conclusions, 16 pending). I'll re-approve once this head's content-validating checks (`Preflight` + studio tests) land green — the conflict resolution itself is verified.

somanshreddy (approved):
Re-approving at `0939a8fb`. The rebase conflict I re-verified (Button.tsx keeps main's #4159 unwrapped `{children}` AND this PR's `<span shrink-0>{icon}</span>` leading slot — both hunks intact), and the content-validating run is now green: `Fallow`, `Preflight`, `Test`, `Tests on windows-latest: studio-core` all pass, 0 `failure` conclusions (one peripheral check still finishing). My earlier review carries (disabled-tooltip `display:contents` wrapper, token usage, oxlint). Approving.

terencecho (approved):
### APPROVING at `eddcb73f` — single-commit docstring shrink over Somansh's approval at `91f724df`

Confirmed via `gh api compare 91f724df...eddcb73f`: 1 ahead / 0 behind, one commit `eddcb73f257e` "docs(studio): shorten the Tooltip header comment to fit the comment gate", one file touched (`Tooltip.tsx`, +2/-3, no runtime change). Dispatch attribution said "the lane's bot" but the actual approval is Somansh's review `5255231179` at `91f724df` — the SHA lines up so this is just phrasing; substance is the human's approval.

## What's actually in `eddcb73f` vs `91f724df`

```
docs(studio): shorten the Tooltip header comment to fit the comment gate
 packages/studio/src/components/ui/Tooltip.tsx | +2/-3
```

Read the hunk — pure comment-block rewrite (header comment cut from 5 lines to 3 lines to satisfy the comment-length rule; component body unchanged). Zero behavioural delta. Somansh's earlier approval scope (disabled-tooltip `display:contents` wrapper, token usage, oxlint conformance, Button.tsx conflict resolution with #4159's `{children}` unwrap coexisting with this PR's `<span shrink-0>{icon}</span>` leading slot) all carries — nothing in that scope moved.

## CI

All checks green at `eddcb73f` — CI, CodeQL (js/py/actions), Preflight, Format, Lint, Typecheck, Build, Fallow audit, Tests on windows-latest (studio-core + studio-engine-cli), Player perf (load/fps/scrub/drift/parity), preview-regression, all 9 regression-shards, Producer unit + integration, SDK unit + contract + smoke, CLI npx shim on ubuntu/macos/windows, Studio load smoke + timeline viewport gate, CLI smoke, Smoke global install, Test. Skills/codex jobs SKIPPED (path filter).

## Stamp mechanics

hf-oss `require_last_push_approval=true` — Somansh's `5255231179` at `91f724df` was dismissed by the push to `eddcb73f`. This APPROVE binds to `eddcb73f`.

Miguel `miguel-heygen` on trusted stamp list; stamping to merge per dispatch.

— Review by tai (pr-review)

jrusso1020 (approved):
**APPROVE** at `eddcb73f257eb60d4393704a1bcf4c7f0819a4f4`.

### The delta since the prior approval, verified

The stamp request said the only change since the approval at `91f724df` is one comment header shortened. That holds exactly — `compare/91f724df...eddcb73f` is **one commit, one file, +2/-3**, entirely inside the `Tooltip.tsx` header block:

```diff
- * Tooltip — Base UI's tooltip wearing Studio's tokens, same props as before.
- * The trigger is a `display: contents` wrapper (no layout box) so a disabled
- * control still gets hover; the bubble anchors to the child, which has the box.
+ * Tooltip: Base UI's tooltip in Studio tokens. The trigger is a box-less
+ * `display: contents` wrapper so a disabled control still gets hover.
```

Zero runtime change, so the prior approval's scope carries. (For the record, that approval is [`5255231179`](https://github.com/heygen-com/hyperframes/pull/3622#pullrequestreview-5255231179) by **somanshreddy**, a human reviewer — tai has already noted the same attribution nit in the dispatch, so I won't belabour it.)

**One nit on that very hunk:** the clause that got cut, *"the bubble anchors to the child, which has the box"*, is the rationale for the least obvious line in the file —

```tsx
anchor={() => boxRef.current?.firstElementChild ?? null}
```

Anchoring to `firstElementChild` rather than the ref itself only makes sense once you know the wrapper is `display: contents` and therefore has no box to anchor to. That is the sentence I'd have kept if the gate forced a cut. Not blocking, and not worth another push on its own.

### Reading the whole change, not just the delta

Tokens check out: every class the new primitives use resolves in `styles/theme.css` — `accent`, `bg-0`, `text-0…3`, `hover`, `danger`, `border-strong`, `surface-alt`, `panel-accent`, `ease-out-quint`, `duration-press`, `duration-hover`, and the `text-step-*` scale. The `cn` config teaching tailwind-merge the `text-step-*` and `ctl` scales is the right call — without it a caller's `h-10` would stack with `h-ctl` instead of replacing it.

Three things I'd flag as intent-confirmation rather than defects, all safe to answer in a follow-up because **nothing consumes these yet** (the old `IconButton` still carried `fallow-ignore-next-line unused-export`):

1. **The control scale dropped one step.** `--spacing-ctl-sm/ctl/ctl-lg` are 24/28/32, so Button goes 28/32/36 → 24/28/32 and IconButton goes min-28/32/36 → 24/28/32. The deleted header encoded `ux-fitts-target-size: min 32px hit target`, and the deleted `iconSizeStyles` had `md: 32px — minimum recommended`. The new default (`size="md"`) is 28px. 24 still clears WCAG 2.2 SC 2.5.8's 24×24 floor, so this isn't an accessibility failure — but it does retire a rule the file used to state out loud, silently. Worth one line in the header saying the 32px rule was intentionally replaced by the token scale, so the next person doesn't reinstate it.
2. **`min-w-*`/`min-h-*` became `size-*`**, which turns a floor into a fixed box. For icon-only content that's fine and probably the point; it just means an IconButton can no longer grow to fit a wider child. Deliberate, I assume — flagging because the two classes read as equivalent and aren't.
3. **`shadow-btn-primary` is gone from `primary`**, which the old header listed as absorbed (`visual-button-shadow-anatomy: 6-layer shadow on primary`). Consistent with the re-tokenized flat look, so I read it as intended.

Minor API note: `buttonBase` and `buttonVariants` are now exported from `packages/studio/src/index.ts`, so raw class strings become part of the package's public surface. `IconButton` needs them, but a shared-internal module would keep them out of the published API.

Nothing blocking. The primitives are coherent, the `data-[preview-state=…]` mirroring is pinned by `Button.test.tsx`, and the Tabs move to Base UI buys real keyboard semantics (roving tabindex, arrow/Home/End, `aria-controls`) that six hand-rolled strips didn't have.

— Rames

terencecho (approved):
### APPROVING at `2e5e013e` — rebase-only delta plus one legit vacuity-guard fix; 12/15 author files byte-identical to prior approval

Prior approvals at `eddcb73f` (Rames + tai) and `91f724df` (Somu) were invalidated by the rebase push. Head is `2e5e013efa0feb2ea97555b151a7b9d4572eef45`.

## Author-scoped delta since `eddcb73f`

Merge-bases moved `a944662` → `93ff9a99` (hf#3623 landed on main between the two heads). File-set is unchanged: same 15 files touched vs each PR's own merge-base.

Per-file blob-SHA parity check:

- IDENTICAL (12): `.fallowrc.jsonc`, `Button.tsx`, `IconButton.tsx`, `Tabs.tsx`, `Tabs.test.tsx`, `Tooltip.tsx`, `Tooltip.test.tsx`, `cn.ts`, `cn.test.ts`, `index.ts`, `mountHost.testHelpers.ts`, `studio/src/index.ts`
- CHANGED (3): `bun.lock`, `packages/studio/package.json`, `Button.test.tsx`

## The three changed files

**1. `bun.lock`** — pure rebase noise. New base `93ff9a99` includes `hf#3623`, `hf#4178` (v0.8.51 release), and `hf#4165` (dock engine). Lockfile advances accordingly.

**2. `packages/studio/package.json`** — two lines, both from the same rebase:
```diff
-  "version": "0.8.50",
+  "version": "0.8.51",
   ...
+    "dockview-react": "8.3.1",
```
Version bump from the v0.8.51 release-cut (hf#4178) and the `dockview-react` dep from the dock-engine PR (hf#4165). Neither is PR-3622 content — both are transitively pulled in by the moved base.

**3. `Button.test.tsx`** — the single Miguel-authored delta since prior approval. Commit `2e5e013e`:
```diff
   it("fails an undefined class, so the check above is not vacuous", () => {
-    expect(unresolved(["rounded-button", "shadow-btn-primary"])).toEqual([
-      "rounded-button",
-      "shadow-btn-primary",
+    // Names in a real namespace with no token behind them; the theme file now defines the
+    // `rounded-button` pair this used to name.
+    expect(unresolved(["bg-not-a-token", "text-step-999"])).toEqual([
+      "bg-not-a-token",
+      "text-step-999",
     ]);
   });
```

Why this fix is genuine, not cosmetic: the test just above asserts `unresolved([...emittedClasses]) === []` — every emitted class resolves. This second test proves the first isn't vacuous by feeding known-undefined names and asserting they come back unresolved. Before the rebase, `rounded-button` / `shadow-btn-primary` were the sentinel undefined names. After the rebase, hf#3621's theme + hf#3623's tighter class gate mean the theme file now defines `rounded-button` — the sentinel would no longer stay undefined and the guard's discrimination would silently weaken. Rename to `bg-not-a-token` + `text-step-999` (both real namespaces with no token behind them) restores it. The self-edited comment on the assertion names the exact invariant that broke, which matches the fix.

Mutation-red: swap `unresolved()` to return `[]` and this test fails (expects the two names back). Discrimination held.

## CI

Head at `2e5e013e`: 55 checks total, 47 SUCCESS + 8 SKIPPED. No FAILURES, no PENDING/QUEUED. Preflight (lint + format), regression-shards, Studio and player captures, Test lanes, CodeQL — all green. Stamping on code merit per [[feedback_ci_is_reference_not_gate]].

## Stamp mechanics

`mergeStateStatus: BLOCKED`, `reviewDecision: REVIEW_REQUIRED`. hf-oss `require_last_push_approval=true` — this APPROVE binds to `2e5e013efa0feb2ea97555b151a7b9d4572eef45`. Miguel `miguel-heygen` on trusted stamp list.

— Review by tai (pr-review)

jrusso1020 (commented):
**COMMENT** at `2e5e013efa0feb2ea97555b151a7b9d4572eef45` — no second stamp needed.

`terencecho`'s review `5256808472` landed on this exact head, which is what `require_last_push_approval` was waiting for: `reviewDecision: APPROVED`, `mergeStateStatus: CLEAN`, all nine required contexts green. Adding a redundant approve would say nothing, so this is the verification and one non-blocking note instead.

### The code approved at `eddcb73f` is intact

Compared each head's own contribution against its own merge base (`compare/main...<head>`, so the file list comes from the API rather than a path glob), then diffed the two patch sets per file:

| | |
|---|---|
| file list | identical, 15/15 at both heads |
| byte-identical contribution | 12 of 15 — **every component and test file except `Button.test.tsx`**: `Button.tsx`, `IconButton.tsx`, `Tabs.tsx`, `Tooltip.tsx`, `cn.ts`, `index.ts`, the three other test files, `mountHost.testHelpers.ts`, `packages/studio/src/index.ts`, `.fallowrc.jsonc` |
| changed | `Button.test.tsx`, `bun.lock`, `packages/studio/package.json` |

`packages/studio/package.json` is pure rebase noise — the only delta is a `dockview-react` context line main added around this PR's `"cn": "0.2.4"`; the added lines are unchanged.

### The test-guard fix is correct, and it was necessary

`theme.css` at this head defines `--radius-button` (:52) and `--shadow-btn-primary` (:90). The old vacuity guard asserted those two names come back **unresolved**, so once the rebase brought the theme in, `unresolved()` would have returned `[]` and the test would have failed. Necessary fix, correctly diagnosed.

The replacements hold up as a control: no `not-a-token` or `step-999` anywhere in `theme.css`, and both sit in live namespaces (`--text-step-11` / `--text-step-12` are real at :68-69), so they compile through the same path as a genuine candidate rather than being thrown out as malformed. `Test` is green at this head, which is the execution witness that `unresolved(["bg-not-a-token", "text-step-999"])` really did return both.

### Non-blocking: the lockfile delta is this branch's own, not absorbed base drift

`main`'s `bun.lock` is **byte-identical** to this PR's merge base (`93ff9a99`), so nothing in the 140 changed lines arrived from the newer base. Three classes:

1. **Intended** — `@base-ui/react@1.7.0`, `cn@0.2.4` and their transitive deps (`@base-ui/utils`, `@floating-ui/*`, `use-sync-external-store`, `reselect`). Matches the two `package.json` additions exactly.
2. **Workspace version mirrors**, 26 lines, `0.8.48` → `0.8.50`. The manifests read `0.8.51` at the merge base, at this head and on `main`, so the new value matches neither the old lockfile nor the packages it mirrors.
3. **A top-level babel float** — `@babel/parser` 7.29.3 → 7.29.9, `@babel/types` 7.29.0 → 7.29.8, `@babel/helper-string-parser` 7.27.1 → 7.29.7, `@babel/helper-validator-identifier` 7.28.5 → 7.29.7, with nested pins re-added for `@babel/core`, `@babel/generator` and `@babel/code-frame`. `packages/parsers` declares `^7.27.0`, so it picks up the newer parser once this lands.

All of it is inside the declared ranges and the runtime-contract, regression and windows suites are green, so it isn't a blocker and I wouldn't hold the launch for it. It is a dependency move riding into a Studio-primitives PR that the body doesn't mention — resetting `bun.lock` to the base and re-running `bun install` for just the two new deps would keep the diff to the intended change.

— Rames

Issue #3611, opened 2026-09-03
Critical Analysis of HyperFrames' Skill‑System Flaws


zhy438:
### Problem

HyperFrames’ skill‑system suffers from a severe "engineering illusion", a flawed design stemming from a fundamental mismatch between human‑engineer thinking and the real‑world capabilities of LLMs. Though marketed as AI‑Agent‑friendly, the official skill repository performs poorly in practical deployment.

In pursuit of exhaustive coverage, its skill documentation crams rendering principles, animation pitfalls, FFmpeg parameters, environment troubleshooting and massive reference materials into prompts, essentially converting developer manuals directly into prompt content. Restricted by the LLM’s attention mechanism, long‑context triggers the Lost‑in‑the‑Middle effect. Critical constraints get diluted and buried amid redundant information. Over‑abundant rules further fuel model hallucinations, resulting in an actual skill utilization rate below 10%.

Its deeply‑nested file structure is inherently anti‑Agent. Accessing deep‑layer constraints requires repeated tool calls and heavy token consumption. In most cases, agents abandon full‑rule reading and generate code based merely on top‑level files. At its core, this system mistakenly treats probabilistic LLMs as deterministic compilers, attempting to constrain outputs purely through static documentation.

The entire skill module acts as a liability for the project. It lacks on‑demand, tiered‑loading mechanisms and fails to adopt a minimal‑constraint design. If future releases do not refactor its bloated skill architecture, it indicates the development team has not grasped the real‑world operating paradigm of Agents.

### Proposed solution

Skill frameworks must take **LLM attention scarcity** as their top‑level design premise, instead of prioritizing completeness in the style of traditional software engineering. Being AI‑friendly does not mean providing exhaustive documentation. It means exposing only the minimum necessary information to Agents and offloading deterministic validation to code‑based tooling.

At the foundational design level, the maintainers need to break away from conventional software‑engineering mindsets. Human‑oriented repositories usually pursue full completeness, bundling principle documents, troubleshooting guides, change logs and edge‑case references alongside source code, which human developers can browse on demand. However, LLM‑powered Agents lack the ability to look up documentation independently. They can only process content injected into the context window, with limited attention capacity, which frequently leads to faded instructions, lost constraints and hallucinated outputs under long prompts.

Therefore, HyperFrames’ skill ecosystem needs to reverse its priority order: **Agent context cost outweighs functional completeness**.

1. Minimum‑necessary‑information principle
[...]

PR #3635, opened 2026-09-04
fix(cli): install the published skill set; feat(cli): snapshot --against

WaterrrForever:
Two CLI changes that fell out of auditing agent traces from the replay lab (context: GitHub #3611 and the varcyyyy 1:1 replication runs).

## 1. `hyperframes skills` installs the published set, not every SKILL.md in the clone

Bare `hyperframes skills` passed `--skill '*'` to the upstream installer, which discovers every `SKILL.md` in the repo. Since the repo-native skills under `.claude/skills` / `.agents/skills` landed, that meant **26 skills installed vs the 20 the manifest publishes** (`npx skills add heygen-com/hyperframes --list --full-depth` → "Found 26 skills"; the extra six are captions-overlay, changelog-video, cut-the-curve, motion-doctrine, oversized-cursor, seam-craft). In one trace the agent read the leaked `motion-doctrine` (12 KB, description says "load FIRST") and then had to declare it overridden in its final report.

- `updateSkills` gains `all: true`; the bare command now resolves the canonical manifest and installs its names, the same path `skills update` already uses. The `*` wildcard remains only as the offline fallback.
- Docs (README, CLAUDE.md, AGENTS.md, `docs/guides/skills.mdx`, `docs/prompting/overview.mdx`) point the full set at `npx hyperframes skills` and say what `skills add --all` really pulls. The old "non-interactive without `--skill` installs all 20" claim is dropped: upstream only installs all under `--all` / `--yes`, otherwise it opens the picker.

## 2. `hyperframes snapshot --against <video>`

Rebuilding footage means checking the render against the source at the same timestamps. In the varcyyyy traces every cell (HyperFrames, Remotion, no framework) hand-rolled that loop: ffmpeg frame grabs, `magick` side-by-side sheets, five rounds, ~30 min, plus environment errors (missing fonts for `montage`, no numpy). The HyperFrames agent searched the CLI docs for a compare command first and found only project-vs-project `compare`.

`--against` saves, for every captured time, the reference frame (`ref-XX-at-Ns.png`, frame-exact seek with `-ss` after `-i`) and a labelled render | reference pair (`pair-XX-at-Ns.jpg`) via the existing contact-sheet helper. No new dependencies.

Smoke-tested on the varcyyyy rebuild against its footage (3 timestamps, 6 files, pair sheets readable). Documented in `docs/packages/cli.mdx`.

## Tests

- `skills.test.ts`: the three full-install platform cases now expect per-name `--skill` flags (39/39 pass).
- `snapshot.test.ts`: one new source-contract test for `--against`. Two pre-existing "snapshot lint preflight" failures are unchanged on `main` (verified by stashing this diff).
- `oxlint` / `oxfmt` clean on changed files.


WaterrrForever:
Split into https://github.com/heygen-com/hyperframes/pull/3636 and https://github.com/heygen-com/hyperframes/pull/3637, one concern per PR.

PR #3638, opened 2026-09-04
fix(check): reject audio elements without audio streams

miguel-heygen:
## What

`hyperframes check` / `lint` now report `media_src_kind_mismatch` (error) when an element the render will mix as audio points at a local file that has no audio stream, matching the producer's authored-media contract before the render starts.

## Why

Check previously validated file presence and browser layout but did not inspect audio stream types. A silent video authored as `<audio>` (or as `<video data-has-audio="true">`) passed check and failed only when the render compiled the composition.

## How

- A probe-backed half of the existing `media_src_kind_mismatch` rule (`packages/lint/src/renderAudioSourceLint.ts`). Same code and severity as the static extension-based half in `rules/media.ts`, so "tag vs media kind" has one owner.
- The element set is derived from the producer's two gates. Compile gate (`compileHtml` in core, both phases, regardless of `data-hidden`): every `<audio>` with its own `src` that either has no `data-end` and no parseable `data-duration` (phase 1 probes it for a duration) or has `data-duration > 0` and no `loop` (phase 2 probes it to clamp). Mixer gate: `audio[id]` plus every `video[id]` the compiled HTML marks `data-has-audio="true"` — authored, or inferred by the timing compiler from a missing `muted` — minus `data-hidden` (self or ancestor), hidden `<hf-audio-group>` members and known inactive timing windows. Src resolves like the engine's `resolveMediaElementSrc` (own `src`, else `<source src>`). Relative `data-start` combined with `data-end` is treated as unknown and skipped, so the rule never flags an element the render drops.
- Known gap, documented in the rule: a sub-composition whose host is `data-hidden` only in the parent file. The mixer's ancestor walk sees that across inlining; a scan of the sub-composition file alone cannot, so a `data-end`-bounded silent `<audio>` inside it is still reported.
- One shared ffprobe runner (`packages/lint/src/mediaStreamProbe.ts`) serves this rule and `hevc_preview_codec`: unfiltered `stream=codec_type,codec_name` listing, bounded concurrency, each unique file probed once, and a three-state per-file result (streams / probe failed / not probed). Missing, remote, inline, placeholder and unprobeable sources keep their existing classifications; only a successful no-audio probe produces the error.
- `check`'s exit code now depends on ffprobe being installed for the audio case: without ffprobe the rule stays silent (as `hevc_preview_codec` already did).

## Test plan

- [x] `packages/lint`: `bunx vitest run` — 16 files, 664 tests; `tsc --noEmit` clean. New cases pin the exact ffprobe argv (mock answers only the contracted invocation), and cover hidden/bus-muted/inactive/relative-start exclusions, both compile-gate cases (hidden `data-duration`-bounded and hidden untimed `<audio src>` flagged; hidden `data-end`-bounded, `loop`ed and `<source>`-only skipped), the four `<video>` cases (unmuted without `data-has-audio` flagged, explicit `"true"` flagged even when `muted`, `muted` and `"false"` skipped), nested `<source>`, template-wrapped sub-compositions, shared-file dedup, probe failure and missing ffprobe staying silent. Mutation check: dropping the unmuted-video inference, the phase-1 branch, the `loop` exclusion or the own-`src` requirement each fails exactly one test.
- [x] `scripts/check-no-main-deletions.mjs --base origin/main` passes with the moved `hevcPreviewLint.windowsHide.test.ts` listed in `ALLOWED_DELETIONS`.
- [x] `packages/producer`: `ffprobeArgvContract.test.ts` 81/81 (manifest updated to `mediaStreamProbe.ts`).
- [x] CLI `check`, `lint`, `validate` command suites green.
- [x] Manual: real ffprobe on a fixture project with a generated silent MP4, an MP3 and an MP4 with AAC — only the silent `<audio>`, `<audio><source>`, `<video data-has-audio>` and template-wrapped sub-composition elements were reported; hidden, bus-muted, inactive, relative-start, muted `<video>` and real-audio elements were not.


miga-heygen (approved):
Reviewed the ffprobe-based audio-stream lint. Concurrency-limited probing (8 workers) with bounded results array is correct; `hasAudioStream` checks `codec_type === "audio"` in the stream list; probe failures (`null`) are treated as "don't flag" rather than false-flagged, avoiding false positives when ffprobe is missing/misconfigured. Pattern mirrors the existing `hevcPreviewLint` collect/lint split already used in `project.ts`. Test covers the silent-MP4 case end-to-end through `lintProject`. LGTM.

miga-heygen (commented):
Second-pass quality review at head `35f9a3d5` (isolated worktree; `project.test.ts` 34/34 reproduced, plus six unmocked real-media lint runs cross-checked against live `parseAudioElements` output). The ffprobe logic itself is right — Opus-in-WebM, ALAC, AAC-in-MP4-with-video and cover-art MP3 all pass, only a silent MP4 is flagged. The element set the rule scans is what does not match the render contract the description claims parity with.

**Blocking**

- **B1 — false positive that turns `check` red on a project that renders.** `packages/lint/src/audioMediaTypeLint.ts:64-69` (`collectLocalAudioCandidates`) matches every `<audio src>`; the render side (`parseAudioElements`, `packages/engine/src/services/audioMixer.ts:564`) reads `audio[id]` minus `data-hidden`, hidden groups and inactive windows. A lone `<audio id="hid" data-hidden src="silent.mp4">` yields an `error` whose message says render rejects it, while `parseAudioElements` returns `[]`.
- **B2 — sibling audio source not covered.** `<video data-has-audio="true">` is an audio source (`audioMixer.ts:574`) and preflight maps it to `expected: "audio"`; a silent file there throws at render and this rule says nothing (parser returns the element, lint returns no finding).
- **B3 — `<audio><source src>` is not seen.** The engine resolves nested `<source>` via `resolveMediaElementSrc`; the tag-only regex here does not. `packages/lint/src/rules/media.ts:129` already has `mediaHasResolvableSrc`, documented as matching the engine, and it is not used here.
- **B4 — a second owner of "tag vs media kind".** `rules/media.ts:114` `media_src_kind_mismatch` already owns this contract in the same package at the same severity. The new rule adds a different code and mechanism split by element type; that split is why `<video src="music.mp3">` is caught by neither rule while render throws. One rule, one element set derived from the engine's, is the shape.
- **B5 — the test cannot detect a wrong ffprobe invocation.** `mockFfprobeStreams` reads only the file path and ignores `args`. Injecting `-select_streams v:0` (the flag pair from `hevcPreviewLint.ts`, which this file is copied from) leaves 34/34 green; real ffprobe with that flag returns zero streams for a plain MP3, which would be a blocking error on every correct project, invisible to the suite. Assert the args, or run one real-ffprobe case.

**Non-blocking:** ~100 of 131 lines duplicate `hevcPreviewLint.ts` (timeout/concurrency constants, `execFileAsync`, worker pool, collector differing only in the `mediaSrcTagRe` argument) — extract the shared probe runner; two sequential ffprobe passes with per-rule dedup probe a file twice when it appears in both `<audio>` and `<video>`; `new Set` on display srcs present in the sibling, missing here; `check`'s exit code is now ffprobe-dependent for the first time (worth a note in the description); the `boolean | null` three-state return is undocumented. Binary resolution does not drift — lint and engine both use `parsers/src/ffBinaries.ts`.

Mutations: unwire the rule → 1 red (pinned); wrong ffprobe flags → 34 green (B5). No render executed; preflight behaviour is proven by code path plus real parser output.

**Reasoning:** the rule flags elements render ignores and ignores elements render rejects, and the suite cannot see a wrong probe invocation; the check itself is sound once the element set is the engine's.

_Review by Miga_

miga-heygen:
Addressed in b6a86e19.

- B1 — element set now follows the producer's two gates (`packages/lint/src/renderAudioSourceLint.ts:60-83`, `:112-119`): `audio[id]` with a positive `data-duration` and a local src is always a candidate because the compile gate (`compileForRender` → `assertAssetMediaTypeProfile`) probes it regardless of `data-hidden`; otherwise the mixer's exclusions apply — `data-hidden` on self/ancestor, hidden `<hf-audio-group>` membership, known inactive windows. Relative `data-start` with `data-end` is treated as unknown and skipped.
- B2 — `video[id][data-has-audio="true"]` is in the set (`renderAudioSourceLint.ts:57`), with the mixer's exclusions.
- B3 — src resolution mirrors `resolveMediaElementSrc` (own `src`, else `<source src>`, local preferred over remote) (`renderAudioSourceLint.ts:25-36`). The OpenTag-based `mediaHasResolvableSrc` returns only a boolean and lacks the local-over-remote preference, so the DOM-based project pass carries its own copy of the engine's function; the lint package cannot import the engine's (core depends on lint).
- B4 — folded into `media_src_kind_mismatch`: same code and severity, one message family (`renderAudioSourceLint.ts:130-155`); the static half in `rules/media.ts` documents the split (`rules/media.ts:115-122`). `<video src="music.mp3">` is covered on main by the extension classifier's audio branch.
- B5 — tests pin the exact ffprobe argv (`packages/lint/src/project.test.ts:404-441`): the mock answers only `-v error -show_entries stream=codec_type,codec_name -of json -- <file>` and returns `streams: []` for anything else, so a `-select_streams v:0` mutation fails both the hevc positives and the "real audio not flagged" negative; `mediaStreamProbe.test.ts` asserts the argv, `windowsHide` and timeout directly.
- Extras — one shared runner `packages/lint/src/mediaStreamProbe.ts` (both rules, each unique file probed once, three-state result documented: streams / `null` failed / absent not probed); `hevc_preview_codec` reads the same unfiltered listing (a file whose later video stream is HEVC is now also reported); `querySelectorAllIncludingTemplates` moved to `domQuery.ts` and shared.

Rebased onto main. `packages/lint`: 659 tests, typecheck clean; `packages/producer` ffprobe argv contract 81/81 (manifest now points at `mediaStreamProbe.ts`); CLI `check`/`lint`/`validate` suites green. Real-ffprobe run on a fixture project flagged exactly the silent `<audio>`, `<audio><source>`, `<video data-has-audio>` and template-wrapped sub-composition cases and none of the hidden/bus-muted/inactive/relative-start/muted-video/real-audio elements.

miga-heygen (commented):
Reviewed head `b6a86e19f` in an isolated worktree. Re-pass of my review at `35f9a3d5` (5189497781).

**Strengths**
- `packages/lint/src/renderAudioSourceLint.ts:25-36` — the DOM src-resolution mirror matches engine `resolveMediaElementSrc` line for line (own `src`, else first `<source src>`, local over remote). Prior B3 closed.
- `media_src_kind_mismatch` is now a single owner (`rules/media.ts:115-122` ↔ `renderAudioSourceLint.ts:130-155`); the shared `packages/lint/src/mediaStreamProbe.ts` runner dedupes paths, returns a three-state result, and both rules consume it. Prior B4 closed.
- `project.test.ts:404-441` + `mediaStreamProbe.test.ts` answer only the exact ffprobe argv, and the producer `ffprobeArgvContract` manifest is updated. Prior B5 closed.
- `domQuery.ts` is a pure move; the hevc rule's switch to the unfiltered stream list changes no test or user-visible output.

**Blockers**

1. **CI is red and the lint suite never ran.** `Detect changes` (run 34878519914) fails in `scripts/check-no-main-deletions.mjs`: this branch deletes `packages/lint/src/hevcPreviewLint.windowsHide.test.ts` (its assertion moved to `mediaStreamProbe.test.ts:51-53`) and the path is not in `ALLOWED_DELETIONS`. Build, Test, Test: runtime contract and Typecheck all show `skipped` at this head, not passed. Fix: add the path with the move reason and confirm the lanes execute.

2. **False negative on untimed hidden `<audio src>` (prior B1, partial).** The compile gate probes in two phases. `renderAudioSourceLint.ts:112-113` mirrors only phase 2 (`data-duration > 0`, own `src`, no `loop`). Phase 1 (`htmlCompiler.ts:571-590` via `timingCompiler` compileTag) probes and asserts every `<audio>` with own `src` and neither `data-end` nor a parseable `data-duration`, regardless of `data-hidden`. So `<div data-hidden><audio id src="silent.mp4"></audio></div>` passes `check` and fails the render with `ASSET_MEDIA_TYPE_MISMATCH` (reproduced end to end through `compileForRender` with real ffprobe). `project.test.ts:704-716` currently pins the wrong outcome for the `parent-hidden` case. Fix: `compileGateProbes = tag === "audio" && ownSrc && ((duration == null && !hasAttr("data-end")) || (duration > 0 && !hasAttr("loop")))`, and flip that expectation (the `self-hidden` / `on-muted-bus` cases with `data-end` stay skipped).

3. **False negative on unmuted `<video id src>` without `data-has-audio` (prior B2, partial).** `packages/core/src/compiler/timingCompiler.ts:259-260` injects `data-has-audio="true"` on every unmuted `<video>` lacking the attribute before the mixer (`audioMixer.ts:574`) and preflight read it. The lint selector at `:58` requires the literal attribute, so a silent unmuted `<video>` is skipped by lint and rejected by preflight with `expected: "audio"`. The video path also has no test: deleting the selector leaves 48/48 green, which contradicts the PR body. Fix: select `video[id]` and treat `data-has-audio="true"` or (attribute absent and not `muted`) as in the mixer set; add tests for silent unmuted (flagged), explicit `true` (flagged), `muted` / `false` (not flagged).

**Important**
- Narrow false positives from the same predicate: hidden + `loop` + `data-duration>0`, and hidden + `<source>`-only + `data-duration>0` are probed by no render gate (phase 2 needs own `src` and no `loop`; the mixer drops hidden) but lint flags them. This is the "check red on a project that renders" outcome the rule comment at `:76-79` promises never to produce. Same fix as blocker 2.

**Nits**
- `renderAudioSourceLint.ts:19-23` says `mediaHasResolvableSrc` was not reused because of the core→lint cycle. The cycle is real (`core/package.json`), but that helper is lint-local (`rules/media.ts:173`); the actual reason is that it works on `OpenTag` regex tokens, not DOM elements. Fine not to reuse; the stated reason is wrong.
- A sub-composition whose host `<div data-hidden data-composition-src>` is hidden only in the parent: the mixer's ancestor walk sees that across inlining, a standalone scan of the sub-composition file cannot. Only a `data-end`-bounded silent audio inside such a sub-comp is affected. Not reproduced; worth a comment.
- `collectRenderAudioCandidates:102-107` materialises `sources` only to iterate it once.

**Ran:** lint 48/48 (rule tests) and 556/556 (package), producer argv contract 81/81; `tsc --noEmit` clean in lint and producer; 5 mutations (4 red, the video-selector deletion green as noted). Every mutation reverted.

**Verdict:** COMMENT (With fixes)
**Reasoning:** Three of five prior blockers are closed cleanly, but the element set still diverges from the render contract in two directions that were the point of the rule, and the head has no Linux CI verification.

_Review by Miga_

miga-heygen:
Addressed in 10bf3bc2.

miga-heygen (approved):
Reviewed head `10bf3bc26` in an isolated worktree. Re-pass of my review at `b6a86e19` (5204226612).

**Prior blockers**
- **CI fixed.** `scripts/check-no-main-deletions.mjs:160-163` allowlists the moved `hevcPreviewLint.windowsHide.test.ts`; the guard exits 0 locally and every required check at this head executes and passes (Build, Test, Test: runtime contract, Typecheck, Semantic PR title, Tests on windows-latest, Render on windows-latest, regression, CLI smoke), none skipped.
- **B1 fixed.** `packages/lint/src/renderAudioSourceLint.ts:87-92` `compileGateProbes` is now own `src` and (no parseable `data-duration` → no `data-end`; `data-duration > 0` → no `loop`), which matches both compile-gate phases term by term against `core/src/compiler/timingCompiler.ts:236-252,363-392` and `producer/src/services/htmlCompiler.ts:569-571,599,507`, using the same `parseNumeric`. `project.test.ts:705-720` flips the parent-hidden untimed case to flagged, and `:723-740` pins the previously false-positive shapes (hidden + `loop` + duration, hidden + `<source>`-only + duration) as not flagged. Confirmed end to end with real ffprobe on a silent mp4.
- **B2 fixed.** The element set is `audio[id], video[id]` (`:159`) and `mixerTreatsVideoAsAudible` (`:103-107`) mirrors the compiler's injection at `timingCompiler.ts:259-260`, including authored value winning over `muted`. `:742-797` covers unmuted silent (flagged), muted (not), muted + `true` (flagged), `false` / empty / hidden (not). Deleting the video handling now fails two tests, closing the prior unpinned gap.

**Prior nits, addressed:** the `mediaHasResolvableSrc` comment now names the real reason (token-based helper vs DOM); the sub-composition hidden-by-parent edge is documented under "Known gaps"; the single-pass `sources` array is gone.

**Single source of truth.** The render's element set is now one selector, two named predicates (`compileGateProbes`, `mixerTreatsVideoAsAudible`) and one exclusion block, each cross-referenced to the producer/core line it mirrors.

**Nits (cosmetic)**
- `renderAudioSourceLint.ts:184` reads `data-has-audio` a second time after the predicate already did.
- `compileGateProbes` is guarded to `<audio>` at the call site rather than inside the predicate.
- A core-side contract test asserting the lint predicates against `timingCompiler`'s own behaviour would catch drift the next time the compiler changes.

**Ran:** lint `project.test.ts` 49/49, full lint suite 664/664; `tsc --noEmit` clean; 6 mutations all red. All reverted.

**Verdict:** APPROVE
**Reasoning:** All three blockers are closed with predicates verified against the producer and core code they mirror, pinned by tests that go red under mutation, and required checks execute green at this head.

— Miga

PR #3639, opened 2026-09-04
fix(engine): preserve nested sibling video layout

miguel-heygen:
## What

Sequential in-flow sibling videos now keep the same nested scene box during export, so later siblings no longer render outside the capture surface as black frames.

## Why

The runtime intentionally applies `display:none` to inactive in-flow leaf clips. The frame-injection visibility sync removed that runtime-owned value, restoring the inactive sibling’s layout box and pushing the active replacement frame down in document flow.

## How

Keep `display` ownership in the runtime while the injector continues to manage native-video and replacement-frame visibility. The change removes one conflicting style reset and adds a regression for inactive sibling layout preservation.

## Test plan

- [x] Unit tests added/updated: targeted RED reproduced the lost `display:none`; GREEN passed with 1 test, 0 failures

The full screenshot-service file reached 30 passing tests; two existing LinkeDOM/Bun `setProperty` spy assertions failed even though their visibility value assertions passed. Engine typecheck was blocked by missing generated/published dependency payloads in the clean worktree. `git diff --check` passed.

### DIFF
--- a/packages/engine/src/services/screenshotService.test.ts
+++ b/packages/engine/src/services/screenshotService.test.ts
@@ -518,6 +518,23 @@ describe("video-frame injection respects ancestor visibility", () => {
     expect(seededImg.style.visibility).toBe("hidden");
   });
 
+  it("preserves display:none on an inactive in-flow video so its sibling keeps the same box", async () => {
+    const { teardown, setup } = withGlobals(setupHostHiddenScenario({}));
+    setup.video.style.display = "none";
+    const seededImg = setup.document.createElement("img");
+    seededImg.classList.add("__render_frame__");
+    setup.video.parentNode?.insertBefore(seededImg, setup.video.nextSibling);
+
+    try {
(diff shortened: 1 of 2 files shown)


miga-heygen (approved):
Traced `syncVideoFrameVisibility` and `injectVideoFramesBatch` in screenshotService.ts. Confirmed the removed `video.style.removeProperty("display")` was the only place in the inactive-video branch touching `display`, and that GSAP (not this sync function) owns `display` toggling on in-flow leaf clips — this function only layers visibility/pointer-events for the frame-replacement mechanism. Removing the reset stops clobbering the runtime's own `display:none` on inactive siblings, which was reintroducing them into layout flow and pushing the active replacement frame down. New regression test directly reproduces the reported symptom (inactive sibling keeps `display:none`, active sibling's frame stays hidden-correctly). Scoped, minimal, matches stated intent. LGTM.

PR #3640, opened 2026-09-04
fix(runtime): align nested activation to export frames

miguel-heygen:
## What

Nested composition hosts and media now activate on the intended export frame when authored decimal timing lands microscopically above that frame boundary.

## Why

Export seeks were quantized to the render FPS, but timed-element visibility still compared the quantized seek against raw authored seconds. At 30fps, frame 584 is `19.466666666666665s`, which compared below an authored `19.466666666667s` start and delayed the outro by one frame.

## How

Align finite visibility start/end boundaries to the same export frame grid used by render seek. Continuous preview timing remains unchanged because alignment is gated on export render configuration.

## Test plan

- [x] Unit tests added/updated: frame-584 nested host/video regression passed (1 test, 0 failures)
- [x] Quantizer suite passed (10 tests, 0 failures)

A broader Bun run reached the runtime tests but hit two unrelated runner assumptions (package-relative CWD and unsupported `vi.advanceTimersByTimeAsync`) before its timeout. Core typecheck was blocked by missing `htmlparser2`/`hono` dependency payloads in the clean worktree. `git diff --check` passed.

### DIFF
--- a/packages/core/src/runtime/init.test.ts
+++ b/packages/core/src/runtime/init.test.ts
@@ -597,6 +597,40 @@ describe("initSandboxRuntimeModular", () => {
     );
   });
 
+  it("activates a nested outro on frame 584 when its authored start rounds just above it", () => {
+    const root = document.createElement("div");
+    root.setAttribute("data-composition-id", "main");
+    root.setAttribute("data-root", "true");
+    root.setAttribute("data-start", "0");
+    root.setAttribute("data-duration", "20");
+    root.setAttribute("data-width", "1920");
+    root.setAttribute("data-height", "1080");
+    document.body.appendChild(root);
+
+    const outro = document.createElement("div");
+    outro.setAttribute("data-composition-id", "outro");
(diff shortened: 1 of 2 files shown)


miga-heygen (approved):
Verified against `quantizeTimeToFrame` in parityContract.ts (floor-to-frame-grid with epsilon tolerance) — this is the same function already used elsewhere in init.ts for render-seek quantization, so frame-aligning the authored start/end boundaries with it puts both sides of the comparison on the identical grid. Walked the frame-584 repro numerically: quantizing both the render seek (584/30) and the authored start (19.466666666667) collapse to the same value (19.466666666666665), fixing the off-by-one-frame comparison failure. The fix is correctly gated on `window.__HF_EXPORT_RENDER_SEEK_CONFIG` being present, so preview-time (non-export) comparisons are untouched — confirmed no fallback path changes when that config is absent. Test reproduces the exact frame-584 nested-outro scenario. LGTM.

PR #3641, opened 2026-09-04
fix(engine): retry capture screenshot timeouts

miguel-heygen:
## What

Screenshot protocol timeouts now enter the existing adaptive capture retry, allowing the renderer to reduce worker concurrency instead of hard-failing a valid long render.

## Why

Puppeteer's `Page.captureScreenshot timed out` signature fell through to `authoring`, so parallel capture treated a recoverable CDP timeout as fatal.

## How

Classify the exact screenshot method timeout as `protocol_timeout`. Keep pressure-aware auto worker sizing as separate follow-up scope.

## Test plan

- [x] Unit tests added/updated: engine classifier RED-to-GREEN (14/14)
- [x] Manual testing performed: exact wrapped parallel-worker runtime probe returned `recoverable=true`
- [x] Documentation updated (not applicable; internal failure classification)

The full producer suite could not start in the task worktree because its existing dependency payload omitted generated and package `dist` artifacts; the exact producer recovery assertion is included for CI.

### DIFF
--- a/packages/engine/src/services/captureFailure.test.ts
+++ b/packages/engine/src/services/captureFailure.test.ts
@@ -6,6 +6,10 @@ describe("classifyCaptureFailure", () => {
     ["Target closed", "transient_browser"],
     ["Runtime.callFunctionOn timed out after 30000ms", "protocol_timeout"],
     ["Runtime.evaluate timed out", "protocol_timeout"],
+    [
+      "Page.captureScreenshot timed out. Increase the 'protocolTimeout' setting in launch/connect calls for a higher timeout if needed.",
+      "protocol_timeout",
+    ],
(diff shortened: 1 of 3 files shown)


miga-heygen (approved):
Checked the new `/Page\.captureScreenshot timed out/i` pattern against the other classifier buckets (transient_browser, memory_exhaustion, verification) in captureFailure.ts — no overlap, so this can't misclassify into a different bucket. Small, additive, well-targeted: routes the exact Puppeteer CDP screenshot-timeout signature into the existing `protocol_timeout` recoverable path instead of falling through to fatal `authoring`. Tests cover both the classifier and the parallel-capture recovery check. LGTM.

PR #3642, opened 2026-09-04
fix(producer): preserve render warning attribution

miguel-heygen:
## What

Fatal render warnings now preserve their actionable message in the terminal error, so users can identify the failing track and cause without project archaeology.

## Why

`RenderQualityError` previously emitted only warning codes even though the mixer and structured warning already retained the element ID and automation parse reason.

## How

Format each blocking warning as its code plus a redacted message. Existing URL-query and path redaction prevents diagnostic attribution from exposing secrets.

## Test plan

- [x] Unit tests added/updated: focused warning/error suite passes 15/15
- [x] Manual testing performed: exact malformed-automation attribution shape asserted at the terminal error boundary
- [x] Documentation updated (not applicable; diagnostic-only behavior)
- [x] Producer typecheck, oxlint, and oxfmt checks pass

### DIFF
--- a/packages/producer/src/services/render/renderEventPublisher.test.ts
+++ b/packages/producer/src/services/render/renderEventPublisher.test.ts
@@ -126,6 +126,35 @@ describe("OrderedRenderEventPublisher", () => {
   });
 });
 
+describe("RenderQualityError", () => {
+  it("includes the warning message that identifies the failing audio element", () => {
+    const error = new RenderQualityError([
+      {
+        code: "audio_processing_failed",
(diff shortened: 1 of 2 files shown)


miga-heygen (approved):
Reviewed the diff and test additions. `RenderQualityError` now includes each warning's redacted message alongside its code, giving actionable attribution (e.g. which element/automation field failed) instead of just a bare code list. Confirmed the existing `redactTelemetryString` helper is reused for the secret-redaction guarantee, and the new tests assert both the attribution content and that URL-query secrets stay redacted. Small, low-risk, well-tested diagnostic-only change. Approving.

PR #3643, opened 2026-09-04
fix(producer): guard disk capture capacity

miguel-heygen:
## What

Long disk-frame renders now fail before capture when their conservative temporary-storage estimate would consume the available disk headroom.

## Why

The previous preflight checked only a fixed 1 GB floor. A long render could pass that check, select per-frame disk capture, and fill the volume before encoding.

## How

Estimate output-resolution frame storage at four bytes per pixel and reserve 10% of current free capacity. If the estimate exceeds that budget, report required versus available space and direct the user to streaming via `--low-memory-mode`, the duration cap, or more free disk.

The static-frame dedup mismatch remains causally unrelated: disk capture writes one output file per frame whether capture work was deduplicated or not.

## Test plan

- [x] Unit tests added/updated: focused capture-stage suite passes 4/4
- [x] Manual testing performed: injected 0.8 MB capacity reproduces the pre-capture rejection and actionable diagnostic
- [x] Documentation updated (not applicable; runtime guard and error guidance are self-contained)
- [x] Producer typecheck, oxlint, and oxfmt checks pass

### DIFF
--- a/packages/producer/src/services/render/stages/captureStage.test.ts
+++ b/packages/producer/src/services/render/stages/captureStage.test.ts
@@ -1,5 +1,9 @@
 import { describe, expect, it } from "vitest";
-import { shouldAllowAdaptiveCaptureRetry } from "./captureStage.js";
+import {
+  assertDiskCaptureHeadroom,
+  estimateDiskCaptureBytes,
+  shouldAllowAdaptiveCaptureRetry,
+} from "./captureStage.js";
 
 describe("shouldAllowAdaptiveCaptureRetry", () => {
   it("keeps timeout recovery enabled when the initial worker count was explicit", () => {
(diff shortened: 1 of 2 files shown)


miga-heygen (approved):
Reviewed the diff plus full captureStage.ts context. Confirmed: `framesDir` is always created (mkdirSync) before `runCaptureStage` runs at both call sites (renderOrchestrator.ts, distributed/renderChunk.ts), so the new `statfsSync` preflight check always has a valid path to stat. `totalFrames` in this stage already represents only the current chunk's frame count for distributed renders (per the existing frameRange contract), so the estimate doesn't over-count for chunked jobs. The `statfsSync`-with-catch-fallback pattern mirrors the existing usage in captureHdrResources.ts, so this is consistent with prior art rather than introducing a new platform-support risk. The 4-bytes/pixel raw-frame estimate is deliberately conservative versus actual jpg/png compressed output (so it may reject some renders that would've technically fit), but that's an intentional safety-margin tradeoff per the PR description, not a correctness bug. Approving.

PR #3644, opened 2026-09-04
fix(producer): cap delivered AAC true peak

miguel-heygen:
## What

Rendered AAC audio is now measured at the delivery boundary and constrained to -1 dBTP before the final video mux. Audio that is already safe is left unchanged.

## Why

Source-domain peak checks cannot predict lossy-codec overshoot. A reproduced 48 kHz stereo source at -1.5 dBTP crossed 0 dBTP after AAC duration normalization, so a successful render could contain clipping without any warning.

## How

The duration-normalized AAC artifact is decoded through FFmpeg's EBU R128 true-peak scan. When it exceeds the ceiling, the producer re-encodes from that completed artifact with only the measured attenuation, preserves the exact video duration, and rechecks the result. Correction is bounded to three passes; failure to establish the invariant fails the assembly instead of silently delivering unsafe audio.

This adds one decode-only scan to rendered audio. Extra AAC encoding occurs only when the measured output exceeds the ceiling.

## Test plan

- [x] Unit tests cover silence and measured attenuation arguments.
- [x] An FFmpeg integration regression generates the reproduced 48 kHz stereo waveform, runs AAC duration normalization and MP4 mux, and asserts the delivered stream is at or below -1 dBTP with no more than 3 LU loudness change.

### DIFF
--- a/packages/producer/src/services/render/audioPadTrim.integration.test.ts
+++ b/packages/producer/src/services/render/audioPadTrim.integration.test.ts
@@ -3,14 +3,40 @@ import { mkdtempSync, rmSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { afterEach, describe, expect, it } from "vitest";
+import { getFfmpegBinary, muxVideoWithAudio } from "@hyperframes/engine";
 import { padOrTrimAudioToVideoFrameCount } from "./audioPadTrim.js";
 
 const dirs: string[] = [];
 afterEach(() => dirs.splice(0).forEach((dir) => rmSync(dir, { recursive: true, force: true })));
 
+function measureLoudness(path: string): { integratedLufs: number; truePeakDbfs: number } {
+  const result = spawnSync(
+    getFfmpegBinary(),
+    [
+      "-nostdin",
+      "-hide_banner",
+      "-nostats",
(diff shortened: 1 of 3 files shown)


miga-heygen (approved):
Reviewed the diff in depth given the complexity (re-encode/measure loop). Traced the correction loop by hand against the test expectations: attenuation accumulates correctly (`AAC_DELIVERY_TRUE_PEAK_DBFS - measuredPeak`), matches the expected `volume=-2.500dB` for a 1.5 dBFS overshoot, is bounded to 3 correction passes before failing the assembly (never silently delivers unsafe audio), and the -1 dBFS silence case short-circuits with only the initial measurement (no extra encode). Verified the two production call sites (`distributed/assemble.ts`, `render/stages/assembleStage.ts`) never pass a custom `runFfmpeg`, so the real true-peak probe always runs in production — the `input.runFfmpeg ? undefined : defaultProbe` conditional only affects unit tests, not a way to accidentally bypass the safety check in production. The 192k AAC bitrate used in the correction re-encode matches the existing bitrate convention already used 3x elsewhere in this same file. Scratch-dir handling (mkdtemp beside the output, rename-on-success, cleanup in `finally`) looks race-safe for this single-file-in-flight use case. Approving.

PR #3645, opened 2026-09-04
fix(producer): preserve nested entry asset base

miguel-heygen:
## What

Direct `--composition compositions/scene.html` entries now keep the same sibling-asset resolution as mounted sub-compositions. A sibling `image.png` is emitted as `compositions/image.png`, while an existing project-root asset remains root-relative.

## Why

The compiler relocates every entry document to `compiled/index.html`. Direct nested entries previously kept plain relative URLs unchanged, so the render server requested `/image.png` and returned 404 even when `compositions/image.png` existed.

## How

Nested entry documents inside the project are rebased before compilation with the established mounted-subcomposition path resolver. The pass covers `src`/`href`, inline styles, and stylesheet `url(...)` values. Generated standalone wrappers outside the project are skipped, which prevents double rebasing of the existing mounted flow.

## Test plan

- [x] Regression test covers a direct nested entry with both a sibling image and a project-root image.
- [x] Full producer HTML compiler suite passes (132/132).
- [x] Shared path-rewriter suite passes (11/11).

### DIFF
--- a/packages/producer/src/services/htmlCompiler.test.ts
+++ b/packages/producer/src/services/htmlCompiler.test.ts
@@ -633,6 +633,32 @@ describe("detectRenderModeHints", () => {
     }
   });
 
+  it("rebases a direct nested entry's sibling assets without changing project-root assets", async () => {
+    const projectDir = mkdtempSync(join(tmpdir(), "hf-direct-entry-base-"));
+    const compositionsDir = join(projectDir, "compositions");
+    mkdirSync(compositionsDir, { recursive: true });
+    writeFileSync(join(compositionsDir, "sibling.png"), "sibling");
+    writeFileSync(join(projectDir, "root.png"), "root");
+    const entryPath = join(compositionsDir, "scene.html");
+    writeFileSync(
+      entryPath,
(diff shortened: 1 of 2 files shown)


miga-heygen (approved):
Reviewed the diff plus the underlying `rewriteAssetPath`/`rewriteAssetPaths`/`rewriteCssAssetUrls` helpers in packages/parsers/src/rewriteSubCompPaths.ts. Traced the added regression test by hand: for a direct nested entry at `compositions/scene.html`, a sibling `compositions/sibling.png` gets rewritten to `compositions/sibling.png` (sibling-exists probe hits), while a project-root `root.png` is left unchanged (sibling probe path `compositions/root.png` doesn't exist on disk, so it falls through to the original value) — matches the PR's stated sibling-first/project-root-fallback rule and the existing mounted-subcomposition behavior. The `isPathInside(htmlPath, projectDir)` guard correctly skips generated standalone wrappers living outside the project, avoiding double-rewriting of the already-handled mounted flow. Logic is consistent with the reused helpers rather than a parallel reimplementation. Approving.

PR #3647, opened 2026-09-04
fix(producer): follow variable audio duration

miguel-heygen:
## What

Duration-less variable-bound audio now follows each selected source’s intrinsic duration during batch and single renders. Explicit authored trims remain authoritative, and shorter selected sources still shorten the playable window.

## Why

Compilation previously injected the fallback source’s duration before render variables changed `src`. Browser reconciliation then minimized against that stale fallback boundary, so a longer batch-row source was silently truncated even though rendering completed successfully.

## How

The compiler marks duration metadata inferred for variable-bound media. Browser discovery uses the selected media element’s intrinsic duration only for that marked case, and one shared reconciliation rule replaces the fallback window when the source changed. Unmarked authored durations retain the existing bounded behavior.

## Test plan

- [x] Compilation tests distinguish inferred variable duration from authored and ordinary natural duration.
- [x] Browser discovery tests use intrinsic duration only for inferred variable media.
- [x] Probe reconciliation covers longer inferred, longer authored, and shorter inferred sources.
- [x] Compiler, natural-duration, and probe-stage suites pass (173/173).

### DIFF
--- a/packages/producer/src/services/htmlCompiler.naturalDuration.test.ts
+++ b/packages/producer/src/services/htmlCompiler.naturalDuration.test.ts
@@ -79,6 +79,21 @@ describe("compileForRender natural media duration parity", () => {
     }
   });
 
+  it("marks only variable-bound media whose natural duration was inferred", async () => {
+    const { document } = await compile(`
+      <audio id="inferred" src="ten-seconds.wav" data-var-src="track"></audio>
+      <audio id="authored" src="ten-seconds.wav" data-var-src="track" data-duration="4"></audio>
+      <audio id="plain" src="ten-seconds.wav"></audio>`);
+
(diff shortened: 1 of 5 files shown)


miga-heygen (approved):
Reviewed the diff and tests. The fix correctly marks variable-bound media whose duration was inferred from the fallback source (`data-hf-inferred-duration`), and only uses browser intrinsic duration for those marked elements — explicit authored durations still take the old bounded/clamped path. The reconciliation rule (`reconcileBrowserMediaEnd`) replaces the fallback window only when both the source changed and the duration was inferred, otherwise falls back to the existing min-clamp behavior, so authored trims and non-variable media are unaffected. Test coverage (longer-inferred, longer-authored, shorter-inferred) exercises exactly the cases described in the bug report. All CI green. LGTM.

PR #3649, opened 2026-09-04
fix(producer): scale high-quality encode timeout

miguel-heygen:
## What

Long high-quality disk-frame renders now receive a 24× source-duration FFmpeg budget instead of the draft/standard 4× budget. The reported 331.273-second render receives 7,950,552 ms rather than being killed at 1,325,092 ms.

## Why

High-quality CPU encoding can remain healthy and make progress beyond the quality-blind wall-clock estimate. Killing it after completed frame capture wastes the most expensive stage and removes the captured frames during normal cleanup.

Related: #2419

## How

The existing duration scaler remains the single timeout owner. High quality multiplies its established 4× budget by six; draft and standard stay unchanged. A larger `FFMPEG_ENCODE_TIMEOUT_MS` operator override still wins.

## Test plan

- [x] Exact 331.273-second high-quality regression receives 7,950,552 ms.
- [x] Standard-quality control remains at 1,325,092 ms.
- [x] Larger operator timeout remains unchanged.
- [x] Encode-stage suite passes (16/16).

### DIFF
--- a/packages/producer/src/services/render/stages/encodeStage.test.ts
+++ b/packages/producer/src/services/render/stages/encodeStage.test.ts
@@ -259,6 +259,65 @@ describe("runEncodeStage config plumbing", () => {
     });
   });
 
+  it("gives the reported long high-quality encode a 24x source-duration budget", async () => {
+    const { runEncodeStage } = await import("./encodeStage.js");
+    const input = makeInput();
+
+    await runEncodeStage(
+      makeInput({
+        job: {
+          ...input.job,
+          config: { ...input.job.config, quality: "high" },
+          duration: 331.273,
+        },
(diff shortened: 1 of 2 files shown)


miga-heygen (approved):
Simple, well-tested change: multiplies the existing 4x source-duration encode timeout budget by 6 (24x total) for high-quality renders only; draft/standard keep the original 4x, and a larger operator override via `FFMPEG_ENCODE_TIMEOUT_MS` still wins since the max-of comparison against `engineCfg.ffmpegEncodeTimeout` is unchanged. Tests cover the exact regression numbers (331.273s -> 7,950,552ms high / 1,325,092ms standard) plus the operator-override-preserved case. CI green. LGTM.

PR #2392, opened 2026-07-14
fix(engine): preserve mono audio level

miguel-heygen:
## Summary
- detect mono sources before audio preparation/extraction
- duplicate mono center audio into both stereo channels explicitly, avoiding FFmpeg's implicit -3 dB rematrix
- preserve the existing `-ac 2` conversion path for stereo/multichannel inputs and probe failures
- add a real-FFmpeg regression test that compares source and mixed mean volume

## Reproduction
A mono WAV at unity volume rendered through CLI 0.7.39 and current 0.7.56 measured 3.0–3.1 dB quieter in the final AAC track. The mixer forced `-ac 2`, invoking FFmpeg's default mono-to-stereo pan law.

## Verification
- regression test failed before fix at -3.1 dB and passes after fix
- audio mixer tests: 13/13 passed
- `bun run --filter @hyperframes/engine typecheck`
- format and lint checks passed
- branch-built CLI exact repro: source mean -27.1 dB / peak -24.1 dB; final stereo AAC mean -27.1 dB / peak -24.0 dB

PR #3650, opened 2026-09-04
fix(engine): preserve mono narration loudness

miguel-heygen:
## What

Mono-only renders now keep a mono delivery channel and preserve source integrated loudness. Mixed sessions still produce stereo when any native-stereo track requires it, with the existing unity mono amplitude preserved in that stereo mix.

## Why

The mixer previously duplicated every mono source into left and right at unity before it knew the final layout. That preserved per-channel amplitude but raised mono-only programme loudness by about 3 LU, turning a measured -21.8 LUFS narration into approximately -18.8 LUFS delivery.

Related: #2392

## How

Prepared tracks retain their channel count until the final mix. Mono is upmixed with the established unity pan only when another input requires stereo; otherwise the AAC sidecar remains mono. Mono float group buses are decoded through their WAVE_FORMAT_EXTENSIBLE subtype so group FX retain the same channel-preserving path.

## Test plan

- [x] Real FFmpeg test preserves mono-only channel count and integrated loudness within 0.5 LU.
- [x] Mixed mono/native-stereo control preserves stereo output and mono mean amplitude.
- [x] Main mixer suite passes (61/61).
- [x] Level and grouping suites pass (16/16), including group FX and over-unity float buses.

### DIFF
--- a/packages/engine/src/services/audioFxRender.ts
+++ b/packages/engine/src/services/audioFxRender.ts
@@ -61,6 +61,11 @@ function readWavChunks(buf: Buffer): {
       head.channels = buf.readUInt16LE(body + 2);
       head.sampleRate = buf.readUInt32LE(body + 4);
       head.bits = buf.readUInt16LE(body + 14);
+      // FFmpeg writes mono pcm_f32le as WAVE_FORMAT_EXTENSIBLE. Its subformat
+      // tag carries the actual PCM kind in the first two GUID bytes.
+      if (head.format === 0xfffe && size >= 40) {
+        head.format = buf.readUInt16LE(body + 24);
+      }
     } else if (id === "data") {
       data = buf.subarray(body, Math.min(buf.length, body + size));
       // The payload is the rest of the file for anything the mixer writes, and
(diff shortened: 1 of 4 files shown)


miga-heygen (changes requested):
The channel-preservation logic and the WAVE_FORMAT_EXTENSIBLE fix in audioFxRender.ts look correct and well-reasoned (mono stays mono until the final mix, unity pan applied only when a stereo track forces the layout), and the level tests you added pass. But CI is currently red for reasons that trace back to this PR's own changes, not flakiness:

1. `Producer: unit tests` fails the repo's `ffprobeArgvContract` invariant test: the new `channelCount()` helper added in `audioMixer.level.test.ts` calls ffprobe as `["-v","error",...,"-of","csv=p=0", path]` without the `--` terminator immediately before the input path. Every other ffprobe call site in the repo is required to terminate options with `--` before the input (argv-injection hardening); this new call site violates that contract and fails `src/utils/ffprobeArgvContract.test.ts`.

2. `regression` / `regression-shards (shard-6)` show a real regression: the `missing-host-comp-id` suite fails its audio comparison (`passed:false, correlation:0.803`), which is below the pass threshold. That suite mixes mono/stereo tracks, so this looks like the new mono/stereo channel-preservation path is actually changing the audio output for that composition versus the reference fixture — worth double-checking whether the reference audio for that suite needs to be regenerated, or whether there's a real behavioral difference for that specific track layout.

Please fix the ffprobe `--` terminator in the test helper and investigate the `missing-host-comp-id` audio regression before merge. Happy to re-review once CI is green.

— Miga

PR #1577, opened 2026-06-19
fix(runtime): guard dotlottie v1 seek against zero duration

calcarazgre646:
## Problem

The dotlottie-web v1 seek branch (`lottie.ts`) computed `(time / duration) * 100` with `duration = anim.duration ?? 1`. Nullish coalescing does not replace `0`, so a player reporting `duration: 0` (not yet loaded, which is exactly the bootstrap-time `discover` to `seek` window) produced `(time / 0) * 100 = NaN`, and `seek(NaN)` blanked the first frame at composition time 0. The v2 branch directly above already guards with `if (totalFrames > 0)`; the v1 sibling did not.

## Change

Mirror the v2 guard: default `duration` to 0 and seek only when it is positive, with the percentage clamped to `[0, 100]`. No change for a loaded player with a known duration.

## Tests

Two cases in `lottie.test.ts`: a v1 player with a known duration still seeks by percentage; a v1 player reporting duration 0 is not seeked (no `seek(NaN)`). Verified load-bearing (reverting the fix makes the zero-duration case call `seek(NaN)`). Full core suite green.

PR #3651, opened 2026-09-04
fix(runtime): wait for a finite dotlottie duration

jrusso1020:
A dotlottie v1 player reporting zero duration could receive `seek(NaN)` at time zero and blank its first frame. Wait for a positive finite duration before seeking, then preserve the existing percentage clamping.

Builds on #1577 by @calcarazgre646, refreshed against current main with invalid-duration and delayed-load coverage.

Validation: regression tests failed before the fix; all 171 adapter tests pass, runtime typecheck passes, and all pre-commit checks pass.

### DIFF
--- a/packages/core/src/runtime/adapters/lottie.test.ts
+++ b/packages/core/src/runtime/adapters/lottie.test.ts
@@ -109,6 +109,30 @@ describe("lottie adapter", () => {
       expect(player.setCurrentRawFrameValue).toHaveBeenCalledWith(59);
     });
 
+    it.each([0, -1, NaN, Infinity, undefined])(
+      "does not seek a dotlottie v1 player with duration %s",
+      (duration) => {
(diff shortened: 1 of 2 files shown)


miga-heygen (approved):
Reviewed at exact `fbffbf7cd`.

Correct fix. When `anim.duration` is `0`, `undefined`, `NaN`, or `Infinity`, the old code computed `seek(NaN)` or `seek(Infinity)`, blanking the first frame. The guard (`Number.isFinite(duration) && duration > 0`) is the right predicate — it rejects all degenerate values and waits for the player to report a real duration before seeking.

Two tests cover both the degenerate cases (parameterized over `0, -1, NaN, Infinity, undefined`) and the delayed-load scenario where duration transitions from 0 to a real value mid-session. The seek assertions (`[50], [0], [100]`) verify percentage clamping still works. Clean `+29/-3`.

— Miga

PR #3652, opened 2026-09-04
fix(fonts): retain non-Latin subsets for bundled weights

jrusso1020:
Bundled fonts contain only the Latin subset, so discarding other subsets at the same weight drops Japanese, Cyrillic, and other scripts. Retain those subsets and declare the bundled Latin range while preserving canonical aliases, source grouping, and bundled glyph precedence for overlapping/text-subset responses.

Builds on #3086 by @akzarma, with regression coverage for current alias supplementation.

Validation: 78 font tests pass; both new regressions fail on unchanged main. Changed files pass oxlint/oxfmt. No distributed render performed.

### DIFF
--- a/packages/producer/src/services/deterministicFonts-bundledSubset.test.ts
+++ b/packages/producer/src/services/deterministicFonts-bundledSubset.test.ts
@@ -1,0 +1,131 @@
+/**
+ * Regression test for the coverage a bundled face claims.
+ *
+ * `scripts/generate-font-data.ts` embeds the `-latin-` subset file of every
+ * canonical family, so an embedded face carries only Google's `latin` subset.
+ * The emitted `@font-face` used to omit `unicode-range` — advertising full
+ * coverage — and the supplementary Google fetch then skipped every subset of a
+ * weight the bundle "covered". A bundled family therefore could not render the
+ * scripts its own subset omits: `Noto Sans JP` weight 400 carries 218
+ * codepoints with no kana and no kanji.
+ *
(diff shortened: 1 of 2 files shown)


miga-heygen (approved):
Reviewed at exact `58f3a339e`. Clean fix for bundled font subset coverage.

**The bug:** embedded fonts carry only the Fontsource `-latin-` subset but their `@font-face` rules omitted `unicode-range`, falsely claiming full coverage. The Google fetch filter then skipped *all* subsets for bundled weights — so families like Noto Sans JP lost their Japanese/CJK faces entirely.

**The fix:**
- Bundled faces now declare `unicode-range: U+0000-00FF, ...` (the exact Latin subset they ship)
- `isBundledSubsetRange()` identifies fetched faces whose range matches the bundled Latin — those are redundant and skipped
- Non-Latin subsets (Japanese, Cyrillic, etc.) and `text=` responses (no range) for the same weight are retained
- Bundled rules emitted last so they win for overlapping Latin glyphs (CSS last-match precedence), preserving deterministic output

**Verified:**
- `isBundledSubsetRange(undefined)` → `false` — text-subset faces without a range are correctly kept
- Normalization (`lowercase` + whitespace strip) handles minor formatting differences
- Test coverage: CJK retention for Noto Sans JP, Latin dedup for Inter + Cyrillic supplement, ordering assertion, `unicode-range` present on every emitted face
- Both regression tests fail on unchanged main (confirmed in PR body)

LGTM. `+163/-6`.

— Miga

PR #1578, opened 2026-06-19
fix(runtime): size a clip from data-end when it has no data-duration

calcarazgre646:
## Problem

`collectRuntimeTimelinePayload`'s clip loop (`timeline.ts`) resolved a clip's duration from `data-duration`, then fell straight through to the inherited (composition) duration when it was absent. So a clip authored with `data-start` + `data-end` but no `data-duration` (a valid, compiler-emitted shape that the linter only warns about) was sized `start..compositionEnd` instead of `start..end`. A `data-start="2" data-end="8"` clip under a 20s composition rendered as an 18s bar (20 - 2) on the Studio timeline instead of 6s, overrunning every later clip on its track.

The scene loop (same file) and `startResolver.resolveDurationForElement` already read `data-end`; the clip loop was the sibling that did not.

## Change

Read `data-end` as the duration source between `data-duration` and the inherited fallback, mirroring the scene loop (`Math.max(0, end - start)`). No change for clips that carry `data-duration`.

## Tests

A clip with `data-start="2" data-end="8"` and no `data-duration` under a 20s composition now reports duration 6, not 18. Verified load-bearing (reverting the fix produces 18). Full core suite green.

PR #3654, opened 2026-09-04
fix(runtime): honor authored clip end times in the timeline

jrusso1020:
A clip authored with `data-start="2" data-end="8"` under a 20-second composition appeared as an 18-second timeline bar. Resolve explicit and preserved end attributes before inferred duration, keeping explicit duration precedence and the parent-local clock for nested clips.

Builds on #1578 by @calcarazgre646, updated for current authored timing metadata and nested composition offsets.

Validation: four regression cases fail before the fix; all 115 timeline/start-resolver/authored-timing/clip-tree tests pass, runtime typecheck passes, and all pre-commit checks pass.

### DIFF
--- a/packages/core/src/runtime/timeline.test.ts
+++ b/packages/core/src/runtime/timeline.test.ts
@@ -130,6 +130,52 @@ describe("collectRuntimeTimelinePayload", () => {
     expect(result.clips[0].kind).toBe("element");
   });
 
+  it.each(["data-end", "data-hf-authored-end"])(
+    "sizes end-only clips from %s before inherited or media duration",
+    (endAttribute) => {
+      document.body.innerHTML = `<div data-composition-id="main" data-duration="20">
+        <div id="text" data-start="2" ${endAttribute}="8"></div>
+        <video id="video" data-start="3" ${endAttribute}="7"></video>
+      </div>`;
+      const video = document.querySelector("video")!;
+      Object.defineProperty(video, "duration", { value: 15 });
+      const result = collectRuntimeTimelinePayload(defaultParams);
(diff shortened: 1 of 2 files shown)


miga-heygen (approved):
Reviewed at exact `4e33a4c7f`.

**What it does:** Clips authored with `data-end` or `data-hf-authored-end` now resolve to the correct duration before falling through to sub-composition, media, or inherited duration. The end attributes use the parent-composition's local clock via `start - (compositionContext.inheritedStart ?? 0)`, matching how authored starts are already resolved.

**Precedence chain verified:** explicit `data-duration` → end attributes (this PR) → sub-composition duration → media duration → inherited composition duration. The insertion point is correct — `parseElementDurationAttr` gates the new block, so explicit duration always wins.

**Timing arithmetic checked:**
- Nested case: parent at offset 10, child `start=2 end=8` → absolute start=12, local start=2, duration=8−2=6 ✓
- Invalid end (`NaN`): `finiteNumber` returns null, falls through to inherited duration ✓
- Preserved fallback: `data-end=3` with `start=4` rejected (3 < 4), `data-hf-authored-end=8` accepted → duration=4 ✓
- Duration precedence: `data-duration=3` with `data-end=10` → parseElementDurationAttr returns 3, end block skipped ✓

`+55/-0`, four well-structured regression tests, all 115 timeline tests pass per the PR. Clean fix.

— Miga

jrusso1020:
Closing after the current-implementation audit. End-only source HTML is now rejected by deprecated_data_end; supported source uses data-duration. The compiler generates data-end from duration, and runtime preserves authored duration when normalizing timing. Those current paths already produce the correct clip length. This patch only extends deprecated-input compatibility, with no supported current caller identified, so it does not belong in the fix batch.

PR #1566, opened 2026-06-18
feat(core): shared gsapCascade helper + server-delete GSAP cascade (WS-E/2)

vanceingalls:
## What

Brief description of the change.

## Why

Why is this change needed?

## How

How was this implemented? Any notable design decisions?

## Test plan

How was this tested?

- [ ] Unit tests added/updated
- [ ] Manual testing performed
- [ ] Documentation updated (if applicable)

PR #3655, opened 2026-09-04
fix(studio-server): cascade GSAP cleanup when deleting subtrees

jrusso1020:
Deleting a parent through Studio's server fallback left its GSAP tweens and descendant tweens in the source. Share subtree cleanup with the HTML parser, matching each element's DOM id and stable id and visiting nested composition-template scripts. Reparse after each removal so multiple tweens are all removed.

If a selector still matches a surviving element, retain its animations to protect repeated composition instances. This intentionally handles direct id/stable-id selectors; broad or computed selectors remain unchanged. SDK undo/redo behavior is unchanged.

Builds on #1566 by @vanceingalls, updated for the parser/server package split.

Validation: 197 targeted parser/source-mutation/file-route tests passed (2 existing todo); parser and studio-server typechecks, oxlint/oxfmt, and commit hooks passed.

### DIFF
--- a/packages/parsers/src/htmlParser.test.ts
+++ b/packages/parsers/src/htmlParser.test.ts
@@ -580,6 +580,19 @@ describe("removeElementFromHtml", () => {
     expect(updated).toContain('id="el2"');
   });
 
+  it("cascades DOM and stable ids for every descendant", () => {
+    const html = `<!doctype html><html><body>
+      <div id="parent"><div id="box" data-hf-id="hf-box"></div></div>
+      <script>const tl = gsap.timeline();
+        tl.to("#parent", { x: 10 }); tl.to("#box", { x: 20 });
+        tl.to('[data-hf-id="hf-box"]', { x: 30 });
+      </script></body></html>`;
+    const updated = removeElementFromHtml(html, "parent");
+    expect(updated).not.toContain("#parent");
+    expect(updated).not.toContain("#box");
+    expect(updated).not.toContain("hf-box");
+  });
+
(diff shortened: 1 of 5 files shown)


miga-heygen (approved):
Reviewed at exact `69cc4ffef`.

GSAP deletion cascade — when an element subtree is deleted, all GSAP tweens targeting any descendant's `#id` or `[data-hf-id]` selector are now removed, not just the root element's.

**Key correctness checks:**

- **Survivor walk prevents false removal**: after removing the element, `walkCompositionDescendants(doc, survivor)` deletes any selector from the removal set that a surviving element still uses. This prevents erasing a tween that targets a shared selector in a different composition instance (e.g. two `<template>` instances with the same `#box`). Test "retains shared selectors used by a surviving composition instance" validates this.

- **Positional re-parse loop preserved**: animation writer IDs are count-based and renumber after each removal. The `for (;;)` loop correctly re-parses after each tween deletion so later matches aren't skipped by stale IDs.

- **Nested composition templates**: `findScriptElementsDeep` + `walkCompositionDescendants` traverse into `<template data-composition-id>` boundaries, so tweens inside nested composition templates are also cascaded. Test "cascades a stable-id deletion into nested composition template scripts" validates this.

- **No-op when element absent**: if `findTargetElement` returns null, the function is never called and the source is returned unchanged. Test "does not strip scripts when the requested element is absent" validates this.

- **Dead comment removed**: the `ponytail` comment in `useElementLifecycleOps.ts` documented the exact gap this PR closes — correctly removed.

`+103/-40`, 5 files. LGTM.

— Miga

PR #3656, opened 2026-09-04
fix(captions): preserve zero plate grain in generated postfx

jrusso1020:
Themes specifying `plate.grain: 0` still received the default FFmpeg noise filter. Preserve zero for both regular and heroless themes and omit noise entirely, while keeping level 5 when grain is absent.

Builds on #3032 by @miguel-heygen, updated for the heroless compiler changes in #3632. Regenerates the skills manifest.

Validation: both zero-grain regressions fail before the fix; all 586 skill tests pass, including six postfx cases and all 25 heroless themes. Syntax, changed-file lint/format, manifest sync, and pre-commit checks pass.

### DIFF
--- a/skills-manifest.json
+++ b/skills-manifest.json
@@ -2,7 +2,7 @@
   "source": "heygen-com/hyperframes",
   "skills": {
     "embedded-captions": {
-      "hash": "1f29e0f6c77da4f5",
+      "hash": "1677ab00946eb80c",
       "files": 142
     },
     "faceless-explainer": {
--- a/skills/embedded-captions/scripts/make-theme.cjs
+++ b/skills/embedded-captions/scripts/make-theme.cjs
@@ -8574,7 +8574,9 @@ if (!bodyInBg || fx.html || fx.js || setp.fgHtml) {
 }
 
 // _postfx.sh: plate reaction after the matte composite (subject+text move as one)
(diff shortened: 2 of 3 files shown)


miga-heygen (approved):
Reviewed at `55cd6dcab`.

The `||` → `??` fix is the right shape: `(dna.plate || {}).grain || 5` treated `grain: 0` as falsy and fell through to 5. Nullish coalescing preserves the explicit zero.

Verified:
- `grain === 0` → `noise = ""` → FFmpeg filter chain ends at `${rgba}[v]` with no trailing comma — correct
- `grain > 0` → `noise = ",noise=alls=${grain}:allf=t+u"` — leading comma keeps the filter chain valid
- `grain === undefined` → `grain ?? 5` = 5 → default noise level preserved
- Heroless path uses `dna.plate?.grain ?? 5` (safe optional chaining + nullish coalescing)
- Non-heroless path passes through `dna.plate || {}` unchanged — grain propagation is the caller's concern there
- All 6 test permutations (0/3/undefined × heroless/hero) cover the matrix; the `doesNotMatch(/noise=/)` assertion for zero-grain and the `doesNotMatch(/,\s*\[v\]/)` trailing-comma guard are both discriminating

`+55/-5`. LGTM.

— Miga

PR #1747, opened 2026-06-26
feat(core): duck() programmatic audio ducking (re-submit of #1337)

mvanhorn:
Re-submit of #1337 along the lines you suggested: instead of new `data-duck`/`data-duck-fade`/`data-role` HTML attributes, this adds a programmatic `duck(musicTrack, voiceTracks, options)` helper that writes deterministic compile-time volume keyframes into the same envelope used by hand-authored fades. No new HTML schema attributes; `htmlCompiler`/`audioElementParser`/the schema docs are untouched.

- `amount` accepts linear gain or dB (default `-12dB`), `fade` ramp defaults to 0.3s, optional GSAP-compatible `timeline` receives the ramps.
- New `packages/core/src/runtime/audioDucking.ts` + tests; exported from `@hyperframes/core`.
- Full core suite green locally; oxlint/oxfmt/fallow clean on the changed files.

Happy to attach a rendered music-under-voiceover demo if useful.

PR #3657, opened 2026-09-04
feat(audio): add deterministic music ducking helper

jrusso1020:
Add `duck(music, voices, { timeline, amount, fade })` to lower music during speech using the existing preview/export volume envelopes. Rebuilds #1747 by @mvanhorn for current timing and gain behavior: zero-fade music restores, trimmed/rate-adjusted voices end correctly, and gains above unity survive.

The helper uses resolved composition timing and leaves the current clip-relative preview cache intact. The docs describe timeline registration and the one-microsecond linear representation of zero-fade steps.

Validation: 131 core tests and 11 PCM envelope tests pass, including real GSAP reverse-seek/probe coverage. Four regressions fail with the original helper. Core build, pre-commit typecheck/lint/format, Mintlify validation and broken-link checks pass. No full video export performed.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -122,6 +122,7 @@
         "@types/jsdom": "^28.0.0",
         "@types/node": "^25.0.10",
         "@vitest/coverage-v8": "^3.2.4",
+        "gsap": "^3.15.0",
         "jsdom": "^29.0.0",
         "tsx": "^4.21.0",
         "typescript": "^5.0.0",
--- a/docs/packages/core.mdx
+++ b/docs/packages/core.mdx
@@ -134,6 +134,37 @@ if (!result.isValid) {
 For the full composition lint result, use [`@hyperframes/lint`](/packages/lint)
 or run `npx hyperframes lint`.
 
+## Lower music under speech
+
+`duck()` adds music volume ramps around voice clips on your paused GSAP timeline:
+
+```ts
+import { duck } from "@hyperframes/core";
+
+const timeline = gsap.timeline({ paused: true });
+const music = document.querySelector<HTMLAudioElement>("#music");
+const voice = document.querySelector<HTMLAudioElement>("#voice");
+if (music && voice) {
+  duck(music, voice, { timeline, amount: "-12dB", fade: 0.3 });
+}
+window.__timelines["my-composition"] = timeline;
+```
+
+Pass media elements with resolved composition timing (`data-start` and
+`data-duration` or `data-end`). When duration comes from loaded media metadata,
+source trimming and playback rate are respected. Register the timeline under
+your composition ID so preview and export can probe its volume changes.
+
+`amount` accepts a linear multiplier or a dB string; `fade` is seconds on each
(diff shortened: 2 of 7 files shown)


jrusso1020:
Found one playback edge case at `67ee6bca2`: a DOM music element with `data-volume="2"` makes the generated GSAP timeline write native `volume=2`, throwing `IndexSizeError` outside the probe wrapper. Reproduced with music 0–10s, voice 2–3s, fade 0.3 and `timeline.totalTime(1.7, true)`. Fix and a real DOM/GSAP regression are in progress; above-unity probe/export gain must remain supported.

miga-heygen (approved):
Reviewed at exact `67ee6bca2`. `+620/-0`, new deterministic music ducking helper.

**What it does:** `duck()` computes overlap windows between music and voice tracks, generates linear volume ramps (keyframes + GSAP timeline writes), and integrates with the existing volume envelope pipeline for both preview and render.

**Verified:**

- **dB/linear conversion:** `amountToGain` correctly handles dB strings (`"-12dB"` → `10^(-12/20)` ≈ 0.251), negative numbers (as dB), positive (as linear gain). `amount: 0` ducks to silence, which is intentional.
- **Interval merging:** voice gaps shorter than `2 × fade` are merged into a single ducked window, preventing rapid up/down ramps that sound like pumping. Test at line 104 confirms this.
- **Zero-fade mode:** microsecond boundary points (`±0.000001`) create step transitions that the piecewise linear envelope can represent without collapsing. The `writeKeyframes` path uses `set()` for these.
- **Element resolution:** reads `data-start`, `data-duration`, `data-end`, `data-volume` from media elements, computes natural duration via `resolveNaturalMediaTimelineDuration` (respects source trimming + playback rate), loop-aware. Does NOT write duck-related schema attributes back to the DOM — verified by the "no schema attributes" test.
- **Timeline writes:** `set()` for instant jumps, `to()` with `ease: "none"` + `overwrite: false` for linear ramps. Won't clobber hand-authored tweens on other properties.
- **Ramp clamping:** fade ramps can't exceed `duration / 2`, and are clipped to the music track's bounds.
- **Edge cases covered by tests:** silent/muted/no-audio voice tracks filtered out, NaN start → discarded, non-overlapping → empty, `data-duration` before `data-end` precedence, partial numeric rejection, looping elements use authored end, reverse seek stability in GSAP probe.
- **Integration:** WAV baking test in `audioVolumeEnvelope.test.ts` confirms the ducking keyframes produce correct sample values through the full render pipeline (boost → duck → restore).
- **`gsap` devDependency** is test-only — production code has no GSAP import.

LGTM — clean design, thorough tests, well-documented.

— Miga

jrusso1020:
Closing after comparing this proposal with the current audio workflow. `skills/hyperframes-audio/scripts/carve.mjs` and Studio already share `analyseCarveDuck`, writing editable automation that follows speech level and pauses. This PR adds a parallel GSAP clip-window ducking API without a demonstrated missing consumer. We should extend the existing audio tools if a concrete gap appears. My initial triage checked for the old helper rather than this capability overlap.

PR #1729, opened 2026-06-26
fix: correct misleading lint hint for crypto.getRandomValues

AGRO-CODEX:
## What

Corrects the `fixHint` message for the `crypto.getRandomValues()` pattern in the `non_deterministic_code` lint rule (`packages/core/src/lint/rules/core.ts`).

## Why

The previous hint said "Remove time-dependent code. Use a seeded PRNG for deterministic renders." but `crypto.getRandomValues()` is not time-dependent — it is a source of randomness. The "time-dependent" label is accurate for `Date.now()`, `new Date()`, and `performance.now()`, but is factually wrong for the crypto case.

## Change

The new hint follows the same pattern as the `Math.random()` entry:

> Replace with a seeded PRNG (e.g. a simple mulberry32) so renders are deterministic across frames.

No logic changes — lint behaviour, severity, and error codes are untouched.

PR #3659, opened 2026-09-04
fix(lint): give randomness guidance for crypto values

jrusso1020:
`hyperframes lint` still labels `crypto.getRandomValues()` as time-dependent code in the current extracted lint package. Give it the same seeded-PRNG guidance as `Math.random()` while retaining clock-specific guidance for `Date.now()`.

Builds on #1729 by @AGRO-CODEX. Validation: 59 core-rule tests pass; the new semantic guidance regression fails on unchanged main. Required lint, format and typecheck hooks pass.

### DIFF
--- a/packages/lint/src/rules/core.test.ts
+++ b/packages/lint/src/rules/core.test.ts
@@ -601,6 +601,25 @@ describe("core rules", () => {
   });
 
   describe("non_deterministic_code", () => {
+    it("gives randomness guidance for crypto and clock guidance for wall time", async () => {
+      const result = await lintHyperframeHtml(`<html><body>
+        <div data-composition-id="c1" data-width="1920" data-height="1080"></div>
+        <script>
+          crypto.getRandomValues(new Uint32Array(1));
+          Date.now();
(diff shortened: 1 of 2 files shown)


miguel-heygen (approved):
Reviewed `51e21c7ccd491b2bf776188eae8a1441bb8a944e`; no actionable findings.

`packages/lint/src/rules/core.ts:481` aligns crypto randomness guidance with `Math.random()` without changing detection or severity. `packages/lint/src/rules/core.test.ts:604` pins seeded-PRNG guidance, rejects the old time-dependent wording, and preserves wall-clock guidance for `Date.now()`.

Traced the rule through the shared lint package and CLI formatter. The successful CI Test job explicitly reports 59 core-rule tests and 548 lint-package tests passing; lint, format, typecheck, build, and CLI smoke also passed. No prior substantive reviews to duplicate. Validation used exact-head source and CI logs; no local dependency installation or test run.

**Verdict:** APPROVE
**Reasoning:** The wording fix is narrowly scoped and the semantic regression test protects both randomness and clock guidance without adding unnecessary abstractions.

— Magi

PR #1579, opened 2026-06-19
fix(shader-transitions): show the from-scene at gravitational-lens start

calcarazgre646:
## Problem

The `gravitational-lens` frag (`registry.ts`) computes `lensed = vec3(r, A.g, b) * horizon`, where `horizon = smoothstep(0., .3, dist / (1. - u_progress*.85 + .001))` darkens the center toward black. That term is not gated by `u_progress`, so at `u_progress = 0` the center is multiplied by ~0 (pure black at the center, fading to 1 only near the corners) while the rest of the math already resolves to the untouched from-scene. The viewer sees a black-hole vignette flash at the transition start (a paused seek at the start, and the first rendered frame, both hit progress 0). The 13 other shaders in the file collapse cleanly to the from-scene at progress 0; this is the only one that does not.

## Change

Ramp the horizon darkening in from progress 0: `lensed = vec3(r, A.g, b) * mix(1., horizon, smoothstep(0., .3, u_progress))`. At progress 0 the factor is 1 (clean from-scene); by progress 0.3 the full lens darkening is in effect, unchanged from before. The progress-1 endpoint is unaffected (the outer `mix(..., smoothstep(.3, .9, u_progress))` already selects the to-scene there).

## Verification

The center darkening factor at progress 0 goes from 0 (black) to 1 (from-scene); progress 0.3+ and the progress-1 endpoint are unchanged. This package has no headless-GL test path, so shader behavior here is verified by the math and visual inspection, consistent with how the other shaders in the file are maintained.

PR #3660, opened 2026-09-04
fix(shader-transitions): preserve gravitational lens start frame

jrusso1020:
The active gravitational-lens shader darkens the source scene at progress zero, producing a black-hole flash before the transition begins. Ramp in horizon darkening through progress 0.3; preserve the existing effect from that point onward.

Builds on #1579 by @calcarazgre646. The current shader renderer still calls this registered shader; the CSS fallback does not replace its WebGL path.

Validation: rendered the production shader and `renderShader` implementation in Chrome WebGL on ANGLE Metal / Apple M4 Max. Main's start differs from the source by up to 183 RGB levels; this change exactly matches both endpoint textures. Progress 0.3, 0.5, 0.75, 0.9 and 1 are byte-identical to main, and reverse/repeated samples match. Visually checked start, early and middle frames. All 12 shader-package tests and required hooks pass.

### DIFF
--- a/packages/shader-transitions/src/shaders/registry.ts
+++ b/packages/shader-transitions/src/shaders/registry.ts
@@ -87,6 +87,7 @@ const shaders: Record<string, ShaderDef> = {
   },
 
   "gravitational-lens": {
+    // Begin with the untouched source; full horizon darkening starts at progress 0.3.
     frag:
       H +
       "void main(){vec4 B=texture2D(u_to,v_uv);" +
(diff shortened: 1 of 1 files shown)


miguel-heygen (approved):
The ramp in `packages/shader-transitions/src/shaders/registry.ts:102` restores the source frame at progress zero and preserves the existing lens from 0.3 onward. No blocking findings.

Independently rendered the exact-head production shader and `renderShader` in Chromium/ANGLE SwiftShader: both endpoints matched their textures exactly; samples at 0.3, 0.5, 0.75, 0.9 and 1 matched the original shader byte-for-byte; reverse/repeated samples were deterministic. This is software WebGL evidence, separate from the author’s reported Apple GPU run. CI checks are passing.

**Verdict:** APPROVE
**Reasoning:** The localized arithmetic fix removes initial darkening without changing the later transition or adding state; direct runtime comparisons verify the intended behavior at `041f7bf`.

— Magi

PR #2407, opened 2026-07-14
fix(skills): consolidate TTS provider behavior and errors

miguel-heygen:
## Summary
- forward requested speech speed to Kokoro
- allow explicit offline TTS providers when cached HeyGen OAuth is expired
- surface concrete HeyGen provider errors from both the shared audio engine and standalone TTS command
- port the still-relevant behavior from the retired hyperframes-media path into media-use
- supersede #2335, #2330, and #1887

## Verification
- media-use and faceless-explainer audio tests: 29 passed
- generated skills manifest refreshed
- diff whitespace check

PR #2335, opened 2026-07-13
fix(media-use): forward Kokoro speech speed

miguel-heygen:
## Summary
- forward the audio request `speed` value to the local Kokoro `hyperframes tts` subprocess
- centralize Kokoro CLI argument construction
- add regression coverage for a non-default `1.15` speed

## Reproduction
A media-use audio request with `provider: kokoro` and `speed: 1.15` reaches `synthesizeOne`, but the spawned command omits `--speed`; generated lines therefore retain the same duration as speed 1.0. The underlying `hyperframes tts` command already supports `--speed`.

## Verification
- `node --test skills/media-use/audio/scripts/lib/tts.test.mjs skills/media-use/audio/scripts/lib/tts.spawn.test.mjs` (21 tests)
- `npm run test:skills` (230 tests)
- oxfmt and oxlint on changed files
- `git diff --check`

PR #3665, opened 2026-09-04
fix(skills): forward Kokoro TTS speed

miguel-heygen:
## What

Kokoro synthesis now receives the resolved TTS speed at the final subprocess boundary. A requested `--speed 1.16` reaches `npx hyperframes tts` as `--speed 1.16` instead of silently using Kokoro’s 1.0 default.

## Why

The faceless adapter and shared audio request preserved speed correctly, but Kokoro argv construction dropped it. Synthesis still exited successfully, leaving narration audio and word timing at the wrong pace.

Related: #2335
Related: #2407

## How

One Kokoro argument builder owns voice, output, speed, and optional language ordering. `synthesizeOne` passes its already-resolved speed into that builder, and the default 1.0 value is emitted explicitly under a tested contract.

## Test plan

- [x] Exact non-default argv includes `--speed 1.16`.
- [x] Default argv explicitly includes `--speed 1`.
- [x] Full provider helper suite passes (13/13).
- [x] Node syntax, oxlint, and oxfmt checks pass.
- [ ] Live Kokoro model synthesis (not required; subprocess boundary is deterministic).

### DIFF
--- a/skills/media-use/audio/scripts/lib/tts.mjs
+++ b/skills/media-use/audio/scripts/lib/tts.mjs
@@ -241,6 +241,22 @@ save(audio, sys.argv[3])
 // [{text,start,end}] array for HeyGen (native), or null for ElevenLabs/Kokoro
 // (caller must transcribeWav). Never throws; failures return { ok:false, error }
 // where `error` states WHY (so the caller can surface it, not a bare "TTS failed").
+export function buildKokoroTtsArgs({ textPath, voiceId, wavRel, lang = "en", speed = 1.0 }) {
+  const args = [
+    "hyperframes",
+    "tts",
+    textPath,
+    "--voice",
+    voiceId,
+    "--output",
+    wavRel,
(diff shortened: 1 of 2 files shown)


miga-heygen (changes requested):
The fix itself is correct: `buildKokoroTtsArgs` now always emits `--speed`, and `synthesizeOne` forwards its already-resolved `speed` into it. Tests cover both the non-default (1.16) and explicit-default (1) cases correctly.

However, CI is red: `Skills: manifest in sync` fails because `skills-manifest.json` wasn't regenerated after editing `skills/media-use/audio/scripts/lib/tts.mjs` (hash mismatch for the `media-use` skill: `00a2d26e22fc1ab1 -> ae2be19b68fe6e84`). Please run `bun run --cwd packages/cli gen:skills-manifest` and commit the updated `skills-manifest.json`, then this is good to merge.

— Miga

PR #1833, opened 2026-07-01
fix(cli): forward GIF and frame-format options in batch renders

dpersek:
## What

Batch renders now carry the parsed `--gif-loop` and `--video-frame-format` options into each row render.

## Why

`hyperframes render --batch` validated these flags, but the batch render base options did not include the parsed values. That meant row renders silently used defaults even when callers requested a finite GIF loop count or a specific source video frame extraction format. Single renders already forwarded both options correctly.

## How

- Added `gifLoop` and `videoFrameFormat` to the shared batch `renderOptionsBase`.
- Added a command-handler regression test that runs a one-row GIF batch render and asserts the mocked render job receives `format: "gif"`, `gifLoop: 3`, and `videoFrameFormat: "png"`.
- Mocked browser resolution in the test so the regression stays focused on option forwarding.

## Test plan

- [x] Unit tests added/updated
- [ ] Manual testing performed
- [ ] Documentation updated (not applicable)

Commands:

- `bun run --filter @hyperframes/cli test -- src/commands/render.test.ts -t "forwards gif loop and video frame format to batch row renders"`
- `bun run --filter @hyperframes/cli test -- src/commands/render.test.ts`
- `bun run --filter @hyperframes/cli typecheck`
- `bunx oxlint packages/cli/src/commands/render.ts packages/cli/src/commands/render.test.ts`
- `bunx oxfmt --check packages/cli/src/commands/render.ts packages/cli/src/commands/render.test.ts`
- `git diff --check`
- `bun run --filter @hyperframes/cli test`

PR #3666, opened 2026-09-04
fix(cli): preserve batch GIF and frame-format options

jrusso1020:
Batch renders validate `--gif-loop` and `--video-frame-format` but currently drop both before creating row jobs. Forwarding them makes a batch such as `--format gif --gif-loop 3 --video-frame-format png` preserve the same encoder and source-frame settings as a single render.

Ports Dustin Persek’s fix and command regression from #1833 onto current main, with co-author credit. The original branch contains unsigned commits, so this successor preserves the contribution in a signed commit.

Validation: the real command-handler regression failed before the fix (both job settings were undefined); all 87 render/batch tests pass after it. CLI typecheck, lint, formatting and commit hooks pass.


miguel-heygen (approved):
Reviewed `7cb72c8478efd8c7cffe0dfeda043700736dc0e4`. No correctness findings.

`packages/cli/src/commands/render/execute.ts:260` and `:268` preserve the validated settings in the shared batch-row options, covering both local and Docker execution. `packages/cli/src/commands/render.test.ts:1492` exercises the command handler and checks the resulting job settings. I traced both values through the producer request conversion to GIF encoding and source-video extraction.

Validation: static review of the complete diff and downstream consumers; no local runtime test rerun. Build, lint, typecheck, and CLI smoke passed; remaining CI was still running with no observed failures at review time. No prior reviews or comments.

**Verdict:** APPROVE
**Reasoning:** The two missing fields are forwarded at the correct shared boundary, with a focused regression covering the reported failure. Pending CI must complete before merge.

— Magi

PR #3669, opened 2026-09-04
fix(producer): isolate font cache temporary writes

jrusso1020:
Font cache writes use a predictable PID/timestamp filename and unconditionally remove it in `finally`. If another file already occupies that name, the exclusive write fails but cleanup deletes the other writer's file. This addresses [code-scanning alert 768](https://github.com/heygen-com/hyperframes/security/code-scanning/768).

Create a private temporary directory atomically inside the cache filesystem, then publish the compressed font with the existing atomic rename. Cleanup removes only the directory created by this invocation. Cache keys, font output, cache reuse, and best-effort behavior when the cache is unavailable remain unchanged.

Validation: reproduced the pre-existing-file deletion with a failing regression test before the fix; all 18 font compression and deterministic font tests now pass. Added coverage for failed cache publication and an unavailable cache directory. Core build, producer typecheck, changed-file lint/format, and all pre-commit checks pass. CI and CodeQL must pass before landing.

### DIFF
--- a/packages/producer/src/services/fontCompression.test.ts
+++ b/packages/producer/src/services/fontCompression.test.ts
@@ -1,5 +1,13 @@
-import { describe, expect, it } from "bun:test";
-import { existsSync, mkdtempSync, readFileSync, rmSync } from "node:fs";
+import { describe, expect, it, spyOn } from "bun:test";
+import {
+  existsSync,
+  mkdirSync,
+  mkdtempSync,
+  readFileSync,
+  readdirSync,
+  rmSync,
+  writeFileSync,
+} from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { compressToWoff2, fontToDataUri } from "./fontCompression.js";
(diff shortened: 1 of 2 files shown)


miguel-heygen (approved):
Reviewed `939bb170e46f95224790c7a7aeb23beea390192b`. No actionable findings.

`fontCompression.ts:67-84` now ties cleanup to the directory created by this invocation while retaining same-filesystem atomic publication and best-effort cache behavior. Both font embedding callers continue through this shared helper.

Validation: all 11 exact-head font-compression tests passed on Linux, including real TTF compression. The collision regression fails against the base implementation. An eight-process same-key publication probe also passed with complete cache contents and no leftover staging directories. Windows behavior was inspected, not executed locally. CI is still running; no failures observed at review time.

— Magi

**Verdict:** APPROVE
**Reasoning:** The fix establishes cleanup ownership without changing cache keys or font output, supported by targeted regression/error/concurrency checks. Outstanding CI remains a landing gate.

PR #3670, opened 2026-09-04
docs: clarify paused roots in project instructions

miguel-heygen:
## What

Project and scaffold instructions now distinguish the paused composition root from scene timelines added to it.

## Why

The old plural instruction could leave nested scenes paused during root seeks. Current runtime repair activates registered composition siblings, but not arbitrary manually nested scenes.

Related: #3428

## How

Align the four compressed instruction copies with the existing core-skill contract. Avoid claiming that every paused child necessarily produces a blank render.

## Test plan

- [x] Current GSAP3.15.0 probe confirms a manually nested paused child does not advance under a root seek.
- [x] Checked canonical core skill and runtime sibling activation; three independent review lenses found no issues.
- [x] Scaffold AGENTS/CLAUDE copies are byte-identical; oxfmt, diff checks and commit hooks pass.
- [ ] New unit tests: not applicable to this documentation-only correction.

No additional operational monitoring required; runtime behavior is unchanged. Check generated project instructions on the next CLI scaffold.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -81,7 +81,7 @@ skills/                 → AI agent skill definitions
 - **Package manager**: bun (not pnpm, not npm for workspace operations)
 - **Commit format**: Conventional commits (`feat:`, `fix:`, `docs:`, `refactor:`, `test:`)
 - **TypeScript**: Avoid `any` and `as T` assertions. Prefer type guards and narrowing.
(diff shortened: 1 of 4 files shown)


miga-heygen (approved):
Docs clarification is accurate — paused children don't advance on root seek. Updated consistently across AGENTS.md, CLAUDE.md, and templates. — Miga

PR #3672, opened 2026-09-04
fix(shader): keep shader handles within their WebGL context

miguel-heygen:
## What

Shader transitions now initialize correctly when a second WebGL context is created.

## Why

A module-global vertex shader belonged to the first context. Reusing it in another context failed program linking. The old lifecycle rewrite is unnecessary for this specific defect.

Related: #3005

## How

Use the existing per-context program creation helper instead of caching a shader globally. Context-loss handling and lazy allocation are unchanged.

## Test plan

- [x] New ownership regression fails before the fix; all13 shader-package tests pass.
- [x] Actual-source Chrome probe: two distinct contexts link and render across alternating calls, with no WebGL errors; rerun on current main a93106aeb.
- [x] Independent review, package typecheck, lint/format/Fallow and commit hooks pass.
- [ ] Physical GPU matrix/full producer render: not run; browser proof uses SwiftShader.

On the next shader integration run, exercise two independently mounted contexts and watch for program-link failures. Expected: both render. Revert this narrow helper change if a new initialization regression appears.

### DIFF
--- a/packages/shader-transitions/src/webgl.test.ts
+++ b/packages/shader-transitions/src/webgl.test.ts
@@ -1,0 +1,44 @@
+import { describe, expect, it, vi } from "vitest";
+import { createProgram } from "./webgl.js";
+
+/** Model WebGL's rule that shader handles belong to the context that created them. */
+function createMockContext() {
+  const shaders = new Set<WebGLShader>();
+  const gl = {
+    VERTEX_SHADER: 0x8b31,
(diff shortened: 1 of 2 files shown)


miga-heygen (approved):
Removing the module-level cached vertex shader fixes cross-context handle reuse. The existing createProgramWithVertex already creates both shaders fresh. Test confirms multi-context safety. — Miga

PR #1836, opened 2026-07-01
fix: dead code cleanup, missing MIME types, partial download leak

fix2015:
A few things I noticed while going through the codebase:

- \`formatSpeed\` had a ternary where both branches returned the exact same string — simplified it
- the file server's MIME type map was missing avif, flac, m4a, mov, and ico, so those formats were being served as \`application/octet-stream\` which breaks browser rendering
- \`parseObjectPositionAxis\` had an unreachable branch — the axis param is typed as \`"x" | "y"\` so the guard is always true
- when a download fails midway in \`downloadToTemp\`, the partial file was left on disk. on the next call it would find that partial file and return it as if it succeeded. now cleans up the partial in the catch block
- added a description to @hyperframes/core package.json — it was empty

PR #3677, opened 2026-09-04
fix(render): serve media assets with registered content types

jrusso1020:
Render file servers currently serve AVIF, FLAC, M4A, MOV and ICO assets as `application/octet-stream`. Add the missing types to the active producer server used by local/distributed renders and to the public engine server helper. Full and partial producer responses now identify the media correctly while preserving its bytes.

Isolates the MIME fix reported by @fix2015 in #1836, with co-author credit. Uses the [registered ICO type](https://www.iana.org/assignments/media-types/image/vnd.microsoft.icon), `image/vnd.microsoft.icon`. The downloader portion is already handled on main by per-attempt partial directories and cleanup.

Validation: all ten new HTTP cases failed before the mapping fix. Afterward, 54 producer file-server tests and five engine file-server tests pass, including binary bodies and partial-response headers. Both package typechecks, lint/format, producer test classification and commit hooks pass.


miguel-heygen (approved):
Reviewed `f60c440bb04188ba5c6b879eb77d56a0d1184853`.

`packages/producer/src/services/fileServer.ts:91` and `packages/engine/src/services/fileServer.ts:27` add the five MIME mappings to both render servers without changing their existing case-insensitive lookup or binary response paths. `packages/producer/src/services/fileServer.test.ts:319` exercises real full/range HTTP responses, and the new engine tests verify full responses and unchanged bytes. @fix2015 receives co-author credit; this ports the reported MIME fix rather than claiming new downloader functionality.

No findings. Independently reproduced all ten new cases failing with the mappings removed; at the reviewed head, 54 producer tests and five engine tests pass. Reused existing dependencies in scratch; the engine run aliases the compiler import to its exact-head HTML helper to avoid unrelated missing generated build assets. No full render/browser codec test was run. Traced local/distributed producer callers, the exported engine helper, existing Studio MIME lookup, and downloader attempt-directory cleanup.

No prior reviews or comments were present at precheck. CI has no failed checks at this check, but pending checks remain; this approval is on code/documentation merits and is not a CI-complete or merge-ready claim.

— Magi

**Verdict:** APPROVE
**Reasoning:** The change matches the current implementation and no blocking regression was found within the stated verification scope. Required CI must still finish successfully.

PR #2735, opened 2026-07-22
fix(cli): reject unsupported MP4 encoders

miguel-heygen:
## Summary

MP4 renders now stop at capability preflight when the selected FFmpeg build has neither `libx264` nor the supported VideoToolbox fallback. Previously, the CLI treated this definitive capability result as advisory and continued until the encoder rejected `-x264-params`; users now get both a distro-aware install command and an actionable `--format webm` alternative before frame capture starts. Inconclusive probe failures remain advisory, and the existing VideoToolbox fallback is unchanged.

Authorized originating source: https://slack.com/archives/C0BGC335AQY/p1784695521573939

## Test plan

- `bun run --cwd packages/cli test -- src/browser/ffmpeg.test.ts src/commands/render.test.ts` (75 passed)
- `bun run --cwd packages/cli typecheck`
- `bunx oxlint` and `bunx oxfmt --check` on changed files
- `bun run build`
- Exact fake-FFmpeg fixture: exit 1 at preflight; install/WebM hint present; no `x264-params`, producer pipeline, or MP4 output

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
Codex

PR #3679, opened 2026-09-04
fix(cli): reject confirmed unsupported SDR encoders

miguel-heygen:
## What

An explicitly SDR MP4 render now stops before loading the producer when FFmpeg confirms that the required H.264 encoder is unavailable.

## Why

The preflight catch treated a definitive unsupported-encoder result as an inconclusive probe and continued. That wasted rendering work before the same failure surfaced later.

Related: #2735

## How

Distinguish confirmed encoder absence with a typed error. Keep unknown/probe-timeout results advisory. Restrict early rejection to forced SDR: auto and HDR may select HEVC after inspecting source media, so this preflight cannot safely reject those paths.

## Test plan

- [x] Regression failed before the change: unsupported SDR render incorrectly completed.
- [x] 91 render/FFmpeg tests pass on current main, including unknown probe, auto/HDR, alternate format and VideoToolbox controls.
- [x] Independent review found no issues; lint, format, Fallow, typechecks and commit hooks pass.
- [ ] No live full encode performed; the regression verifies rejection before producer loading.

On the next unsupported forced-SDR render, expect an actionable encoder error before capture begins. Auto/HDR and inconclusive probes retain fallback behavior. Revert this classification if valid supported encoders are rejected.

### DIFF
--- a/packages/cli/src/browser/ffmpeg.test.ts
+++ b/packages/cli/src/browser/ffmpeg.test.ts
@@ -2,7 +2,7 @@ import { execFileSync } from "node:child_process";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { afterEach, describe, expect, it, vi } from "vitest";
-import { findFFmpeg, findFFprobe } from "./ffmpeg.js";
+import { findFFmpeg, findFFprobe, H264EncoderUnavailableError } from "./ffmpeg.js";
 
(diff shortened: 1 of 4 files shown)


miga-heygen (approved):
Typed H264EncoderUnavailableError correctly gates early rejection to force-sdr MP4 only, leaving HDR/WebM/PNG paths through. Tests cover all format and mode combinations. — Miga

PR #3680, opened 2026-09-04
fix(engine): isolate chunked encode temporary files

jrusso1020:
Chunked encodes sharing an output directory currently reuse `chunk-encode/chunk_0000.mp4` and `concat-list.txt`. Concurrent calls can overwrite each other's inputs, and a pre-existing directory can supply files or links at these predictable write locations. This addresses [code-scanning alert 857](https://github.com/heygen-com/hyperframes/security/code-scanning/857).

Allocate each invocation's intermediate directory atomically with `mkdtempSync` beside the output. Keep the existing output, codec/timeout/concat behavior and caller-owned cleanup: the producer removes its work directory normally and retains artifacts for debug/KEEP_TEMP inspection. No workflows or public options change.

Validation: the new concurrent-encode regression fails on main because both calls use the same chunk path, and passes after the fix. It also checks separate concat manifests and preservation of a pre-existing concat file. All 102 chunk-encoder and provenance tests pass, including real FFmpeg renders; core/parser builds, engine typecheck, and changed-file lint/format pass. GitHub CI and CodeQL remain landing gates.

### DIFF
--- a/packages/engine/src/services/chunkEncoder.test.ts
+++ b/packages/engine/src/services/chunkEncoder.test.ts
@@ -1,7 +1,7 @@
 import { EventEmitter } from "node:events";
-import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
-import { join } from "node:path";
+import { dirname, join } from "node:path";
 import { afterEach, describe, it, expect, vi } from "vitest";
 import { ENCODER_PRESETS, getEncoderPreset, buildEncoderArgs } from "./chunkEncoder.js";
 import { renderProvenanceArgs } from "../utils/renderProvenance.js";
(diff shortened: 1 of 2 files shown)


miguel-heygen (approved):
Reviewed at `3f98d808b31b6e66ea535e48f57a14fb1bb09b39`; no blockers.

The isolation fix is correctly placed. `mkdtempSync(join(dirname(outputPath), "chunk-encode-"))` creates a private per-invocation directory atomically (`packages/engine/src/services/chunkEncoder.ts:567-570`), and both chunk paths and `concat-list.txt` derive exclusively from it (`:591`, `:634-636`). Concurrent encodes sharing a parent can no longer overwrite or follow files at the legacy predictable locations.

The regression is non-vacuous: it holds two encodes open concurrently, proves distinct chunk and manifest directories, verifies each manifest names its own chunk, and confirms a pre-existing legacy manifest is untouched (`packages/engine/src/services/chunkEncoder.test.ts:232-273`).

I also audited cleanup rather than assuming the new unique directories disappear: the producer registers recursive work-directory removal while preserving debug/`KEEP_TEMP` behavior (`packages/producer/src/services/renderOrchestrator.ts:2049-2055`), and distributed chunk rendering removes its work directory after artifacts land (`packages/producer/src/services/distributed/renderChunk.ts:1033-1041`). The engine's direct export retains its existing caller-owned intermediate-cleanup contract.

All required CI, Windows rendering, nine regression shards, and CodeQL are green on this exact head. No public API, option, or output behavior changed.

**Verdict:** APPROVE
**Reasoning:** Intermediate paths are atomically isolated per invocation, the concurrency/pre-existing-file failure modes are pinned, and the existing cleanup/debug ownership remains intact.

— Magi

PR #3681, opened 2026-09-04
fix(studio): preserve timeline DOM identity during hydration

miguel-heygen:
## What

A stale image entry no longer takes a scene element's DOM identity. Clips identified only by their persistent hfId can be moved and trimmed.

## Why

The positional fallback could bind an image to a div, replacing the legitimate clip identity. The capability gate also rejected hfId-only targets even though the editing path supports them.

Related: #3334, #3007

## How

Require compatible tags for identity/fallback binding and share one lazy candidate snapshot within each hydration pass. The snapshot never survives a reload.

## Test plan

- [x] Current DOM regressions fail before the fix; 100 targeted tests pass.
- [x] Real Studio master and standalone flows: stale manifest, hfId-only move/trim, persistence, undo and reload; the other clip remains unchanged.
- [x] 300 fallback clips use one candidate scan; direct identities use zero. No general CPU claim.
- [x] Independent review, typecheck, lint/format/Fallow and commit hooks pass.

On the next Studio smoke pass, move/trim an hfId-only clip and switch/reload projects. Watch for wrong-target edits or lost identities; revert this binding change if either appears.

### DIFF
--- a/packages/studio/src/player/components/timelineAuthoredMoveTarget.ts
+++ b/packages/studio/src/player/components/timelineAuthoredMoveTarget.ts
@@ -10,6 +10,7 @@ export function canMoveTimelineElement(element: TimelineElement): boolean {
     kind: element.kind,
     duration: element.duration,
     domId: element.domId,
+    hfId: element.hfId,
     selector: element.selector,
     compositionSrc: element.compositionSrc,
     playbackStart: element.playbackStart,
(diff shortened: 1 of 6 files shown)


miga-heygen (approved):
hfId identity threading, lazy candidate snapshot, cross-tag rejection all look correct. Tests cover move, trim, positional recovery, and scan efficiency. — Miga

PR #3683, opened 2026-09-05
fix(skills): accept durationless transition roots

miguel-heygen:
## What

Product-launch and PR-to-video transition injection now accepts a valid frame root without `data-duration`, inserts the outgoing transition target duration, and continues. A genuinely missing or mismatched composition root still fails hard.

## Why

Composition root duration is optional and can be derived from its timeline. Both injectors checked duration before composition identity, so a durationless but correctly identified worker frame was skipped and then misreported as missing `data-composition-id`.

## How

Both workflows now use the already-settled faceless-explainer ordering: identify the matching root first, replace its duration when present, or insert the target duration when absent. Non-root timed elements retain the existing tail-extension rules.

## Test plan

- [x] Product-launch inserts 3.5 seconds on a valid durationless outgoing root.
- [x] PR-to-video inserts 3.5 seconds on the identical fixture.
- [x] Both workflows retain the hard failure for a mismatched root.
- [x] Shared transition regression passes (4/4).
- [x] Existing PR-to-video frame-contract suite passes (9/9).
- [x] Node syntax, oxlint, and oxfmt checks pass.

### DIFF
--- a/skills/pr-to-video/scripts/transitions.mjs
+++ b/skills/pr-to-video/scripts/transitions.mjs
@@ -117,15 +117,17 @@ function extendFrameTail(hyperframesDir, frame, baseDuration, targetDuration, di
   let extended = 0;
   const rewritten = html.replace(/<([A-Za-z][\w:-]*)\b([^>]*)>/g, (tag, name, attrs) => {
     const durationMatch = attrs.match(/\bdata-duration="([\d.]+)"/);
-    if (!durationMatch) return tag;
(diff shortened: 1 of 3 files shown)


miga-heygen (approved):
Good edge case fix — a valid frame root without data-duration should get the attribute inserted rather than bailing. The reorder of the duration/composition match checks is clean. Regression test covers both the fix and the existing rejection path. — Miga

PR #2262, opened 2026-07-11
fix(scoping): rewrite prefix/substring composition-id selectors in place

Thomaswebstich:
## Problem

A sub-composition whose `<style>` scopes rules by a **prefix/substring** composition-id selector — `[data-composition-id^="x"]` (also `*=`, `$=`) — has those rules **silently dropped** when mounted in a multi-composition document.

`scopeSelector()` (in `compositionScoping.ts`) rewrites the **exact** `[data-composition-id="x"]` root selector *in place* to the instance scope, but any prefix/substring form falls through to the prepend branch and becomes a two-level selector:

```
[data-composition-id="x__hf2"] [data-composition-id^="x"] .thing
```

That now requires a **nested** composition-id element — but `compositionLoader` strips the inner root's `data-composition-id` at mount, so nothing matches and every class-scoped rule in that sub-comp stops applying (class-styled elements render unstyled). Single-mount renders keep the inner structure, so they're unaffected — the drop only shows when the loader renames/flattens (e.g. duplicate instances in one document).

This bites any authoring pipeline that rewrites `="x"` to `^="x"` so a sub-comp's own root rules survive the duplicate-instance rename (`x` → `x__hf2`).

## Fix

Widen the composition-id match to accept the substring operators (`^=`, `*=`, `$=`) of the authored id, so the authored-root selector is rewritten in place to the instance scope in every form — exactly like the exact `=` case — instead of being prepended.

One-line change to the selector regex, plus a regression test covering `^=`/`*=`/`$=`.

## Test

`packages/core/src/compiler/compositionScoping.test.ts` — new case asserts the prefix/substring forms are consumed in place (and never left intact behind a prepended scope). Full scoping suite passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3684, opened 2026-09-05
fix(core): scope composition root pattern selectors per instance

jrusso1020:
Duplicate sub-compositions lose styles authored with `[data-composition-id^="scene"]`, `*=` or `$=`: the scoper prepends an instance selector instead of replacing the authored root. After mounting strips the inner root ID, these rules match nothing. Recognize those operators when their value is the authored root ID, preserving the existing composition-box remap and exact-match behavior.

Successor to Thomaswebstich’s #2262, with co-author credit. Different nested-composition IDs remain scoped descendants. The pre-existing exact-only timing-attribute normalization is unchanged.

Validation: headless Chrome measured root padding, descendant borders/colors and outside isolation for two instances through both `bundleToSingleHtml` and `loadInlineTemplateCompositions`. Before the fix, exact `=` passed and all six pattern cases failed; all eight cases pass afterward. All 199 scoper/compiler/loader tests, core/runtime typechecks, lint/format and commit hooks pass.

### DIFF
--- a/packages/core/src/compiler/compositionScoping.test.ts
+++ b/packages/core/src/compiler/compositionScoping.test.ts
@@ -106,6 +106,31 @@ body { margin: 0; }
     expect(wrapped).not.toContain("requestAnimationFrame");
   });
 
+  it.each(["=", "^=", "*=", "$="])(
+    "scopes %s authored-root selectors to the duplicate instance and its box",
+    (operator) => {
+      const scope = '[data-composition-id="scene__hf2"]';
+      const scoped = scopeCssToComposition(
+        `[data-composition-id${operator}"scene"] { padding: 13px; }
+[data-composition-id${operator}"scene"] .item { border-width: 3px; }`,
+        "scene",
+        scope,
+      );
+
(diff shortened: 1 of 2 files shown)


miguel-heygen (approved):
The one-line matcher expansion at `packages/core/src/compiler/compositionScoping.ts:142` is correctly constrained to `^=`, `*=`, and `$=` forms whose value exactly equals the authored composition id. It therefore follows the renamed instance through the shared compiler/runtime scoper without consuming selectors for other nested compositions.

The matrix at `packages/core/src/compiler/compositionScoping.test.ts:109` covers exact plus all three pattern operators for both root-box and descendant forms, while `packages/core/src/compiler/compositionScoping.test.ts:128` pins the nested-composition non-match. Contributor credit is preserved. Live checks currently have no failures; remaining jobs should gate merge normally.

**Verdict:** APPROVE
**Reasoning:** The regex change is narrowly scoped, preserves existing exact-match behavior, and directly covers the loader/compiler transformation contract that caused the duplicate-instance regression.

— Magi (Home)

PR #1880, opened 2026-07-03
fix(capture): address review nits from #1879 (rgba-only transparent check, glass dedup key)

xuanruli:
Follow-up to #1879 (merged before these nits landed). Addresses the three review nits from @miga-heygen.

- **`isFilledEl` transparent check** — require `rgba(` (not `rgba?`) for the zero-alpha test. `rgba?\([^)]*,\s*0\s*\)` matched the blue channel of `rgb(r, g, 0)`, so a solid `rgb(0, 0, 0)` nav CTA was misread as transparent and dropped. Only `rgba()` carries an explicit alpha.
- **Button dedup key** — include `backdropFilter`, mirroring the card dedup key, so a frosted-glass button variant doesn't collapse into its non-frosted sibling.
- **`rgbToHex` comment** — clarify that RGB channels stay comma-separated (as `getComputedStyle` returns them); only the alpha may follow a comma or slash.

Tested: `bun run build`, `oxlint`/`oxfmt` clean, `vitest run src/capture/` 54/54.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3685, opened 2026-09-05
fix(capture): retain opaque and glass button styles

jrusso1020:
`hyperframes capture` drops opaque navigation CTAs whose RGB blue channel is zero, including black buttons, and merges frosted buttons with otherwise identical plain variants. Reuse the existing color parser's transparent result and include backdrop blur in the button style key. Transparent navigation links remain excluded, gradients remain included, and identical glass variants still deduplicate.

Refreshes #1880 with credit to @xuanruli in the signed commit. The active capture path writes these styles to `extracted/design-styles.json` before DOM mutation.

Validation: five DOM/VM regressions (three fail before the fix); real Chrome 152.0.7977.77 extraction before/after, including modern authored RGB/alpha syntax and computed-style serialization; 198 capture tests; full workspace build; CLI typecheck; formatting, lint and required commit hooks. CI tests use Happy DOM and do not require Chromium.

### DIFF
--- a/packages/cli/src/capture/designStyleExtractor.buttons.test.ts
+++ b/packages/cli/src/capture/designStyleExtractor.buttons.test.ts
@@ -1,0 +1,68 @@
+import { describe, expect, it } from "vitest";
+import { runInNewContext } from "node:vm";
+import { Window } from "happy-dom";
+import type { Page } from "puppeteer-core";
+import { extractDesignStyles } from "./designStyleExtractor.js";
+
+async function extractButtons(body: string) {
+  const window = new Window();
+  window.document.body.innerHTML = body;
+  for (const element of window.document.querySelectorAll("button, a")) {
+    element.getBoundingClientRect = () => new window.DOMRect(0, 0, 100, 40);
+  }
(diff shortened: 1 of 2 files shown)


miguel-heygen (approved):
The opaque/transparent split is correctly repaired at `packages/cli/src/capture/designStyleExtractor.ts:130`: reusing `rgbToHex` avoids treating an RGB blue channel of zero as alpha while retaining the zero-alpha sentinel. Adding `backdropFilter` to the visual key at `packages/cli/src/capture/designStyleExtractor.ts:147` also prevents glass/plain variants from collapsing.

The focused matrix at `packages/cli/src/capture/designStyleExtractor.buttons.test.ts:31` pins opaque black/amber, transparent navigation, gradient fill, and glass deduplication. Contributor credit is preserved. Live checks currently have no failures; remaining jobs are still pending and should gate merge normally.

**Verdict:** APPROVE
**Reasoning:** The two behavior changes match the reported failures, preserve the surrounding capture contract, and have direct regression coverage.

— Magi (Home)

PR #2005, opened 2026-07-07
feat(url-to-frame-md): website capture → frame.md design system + showcase renderer

xuanruli:
## What

A toolkit (`url-to-frame-md/`) that turns a `hyperframes capture` of any website into a **video-first design spec** (`frame.md`) and a **brand-dressed showcase board** (`showcase.png`).

- **`build-frame-from-capture.mjs`** — capture → `frame.md`: palette (incl. vivid secondary brand accents), two type ramps, spacing, radii, shadows, and video-usable components. Faithful to the capture.
- **`verify-frame.mjs`** — L0/L1/L2 machine gates (schema · internal coherence · fidelity to the captured site).
- **`render-showcase.mjs`** — `frame.md` → `showcase.png` (composed 16:9 frames + a 9:16 portrait section) via headless Chrome. Demo avatar/hero load from hosted URLs; overridable with `--avatar` / `--hero`.
- **`reextract-design.mjs`** + **`lib/`** — offline re-extract against a saved capture; shared color/font/token helpers.

## Why

Turns a captured site's design system into reusable, brand-faithful video frames. Two notable pieces:
- **Accent selection rejects injected-widget strays** — a candidate must be a palette member, paint a background, or recur (`count≥5`); otherwise it's dropped. This stopped a recurring cobalt `#3860BE` (a consent/focus-ring artifact) from burying real brand colors on unrelated sites.
- **Vivid secondary brand hues** that live in gradients/graphics (which usage stats miss) are surfaced as `accent-2` / `accent-3`.

## How

Pure Node; reads `extracted/tokens.json` + `design-styles.json`. Every hex/size is faithful to the capture — `verify-frame.mjs` L2 enforces this. No new dependencies.

## Test plan

- [x] Manual: ran capture → build → verify → render across 20+ sites (Anthropic, Linear, Stripe, HeyGen, Figma, GitLab, Aesop, Miro, Zed, …) — all pass `verify-frame` (L0/L1/L2 green).
- [x] `oxlint` + `oxfmt` clean on all scripts.
- [ ] Unit tests (N/A — standalone toolkit scripts).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1883, opened 2026-07-03
feat(capture): capture box-shadows by computed style, not class-name selectors

xuanruli:
## Problem

The shadow sampler in `designStyleExtractor.ts` selected elements by class-name substring (`[class*='card']`, `button`, `[class*='panel']`, …). Sites with obfuscated / utility class names carry their card elevation on plain `div`s that never match those selectors, so `shadows: []` came back empty — and the showcase renderer then faked a heavy `0 10px 34px` Material drop shadow, the *opposite* of e.g. Vercel's near-invisible hairline. (Surfaced by reviewers on Stripe + Vercel.)

## Fix

Sweep a broad structural set of visible, **card-sized** boxes and collect their real computed `box-shadow`; the shadow that recurs on the most boxes is the brand's elevation signature. Skips tiny (icon/inline) elements and full-bleed page wrappers so a body/section shadow can't win.

## Verified (live re-capture)

- Stripe: `[]` → **5** shadows, incl. the signature `rgba(50, 50, 93, 0.12) 0 16px 32px`
- Vercel: `[]` → **3**
- `oxlint`/`oxfmt` clean, `bun run build` ✓, `vitest run src/capture/` 54/54

Output shape unchanged (`shadows: Array<{value, count}>`), backward compatible. Follow-ups (separate): `cards: []` shares the same class-selector root cause; ElevenLabs glass/aura live on pseudo-elements/canvas.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3686, opened 2026-09-05
fix(capture): sample computed shadows on ordinary boxes

jrusso1020:
Capture misses box shadows on ordinary or utility-class containers because it samples only named cards, buttons, and similar selectors. Add a bounded computed-style sample of structural boxes while preserving existing semantic samples, including small buttons and late-page headers. Deduplicate overlapping samples; apply tiny/full-page exclusions only to newly sampled boxes.

Successor to #1883 by @xuanruli; original authorship preserved. The separate toolkit in #2005 was closed and re-homed, while this extractor remains part of the active capture command.

Validation: 196 capture tests pass, including three new regressions; CLI typecheck, formatting/lint, and commit hooks pass. Chrome 152 reproduces the missed shadows before the fix and verifies coverage, deduplication, hidden/tiny/full-page exclusions, and preservation of late semantic samples afterward. A 10,000-div fixture adds only 1,600 computed-style reads (800 new candidates). No Chrome dependency added to CI tests.

### DIFF
--- a/packages/cli/src/capture/designStyleExtractor.shadows.test.ts
+++ b/packages/cli/src/capture/designStyleExtractor.shadows.test.ts
@@ -1,0 +1,72 @@
+// @vitest-environment happy-dom
+import type { Page } from "puppeteer-core";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { extractDesignStyles } from "./designStyleExtractor.js";
+
+// Use the complete browser expression; happy-dom supplies CSS, and these
+// fixtures supply the layout measurements that a browser normally computes.
+async function shadows() {
+  const evaluate: Page["evaluate"] = async (script, ..._args) => {
+    if (typeof script !== "string") throw new Error("Expected a browser expression");
+    return window.eval(script);
+  };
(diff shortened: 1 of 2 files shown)


miguel-heygen (approved):
The sampler extension is bounded and preserves the old priority surface: semantic elements are retained independently at `packages/cli/src/capture/designStyleExtractor.ts:377`, structural candidates are capped and deduplicated at `packages/cli/src/capture/designStyleExtractor.ts:384`, and tiny/full-page exclusions apply only to the new candidates at `packages/cli/src/capture/designStyleExtractor.ts:391`.

The regressions at `packages/cli/src/capture/designStyleExtractor.shadows.test.ts:32` cover ordinary/utility boxes, overlap counts, dense-prefix behavior, and the hidden/tiny/full-page boundaries. Contributor authorship is intact. I also verified the merge tree with #3685 is clean and contains both suites. Live checks currently have no failures; remaining jobs should gate merge normally.

**Verdict:** APPROVE
**Reasoning:** The change broadens shadow discovery without dropping existing semantic samples or making the scan unbounded, and the boundary cases are directly pinned.

— Magi (Home)

PR #3687, opened 2026-09-05
fix(engine): use supported VideoToolbox quality options

miguel-heygen:
## What

VideoToolbox GPU renders now use supported bitrate rate control instead of FFmpeg’s rejected generic `-q:v` path. The reported 1920×1080, 24fps high-quality case derives `-b:v 12M` in both disk and streaming encoders.

## Why

The usability probe previously tested only the bare encoder, then production added `-q:v 70`. Some macOS VideoToolbox builds reject that option with exit 187, so a valid `--gpu` render failed even though software encoding and explicit bitrate succeeded.

## How

One VideoToolbox rate-control helper owns explicit and derived bitrate selection. Derived bitrate scales with pixel rate and the existing quality tiers. Disk encoding, streaming encoding, and the GPU usability probe all use the same supported option family; explicit `--bitrate` remains authoritative.

## Test plan

- [x] Disk high-quality 1080p24 emits `-b:v 12M` and no `-q:v`.
- [x] Streaming high-quality 1080p24 emits the identical rate control.
- [x] VideoToolbox probe exercises `-b:v` and `-allow_sw`.
- [x] Existing explicit-bitrate behavior remains covered.
- [x] GPU, disk, and streaming suites pass (179/179).
- [ ] macOS VideoToolbox integration (not available on this Linux host).

### DIFF
--- a/packages/engine/src/services/chunkEncoder.test.ts
+++ b/packages/engine/src/services/chunkEncoder.test.ts
@@ -891,6 +891,27 @@ describe("buildEncoderArgs GPU preset mapping", () => {
     expect(presetArg(args)).toBe("p5");
   });
 
+  it("uses a supported derived bitrate for high-quality VideoToolbox", () => {
+    const args = buildEncoderArgs(
+      {
+        fps: { num: 24, den: 1 },
+        width: 1920,
+        height: 1080,
+        codec: "h264",
+        preset: "slow",
+        quality: 15,
+        useGpu: true,
+      },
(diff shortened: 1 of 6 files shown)


miga-heygen (approved):
Replacing the unsupported qscale path with derived bitrate is the right call for VideoToolbox — the three-tier bits-per-pixel mapping is applied consistently across chunk, streaming, and probe paths. — Miga

PR #3688, opened 2026-09-05
fix(keyframes): resolve helper-returned DOM targets

miguel-heygen:
## What

`hyperframes keyframes` now resolves simple lexical helpers that return DOM lookups, so `const target = scoped("drag")` surfaces as `#drag` instead of `__unresolved__`. Truly runtime-only targets are no longer forwarded as concrete onion-shot selectors.

## Why

GSAP received the real DOM node and rendered motion correctly, but static introspection understood only direct lookup calls. The shot path then treated its internal unresolved marker as a selector and falsely reported that nothing animated under the requested scope.

## How

The parser indexes lexically scoped function declarations and function-valued variables when they have one direct return. At a helper call it binds literal arguments to parameters and reuses the existing DOM-query selector resolver on the return expression. Opaque helpers retain `hasUnresolvedSelector`; the CLI filters that marker and reports a static-analysis limitation with direct-selector and `--ghost` guidance.

## Test plan

- [x] `scoped(id) { return document.getElementById(id) }` resolves `scoped("drag")` to `#drag`.
- [x] Unknown runtime targets remain unresolved but are excluded from shot requests.
- [x] Parser and keyframes suites pass (225/225).
- [x] Parser package typecheck passes.
- [x] Changed files pass oxlint and oxfmt.
- [ ] Full CLI workspace typecheck: unavailable in the reused checkout because unrelated AWS/GCP/producer dependencies and generated types are missing or mismatched; no reported error points to the changed keyframes file.

### DIFF
--- a/packages/cli/src/commands/keyframes.test.ts
+++ b/packages/cli/src/commands/keyframes.test.ts
@@ -192,6 +192,20 @@ describe("keyframes runtime surfacing", () => {
 
     expect(selectors).toEqual(expect.arrayContaining([".dot", ".chip"]));
   });
+
+  it("does not forward unresolved static targets as concrete shot selectors", () => {
+    const html = wrap(`
+      const tl = gsap.timeline({ paused: true });
+      tl.to(runtimeOnlyTarget(), { x: 240, duration: 1 });
+      tl.to("#drag", { x: 240, duration: 1 });
+    `);
+
+    const selectors = collectShotSelectors([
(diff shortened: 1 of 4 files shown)


miga-heygen (approved):
Helper-returned DOM target resolution is well structured — single-return constraint, parameter inlining into scope, and consistent plumbing through all three resolution call sites. Test proves the scoped(id) → getElementById pattern resolves correctly. — Miga

PR #3692, opened 2026-09-05
fix(lint): flag autoAlpha on clip elements

miguel-heygen:
## What

- Report `gsap_animates_clip_element` when GSAP animates `autoAlpha` on a framework-owned `.clip` element.
- Keep `autoAlpha` animations on descendants valid.

## Why

GSAP `autoAlpha` writes both opacity and visibility. On a scene root it can override the framework's clip visibility window and leave a blank boundary frame, even though the fractional scene windows themselves are gap-free.

## How

- Add `autoAlpha` to the existing clip visibility conflict check.
- Clarify the diagnostic and fix hint to explain the visibility conflict.
- Cover both the invalid clip-root target and valid child target.

## Test plan

- RED/GREEN focused `autoAlpha` regression tests.
- Full GSAP lint-rule suite: 167 passed.
- Lint package typecheck.
- Oxlint and formatting checks on changed files.

### DIFF
--- a/packages/lint/src/rules/gsap.test.ts
+++ b/packages/lint/src/rules/gsap.test.ts
@@ -387,6 +387,53 @@ describe("GSAP rules", () => {
     expect(finding?.message).toContain("visibility");
   });
 
+  it("ERRORS when GSAP animates autoAlpha on a clip element", async () => {
+    const html = `
+<html><body>
+  <div data-composition-id="c1" data-width="1920" data-height="1080">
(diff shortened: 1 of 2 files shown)


miga-heygen (approved):
Clean addition — autoAlpha writes both opacity and visibility, so it belongs in the clip-element guard alongside the existing visibility/display checks. Tests cover both the positive and negative cases. — Miga

PR #3693, opened 2026-09-05
fix(producer): reserve AAC correction headroom

miguel-heygen:
## What

AAC true-peak correction now reserves 0.5 dB of codec headroom, allowing valid mixes to converge within the existing three-pass bound. Audio already at or below the `-1 dBFS` delivery ceiling remains unchanged, and exhausted corrections now report each measured peak with its applied attenuation.

## Why

The released fail-closed validator correctly blocks unsafe output, but its correction encoded toward the exact acceptance boundary. Each lossy AAC retry could regenerate a smaller peak above that boundary and exhaust all three passes.

Related: #3644.

## How

- Derive a `-1.5 dBFS` correction target from the unchanged `-1 dBFS` delivery ceiling.
- Keep the measured post-encode safety check and three-pass failure bound intact.
- Include the bounded measurement history in terminal non-convergence errors.

## Test plan

- [x] Unit regression models partial AAC gain response and converges on the third correction pass.
- [x] Unit regression verifies terminal errors retain all peak and attenuation measurements.
- [x] Existing real-FFmpeg regression remains at or below `-1 dBTP` with no more than `3 LU` change.
- [x] Producer typecheck, oxlint, formatting, and diff checks pass.

### DIFF
--- a/packages/producer/src/services/render/audioPadTrim.test.ts
+++ b/packages/producer/src/services/render/audioPadTrim.test.ts
@@ -14,6 +14,9 @@
  */
 
 import { describe, expect, it, mock } from "bun:test";
+import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 import {
   buildPadTrimAudioArgs,
   buildPadTrimAudioPlan,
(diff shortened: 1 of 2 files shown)


miga-heygen (approved):
AAC correction headroom is the right approach — lossy re-encodes regenerate content-dependent peaks, so undershooting the ceiling converges faster. Convergence test and exhaustion diagnostics are solid. — Miga

PR #3694, opened 2026-09-05
fix(lint): flag hidden-style opacity guards

miguel-heygen:
## What

`lint` and `check` now reject broad CSS guards that turn renderer-owned `visibility: hidden` into `opacity: 0`. This prevents successfully extracted video frames from becoming transparent while keeping guards explicitly scoped to sub-composition hosts valid.

## Why

The renderer hides each native video before it copies the video's computed opacity to the visible replacement image. A broad `[style*="visibility: hidden"]` opacity rule observes that runtime mutation, sets the native opacity to zero, and causes the copied frame to remain transparent for every rendered frame.

## How

- Reuse the existing parsed CSS selector pass and resolve nested selectors before evaluating them.
- Inspect the selected compound, so an ancestor scope does not make an arbitrary media target safe.
- Allow the guard only when that selected compound is explicitly constrained to `data-composition-src` or `data-composition-file`.
- Preserve engine host filtering and authored media-opacity semantics.

## Test plan

- [x] RED/GREEN coverage for broad, media-targeted, and root-stylesheet/nested-video cases.
- [x] Safe host-scoped and non-opacity controls remain accepted.
- [x] Full lint package: 553 tests passed.
- [x] Lint package typecheck, oxlint, formatting, and diff checks passed.

### DIFF
--- a/packages/lint/src/rules/core.test.ts
+++ b/packages/lint/src/rules/core.test.ts
@@ -797,6 +797,78 @@ describe("core rules", () => {
     });
   });
 
+  describe("runtime_hidden_style_opacity", () => {
+    const comp = (css: string, extraMarkup = "") => `
+<html><head><style>${css}</style></head><body>
+  <div id="root" data-composition-id="main" data-width="1920" data-height="1080">
+    <video id="footage" src="clip.mp4" data-start="0" data-duration="5" muted playsinline></video>
+    ${extraMarkup}
+  </div>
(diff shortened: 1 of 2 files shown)


miga-heygen (approved):
Hidden-style opacity guard lint rule with proper CSS selector parsing. Correctly exempts sub-composition host scoping and non-opacity declarations. — Miga

PR #3696, opened 2026-09-05
fix(runtime): preserve fractional final-frame visibility

miguel-heygen:
## What

Fractional-duration renders now keep every producer-emitted final sample visible, including repeated nested sub-composition tails. Exact authored ends remain exclusive, and decimal representations of rational frame boundaries still activate on their intended frame.

## Why

The producer correctly ceilings a `79.402s` render at 60fps to 4,765 frames, whose last sample is `79.4s`. Runtime visibility previously floored every authored end to the prior frame boundary, collapsing `79.402s` onto `79.4s` and hiding the root, host, and descendants on that valid sample.

A second field report amplified the same equation across 14 nested sub-compositions. Its bad indices—483, 1211, 1762, 2395, 2872, 3129, 3512, 3817, 4263, 4903, 5317, 5828, 6732, and 9760—are exactly the cumulative final frames of segment lengths that sum to 9,761.

Related: #3640.

## How

- Separate frame seek quantization from authored boundary normalization.
- Snap a boundary only when decimal noise places it within `0.001` frame of an exact rational boundary.
- Preserve genuinely fractional values under the existing half-open visibility comparison.
- Lock the 14-tail arithmetic in the producer and the nested host/descendant visibility at the shared pre-capture seek boundary used by both screenshot and drawElement.

## Test plan

- [x] Runtime regression covers the `79.402s` root, host, and CTA at frame 4,764.
- [x] Repeated nested regression covers all 14 reported cumulative tail frames and hides each host/child on the next sample.
- [x] Producer regression derives the exact 14 segment lengths, tail indices, and 9,761-frame total.
- [x] Exact-boundary, decimal-noise, frame-584 activation, and half-open controls pass.
- [x] Runtime init suite: 98/98 tests passed; producer probe-stage suite: 38/38 tests passed.
- [x] Core/runtime and producer typechecks plus changed-file format/lint/diff checks passed.
- [ ] Re-render the private original project through screenshot and drawElement and compare raw/encoded tail frames after merge/release. The report supplied no source archive, raw frames, output hashes, or authored durations, so this cannot be completed from the artifact.

### DIFF
--- a/packages/core/src/inline-scripts/parityContract.test.ts
+++ b/packages/core/src/inline-scripts/parityContract.test.ts
@@ -1,5 +1,9 @@
 import { describe, it, expect } from "vitest";
-import { quantizeTimeToFrame, MEDIA_VISUAL_STYLE_PROPERTIES } from "./parityContract.js";
+import {
+  quantizeTimeToFrame,
+  snapTimeToFrameBoundary,
+  MEDIA_VISUAL_STYLE_PROPERTIES,
+} from "./parityContract.js";
 
 describe("quantizeTimeToFrame", () => {
   it("quantizes a time to the nearest frame boundary at 30fps", () => {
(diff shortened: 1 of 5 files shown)


miga-heygen (approved):
snapTimeToFrameBoundary correctly preserves genuinely fractional authored timing while only snapping epsilon-close noise. Field-data regression with 14 cumulative segments confirms the fix. — Miga

PR #3698, opened 2026-09-05
fix(check): clip overlap geometry to visible fragments

miguel-heygen:
## What

`content_overlap` no longer reports text that an overflow ancestor fully hides. Partially clipped text is measured by its painted area, so genuine visible collisions still cross the overlap threshold.

## Why

DOM Range rectangles remain geometrically present outside an overflow clip. Comparing those raw rectangles produced false-positive overlap warnings for translated or otherwise clipped text.

## How

Text fragments are intersected with each overflow ancestor in viewport coordinates before overlap analysis. Clipping is axis-aware for `hidden`, `clip`, `auto`, and `scroll`, while `data-layout-allow-overlap` remains a local opt-out.

## Test plan

- `bunx vitest run src/commands/layout-audit.browser.test.ts --reporter=dot` (89 tests passed)
- `node --check packages/cli/src/commands/layout-audit.browser.js`
- `bunx oxlint packages/cli/src/commands/layout-audit.browser.js packages/cli/src/commands/layout-audit.browser.test.ts`
- `bunx oxfmt --check packages/cli/src/commands/layout-audit.browser.js packages/cli/src/commands/layout-audit.browser.test.ts`
- `git diff --check`

The package-wide typecheck is unavailable in this reused checkout because optional AWS, GCP, and producer dependencies are absent and existing cross-package type drift is present. It did not report errors in the changed browser audit or regression test.

### DIFF
--- a/packages/cli/src/commands/layout-audit.browser.js
+++ b/packages/cli/src/commands/layout-audit.browser.js
@@ -245,10 +245,8 @@
     return rects;
   }
 
-  function textRectFor(element, directOnly) {
-    const rects = textClientRects(element, directOnly);
+  function unionRects(rects) {
     if (rects.length === 0) return null;
-
     const union = rects.reduce(
       (acc, rect) => ({
         left: Math.min(acc.left, rect.left),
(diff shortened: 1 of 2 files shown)


miga-heygen (approved):
Overflow-clipped text rects for overlap detection. Clean extraction of unionRects and clipsOverflowValue helpers with per-axis clipping. Tests cover all overflow modes. — Miga

PR #3699, opened 2026-09-05
fix(lint): flag repeated fromTo state leaks

miguel-heygen:
## What

Lint now warns when repeated future `fromTo` calls can leak the last authored start state into earlier frames. Explicit `immediateRender: false` and an earlier timeline baseline remain clean.

## Why

GSAP applies default `fromTo` start values while the timeline is built, before the tween's scheduled position. When several calls target the same element, the last authored start state can become visible before the first tween, while existing lint reports no cause.

A standalone `gsap.set` is not accepted as proof of safety. An earlier global set can be overwritten by later `fromTo` construction, and a set for unrelated properties does not establish the affected state. The warning therefore clears only for explicit opt-out or an earlier `tl.set` baseline that participates in deterministic timeline seeking.

## How

- Preserve `immediateRender` as explicit true, explicit false, or the GSAP method default.
- Group default or explicitly immediate-rendered `fromTo` writers by stable selector.
- Require an earlier timeline set for that selector, or explicit `immediateRender: false` on the repeated writers.
- Keep render seeking and GSAP runtime behavior unchanged.

## Test plan

- [x] GSAP lint suite: 175 tests passed.
- [x] Regression cases cover omitted options, explicit false, earlier timeline baseline, earlier standalone set, unrelated standalone properties, later standalone set, deferred callback set, and late timeline baseline.
- [x] Lint package typecheck passed.
- [x] Changed-file oxlint, formatting, and diff checks passed.
- [ ] Manual testing performed (not applicable; lint-only behavior).

### DIFF
--- a/packages/lint/src/rules/gsap.test.ts
+++ b/packages/lint/src/rules/gsap.test.ts
@@ -2932,33 +2932,93 @@ describe("SVG draw-on rules", () => {
     expect(finding).toBeUndefined();
   });
 
-  it.each([
-    ["timeline", 'tl.set("#ring", { opacity: 0, scale: 1 }, 0);'],
-    ["standalone", 'gsap.set("#ring", { opacity: 0, scale: 1 });'],
-  ])(
-    "gsap_repeated_fromto_without_baseline: accepts an earlier %s set baseline",
-    async (_kind, baseline) => {
-      const html = `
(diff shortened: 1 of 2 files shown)


miga-heygen (approved):
Repeated fromTo state leak detection with proper tri-state immediateRender tracking. 10 test cases covering baselines, standalone sets, source ordering, and deferred callbacks. — Miga

PR #3700, opened 2026-09-05
fix(render): bound and recover capture stalls

miguel-heygen:
## What

Wedged capture operations now fail on bounded worker phases and retry through the existing fresh screenshot plan. Sequential watchdogs name their resolved mode, while parallel drawElement workers report whether they stopped during browser launch/probing, session initialization, first-frame capture, or pipelined frame encoding.

## Why

The original 60-second parallel watchdog measured only completed frames. A stall at `0/N` could therefore mean browser launch, Chrome probing, drawElement initialization, first capture, or the in-page encoder, and logs did not identify the executable/build, CanvasDrawElement result, GPU backend, or worker.

Router-pinned parallel fallback was already correct: it aborts peers, closes every worker session, and retries once via screenshot. This change reduces the default drawElement worker-operation bound to 30 seconds and makes that recovery attributable; it does not add another retry layer.

## How

- The existing drawElement serial path retains its 15-second renderer-specific deadline.
- Sequential screenshot/BeginFrame/drawElement stalls retain the conservative 60-second no-progress bound and exactly-once fresh-session retry.
- Parallel drawElement streaming wraps browser launch/probe, session initialization, capture, and pipelined encoding with `HF_DE_PARALLEL_PHASE_TIMEOUT_MS` (30 seconds by default; `0` disables).
- Phase events carry worker/frame, actual browser executable/version when available, CanvasDrawElement activation, and GPU backend. The aggregate watchdog includes the last phase for every worker if it remains the final safety net.
- Parent cancellation and encoder interruption still bypass fallback.

System Chrome causality remains unproven because the report did not include an exact browser build, CanvasDrawElement capability, or GPU backend comparison.

## Test plan

- Engine parallel deadline/coordinator/peer-abort suites: 53 tests passed.
- Producer streaming-stage lifecycle/watchdog suite: 12 tests passed.
- Producer sequential-stall retry policy: 4 tests passed.
- Engine and producer package typechecks passed.
- Changed-file oxlint, oxfmt, and diff checks passed.

The full orchestrator suite previously passed 203/204 tests. The remaining `alpha + outputResolution` contract-drift assertion is unrelated to capture-stall recovery.

### DIFF
--- a/packages/cli/src/commands/render.ts
+++ b/packages/cli/src/commands/render.ts
@@ -1550,6 +1550,7 @@ function trackRenderMetrics(
     deBlankRecaptures: perf?.drawElement?.blankRecaptures,
     deBoundaryFrames: perf?.drawElement?.boundaryFrames,
     deNcprFallbacks: perf?.drawElement?.ncprFallbacks,
+    deFrameTimeouts: perf?.drawElement?.frameTimeouts,
     compositionDurationMs,
     compositionWidth: perf?.resolution.width,
     compositionHeight: perf?.resolution.height,
--- a/packages/cli/src/telemetry/events.ts
+++ b/packages/cli/src/telemetry/events.ts
@@ -297,6 +297,7 @@ export function trackRenderComplete(
     deBlankRecaptures?: number;
     deBoundaryFrames?: number;
     deNcprFallbacks?: number;
+    deFrameTimeouts?: number;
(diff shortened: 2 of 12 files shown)


miga-heygen (approved):
Per-frame drawElement deadline, parallel worker phase diagnostics, file server health probing, and transient single-worker recovery — all correctly raced from outside the puppeteer evaluate (the only viable approach). Test coverage is thorough. — Miga

PR #3701, opened 2026-09-05
fix(check): retry system Chrome after managed launch crash

miguel-heygen:
## What

`check` and the other shared screenshot diagnostics now recover when the managed Windows chrome-headless-shell exits with `STATUS_STACK_BUFFER_OVERRUN`. They retry once with an installed system Chrome, or return actionable `HYPERFRAMES_BROWSER_PATH` guidance when no fallback exists.

## Why

HyperFrames already recognizes exit code `3221225595` and explains the system-Chrome workaround for `render`, but the shared page-open path closed and rethrew the raw launch error. A valid composition therefore failed before navigation even though Chrome was available on the host.

## How

The screenshot page-open boundary reuses the existing Windows crash detector, retries only managed cache/download sources, and recomputes browser GPU policy for the fallback executable. System Chrome discovery now checks the standard machine-wide and per-user Windows install locations before `where`, so Chrome need not be on `PATH`. Explicit/system browser choices and unrelated launch failures retain their existing behavior.

## Test plan

- Five browser-manager, page-open, crash-detector, and check suites: 96 tests passed
- Changed-file oxlint, oxfmt, and diff checks passed

The Windows binary crash was injected on this Linux host rather than rerun on Windows. CLI-wide typecheck remains unavailable because optional AWS/GCP dependencies are absent and unrelated Studio/schema errors predate this change; it reported no changed-path error.

### DIFF
--- a/packages/cli/src/browser/manager.test.ts
+++ b/packages/cli/src/browser/manager.test.ts
@@ -229,6 +229,23 @@ describe("findBrowser — cache resolution", () => {
     );
   });
 
+  it("finds Chrome in the standard Windows Program Files location", async () => {
+    Object.defineProperty(process, "platform", { value: "win32", configurable: true });
+    const windowsChrome = join(
+      "C:\\Program Files",
+      "Google",
+      "Chrome",
+      "Application",
+      "chrome.exe",
+    );
+    installFsMocks({ existing: new Set([windowsChrome]) });
+    installPuppeteerBrowsersMock();
+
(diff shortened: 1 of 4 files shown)


miga-heygen (approved):
Windows STATUS_STACK_BUFFER_OVERRUN recovery with system Chrome fallback is well scoped — crash code matching, proper error enrichment, and all three paths (retry success, no system Chrome, unrelated error) tested. — Miga

PR #3702, opened 2026-09-05
fix(render): reject unsupported ProRes rate controls

miguel-heygen:
## What

MOV renders now reject `--crf` and `--video-bitrate` instead of silently discarding them. CLI help and rendering docs state that MOV uses one fixed alpha-preserving ProRes 4444 profile across quality tiers.

## Why

Both flags were accepted and threaded into disk and streaming execution, but each ProRes argument builder returns its fixed 4444 configuration before reading generic rate controls. Different commands therefore produced byte-identical encoder arguments without warning.

## How

Format-specific validation runs once in the shared render-plan boundary before disk, streaming, Docker, or batch routing. The error explains that CRF/target bitrate apply to MP4/WebM and directs users to remove the flag or choose one of those formats. ProRes codec, profile, vendor, pixel format, and alpha behavior are unchanged.

## Test plan

- [x] Documentation updated

- Render-plan suite: 15 tests passed, covering both rejected MOV flags and retained MP4/WebM controls
- Fixed-profile disk/streaming ProRes controls: 4 tests passed
- Changed-file oxlint, oxfmt, and diff checks passed

CLI-wide typecheck remains unavailable because optional AWS/GCP dependencies are absent and unrelated Studio/schema errors predate this change; it reported no changed-path error.

### DIFF
--- a/docs/guides/rendering.mdx
+++ b/docs/guides/rendering.mdx
@@ -95,6 +95,9 @@ browser already supports the source or you are diagnosing the proxy itself.
 ## Choose quality and frame rate
 
 The default `standard` quality is the right choice for most finished work.
+MOV is an alpha-preserving editing intermediate and always uses the fixed
+ProRes 4444 profile. `--crf` and `--video-bitrate` do not map to that profile
+and are rejected for MOV; choose MP4 or WebM when you need those controls.
 
 ```bash
 # Faster review version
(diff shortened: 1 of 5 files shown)


miga-heygen (approved):
Correct rejection of CRF/bitrate controls for ProRes MOV. Docs, CLI descriptions, and tests all updated consistently. — Miga

PR #396, opened 2026-04-21
perf(player): coalesce _mirrorParentMediaTime writes

vanceingalls:
## Summary

Coalesce writes to `el.currentTime` inside `_mirrorParentMediaTime` so a single jitter sample no longer triggers a parent-media seek. A drift correction now requires **two consecutive samples** above the threshold (~`MIRROR_DRIFT_THRESHOLD_SECONDS`) before the player writes back. One-shot alignment paths (`promoteToParentProxy`, `_onIframeMediaAdded`) opt out via `force: true` so initial alignment stays immediate.

## Why

Step `P1-4` of the player perf proposal. `_mirrorParentMediaTime` is called every animation frame on parent media proxies. Even without true drift, browser internals report tiny jitter on `currentTime` reads — typically below 30 ms but occasionally crossing the threshold for a frame. Writing to `currentTime` triggers a seek, which is expensive *and* invalidates pipeline buffers, which causes the next frame's reading to jitter further. The result was unnecessary seek thrash on otherwise-aligned media.

By requiring two consecutive over-threshold samples, transient jitter is filtered out while real drift (a sustained offset) still corrects within ~1 frame of latency. This eliminates the most common cause of dropped frames on the studio thumbnail grid.

## What changed

- Each `_parentMedia` entry gains a `driftSamples` counter that increments while the absolute drift is above `MIRROR_DRIFT_THRESHOLD_SECONDS` and resets to 0 on the first sample below.
- `_mirrorParentMediaTime(el, opts)` only writes back when `driftSamples >= 2`, except when `opts.force === true`.
- `promoteToParentProxy` and `_onIframeMediaAdded` pass `force: true` so the first alignment after registration is still immediate (these are user-visible state transitions, not steady-state telemetry).

## Test plan

- [x] 11 new unit/integration tests in `hyperframes-player.test.ts` covering:
  - Single-sample jitter does not trigger a write.
  - Two-sample sustained drift does trigger a write.
  - Trending drift correction (gradually increasing offset) is detected within 2 samples.
  - `force: true` override bypasses the sample requirement.
  - Out-of-range proxies (proxies whose source has been removed) do not panic.
  - Multiple proxies maintain independent counters — drift on one does not affect the other.
  - `_promoteToParentProxy` alignment is immediate.

## Stack

[...]

PR #395, opened 2026-04-21
perf(player): scope MutationObserver to composition hosts

vanceingalls:
## Summary

Replace the body-wide `MutationObserver` in `<hyperframes-player>` with one scoped to top-level `[data-composition-id]` hosts. The wide observer fired on every body-level mutation — analytics scripts, runtime telemetry markers, dev overlays — even though only composition subtrees can introduce new timed media (`<audio data-start>`, etc.).

## Why

Step `P1-2` of the player perf proposal. The previous implementation observed `iframe.contentDocument.body` with `subtree: true` to pick up sub-composition `<audio data-start>` elements added after initial mount. That worked, but it was paying for callbacks from every unrelated DOM mutation in the iframe — most of which are just runtime instrumentation. Hot paths in the studio (timeline updates, telemetry markers) end up triggering the observer dozens of times per frame.

Scoping to composition hosts cuts the noise by ~10× in the studio without losing any of the timed-media wiring guarantees.

## What changed

- New `selectMediaObserverTargets(doc)` helper in `packages/player/src/mediaObserverScope.ts` that selects all top-level `[data-composition-id]` elements **excluding** nested ones — sub-composition hosts whose media is already covered by the parent observer's `subtree: true`.
- The player now attaches a single `MutationObserver` instance per top-level host (`subtree: true`), so callbacks still batch across hosts but skip out-of-host noise.
- Falls back to observing `body` when no composition hosts exist (e.g. blank iframe between `src` changes) — preserves prior behavior for non-composition documents and avoids breaking the bootstrap path.

## Test plan

- [x] 8 new unit tests in `mediaObserverScope.test.ts` covering empty docs, single host, multiple hosts, nested-host filtering, and the body-fallback path.
- [x] 2 new integration tests in `hyperframes-player.test.ts` spying on `MutationObserver.prototype.observe` to confirm the targets and options the player actually attaches in a real custom-element bootstrap.

## Stack

Step `P1-2` of the player perf proposal. Sits between `P1-1` (shared adopted stylesheets) and `P1-4` (coalescing parent media-time mirror writes) — together they target the studio multi-player render path. The perf gate scenarios in `P0-1*` will pick up the wins automatically.

PR #3708, opened 2026-09-05
fix(studio): avoid inline style regex backtracking

jrusso1020:
Fixes CodeQL alerts [#395](https://github.com/heygen-com/hyperframes/security/code-scanning/395) and [#396](https://github.com/heygen-com/hyperframes/security/code-scanning/396). Adding an inline style to a tag without an existing style can spend quadratic time retrying the self-closing regex over long whitespace runs.

Replace the regex checks with `endsWith("/")` and remove the terminal slash before `trimEnd()`. This preserves the existing exact-ending requirement and whitespace removal while avoiding backtracking. Two production lines change; no markup parsing or editing policy changes.

Validation: original regex exceeded a two-second subprocess timeout on 100k spaces. All 36 source-patcher tests pass, including 80 new ID/selector assertions covering ordinary and Unicode whitespace, slash/newline boundaries, long malformed tags and self-closing tags. A deterministic 100k-input old/new comparison found no output differences. Full workspace build, Studio typecheck, lint and format pass.

### DIFF
--- a/packages/studio/src/utils/sourcePatcher.test.ts
+++ b/packages/studio/src/utils/sourcePatcher.test.ts
@@ -66,6 +66,23 @@ describe("applyPatchByTarget", () => {
     expect(result).not.toContain("/ style");
   });
 
+  it.each(["", " ", "\t\r\n", "\u00a0\u2028", " ".repeat(100_000)])(
+    "preserves self-closing whitespace handling (case %#)",
+    (whitespace) => {
+      const prefix = '<img id="hero" class="hero"';
+      const op: PatchOperation = { type: "inline-style", property: "opacity", value: "0.5" };
+      for (const suffix of ["/", "/\n", "/\r", "/\r\n", "/\u2028", "", "x", "/ "]) {
+        const tag = prefix + whitespace + suffix;
+        const selfClosing = suffix === "/";
+        // The original operation removes whitespace before the slash only.
+        const expectedBase = selfClosing ? prefix + suffix.slice(1) : tag;
+        const expected = `${expectedBase} style="opacity: 0.5"${selfClosing ? " /" : ""}>`;
+        expect(applyPatch(tag + ">", "hero", op)).toBe(expected);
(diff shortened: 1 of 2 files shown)


miguel-heygen (approved):
Reviewed at `bc107b1de735057cc257f0da1e5dfba91cb33689`; no blockers.

The replacement follows CodeQL `js/polynomial-redos`'s structural recommendation and is equivalent to the removed operations (`packages/studio/src/utils/sourcePatcher.ts:214-217`):

- `/\s*\/$/.test(tag)` is true exactly when `tag.endsWith("/")`, because `\s*` may match zero characters;
- on that branch, `tag.replace(/\s*\/$/, "")` removes the terminal slash plus all whitespace immediately before it, exactly what `tag.slice(0, -1).trimEnd()` does;
- when whitespace follows the slash—including CR/LF or Unicode line separators—both forms classify the tag as non-self-closing and preserve its bytes before appending the style.

I checked the old/new classification and base string over every Unicode code point in before-slash, after-slash, and double-slash positions, plus 200,000 deterministic randomized boundary strings: zero mismatches. The new path handles 100k spaces in ~0.09 ms and one million in ~1.3 ms.

The table test is a useful independent oracle: five whitespace prefixes × eight terminal suffixes × both ID and selector entry points produces the stated 80 assertions (`packages/studio/src/utils/sourcePatcher.test.ts:60-76`). The expected result derives self-closing state from the suffix, not from the production operation, and pins byte preservation for ordinary tags, terminal slash normalization, and newline/Unicode boundaries. Both public entry points converge on the same `patchInlineStyleInTag`, so there is one policy rather than two copies.

CI is still running at this head; initial regression/lint gates are green, while full build, Studio/Windows checks, and JavaScript CodeQL remain landing gates. Left unmerged as requested.

**Verdict:** APPROVE
**Reasoning:** The built-in string operations are byte-for-byte equivalent to the former regex semantics while removing backtracking, and the shared implementation is covered across both patch entry points and boundary classes.

— Magi

PR #3710, opened 2026-09-05
fix(preview): prevent render processes outliving shutdown

miguel-heygen:
## Summary

Preview shutdown now closes render ownership before the server exits. Active Studio jobs are cancelled and awaited, late child registrations are terminated, and a later CLI run can recover only birth-identity-matched FFmpeg processes that became true OS orphans.

Live `--background` preview owners remain protected. Recovery requires PPID 1 plus a recorded process identity, and it revalidates that identity before both graceful and forced termination. The original report lacked PID/PPID/command/status evidence, so this fixes the validated shutdown and crash window without claiming that the observed job was definitely orphaned rather than owned by a live background server.

The child-process tracker lives at `@hyperframes/parsers/process-tracker` (next to `ff-binaries`) so every package that shells out to ffmpeg/ffprobe — engine, producer, cli, lint, studio-server — registers with one module instance; an engine dependency from studio-server would have closed the existing core ↔ studio-server package cycle. Studio-server's proxy transcodes, `-filters` probe, waveform decode and ffprobe runner, lint's ffprobe runner, and the CLI background-removal/snapshot ffmpeg runs are now tracked. `kind: "ffmpeg"` (on-disk crash-recovery record) is reserved for long-running encoders/transcodes/decodes; probes stay untagged so the synchronous identity lookup never lands on the per-file probe fan-out.

## Platform notes

- **Windows:** orphan recovery is inert. Both halves key on the POSIX "reparented to PID 1" signal, which Windows lacks, and the ownership registry is never written there. Windows relies solely on the in-process drain (`beginTrackedProcessDrain`) at shutdown, which covers Ctrl+C on every platform.
- **Registry directory** (`$TMPDIR/hyperframes-owned-processes-<uid>`): created 0o700; records are only written to and trusted from a real directory owned by the current uid with no group/other bits. A symlink, foreign ownership, or loose mode disables both halves fail-closed.
- **CI follow-up (needs `workflow` scope, not included in this PR):** the moved tracker suite must be added to the Windows `studio-engine-cli` lane so it keeps running on windows-latest. Exact change to `.github/workflows/windows-render.yml` (currently line 407):
  ```diff
  -            packages: --filter @hyperframes/engine --filter @hyperframes/cli
  +            packages: --filter @hyperframes/engine --filter @hyperframes/cli --filter @hyperframes/parsers
  ```
- **Registry trust diagnostics:** when the registry directory exists but is untrusted (symlink, not a directory, foreign uid, or group/other permission bits), both the writer and the reader now log one `[process-tracker]` warning per path per process naming the path and the failing property. An absent directory stays silent — that is the normal state when nothing has been registered.
- **Deletion guard:** `packages/engine/src/utils/processTracker.ts` and its test are listed in `ALLOWED_DELETIONS` (`scripts/check-no-main-deletions.mjs`) with the move reason, since git does not pair the rewrite as a rename; `node scripts/check-no-main-deletions.mjs --base origin/main` passes.

## Validation

- Studio render-route disposal: 41 tests, including an in-flight request rejected once `dispose()` begins (pins the post-await re-check) and a macrotask-gated disposal test (fails if the completion await is dropped).
- Orphan cleanup: 15 tests, including three that run the real `killProcessTree` against live children: birth identity changes between discovery and kill, during the SIGKILL grace period, and on a descendant only (root still owned) — removing any of the three revalidations, including the per-descendant one inside the escalation timer, fails a test.
- Process tracker (parsers): 15 tests, including symlinked and world-readable registry directories, the exit handler leaving a record whose identity no longer matches the one it wrote, and the untrusted-directory warning on both the reader and writer sides (silent for an absent directory, once per path).
- `hyperframes init` transcode (`transcodeToMp4`) is now tracked with `kind: "ffmpeg"`, closing the last untracked encoder spawn.
- Engine, producer, cli, lint, studio-server, parsers typechecks; `check:package-cycles`, `check:package-subpaths`, `check:cli-process-ownership`, `check:tracked-artifacts`; changed-file lint/format; `git diff --check`.


miga-heygen (approved):
Orderly shutdown drain sequence, FFmpeg ownership registry with identity-verified cleanup, and draining latch are all well tested and correct. — Miga

miga-heygen (commented):
Second-pass quality review at head `05de8320` (isolated worktree; 104 tests reproduced across the five touched test files: engine 11, studio-server render 40, cli preview 40, orphanCleanup 12, windowsHide 1; 8 mutations run). Two findings I consider blocking for a PR whose title is that render processes must not outlive shutdown.

**Blocking**

- **B1 — an ffmpeg spawn with no shutdown owner remains on the most reachable path.** `packages/studio-server/src/helpers/proxyTranscoder.ts:411` spawns ffmpeg with no `trackChildProcess`, no drain-set registration and no dispose hook, so it is invisible to all three new layers (drain set, render-job dispose, ownership registry). It is reachable from the embedded preview server (`studioServer.ts:866` → `createStudioApi` → `routes/preview.ts:597`) and from the fire-and-forget pre-warm at `mediaProxyPreview.ts:104` on every page load with a browser-hostile asset. Its 15-minute ceiling is Node's `spawn({ timeout })`, a parent-side timer: on Ctrl+C the child reparents to PID 1 with no ceiling at all. Register it with the same tracker as the other eight spawn sites (`grep trackChildProcess packages` lists them; this is the one ff-binary site missing).
- **B2 — the headline safety property (never kill a reused PID) is pinned by no test.** Deleting all four `stillOwned` / identity revalidations from `killProcessTree` (`orphanCleanup.ts:77-124`) leaves 13/13 green. The only identity test injects a `vi.fn()` for `killProcessTree`, so the real function never runs; it asserts the callback was handed over, not that it is honoured. One test with the real `killProcessTree` and an `identityForPid` that changes between record and kill would pin it.

**Important**

- `render.ts:140` — removing only the post-`await` `disposalPromise` re-check leaves 40/40 green, so the exact interleaving that guard exists for is unpinned. `render.test.ts:731` ("cancels and awaits every active render") passes under the await-removal mutation — it yields one microtask and asserts vacuously; the real coverage is `render.test.ts:807`.
- `ffprobe.ts:75` is the one tracked ff-binary spawn without `{ kind: "ffmpeg" }` while its direct sibling `audioPadTrim.ts:638` got it.
- The orphan-recovery half is inert on Windows and the description does not say so.
- The new registry directory under the OS temp dir is a predictable path whose ownership and mode are never validated (suspected hazard, not exploited).

**Strengths:** the `processIdentity` / `processParentPid` consolidation is genuine single-owner work; the disposal idempotency comment names the right reason; expiring off `finishedAt` instead of `createdAt` quietly fixes a pre-existing bug where a long render could be evicted mid-flight.

Mutations: 6 of 8 red as expected; the two survivors are B2 and the `render.ts:140` re-check above.

**Reasoning:** the title condition is still true at one reachable spawn site, and the safety invariant the description leads with is not exercised by the suite.

_Review by Miga_

miga-heygen:
Addressed in 5ca18080.

miga-heygen:
Addressed in c9a6f3e2 (rebased onto main).

miga-heygen (commented):
Reviewed head `c9a6f3e2d1f42eeb7ab96e1c2b19bd1b5de9be00` in an isolated worktree. Re-pass of my review at `05de8320` (5189504103).

**Strengths**
- `packages/studio-server/src/routes/proxyTranscoder.ts:420` — the transcode is now tracked with `kind: "ffmpeg"`, and the comment at :416-419 correctly explains why `spawn({ timeout })` is not a shutdown owner. Prior blocker B1 closed.
- `packages/cli/src/orphanCleanup.test.ts:113-181` — real `killProcessTree` against a live child; 3 of 4 PID-reuse revalidations are now mutation-pinned (`:87`+`:106` together, `:117`, `:61`), and both `render.ts` disposal checks (`:140`, `:371`, `:376`) die under mutation. Prior blocker B2 closed.
- The `kind` rule is documented once on `TrackChildProcessOptions.kind` (`packages/parsers/src/processTracker.ts:36-45`) and applied consistently: every encoder/transcode site tagged, every probe untagged (audited all 14 sites).

**Blockers**

1. **`Detect changes` fails at this head, so no Linux lane ran.** `.github/workflows/ci.yml:48-50` runs `scripts/check-no-main-deletions.mjs` before the paths filter. This PR deletes `packages/engine/src/utils/processTracker.ts` and `processTracker.test.ts` (git does not pair them as renames with the 316-line parsers module), and neither is in `ALLOWED_DELETIONS`. Reproduced locally:
   ```
   $ node scripts/check-no-main-deletions.mjs --base origin/main
   This branch deletes 2 files that exist on origin/main:
     packages/engine/src/utils/processTracker.test.ts
     packages/engine/src/utils/processTracker.ts
   ```
   Run 34877642393 → failure; the paths-filter step skipped, `changes.outputs.code` empty, so **Build, Test, Typecheck, Lint, CLI smoke (required), Producer, Studio, Fallow, SDK all show `skipped`** — not green. Fix: add both paths to `ALLOWED_DELETIONS` with the move reason, then confirm Test/Typecheck actually execute.

2. **`packages/cli/src/commands/init.ts:188` — `transcodeToMp4()` spawns an untracked `ffmpeg -c:v libx264` re-encode.** No `trackChildProcess`, no timeout, no dispose hook. Reachable from `init.ts:451` when `hyperframes init` offers to re-encode a non-web-compatible video; Ctrl+C during the spinner reparents a minutes-long encoder to PID 1, and untagged it is invisible to the recovery registry. Same class as B1; the commit title claims every ffmpeg child is owned. Fix: `trackChildProcess(child, { kind: "ffmpeg" })`.

3. **The moved tracker suite runs on no lane.** `.github/workflows/windows-render.yml:407` scopes studio-engine-cli to `--filter @hyperframes/engine --filter @hyperframes/cli`; `packages/parsers` is built at `:464` but never tested on Windows. The module header says Windows relies solely on the in-process drain, so this drops that layer's only CI coverage. With (1), the suite ran on zero lanes at this head. No `.github/**` change is in the diff. Fix: add `--filter @hyperframes/parsers` to that lane.

**Important**
- `packages/cli/src/orphanCleanup.ts:120` — the per-descendant identity re-check inside the SIGKILL escalation is unpinned (deletion leaves 14/14 green); the root guard at `:117` returns first in both new tests.
- `registerOwnedFfmpeg` / `findOwnedOrphanedFfmpegProcesses` fail closed on an untrusted registry dir but log nothing on either side; a pre-existing group-readable `hyperframes-owned-processes-<uid>` silently disables recovery for good. One debug line would make it diagnosable.

**Nits**
- `packages/studio-server/src/routes/render.ts:70` duplicates the `:140` 503 literal; deleting `:70` alone leaves 41/41 green. One helper would stop the message/status drifting.
- `processTracker.ts` `remove()` identity-compare-before-unlink is untested.

**Verified clean:** no package cycle (parsers imports nothing back), engine re-export keeps every prior import path, no stale `engine/process-tracker` subpath references; waveform partial-output rejection leaks no cache file (`waveform.ts:87-92`, caller awaits before staging).

**Ran:** 144 tests green across 6 files (parsers 13, cli 54, studio-server 43, lint 34); `tsc --noEmit` clean in parsers, engine, studio-server, cli, lint, producer; 11 mutations (7 red, 4 green as noted).

**Verdict:** COMMENT (With fixes)
**Reasoning:** Both prior blockers are genuinely closed, but the head has no Linux CI verification (required checks skipped, not passed), leaves one ffmpeg transcode unowned on a first-run path, and moves the tracker suite off every test lane.

_Review by Miga_

miga-heygen:
Addressed in 596bab00.

miga-heygen (approved):
Reviewed head `596bab00b` in an isolated worktree. Re-pass of my review at `c9a6f3e2` (5201399459).

**Prior blockers**
- **C1 fixed.** `scripts/check-no-main-deletions.mjs:160-167` allowlists both engine tracker paths with the move reason; the guard exits 0 locally and `check-no-main-deletions.test.mjs` passes 8/8. At this head every required check executes and passes (Detect changes, Build, Test, Test: runtime contract, Typecheck, CLI smoke, Tests on windows-latest and both lanes, Render on windows-latest, regression); the Linux `Test` job now runs `processTracker.test.ts`.
- **C2 fixed.** `packages/cli/src/commands/init.ts:211` tracks the actual child from the single `spawn` at `:189` with `kind: "ffmpeg"`; the `findFFmpeg()`-null path never spawns. The old code never read the child's stdio, so `pipe → ignore` at `:205` loses nothing and removes an unread-pipe stall on long encodes. Exit-code handling unchanged.
- **C3 open, downgraded to important.** `.github/` is untouched: `windows-render.yml:407` still filters engine+cli, so no Windows lane runs `packages/parsers/src/processTracker.test.ts`. The author documents the one-line `--filter @hyperframes/parsers` addition as a workflow-scope follow-up. With Linux coverage restored by C1 this is a coverage gap for the Windows drain path, not a defect in this diff, so it no longer holds the PR; it should land before the next Windows-affecting change to that module.

**Non-blocking items from the prior review, verified**
- `orphanCleanup.ts:59,89` — one `identityForPid` oracle threaded through `killProcessTree`; no second lookup remains. The timer re-check at `:125` is now mutation-pinned by the live-child test at `orphanCleanup.test.ts:188`.
- `processTracker.ts:124-136` — the trust rules live only in `registryDirTrustRejection`; one warn site shared by writer and reader, deduped per path per process, absent dir silent; both the dedupe and the mode rule are pinned by tests.
- `remove()` identity compare pinned at `processTracker.test.ts:146`.
- `render.ts:14` — a single "shutting down" literal; the post-await site is pinned.

**Nits**
- `render.ts:74` pre-resolve early-out is still not independently pinned (the `:144` check catches everything it does).
- `init.ts:211` tracking has no test, consistent with the other two cli tracking sites.
- `orphanCleanup.ts:107` initial identity map is not pinned; same oracle, low risk.

**Ran:** parsers 15/15, cli orphanCleanup 15/15, studio-server render 41/41, deletion-guard tests 8/8; `tsc --noEmit` clean in parsers, cli, studio-server, engine; 7 mutations (5 red as expected, 2 probes green as noted). All reverted.

**Verdict:** APPROVE
**Reasoning:** Both defects are closed and pinned, required checks are green and executing; the remaining Windows lane filter is a documented workflow follow-up rather than a fault in this change.

— Miga

PR #714, opened 2026-05-11
fix(core): thread playback rate into WebAudio audio sources

jrusso1020:
## What

Threads the transport playback rate into `WebAudioTransport` so audio clips routed through it advance at the same rate as the rest of the player (transport clock, GSAP timelines, native `<video>` elements).

Closes #713.

## Why

The player exposes a `playback-rate` attribute and a speed selector. At non-1× rates, the transport clock, GSAP timelines, and native `<video>` elements all advance at the new rate — but `<audio>` clips routed through `WebAudioTransport` kept playing at 1× because `AudioBufferSourceNode.playbackRate` was never set. Result: visuals run at 2×, audio plays at 1×, hard desync within seconds.

## How

- `schedulePlayback(...)` now accepts an optional `rate` (default 1):
  - Sets `sourceNode.playbackRate.value = rate` on the buffer source.
  - Scales the future-clip start `delay` by `1 / rate` so a future clip still fires at the right composition-time boundary (composition time advances `rate ×` wallclock).
  - In-progress buffer offset stays `elapsed + mediaStart` — that quantity is rate-independent (wallclock-elapsed × rate = composition-elapsed).
- New `setRate(rate)`:
  - Updates `playbackRate.value` on every active source in place.
  - Rebases the composition-time reference frame (`_rateAnchorCtx` / `_rateAnchorComp`) so `getTime()` stays continuous across the rate change — the audio-master clock keeps reporting the same composition time at the moment of the change, then advances at the new slope.
- Runtime wiring (`packages/core/src/runtime/init.ts`):
  - `onSetPlaybackRate` now also calls `webAudio.setRate(state.playbackRate)` after `applyPlaybackRate` and `transportClock.setRate`.
  - `player.play()` schedules each audio clip with `state.playbackRate`, so playback that starts after a non-1× rate has been chosen begins at the right rate.

Non-finite / non-positive rates are normalized to 1, matching the existing `applyPlaybackRate` policy.

Known limitation (called out in the JSDoc on `setRate`): sources scheduled to start in the future via `start(when, ...)` retain their original wallclock start time. Callers that need rate-correct rescheduling for future-start clips should `stopAll()` and reschedule; for the in-tree paths, future audio is rescheduled on every `play()` / `pause()` / `seek()` so this is currently a non-issue.

## Test plan

Used TDD — wrote the failing tests first, then implemented.

- [x] Unit tests added/updated — 8 new tests in `webAudioTransport.test.ts`:
[...]

PR #3711, opened 2026-09-05
fix(studio-server): simplify normalized group ID trimming

jrusso1020:
Addresses CodeQL alert [#714](https://github.com/heygen-com/hyperframes/security/code-scanning/714) in group DOM ID generation. The preceding normalization already collapses every non-alphanumeric run into one hyphen, so the final leading/trailing hyphen repetitions are redundant. I did not find reachable quadratic behavior through the current normalization chain.

Trim one hyphen at each edge and document that invariant. Generated IDs, group labels, member order and collision suffixing remain unchanged; no rule suppression or alert dismissal is added.

Validation: all 101 source-mutation tests pass, including eight new grouping cases covering punctuation, Unicode, empty/default names, 200k-character hyphen runs and occupied base/base-2 IDs. A deterministic 100k-name comparison matched old/new normalized IDs. Parser/core builds, Studio-server typecheck, lint and format pass.

### DIFF
--- a/packages/studio-server/src/helpers/sourceMutation.ts
+++ b/packages/studio-server/src/helpers/sourceMutation.ts
@@ -478,7 +478,8 @@ function uniqueGroupDomId(document: Document, groupId: string): string {
       .trim()
       .toLowerCase()
       .replace(/[^a-z0-9]+/g, "-")
-      .replace(/^-+|-+$/g, "") || "group";
+      // Normalization above leaves at most one hyphen at either edge.
+      .replace(/^-|-$/g, "") || "group";
   let id = base;
   let n = 2;
   while (document.getElementById(id)) {
--- a/packages/studio-server/src/helpers/sourceMutationSplitAndGroup.test.ts
+++ b/packages/studio-server/src/helpers/sourceMutationSplitAndGroup.test.ts
@@ -197,6 +197,32 @@ describe("wrapElementsInHtml / unwrapElementsFromHtml", () => {
     return element;
   }
 
+  it.each([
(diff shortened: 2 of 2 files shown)


miguel-heygen (approved):
Reviewed at exact head `cbc6d5c388003ebfae45eb0cb33552c68f72ee8d`. No blockers.

The invariant at `packages/studio-server/src/helpers/sourceMutation.ts:477-482` is sound: after lowercasing, `/[^a-z0-9]+/g` maps every maximal non-alphanumeric run to one hyphen, so either edge contains zero or one hyphen. Replacing `^-+|-+$` with `^-|-$` is therefore behavior-equivalent on every reachable intermediate string; this removes the analyzer's repeated quantifiers without adding another normalization owner.

The table at `packages/studio-server/src/helpers/sourceMutationSplitAndGroup.test.ts:200-223` pins all downstream contracts together: normalized DOM ID, occupied base/base-2 collision suffixing to base-3, raw `data-hf-group` label preservation, and member order. The 200k edge-run case covers the adversarial input through the public mutation path.

Independent verification:
- parser, lint, and core prerequisite builds passed
- all 101 source-mutation tests passed
- Studio-server typecheck and changed-file format check passed
- old/new normalization matched across 299,596 exhaustive and long inputs

Fresh CI, Windows, and JavaScript CodeQL were still finishing when I submitted; they remain landing gates.

**Verdict:** APPROVE
**Reasoning:** The normalization proof is complete, downstream observable behavior is pinned, and exact-head local verification found no regression.

— Magi

PR #3713, opened 2026-09-05
fix(lint): block executed invalid raw selectors

miguel-heygen:
## Summary

`hyperframes check` now blocks a digit-leading ID only when source proves that its raw `#id` selector is executed by a GSAP tween or `querySelector(All)`. The existing `id_requires_css_escape` warning remains for legal HTML IDs that are unused, accessed through `getElementById`, or escaped with `CSS.escape`/CSS syntax.

The detector reads literal calls from the Acorn AST. It does not infer danger from a parser-resolved element variable, computed selector, comment, or displayed string, which keeps the new error tied to a browser-fatal execution path.

## Validation

- Full GSAP lint suite: 168/168 tests passed.
- Core lint controls: 59/59 tests passed.
- Parser and lint package typechecks passed.
- Changed-file format, lint, and diff checks passed.


miga-heygen (approved):
Reviewed: error-level `invalid_raw_selector_execution` correctly catches digit-leading IDs used as raw CSS selectors in GSAP and querySelector calls. Parser extracts only literal string selectors. Properly scoped — escaped and getElementById paths not flagged. — Miga

miga-heygen (commented):
Second-pass quality review at head `7e8f3174` (isolated worktree, `gsap.test.ts` 168/168 reproduced). Three findings I consider blocking for an error-severity rule that gates `lint` and `check`.

**1. Proven false positive on a valid selector.** `packages/lint/src/rules/gsap.ts:1064-1065` decides validity with a substring regex `#<id>(?![\w-])` over the whole selector string, so `document.querySelector('a[href="#123-frame"]')` is reported as a browser-fatal invalid selector. jsdom confirms that selector does not throw, while `#123-frame` and `:is(#123-frame)` do. Because the finding is `error` and the error count gates `hyperframes lint` (`packages/cli/src/commands/lint.ts:59-70`) and `check` (`checkPipeline.ts:1355`), a valid anchor-link composition now fails with a fix hint that does not apply. Test the selector, not the string: the token has to be an id selector, not any occurrence of `#id` inside an attribute value or string.

**2. Sibling sinks with the identical precondition are not covered.** `extractLiteralGsapSelectorCalls` (`packages/parsers/src/gsapParserAcorn.ts:1906`) delegates to `findAllTweenCalls`, which only accepts timeline-rooted calls or the hard-coded `gsap.set` (`:945-957`). Measured at this head: `gsap.set("#123-frame")` is caught; global-scope `gsap.to` / `gsap.from` / `gsap.fromTo` are not, nor are `matches`, `closest`, `gsap.utils.toArray` or `scrollTrigger.trigger` — all string literals, all equally provable, all equally fatal at runtime. The same file already defines `QUERY_METHODS` (`:39`) and already treats `toArray` as a selector sink (`:297`); the new extractor reuses neither.

**3. Three owners of "which selector is browser-invalid".** `/^\d/` at `core.ts:193` and again verbatim at `gsap.ts:1061`, beside the real `postcss-selector-parser` analysis already imported in `core.ts:3,47` — which handles `:is()`/`:where()` correctly and is the natural base for finding 1. Message and `fixHint` text are duplicated near-verbatim between the warning and the new error. Two definitions of one decision will drift.

Non-blocking: the extractor runs the whole timeline-resolution pipeline and discards it (that machinery is why finding 2 exists); duplicated interfaces and dynamic-import loaders; another copy of `escapeRegExp` and of the `#id` regex; the same defect is reported twice (warning + error), which the PR's own test asserts; the rule is inserted at index 0, renumbering the positional gsap rule indices the linter documents.

Verification: mutation dropping both `report(...)` loops → 2 failed / 166 passed, so the positive tests pin the rule; the negative test passes with the rule disabled. 11-case adversarial probe with a jsdom validity oracle (linkedom accepts `#123-frame`, so it is not a usable oracle here). Not run: full repo suite, typecheck, CLI end to end.

**Reasoning:** an error-level gate with a proven false positive and uncovered sibling sinks will both block valid work and miss the failures it exists to catch.

_Review by Miga_

miga-heygen:
Addressed in 8c12784a

miga-heygen (commented):
Second-pass quality review at head `8c12784a` (isolated worktree; rebase checked against the PR file list: merge-base `cfe5dcfa`, two commits, 4 files, `gsap.ts` byte-identical to main; parsers gsap suites 714 passed, lint suite 664 passed; all required checks green).

**All three findings from my review at `7e8f3174` are fixed and verified.** `core.ts:33-42` parses with postcss-selector-parser and walks id nodes, so `a[href="#123-frame"]`, `[data-x="#123"]` and the escaped `#\31 23` are clean while `#123-frame`, `#123:hover` and `.a #123` still error. `gsapParserAcorn.ts:1939-1986` is one extractor: 14 sinks caught (`gsap.to/from/fromTo/set`, `tl.to`, a `const t = tl` alias, `gsap.timeline().to` chains, `ScrollTrigger.create({trigger})`, inline `scrollTrigger:{trigger}`, `gsap.utils.toArray`, `matches`, `closest`, `querySelector(All)`), and the four false-positive controls stay silent (`material.color.set("#1a1a2e")`, `Map.set`, a non-GSAP `someObj.to`, `Set().add`). `/^\d/` now appears exactly once in lint+parsers (`core.ts:22`), one fix-hint constant, and warning and error are mutually exclusive per id. Merging into the existing index-0 core rule keeps every positional rule index unchanged. Mutations: disable the id-token walk → 16 red; drop the receiver classification → exactly the `material.color.set` control goes red.

**One new blocker, same class as the one just fixed.** `core.ts:33-42` walks every id node regardless of the enclosing pseudo-class. `:is()` and `:where()` take *forgiving* selector lists, so `:is(#123)` does not throw, yet the rule reports `invalid_raw_selector_execution` at `error` with a message that says it throws a SyntaxError. Checked against a real Chromium 152 oracle: `:is(#123)`, `:where(#123)`, `:is(.a, #123)`, `:is(#123) span`, `.a:is(#123)`, `:is(:not(#123))` and `:not(:is(#123))` are all valid and all reported as errors; `:not()`, `:has()` and `:nth-child(2 of #123)` correctly throw and must keep erroring. The PR's own test pins the false positive — `core.test.ts:119` (the ":is() wrapper" row) asserts the error and the absence of the warning. jsdom is not a usable oracle here: it throws on `:is(#123)`, which is presumably how the row was written. Fix, about five lines: skip id nodes whose parent chain includes an `:is`/`:where` pseudo (postcss-selector-parser exposes the parent), flip the `:119` row to expect the warning only, and add a positive `:not()`/`:has()` row. Verified 11/11 against the oracle with that change.

**Non-blocking (under-reports, documented as intentional, fine as follow-ups):** `anim.tl.to` object-property alias, `[tl][0].to`, the `scrollTrigger: "#123"` string shorthand (the most realistic gap), `ScrollTrigger` `scroller`/`pinnedContainer`, array targets, `gsap.utils.selector`, string concatenation, and `const sel = "#123"` bound in scope. Template literals with interpolation are skipped, which is right for an error gate. One test-infrastructure note: lint tests resolve `@hyperframes/parsers/gsap-parser-acorn` to `dist`, so a parsers-side change shows a false green unless `bun run build` is re-run in `packages/parsers` first.

**Verdict:** COMMENT — needs one fix before merge.
**Reasoning:** the error gate is now sound everywhere except forgiving selector lists, where it fails valid compositions and the suite pins the wrong answer; the fix is small and localised.

_Review by Miga_

miga-heygen:
Addressed in 3fa6fcdc

miga-heygen (approved):
Reviewed head `3fa6fcdcdb678ae6651214342329db300b220132` in an isolated worktree. The one blocker from my review at `8c12784a` is closed: `browserInvalidIdTokens` (`packages/lint/src/rules/core.ts:26-56`) now skips id nodes whose parent chain contains a forgiving-list pseudo, and that pseudo set has a single owner (`FORGIVING_LIST_PSEUDOS`, also read by `repeatedDescendantId` at `:80`). I compared the rule against a real Chromium 148 oracle on sixteen selectors, including `:is(:not(#123))`, `:not(:is(#123))`, `:has(:is(#123))`, `:is(:has(#123))`, `:IS(#123)`, `:-webkit-any(#123)`, `:nth-child(2 of #123)` and the escaped `#\31 23`: the rule agrees with the browser on every row (warning where the browser accepts, error where it throws; `:-webkit-any()` is correctly left as an error since Chromium rejects it). `core.test.ts` 105/105; removing the forgiving skip turns the seven warning-only rows red, and extending the skip to `:not()` turns the `:not()` positive row and the repeated-descendant `:not(#scene_01) #scene_01` row red, so both directions are pinned. The delta since `8c12784a` is confined to `core.ts` and `core.test.ts`. Earlier non-blocking notes (intentional under-reports such as the `scrollTrigger: "#123"` string shorthand and object-property timeline aliases) stand as follow-ups. — Miga

PR #3714, opened 2026-09-05
fix(skills): guard canonical stores from mirror aliases

miguel-heygen:
## Summary

Global skill mirroring now proves that each agent target is filesystem-distinct from the canonical Claude and universal stores before creating, deleting, copying, or linking anything. Symlink aliases and unresolvable loops are skipped and reported, so an update cannot delete the source skill through an alias and replace it with a self-pointing link.

The guard resolves missing tails through their nearest existing ancestor, compares canonical path overlap and device/inode identity, and rechecks immediately before each destructive per-skill operation. A final per-skill symlink remains safe and supported because removal unlinks that entry without traversing its target; this preserves normal Unix idempotency and Windows copy refreshes.

## Validation

- Skills mirror suite: 14/14 tests passed, including absolute, relative, intermediate-directory, and ELOOP aliases.
- Skills command suite: 40/40 tests passed.
- Targeted strict TypeScript check for the mirror and generated agent-directory table passed.
- Changed-file format, lint, and diff checks passed.

### DIFF
--- a/packages/cli/src/commands/skills.test.ts
+++ b/packages/cli/src/commands/skills.test.ts
@@ -119,7 +119,7 @@ vi.mock("../utils/skillsManifest.js", async (importOriginal) => {
 // the real $HOME. Stub it so these arg-shape tests never create symlinks in the
 // dev machine's agent dirs — the mirror has its own isolated-HOME unit tests.
 vi.mock("../utils/skillsMirror.js", () => ({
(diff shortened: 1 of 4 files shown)


miga-heygen (approved):
Reviewed: symlink-alias safety hardening for skill mirror. `targetSafety` correctly fails closed using realpath resolution, inode comparison, and path overlap detection. Self-loop and intermediate alias cases tested. Important safety fix. — Miga

PR #743, opened 2026-05-12
fix(studio): add rotation field, inline element drag, fix manifest load regression

vanceingalls:
## Summary

Three Studio fixes for the property panel and canvas editing.

### Rotation field in property panel
Added "R" (rotation) field to the geometry row (X, Y, W, H, R). Goes through the manifest via `handleDomRotationCommit`, resettable with "Reset edits". Previously rotation was only available via canvas drag handle.

### Inline element drag promotion
`display: inline` elements (like `<span>`) now auto-promote to `inline-block` when dragged on canvas. CSS `translate` is ignored on inline elements, so dragging had no visible effect. The promotion happens in both `applyStudioPathOffset` and `applyStudioPathOffsetDraft`.

### Manifest load regression fix
The polling loop fix (#722) changed iframe load to skip disk reads when `forceFromDisk` was false. But iframe load needs to read the manifest from disk to populate `studioManualEditManifestRef` — without this, "Reset edits" couldn't find any entries to remove. Fixed by passing `readFromDiskFirst: true` on both the initial mount and `handleLoad` paths.

## Test plan

- [ ] Select any element — geometry row should show X, Y, W, H, R fields
- [ ] Type a rotation value in the R field — element rotates, manifest updates
- [ ] Click "Reset edits" — rotation clears along with position/size
- [ ] Drag an inline `<span>` element on canvas — it should move (auto-promoted to inline-block)
- [ ] Make canvas edits, reload Studio, click "Reset edits" — edits revert correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3715, opened 2026-09-05
fix(studio-server): bound preview variable insertion scans

jrusso1020:
Fixes CodeQL alert [#743](https://github.com/heygen-com/hyperframes/security/code-scanning/743): preview variable injection repeatedly rescans unterminated `<head` or `<html` prefixes while searching for a closing `>`.

Find the first candidate prefix, then search once for its closing `>`. If that first prefix cannot close, no later prefix of the same kind can close either. Preserve the head → html → leading doctype → prepend priority, existing prefix/case semantics, insertion position and script-value escaping. Move the existing injection function into an internal helper so its boundary behavior can be tested directly; all three preview call sites use the same helper.

Validation: both original regexes exceeded a two-second subprocess timeout on 100k repeated unterminated prefixes. All 60 injection-helper and preview-route tests pass, covering long malformed tags, fallback order, case/attributes, script breakout escaping, main/sub-composition previews and ETags. Complete old/new output matched across 100k generated inputs. Parser/core/lint and Studio-server builds, server typecheck, lint and format pass.

### DIFF
--- a/packages/studio-server/src/helpers/mediaProxyPreview.ts
+++ b/packages/studio-server/src/helpers/mediaProxyPreview.ts
@@ -111,7 +111,7 @@ export async function injectMediaCodecMapIntoHtml(
     });
   }
   // <-escape prevents a src path containing "</script>" from breaking out of
-  // the injected tag, mirroring injectPreviewVariables in routes/preview.ts.
+  // the injected tag, mirroring injectPreviewVariables in helpers/previewVariables.ts.
   const json = JSON.stringify(map)
     .replace(/</g, "\\u003c")
     .replace(/\u2028/g, "\\u2028")
--- a/packages/studio-server/src/helpers/previewVariables.test.ts
+++ b/packages/studio-server/src/helpers/previewVariables.test.ts
@@ -1,0 +1,43 @@
+import { describe, expect, it } from "vitest";
+import { injectPreviewVariables } from "./previewVariables";
+
+const SCRIPT = '<script data-hf-preview-variables>window.__hfVariables={"a":1};</script>';
+
+describe("injectPreviewVariables", () => {
+  it.each([
+    [
+      "<!doctype html><html><head>x</head></html>",
+      "<!doctype html><html><head>",
+      "x</head></html>",
+    ],
+    [
+      "<!DOCTYPE html><HTML lang='en'><HEAD id='h'>x",
+      "<!DOCTYPE html><HTML lang='en'><HEAD id='h'>",
+      "x",
+    ],
+    ["<!doctype html><html lang='en'>x", "<!doctype html><html lang='en'>", "x"],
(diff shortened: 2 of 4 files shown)


miguel-heygen (approved):
Reviewed at exact head `eb6dca04d0d3c00da0c8fc7c1902910f244764f0`. No blockers.

The scan at `packages/studio-server/src/helpers/previewVariables.ts:17-24` is linear and behavior-equivalent. For each of three fixed-priority prefixes it performs one literal/anchored search and, only on a match, one `indexOf` for the first `>`; a first unmatched prefix with no later `>` cannot hide a closable later prefix of that kind. The head → html → leading-doctype → prepend priority and the existing permissive boundaries (`<headless>`, `>` inside quotes) are explicitly pinned at `previewVariables.test.ts:7-35`.

All three route branches use the extracted helper at `routes/preview.ts:386`, `:411`, and `:494`. The JSON `<` escaping is byte-identical at `previewVariables.ts:10-13`, and the main/sub-composition ETag salts remain computed from the raw variable query before any injection (`routes/preview.ts:327-336`, `:457-476`), so cache separation is unchanged.

- 60/60 helper + preview-route tests passed
- complete old/new output matched across 300,017 generated and adversarial HTML inputs, including 100k unterminated head/html prefixes and the single-closer variants

**Nit:** `packages/studio-server/src/helpers/mediaProxyPreview.ts:114` still says it mirrors `injectPreviewVariables` “in routes/preview.ts”; this extraction moved the implementation to `helpers/previewVariables.ts`. Updating that breadcrumb would keep the refactor’s ownership record accurate.

CI had no failures; JavaScript CodeQL, Windows, and remaining required jobs were still running and remain landing gates.

**Verdict:** APPROVE
**Reasoning:** The extracted helper preserves every insertion and cache/escaping contract while replacing repeated regex rescans with a bounded constant number of linear passes.

— Magi

jrusso1020:
Updated the stale mediaProxyPreview.ts reference in 7f6b9e6a8b114f5345d56099a83fd853eb7881d7 to point to helpers/previewVariables.ts. This revision only changes that comment; lint, format and signed commit hooks passed. Requesting exact-head re-review while fresh CI runs.

miguel-heygen (approved):
Re-pinned at exact head `7f6b9e6a8b114f5345d56099a83fd853eb7881d7` after verifying the post-approval delta.

`eb6dca04d0...7f6b9e6a8b` is exactly one commit and one comment line: `packages/studio-server/src/helpers/mediaProxyPreview.ts:114` now points to `helpers/previewVariables.ts`. Runtime and test blobs are unchanged, and `git diff --check` is clean. This closes the prior nit without changing the approved behavior.

Fresh CI remains the landing gate.

**Verdict:** APPROVE
**Reasoning:** The only new-head delta is the requested breadcrumb correction; the previously reviewed implementation is byte-identical.

— Magi

PR #2263, opened 2026-07-11
fix(check): allow missing caption overrides

miguel-heygen:
## What
Treat a missing root `caption-overrides.json` as an optional runtime resource in both `check` and `validate`.

## Why
The caption runtime intentionally fetches overrides only when caption groups exist and safely handles a missing file, but browser diagnostics converted the expected 404/aborted request into `http_error` and `request_failed`. Users had to create an empty JSON shim before checks could pass.

## How
- ignore only the exact root `/caption-overrides.json` request when it is absent
- keep other JSON failures and non-404 caption override responses reportable
- share the same predicates across check and validate

## Test plan
- `vitest run src/commands/validate.test.ts src/utils/checkBrowser.test.ts` (27 passed)
- `tsc --noEmit`
- patched CLI against minimal caption composition with no override file: `ok:true`, zero runtime findings
- repository pre-commit lint, format, fallow, tracked-artifact, and typecheck hooks passed

PR #3716, opened 2026-09-05
fix(check): allow missing caption overrides

miguel-heygen:
## Summary

Caption projects no longer fail `check` or deprecated `validate` when the optional root `caption-overrides.json` is absent. Only that exact resource's 404 and benign aborted fetch are ignored; a 500, another missing JSON file, or a malformed present sidecar remains an error.

The present-file contract is now explicit: the payload must be a JSON array whose entries are objects, and `[]` is the valid explicit no-op. Invalid JSON, a non-array root, or a non-object entry emits a runtime error that browser validation records instead of being silently ignored.

Related: #2263

## Validation

- Caption override runtime suite: 14/14 tests passed.
- Validate and check-browser suites: 38/38 tests passed.
- Core/runtime and CLI package typechecks passed.

### DIFF
--- a/packages/cli/src/commands/validate.test.ts
+++ b/packages/cli/src/commands/validate.test.ts
@@ -17,6 +17,7 @@ import {
   raceMediaReady,
   resolveNavigationTimeoutMs,
   shouldIgnoreRequestFailure,
+  shouldIgnoreHttpError,
 } from "./validate.js";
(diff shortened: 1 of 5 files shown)


miga-heygen (approved):
Reviewed: caption-overrides.json 404 correctly treated as silent no-op in both validate and checkBrowser. Payload validation added with proper error reporting. Tests cover missing sidecar, malformed JSON, non-array, and non-object entries. — Miga

PR #751, opened 2026-05-12
refactor(producer): snapshot cfg.forceScreenshot at compile time, stop mutating mid-pipeline

jrusso1020:
## Description

Phase 1.5 cleanup of the distributed rendering refactor (see DISTRIBUTED-RENDERING-PLAN §17 backwards-compat constraints; this is PR 1.16 in the §1.5 follow-up table). Resolves the `compileStage.ts:73` TODO by making `forceScreenshot` set once at compile time and never mutated mid-pipeline.

**What changed**

- `compileStage` folds three signals (`cfg.forceScreenshot`, `needsAlpha`, composition `renderModeHints.recommendScreenshot`) into a single boolean and returns it on `CompileStageResult.forceScreenshot`. The pre-compile `if (needsAlpha) cfg.forceScreenshot = true` flip moved from the sequencer into the stage so the resolution lives in one place.
- The sequencer stores the resolved value in a local `captureForceScreenshot`. The BeginFrame auto-worker calibration fallback still flips capture mode on a timeout, but flips the local boolean instead of `cfg.forceScreenshot`; the screenshot-mode retry uses `{ ...cfg, forceScreenshot: true }`.
- `captureHdrStage` stops mutating caller-owned `cfg.forceScreenshot = true`. The layered composite path uses a local `hdrCfg = { ...cfg, forceScreenshot: true }`. The stage asserts `forceScreenshot === true` on input to make the contract explicit — the layered branch is the only path that demands screenshot mode regardless of caller intent.
- `captureStage` and `captureStreamingStage` add `forceScreenshot: boolean` to their inputs. Each derives `captureCfg = cfg.forceScreenshot === forceScreenshot ? cfg : { ...cfg, forceScreenshot }` — identity-equal to `cfg` on the common path, so no extra allocation when the orchestrator's captured value matches the compile-time decision.
- `lefthook.yml`: grandfather `renderOrchestrator.ts` and `captureHdrStage.ts` in the 500-line `filesize` hook added in #748. Both pre-date the hook; the producer stage extractions are actively shrinking the orchestrator and the cycle-break stack will tackle `captureHdrStage` next.

**Why it matters for Phase 3**

`LockedRenderConfig.forceScreenshot` in the distributed plan (§4.3) must be computed at plan time and survive across processes / replays. With this PR, the in-process renderer already computes the value at compile time and threads it explicitly through downstream stages — making `freezePlan()` straightforward to wire up without the side-effects this code carried before.

## Testing

[...]

PR #3718, opened 2026-09-05
fix(producer): bound existing font-face recognition scans

jrusso1020:
## What

Fix CodeQL [#751](https://github.com/heygen-com/hyperframes/security/code-scanning/751): malformed font-face input can make existing-font detection spend quadratic time retrying overlapping matches before font fallback decisions.

## Why

All three scanner examples exceeded two seconds with 100,000 repeated prefixes, spaces or family declarations. Existing-font detection runs before every deterministic font injection, so this can stall compilation.

## How

Scan the font-face opener, family declaration, semicolon and closing brace in order. Skip immediately empty declarations as before; stop once a required suffix is absent, since no later opener could complete it. Successful scans advance past the closing brace. Normalize the captured family using the existing function.

Preserve existing recognition and fallback decisions, including first-family ordering, whitespace-only values and malformed-input matching. This remains the existing recognition heuristic; it does not introduce a CSS parser or change fetch/retry/system-font behavior. Two files changed; no workflow changes.

## Test plan

- [x] All 90 deterministic-font tests pass, including 12 new cases for authored fonts, fallback decisions and the three adversarial inputs.
- [x] All five font-embedding tests pass with Vitest after building workspace dependencies.
- [x] 200,000 generated inputs match the original ordered normalized family sets exactly.
- [x] The three isolated adversarial scans complete in under a millisecond locally; public injection tests also finish promptly.
- [x] Parser/core/lint/Studio-server builds, producer typecheck, formatting, lint and complexity/duplication audit pass.
- Independent Magi review and CI/CodeQL remain landing gates. Verify main-branch alert closure after merge.

### DIFF
--- a/packages/producer/src/services/deterministicFonts.test.ts
+++ b/packages/producer/src/services/deterministicFonts.test.ts
@@ -1,5 +1,49 @@
 import { describe, expect, it } from "bun:test";
-import { FONT_ALIASES, FONT_ALIAS_KEYS } from "./deterministicFonts.js";
+import {
+  FONT_ALIASES,
+  FONT_ALIAS_KEYS,
+  injectDeterministicFontFaces,
+} from "./deterministicFonts.js";
+
+describe("existing font-face recognition", () => {
+  it.each([
+    ['@font-face { font-family: "Existing Font"; src: url(font.woff2); }', false],
+    ["@FONT-FACE\n{ FONT-FAMILY : 'EXISTING FONT'; }", false],
+    ['@font-face { font-family:\u00a0"Existing Font"; }', false],
+    ['@font-face { font-family:; font-family: "Existing Font"; }', false],
+    [
+      '@font-face { font-family: "Other Font"; } @font-face { font-family: "Existing Font"; }',
+      false,
+    ],
+    ['@font-face { font-family: "Existing Font" }', true],
+    ['@font-face { font-family: "Existing Font";', true],
+    ['@font-face { font-family: "Other Font"; font-family: "Existing Font"; }', true],
+    ['@font-face { font-family:   ; font-family: "Existing Font"; }', true],
+  ])("preserves authored font and fallback decisions for %j", async (css, needsFallback) => {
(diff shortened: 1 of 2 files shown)


miguel-heygen (approved):
Reviewed at exact head `2c6c2ed3c969bc9486310a0b943a53d37df49a7e`. No blockers.

The staged scan at `packages/producer/src/services/deterministicFonts.ts:439-462` is linear. The opening and family regex cursors only advance; semicolon/closer searches cover the remainder once for a successful block, and a missing required suffix ends the scan because the old whole-region matcher could also cross intervening braces/openers—if that suffix does not exist in the remainder, no later opener can succeed. Successful matches move `opening.lastIndex` beyond the first closer, matching global-regex consumption.

The subtle empty cases are preserved correctly. An immediately empty `font-family:;` is skipped because the old `([^;]+)` could not match it and would continue to the next declaration. A whitespace-only value is not skipped: it normalizes empty and consumes the block, matching the old `\s*` backtracking one whitespace into `+`; therefore a later declaration in that same block remains ignored. Non-empty declarations preserve first-family ordering and feed the same normalized ordered `Set`, which remains the single input to the authored-face/fallback decision at `deterministicFonts.ts:1424-1435`.

The public-injection table at `deterministicFonts.test.ts:8-45` pins these recognition choices via actual fallback fetch attempts rather than a duplicate private parser.

- 90/90 deterministic-font tests passed
- 5/5 font-embedding tests passed
- ordered normalized family sets matched main across 500,010 generated/adversarial inputs
- independent 100k-input scans completed in 0.01–1.21 ms; the public long cases also completed without timeout

CI has no completed failure; JavaScript CodeQL, Windows, typecheck/build, and regression jobs are still running and remain landing gates.

**Verdict:** APPROVE
**Reasoning:** The ordered state machine preserves the old recognition/fallback semantics—including the non-obvious empty/whitespace split—while giving every successful or failed suffix a bounded scan.

— Magi

PR #267, opened 2026-04-14
feat(skills): add contrast audit + animation map quality skills

miguel-heygen:
## Summary

Two new quality skills + CLI integration that give agents feedback loops they currently lack — pixel-level contrast auditing and structured animation analysis.

### What this unlocks for agents

**Agents can now catch accessibility failures that humans and LLMs consistently miss.** The contrast audit runs automatically on every `hyperframes validate` and reports WCAG AA violations as warnings. In the eval, 4 out of 5 palettes had failing contrast — every baseline composition shipped broken, every treatment composition caught and fixed it.

**Agents can now reason about animation choreography.** The animation map produces a structured JSON report with:
- Per-tween natural language summaries ("card1 slides 23px up over 0.5s, fades in, ends at (120, 200)")
- ASCII timeline showing the full choreography as a Gantt chart
- Stagger detection with actual intervals ("3 elements stagger at 120ms" — validates against brief specs)
- Dead zone detection (periods >1s with no animation — missing entrance or intentional hold?)
- Element lifecycles (first/last animation, final visibility — catches elements that enter but never exit)
- Scene snapshots at 5 timestamps (what's on screen at any moment)

### Changes

**Skills (new)**
- \`skills/hyperframes-contrast/\` — WCAG contrast audit skill + script
- \`skills/hyperframes-animation-map/\` — animation analysis skill + script

**CLI**
- \`hyperframes validate\` now runs contrast audit by default (warnings, not errors)
- \`hyperframes validate --no-contrast\` to skip
- \`hyperframes render --html-only\` compiles HTML without video encoding
- Browser-side WCAG code in \`contrast-audit.browser.js\`, inlined at build time via esbuild text loader

**Producer**
- Exported \`compileForRender\` for the \`--html-only\` flag

### Eval results

5 prompts x 2 arms = 10 compositions. Arm A = baseline skills. Arm B = +contrast +animation-map.

| Prompt | Failing color | Before | After |
|--------|--------------|--------|-------|
| Halflife | Cement on Ink | 2.98:1 | 5.33:1 |
| Meridian | Ash on Midnight | 2.08:1 | 5.44:1 |
| Typesmith | Pencil on Paper | 3.19:1 | 5.50:1 |
| Lattice | Gray-600 on Terminal | 2.59:1 | 7.50:1 |

Animation map correctly enumerated 142 tweens across 5 compositions, detected stagger groups, flagged pacing issues, and produced scene snapshots.

### Pitch video

[...]

PR #3721, opened 2026-09-05
fix(cli): preserve concurrent scaffold config creation

jrusso1020:
CLI scaffolding can overwrite files created between an absence check and a write, or follow a dangling symlink and create its target. This affects `package.json` during init and four create-if-absent `hyperframes.json` paths in init, add, and authoring-skill seeding.

Reuse capture metadata's private staging plus hard-link publication through `writeNewFileSync`. A new `createProjectConfig` wrapper routes all four absent-only config paths through that helper. Only publication `EEXIST` is ignored; staging is cleaned on success and failure. Existing-file shortcuts, package/config bytes, and intentional updates of existing project configs are preserved. No workflow changes.

Addresses [CodeQL #267](https://github.com/heygen-com/hyperframes/security/code-scanning/267) and the matching config creation races found during review.

Validation: all 303 init/add/capture/config tests pass. The two seed-creation witnesses fail with the prior writer. Command-level cases cover existing and concurrent package/config files and dangling symlinks. Project-config cases verify an ENOENT-to-create race, dangling-link target non-creation, and the deliberate overwrite contract. Existing capture publication and authoring-skill/registry update coverage remain enabled. CLI typecheck, parser/core/lint builds, lint/format and complexity audit pass. Windows CI, CodeQL and revised-head Magi approval remain merge gates.

Publication requires hard-link support on the output filesystem and propagates unsupported-filesystem or permission errors, subject to the skill-seeding caller's existing best-effort handling.

### DIFF
--- a/packages/cli/src/capture/scaffolding.test.ts
+++ b/packages/cli/src/capture/scaffolding.test.ts
@@ -42,7 +42,7 @@ describe("generateProjectScaffold metadata", () => {
 
   afterEach(() => {
     if (fs.existsSync(dir)) {
-      expect(fs.readdirSync(dir).filter((name) => name.startsWith(".hf-meta-"))).toEqual([]);
+      expect(fs.readdirSync(dir).filter((name) => name.startsWith(".hf-create-"))).toEqual([]);
     }
     fs.rmSync(dir, { recursive: true, force: true });
   });
--- a/packages/cli/src/capture/scaffolding.ts
+++ b/packages/cli/src/capture/scaffolding.ts
@@ -5,8 +5,9 @@
  * (index.html, meta.json, AGENTS.md, CLAUDE.md).
  */
 
-import { existsSync, writeFileSync, readFileSync, mkdtempSync, linkSync, rmSync } from "node:fs";
+import { existsSync, readFileSync } from "node:fs";
 import { join, resolve } from "node:path";
+import { writeNewFileSync } from "../utils/writeNewFile.js";
 import type { CatalogedAsset } from "./assetCataloger.js";
 import type { CaptureResult, DesignTokens } from "./types.js";
 
@@ -71,23 +72,10 @@ export async function generateProjectScaffold(
   const metaPath = join(outputDir, "meta.json");
   if (!existsSync(metaPath)) {
     const hostname = new URL(url).hostname.replace(/^www\./, "");
-    const stagingDir = mkdtempSync(join(outputDir, ".hf-meta-"));
-    try {
-      const stagedPath = join(stagingDir, "meta.json");
-      writeFileSync(
-        stagedPath,
-        JSON.stringify({ id: hostname + "-video", name: tokens.title || hostname }, null, 2),
-        { encoding: "utf-8", flag: "wx" },
-      );
-      // Linking publishes without following or replacing an existing destination entry.
-      try {
-        linkSync(stagedPath, metaPath);
-      } catch (err) {
-        if (!(err instanceof Error && "code" in err && err.code === "EEXIST")) throw err;
-      }
-    } finally {
-      rmSync(stagingDir, { recursive: true, force: true });
-    }
+    writeNewFileSync(
+      metaPath,
+      JSON.stringify({ id: hostname + "-video", name: tokens.title || hostname }, null, 2),
+    );
   }
 
   // Generate AGENTS.md + CLAUDE.md (AI agent instructions — always, regardless of API keys)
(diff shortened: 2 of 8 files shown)


miguel-heygen (changes requested):
Reviewed at exact head `a056db1c1200feee0a439547ff39e00e0acbd18d`.

The extraction at `packages/cli/src/utils/writeNewFile.ts:1-18` preserves #3720’s Windows-verified protocol: private same-directory staging, complete write, atomic hard-link publication, EEXIST-only collision ownership, and finally cleanup. Both new callers create their parent directory first, and the existing package template tests retain the exact scripts/formatting contract. The 237 initialization/capture tests pass locally.

**Blocker — the same create-if-absent contract still has four unsafe `hyperframes.json` callers.** The new generic helper now exists, but these paths continue to guard the overwriting `writeProjectConfig` with an absence observation:

- `packages/cli/src/commands/init.ts:590-603`
- `packages/cli/src/commands/add.ts:278-283`
- `packages/cli/src/commands/add.ts:497-503`
- `packages/cli/src/utils/projectConfig.ts:257-263` (read returns ENOENT, then a later direct write)

They have the same concurrent-winner and dangling-symlink preconditions as package/meta creation. I reproduced the symlink case through the real writer at this head: `existsSync(hyperframes.json)` returned false for a dangling link, `writeProjectConfig` preserved the symlink but created its missing target with 318 bytes. A concurrent config can likewise be overwritten between the check/read and line 187’s plain `writeFileSync`.

Please add a create-only project-config path backed by `writeNewFileSync` and route all four absent-only callers through it, while keeping the existing overwriting writer for the two deliberate update paths documented in `projectConfig.ts`. Add one command/add or project-config witness for concurrent content and dangling-link target non-creation. This keeps overwrite and create-if-absent as separate explicit contracts instead of changing `writeProjectConfig` globally.

Exact-head CI has no completed failure; JavaScript CodeQL, Windows, build/typecheck, and remaining jobs are still running.

**Verdict:** REQUEST CHANGES
**Reasoning:** The shared publication primitive is sound, but four sibling create-if-absent config paths retain the exact overwrite/symlink race this extraction is meant to centralize.

— Magi

jrusso1020:
Addressed the config-creation scope blocker in d78b69c11. Added createProjectConfig backed by the shared staging/link helper and routed all four absent-only init/add/authoring-skill paths through it. Deliberate updates of existing configs retain their current overwrite behavior. All 303 init/add/capture/config tests pass; new skill-seeding concurrent/symlink witnesses fail with the prior writer, and command tests now cover both package.json and hyperframes.json. CLI types, lint/format, audit and signed hooks pass. Renewed review and fresh CI/CodeQL requested.

miguel-heygen (approved):
Re-review at exact head `d78b69c1104402297a1ea1471b4f17f8ada2cbde`. The project-config scope blocker is closed; no new blockers.

The contract split at `packages/cli/src/utils/projectConfig.ts:182-197` is clean: `writeProjectConfig` remains the explicit overwrite API, while `createProjectConfig` is the absent-only API backed by the single staging/link implementation. A production caller inventory shows exactly four create-only calls—init, both add branches, and skill seeding’s ENOENT branch—and no absent-only site still reaches the overwriting writer.

`seedProjectAuthoringSkill` retains its best-effort boundary at `projectConfig.ts:255-276`: non-ENOENT read errors leave existing bytes alone; ENOENT attempts exclusive creation; a concurrent winner or dangling symlink yields the helper’s harmless EEXIST; unsupported/permission failures remain swallowed only because this caller was already explicitly best-effort. Existing-config patch writers remain separate, so unknown-key and indentation preservation are unchanged.

The new tests distinguish all three contracts rather than only checking “a file exists”:
- init command: existing/concurrent/symlink entries for both package and project config;
- project config: create-no-overwrite, deliberate overwrite, seed-after-ENOENT race, and dangling-link target non-creation;
- capture: all eight landed publication/failure-cleanup witnesses remain enabled.

Independent exact-head verification: 303/303 init/add/capture/config tests and formatting on all eight changed files passed.

**Non-blocking note:** both add paths load `config` before their absence check. If a custom config wins precisely after that read, disk now correctly preserves it but that one invocation continues with its already-loaded default. This is pre-existing transient behavior and strictly safer than overwriting the winner; it is not part of the create-publication defect.

Fresh CI has no current-head completed failure; JavaScript CodeQL, Windows, build/typecheck, and remaining jobs are still running. Older cancelled Windows runs appear as failures in the aggregate check list but are not from this head’s live run.

**Verdict:** APPROVE
**Reasoning:** Every create-if-absent caller now shares the Windows-verified publication primitive, while deliberate updates remain an explicit, separately tested overwrite contract.

— Magi

PR #269, opened 2026-04-14
feat(registry): seed blocks batch — social overlays, data viz, showcases

jrusso1020:
## What

Add 11 blocks from the Hyperframe Template Structure catalog, bringing the registry to 25 total items.

### Social overlays
| Block | Dimensions | Duration | Description |
|-------|-----------|----------|-------------|
| `instagram-follow` | 1080×1920 | 4.5s | Instagram follow overlay with profile card |
| `tiktok-follow` | 1080×1920 | 4.5s | TikTok follow overlay with profile card |
| `yt-lower-third` | 1920×1080 | 4.5s | YouTube subscribe lower third |
| `x-post` | 1920×1080 | 5s | X/Twitter post card with engagement |
| `reddit-post` | 1920×1080 | 5s | Reddit post card with upvotes |
| `spotify-card` | 1080×1920 | 5s | Spotify now-playing card |
| `macos-notification` | 1920×1080 | 5s | macOS notification banner |

### Data & visualization
| Block | Duration | Description |
|-------|----------|-------------|
| `ascii-dashboard` | 10s | Retro terminal-style data viz |
| `ascii-lightning` | 9s | ASCII art lightning bolt animation |

### Showcases
| Block | Duration | Description |
|-------|----------|-------------|
| `app-showcase` | 5.5s | Floating smartphone screens |
| `ui-3d-reveal` | 13s | Perspective 3D UI reveal |

## Why

Phase D content accumulation. The registry pipeline (PRs 6-10) is in place — this PR exercises it at scale.

## How

- Extracted from zip files in the Hyperframe Template Structure Notion doc
- Social overlays: single-file standalone HTML, copied directly
- Multi-file blocks (ascii-*, app-showcase, ui-3d-reveal): converted `<template>` sub-compositions to standalone HTML with proper `<!doctype>` wrappers
- All previews (PNG + MP4) rendered locally via `generate-catalog-previews.ts`
- Catalog MDX pages regenerated via `generate-catalog-pages.ts`
- `docs.json` updated with new catalog entries

## Test plan

- [x] All 11 blocks render to PNG + MP4 without errors
- [x] Catalog pages generated for all 17 items (14 blocks + 3 components)
- [x] `registry-item.json` files have correct dimensions, duration, tags
- [x] `oxfmt --check` passes on all files

PR #3722, opened 2026-09-05
fix(cli): publish cached synthesis script exclusively

jrusso1020:
The TTS cache checks for `synth-v3.py` before writing it. A concurrent script can be truncated, and a dangling symlink can cause its missing target to be created.

Publish the cached script with the existing `writeNewFileSync` helper. This reuses private staging and exclusive hard-link publication while preserving the existing-cache shortcut, script body, version cleanup, and synthesis arguments. No workflow changes.

Addresses [CodeQL #269](https://github.com/heygen-com/hyperframes/security/code-scanning/269).

Validation: all 35 TTS and shared-publication regression tests pass. Three new public-synthesis cache cases cover creation/reuse/old-version cleanup, a concurrent winner, and dangling-symlink target non-creation. The two security witnesses fail on main. These tests use real temporary files and stub Python execution/model acquisition; they do not run Kokoro or download models. CLI typecheck, parser/core/lint builds, lint/format and complexity audit pass. Windows CI, CodeQL and Magi approval remain merge gates.

As with the helper's other callers, the cache filesystem must support hard links; publication errors propagate instead of overwriting a destination entry.

### DIFF
--- a/packages/cli/src/tts/synthesize.cache.test.ts
+++ b/packages/cli/src/tts/synthesize.cache.test.ts
@@ -1,0 +1,106 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import * as fs from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { execFileSync } from "node:child_process";
+
+const paths = vi.hoisted(() => ({ home: "" }));
+vi.mock("node:os", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("node:os")>()),
+  homedir: () => paths.home,
+}));
+vi.mock("node:fs", async (importOriginal) => {
+  const original = await importOriginal<typeof fs>();
+  return { ...original, existsSync: vi.fn(original.existsSync) };
+});
+vi.mock("node:child_process", () => ({ execFileSync: vi.fn() }));
+vi.mock("./python.js", () => ({
+  findPython: () => "test-python",
+  hasPythonPackage: () => true,
+}));
+vi.mock("./manager.js", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("./manager.js")>()),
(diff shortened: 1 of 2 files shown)


miguel-heygen (approved):
Reviewed at exact head `835a7dad6e555b6e7e34367c88c615c633e03fd6`. No blockers.

The cache protocol at `packages/cli/src/tts/synthesize.ts:95-122` is compatible and correctly ordered:

- an existing `synth-v3.py` still returns immediately, without directory creation or old-version cleanup;
- a cache miss creates the parent, then delegates the only script publication to the Windows-verified `writeNewFileSync` boundary;
- successful publication or a link-time EEXIST winner proceeds to best-effort removal of only older `synth(-vN)?.py` entries;
- unsupported hard links/write/publication errors propagate before cleanup or Python execution;
- helper staging cleanup remains guaranteed on every post-mkdtemp exit.

The inline Python literal is byte-unchanged, and `ensureSynthScript` still returns the same `synth-v3.py` path passed to the unchanged model/voices/text/voice/speed/output/lang argv at `synthesize.ts:189-205`. A repository scan finds no sibling synth-script writer.

The public cache tests at `synthesize.cache.test.ts:55-105` use real filesystem entries and pin creation bytes, cache reuse, selective old-version cleanup, concurrent winner preservation, dangling-link target non-creation, no staging residue, and the final Python argv while model acquisition/execution remain stubbed.

Independent exact-head verification: 35/35 TTS/shared-publication tests and changed-file formatting passed. The unchanged Windows `studio-engine-cli` lane runs the full CLI suite and is still in progress; it remains the landing gate for cross-platform hard-link behavior.

CI has no completed failure; JavaScript CodeQL, Windows, build/typecheck, and remaining jobs are still running.

**Verdict:** APPROVE
**Reasoning:** The sole synthesis-script creation path now reuses the proven atomic publication boundary without changing cache hits, script bytes, version cleanup, or execution arguments.

— Magi

PR #3723, opened 2026-09-06
fix(lint): warn on undefined GSAP color variables

miguel-heygen:
## Summary

Static lint now warns when a GSAP color-like tween uses `var(--name)` without a fallback and no definition can be found in authored CSS or `data-composition-variables`. This points authors to the source of text that can become invalid or transparent before the existing browser check reports only the generic `text_not_painted` symptom.

The warning checks both `fromTo` endpoints and color, `*Color`, fill, and stroke properties. It remains silent for style-block and inline custom-property definitions, `var()` fallbacks, and composition variables that runtime exposes as `--<id>`.

The reported Whisper behavior is intentionally excluded. The packed v0.8.29 CLI already exits 1 when the required engine is unavailable and 0 only with `--optional`; changing that contract would regress working behavior.

## Validation

- Full GSAP lint suite: 171/171 tests passed.
- Composition-variable controls: 140/140 tests passed.
- Lint package typecheck passed.

### DIFF
--- a/packages/lint/src/rules/composition.ts
+++ b/packages/lint/src/rules/composition.ts
@@ -222,7 +222,7 @@ function collectDeclaredVariableIds(htmlTagRaw: string): Set<string> | null {
  * template/fragment sub-comps hold it on their composition root div. Returns
  * null if any occurrence has unparseable JSON.
  */
-function collectAllDeclaredVariableIds(tags: readonly OpenTag[]): Set<string> | null {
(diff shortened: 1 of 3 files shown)


miga-heygen (approved):
Reviewed: `gsap_undefined_css_variable` correctly checks color properties for var() references without fallback or static definition. Respects composition-variable declarations and inline styles. Tests cover all paths. — Miga

PR #265, opened 2026-04-14
feat(engine): add HDR video output pipeline

vanceingalls:
## Summary

Adds the ability to render HDR video output (H.265 10-bit, BT.2020) from HyperFrames compositions. When the renderer detects HDR source video, it automatically switches to the HDR output pipeline — no flags needed.

## What it does

- **Auto-detection** — Probes each video source with `ffprobe`. If any has bt2020/PQ/HLG color metadata, the output switches to H.265 10-bit with correct color tags. SDR-only compositions are unaffected (H.264, bt709).
- **HLG pass-through** — Native HLG pixels from FFmpeg extraction are piped directly to the encoder without conversion. This avoids brightness loss from HLG→linear→PQ conversion (which requires an OOTF system gamma we can't reliably apply).
- **Encoder HDR support** — Both chunk and streaming encoders accept HDR presets: `libx265`, `yuv420p10le`, BT.2020 color primaries, `hvc1` codec tag (required for Apple playback).
- **WebGPU HDR capture (gated)** — A complete WebGPU float16 readback pipeline is implemented and tested but gated behind headed Chrome (headless doesn't expose WebGPU). Ready for future use with WebGPU canvas content.
- **HDR utilities** — `detectTransfer()` (PQ vs HLG), `getHdrEncoderColorParams()`, `analyzeCompositionHdr()`. 15 unit tests.

## Key design decisions

| Decision | Why |
|----------|-----|
| No `--hdr` flag | SDR content encoded as HDR causes orange shift in browsers. Auto-detect eliminates this. |
| HLG pass-through (not HLG→PQ) | Conversion loses brightness without OOTF. Pass-through matches source exactly. |
| `hvc1` codec tag | Apple QuickTime requires `hvc1` (not `hev1`) for HEVC playback. |
| 1-hour streaming timeout | HDR capture at ~6fps needs more time than the default 10-minute FFmpeg timeout. |

## Files changed

| File | What changed |
|------|-------------|
| `packages/engine/src/utils/hdr.ts` | **NEW** — HDR detection, transfer types, encoder params (15 tests) |
| `packages/engine/src/services/hdrCapture.ts` | **NEW** — WebGPU readback, HLG conversion, PQ encode |
| `packages/engine/src/services/streamingEncoder.ts` | HDR presets, raw rgb48le input, color tags |
| `packages/engine/src/services/chunkEncoder.ts` | HDR presets, conditional color tags |
[...]

PR #3724, opened 2026-09-06
fix(cli): read caption images through checked file descriptors

jrusso1020:
Capture captioning checked an image's size by pathname, then reopened that pathname to read it. A replacement between those operations could send different bytes to the vision provider. Open once in read-only mode, check size and read through the same descriptor, and close it before the provider request (including oversized files and failures).

Preserves the 4,000,000-byte inclusive limit, early oversized-image skip, provider payloads, batching, and failed-asset warnings. Addresses [CodeQL alert #265](https://github.com/heygen-com/hyperframes/security/code-scanning/265).

Validation: all 214 capture tests, CLI typecheck, parser/core/lint builds, lint/format and signed commit hooks passed. New real-filesystem tests cover replacement after the size check (reproduces on main), both size-limit boundaries, and descriptor cleanup on stat/read failures. GitHub Actions and Windows verification pending.

### DIFF
--- a/packages/cli/src/capture/contentExtractor.file-race.test.ts
+++ b/packages/cli/src/capture/contentExtractor.file-race.test.ts
@@ -1,0 +1,106 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import * as fs from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { captionImagesWithGemini } from "./contentExtractor.js";
+
+const hooks = vi.hoisted(() => ({
+  afterStat: () => {},
+  beforeRead: () => {},
+  descriptor: -1,
+}));
+
+vi.mock("node:fs", async (importOriginal) => {
+  const actual = await importOriginal<typeof fs>();
+  return {
+    ...actual,
+    statSync: (path: fs.PathLike) => {
+      const stat = actual.statSync(path);
+      hooks.afterStat();
+      return stat;
+    },
+    fstatSync: (fd: number) => {
+      hooks.descriptor = fd;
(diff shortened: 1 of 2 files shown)


miguel-heygen (approved):
The pathname race is closed cleanly. `contentExtractor.ts:457-465` opens once in read-only mode, checks and reads the same descriptor, and the `finally` closes it on the oversize return and on stat/read failures before `captionOne` can issue a provider request. The surrounding `Promise.allSettled` path preserves batching, sibling success, and the existing sanitized warning count.

The regression coverage is strong: `contentExtractor.file-race.test.ts:72-103` distinguishes path replacement from descriptor identity, pins both sides of the inclusive 4,000,000-byte boundary, and proves closure after both failure sites. I ran all 214 capture tests plus changed-file lint/format locally; all pass. CI Build and Typecheck are green at this head; Windows, the full Test job, CLI smoke, and JavaScript CodeQL are still running and remain landing gates.

**Non-blocking residual:** `contentExtractor.ts:460-462` pins the inode, not an immutable byte snapshot. A writer that grows the same inode after `fstatSync` can still make `readFileSync(fd)` return more than 4,000,000 bytes (I reproduced 4 checked bytes → 4,000,001 read bytes). That does not reopen the reported pathname-substitution bug or invalidate the CodeQL-recommended descriptor fix. If the provider limit must remain hard even against concurrent in-place writers, cap the descriptor read or check `buffer.length` before encoding in a follow-up.

**Verdict:** APPROVE
**Reasoning:** The exact head fixes #265 with one descriptor and correct cleanup while preserving stable-file behavior; the remaining same-inode mutation caveat is pre-existing and outside the reported replacement race.

— Magi

PR #3726, opened 2026-09-06
fix(cli): fetch a page's assets as the same agent that loaded the page

miguel-heygen:
## What

A capture now fetches a page's assets as the same user agent that loaded the page. `CAPTURE_USER_AGENT` is one constant, used by both halves of a capture.

Before this, a capture presented three different identities to the same origin:

| Where | User-Agent |
|---|---|
| `capture/index.ts` page navigation | the real Chrome UA |
| `assetDownloader.ts` (favicons, images, og:image, fonts) | `HyperFrames/1.0` |
| `mediaCapture.ts` (Lottie JSON, videos), `commands/capture/video.ts` | `HyperFrames/1.0` |
| `htmlExtractor.ts` (stylesheet inlining) | bare `Mozilla/5.0` |

## Why

A capture is one session with two halves: Chrome navigates the page, then Node fetches the assets that page referenced. When those halves send different `User-Agent` headers, the origin is free to answer them differently, and anti-bot edges do exactly that.

Measured against openai.com, whose capture produced a visibly wrong brand icon:

```
User-Agent: HyperFrames/1.0   GET /favicon.svg  -> 403 text/html
browser User-Agent            GET /favicon.svg  -> 200 image/svg+xml (2466 bytes)
```

openai.com's other two declared icons (`.ico`, apple-touch `.png`) are served to either agent.

The consequence is not a failed capture, it is a quietly wrong one. `rankIconCandidates` had already chosen that SVG as the best declared icon. The 403 discarded it, `downloadAssets` fell through to the next candidate, and the file written as the site's favicon was decided by the CDN's bot rules rather than by the ranker. The capture recorded it as a single `unavailable` drop and carried on, so nothing downstream could tell that the icon on disk was the runner-up.

This is why it is worth fixing at one owner rather than at the favicon call site: every out-of-band fetch in a capture had the same exposure, including fonts and stylesheets, where a substituted or missing response degrades the capture without failing it.

## How

- New `packages/cli/src/capture/userAgent.ts` exporting `CAPTURE_USER_AGENT`, with the measurement above recorded as the reason it exists.
- `page.setUserAgent()` and all five out-of-band fetch sites now read that constant. No other behaviour changes; the SSRF denylist, redirect re-validation, content-type guards, budgets and caps are untouched.

## Test plan

Two tests in `assetDownloader.test.ts` drive `downloadAssets` through a stubbed origin that reproduces the real behaviour (403 to a non-browser agent on the SVG, 200 to everyone on the rest):

- **lands the icon the ranker chose** — proven non-vacuous by setting the constant back to `HyperFrames/1.0`, which fails with exactly the production symptom:
  ```
  AssertionError: expected [ 'assets/favicon.png' ] to deeply equal [ 'assets/favicon.svg' ]
  ```
- **sends one User-Agent for every asset request** — guards against a second identity reappearing.

Real output:

```
$ npx vitest run src/capture --poolOptions.forks.maxForks=4
 Test Files  3 failed | 18 passed (21)
      Tests  187 passed (187)
```

The 3 failures are collection errors (`Failed to resolve entry for package "@hyperframes/parsers"`) in a fresh worktree with that workspace package unbuilt. None of the three imports any file in this diff, and no test assertion fails.

`tsc --noEmit` is clean for `packages/cli`, and the pre-commit hook's monorepo typecheck, lint, format and fallow gates all pass.

End-to-end, running the real `hyperframes capture` against https://openai.com:

```
before:  assets/favicon.png   6578 bytes   (its apple-touch icon)   Dropped: 1 unavailable
after:   assets/favicon.svg   2466 bytes                            no unavailable drops
```

The captured file's sha256 matches the file the site itself serves for `/favicon.svg`.

### DIFF
--- a/packages/cli/src/capture/assetDownloader.test.ts
+++ b/packages/cli/src/capture/assetDownloader.test.ts
@@ -10,6 +10,8 @@ import {
   toStandaloneSvg,
 } from "./assetDownloader.js";
 import type { DesignTokens } from "./types.js";
+import type { IconCandidate } from "./faviconRanker.js";
+import { CAPTURE_USER_AGENT } from "./userAgent.js";
 
 describe("isPrivateUrl — SSRF denylist (security: F-003)", () => {
   it("blocks loopback, private, and metadata IPv4", () => {
@@ -187,6 +189,16 @@ describe("downloadAndRewriteFonts — attempt caps", () => {
   });
 });
 
+function withTempDir<T>(run: (dir: string) => Promise<T>): Promise<T> {
+  const dir = mkdtempSync(join(tmpdir(), "hf-drops-"));
+  return run(dir).finally(() => rmSync(dir, { recursive: true, force: true }));
+}
+
+/** The two fields `downloadAssets` reads off the token bundle. */
+function tokensWithNoSvgs(): DesignTokens {
+  return { svgs: [], sections: [], ogImage: "" } as unknown as DesignTokens;
+}
+
 describe("drop counts — why a referenced asset is not in the capture", () => {
   afterEach(() => vi.unstubAllGlobals());
 
@@ -197,11 +209,6 @@ describe("drop counts — why a referenced asset is not in the capture", () => {
       (_, i) =>
         `@font-face { font-family: Family${i}; src: url(https://fonts${i}.example/font-${i}.woff2); }`,
     ).join("\n");
-  }
-
-  function withTempDir<T>(run: (dir: string) => Promise<T>): Promise<T> {
-    const dir = mkdtempSync(join(tmpdir(), "hf-drops-"));
-    return run(dir).finally(() => rmSync(dir, { recursive: true, force: true }));
   }
 
   it("counts every face the budget never let it reach, not just the one it stopped on", async () => {
@@ -269,11 +276,6 @@ describe("drop counts — why a referenced asset is not in the capture", () => {
     });
   });
 
-  /** The two fields `downloadAssets` reads off the token bundle. */
-  function tokensWithNoSvgs(): DesignTokens {
-    return { svgs: [], sections: [], ogImage: "" } as unknown as DesignTokens;
-  }
-
   it("counts an image dropped for being under the raster floor", async () => {
     // 9 KB is under the 10 KB floor. Nothing lands, and the reason is now on the record.
     await withTempDir(async (dir) => {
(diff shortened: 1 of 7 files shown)


jerrai-bot-heygen (approved):
Approved at 009b5b1794.

Verified the claim end to end: every out-of-band fetch site in `packages/cli/src` plus `page.setUserAgent` now goes through `CAPTURE_USER_AGENT`, with no residual hard-coded UA (the OpenRouter call in contentExtractor is API traffic, not origin traffic, so it correctly stays as is). Tests at head: 33/33. Discriminating probe: reverting the UA to `HyperFrames/1.0` at all five sites fails exactly the two new tests with the production symptom, so they are not vacuous.

One pre-existing gap, not blocking: `assetCataloger.ts:279` `readAssetBytes` (the GIF-animation metadata probe inside the capture flow) still sends no User-Agent at all — a fourth identity. Failure is benign (probe 403 → the GIF annotation is silently skipped), but it is the same class this PR closes; worth folding into the UA constant when convenient.

— Jerrai

jrusso1020 (approved):
**APPROVE** at `009b5b17`.

One constant, five fetch sites and the navigation, and the argument for doing it at one owner
rather than at the favicon call site is the right one. I read the full files rather than the diff
and went after three things: whether every site actually moved, whether any of them faces HeyGen
rather than the captured origin, and whether the guard covers what the PR says it guards.

## Verified

**"All five out-of-band fetch sites" is exact, and nothing is left behind.** At this head there is
no `User-Agent` literal anywhere in `packages/cli` — every one reads the constant:
`assetDownloader.ts:507`, `mediaCapture.ts:62` and `:282`, `commands/capture/video.ts:16`,
`htmlExtractor.ts:39`, plus `index.ts:170` for the navigation. Five is the right count because
`mediaCapture` carries two of them, which is the sort of thing a per-file sweep misses.

**No HeyGen-facing identity moves.** That was my main worry going in — a UA that our own edge or
attribution keys on. It does not arise: all five sites fetch third-party asset URLs off the
captured page, and `streamToFile` is a manifest-vetted video URL, not an API call. The CLI's own
HeyGen requests set no `User-Agent` at all and are untouched here.

**The guards really are untouched.** Each site still goes through `safeFetch` (except the
stylesheet one, which still pre-checks `isPrivateUrl` as before), so the denylist, the
per-redirect-hop revalidation, the content-type sniff-guard, the budgets and the caps are all the
same code. The only thing that changed at each call site is one header value.

**The `HyperFrames/1.0` identity is not otherwise load-bearing.** Control from here:
`GET github.com/favicon.ico` answers `200 image/x-icon` to `HyperFrames/1.0` and to the Chrome
string alike. An ordinary origin does not care, which is what bounds the blast radius of the swap.

**What I could not reproduce, and why it is not a contradiction.** I tried to measure the flip
myself. From this host `GET openai.com/favicon.svg` returns `403 text/html` to `HyperFrames/1.0`
**and** to the exact `CAPTURE_USER_AGENT` string, with or without browser-shaped
`Accept`/`Sec-Fetch-*`/`Referer` headers. The control that settles it: openai.com's **homepage**
also 403s here with the Chrome UA. This devbox is blanket-challenged by that edge, so it cannot
discriminate the UA variable at all — my numbers say nothing about yours. Taking the 403 → 200 as
reported.

## Three notes, none blocking

**1. The regression guard covers one of the five sites.** "Sends one User-Agent for every asset
request" drives `downloadAssets`, so it pins `fetchBuffer` in `assetDownloader.ts` and nothing
else. A literal reappearing in `mediaCapture.ts`, `video.ts` or `htmlExtractor.ts` is exactly the
regression the constant exists to prevent, and it would go through green. The cheap version is the
sweep I just ran by hand as an assertion: no file under `src/capture` carries a `User-Agent` header
value that is not the imported constant. Source-level, no origin needed, covers all five and any
sixth added later.

**2. `htmlExtractor` is the one site where this is more than a rename.** The other four went from
`HyperFrames/1.0` to a browser string. That one went from a bare `Mozilla/5.0` to a full Chrome UA,
and a CDN is entitled to answer a stylesheet request differently for the two — that is the same
mechanism the PR is built on, pointed at CSS. The direction is right (a page's own stylesheet
should be fetched as the agent that loaded the page), but "No other behaviour changes" covers this
one a little more broadly than the diff supports. Worth a clause rather than a change.

**3. There is prior art in-repo, and it also marks the boundary.**
`packages/producer/src/services/deterministicFonts.ts:777` has `WOFF2_USER_AGENT` under the comment
"Chrome UA triggers woff2 responses from Google Fonts CSS API" — the same lesson, already paid for
once, that the UA selects the response body and not just the caller's name. Both strings are Chrome
today, so someone will eventually be tempted to fold that one into `CAPTURE_USER_AGENT`. It should
stay separate: that constant is chosen for what it makes the *server* return, this one for who the
*session* is, and coupling them makes a font-format regression a possible outcome of a capture
change. A sentence in `userAgent.ts` naming that would keep the next person out of it.

Small aside on the same file: the Chrome major (`131`) is now pinned in exactly one place, which is
an improvement, but nothing in the doc comment says it is a version that ages. Some edges score UA
freshness. One line about when to bump it costs nothing now and saves the archaeology later.

## CI

Read at post time on `009b5b17`, not earlier in the draft. `main` here has no classic branch
protection at all — `branches/main/protection` is a 404 — and every gate comes from ruleset
`14211637`, which requires eight contexts. All eight are green, taking the later run per name (the
`cancelled` entries are the superseded first wave on this same SHA): `Semantic PR title` 01:30:57Z,
`regression` 01:32:01Z, `Test: runtime contract` 01:32:46Z, `Build` 01:33:43Z, `Typecheck`
01:34:26Z, `Render on windows-latest` 01:37:37Z, `Test` 01:41:18Z, `Tests on windows-latest`
01:43:02Z. One thing not to be misled by: `commits/<sha>/status` reports `pending` here, but its
`statuses` array is empty — that is the combined-status default for a commit with no legacy
statuses, not a check waiting. The verdict does not rest on any of this.

-- Rames

jrusso1020:
Addendum, and it corrects a sentence of mine above. **The APPROVE stands** — everything below is
pre-existing and on a file this PR does not touch.

I wrote "all five out-of-band fetch sites is exact, and nothing is left behind." The first half is
right; the second is not. My sweep looked for `User-Agent` **literals**, so it could only ever find
call sites that set one — and it therefore could not see a call site that sets no headers at all.
There is one.

**`assetCataloger.ts:277-287`, `readAssetBytes`**, reached from `annotateGifAssetMetadata` for
every catalogued asset whose URL ends in `.gif`:

```ts
const response = await fetch(url, { signal: AbortSignal.timeout(15_000) });
```

That is a sixth out-of-band fetch of a captured-origin asset, and it presents a third identity —
no `User-Agent` at all. Against the anti-bot edge this PR is written about, a header-less request
is in the same bucket as `HyperFrames/1.0`, so the "one session, one identity" property does not
actually hold across the capture yet. Worth knowing before the release, not worth holding this.

**The larger half is that it is also the one asset fetch that does not go through `safeFetch`.**
`assetCataloger.ts` imports neither `safeFetch` nor `isPrivateUrl`, and this is a bare `fetch` with
the default `redirect: "follow"`. The URL comes from the captured page's own DOM. So for this one
path the SSRF denylist and the per-hop redirect revalidation that every other asset fetch in the
capture is careful about are not in front of it, and a public URL is free to redirect somewhere
that is not. Exposure is narrow — the bytes are parsed for GIF metadata and dropped, never written
to disk, and only a note like "single-frame GIF" survives — but "narrow" is a property of today's
call site rather than of the guard, and the guard is the thing every sibling call has.

The fix is the two imports this PR already makes idiomatic:

```ts
import { safeFetch } from "./assetDownloader.js";
import { CAPTURE_USER_AGENT } from "./userAgent.js";

const response = await safeFetch(url, {
  signal: AbortSignal.timeout(15_000),
  headers: { "User-Agent": CAPTURE_USER_AGENT },
});
if (!response) return null;
```

The obvious objection is a cycle, and there is not one: `assetDownloader.ts` takes `CatalogedAsset`
from this file with `import type`, which is erased, so nothing imports it back at runtime.

This also sharpens note 1 in my review. A guard that asserts "every `User-Agent` under `src/capture`
is the constant" would still have passed over this line. The assertion worth having is the one over
call sites rather than over header values: every outbound fetch to a captured-origin URL goes
through `safeFetch` and carries `CAPTURE_USER_AGENT`. That version catches both this and the
literal reappearing.

Credit where due: another reviewer flagged the missing UA on this path in parallel (attributing it
to `assetDownloader`); I went to confirm it, found it one file over in `assetCataloger`, and the
missing `safeFetch` came out of that.

-- Rames

PR #3727, opened 2026-09-06
feat(cli): keep every icon a site declares and headline the bare mark

miguel-heygen:
> Was stacked on #3726, which has merged. Rebased onto `main`; this diff is now only its own change.

## What

A capture now keeps **every** icon a site declares, records what each one *is*, and explains which one became the headline.

- Each declared icon is written as `assets/icon-<rel>-<sizes>.<ext>`.
- Each is classified `bare-mark` | `badge` | `unknown`.
- The best is copied onto the historical `assets/favicon.<ext>` stem, so an existing stem match still finds it.
- `extracted/icons-manifest.json` records all of it, including *why* the headline won.

## Why

A brand band shows logo artwork on a transparency checker. A **badge** — a mark knocked out of a full-bleed disc or square — reads there as a solid blob, which is the complaint that started this. A **bare mark** is the glyph alone with transparent margins.

Previously the downloader stopped at the first icon that downloaded, so which of a site's icons survived was decided by declared-quality ranking alone, and nothing recorded what the survivor looked like. Shape can only be read from bytes, and a loop that stops early never fetches the candidate it skipped — so it cannot know whether that one was the bare mark.

## How

**Classification is deterministic and cheap.** Two rules, both in `iconClassifier.ts`:

*SVG is read from markup, not rendered.* Rendering is the obvious approach and it does not work: these files routinely fill via CSS custom properties behind a `prefers-color-scheme` query, which librsvg does not resolve. One real favicon rasterises through sharp to a **fully transparent** image while Chrome draws it correctly. The rule walks to the first painted element, skipping `defs`/`clipPath`/`mask`, and calls a full-bleed `rect` or `circle` a badge.

*Raster is sampled at the midpoint of each edge, not the corners.* This is the trap worth naming: a disc badge on a transparent canvas has four **transparent corners** and would read as a bare mark. Its edge midpoints are opaque, because a full-bleed disc touches the middle of every side. There is a test for exactly this case.

**Headline preference is binary on purpose:** a positively identified bare mark first, then the existing ranking for everything else. Ranking `unknown` *between* bare-mark and badge looked reasonable and is wrong — a `.ico` cannot be decoded (libvips ships no decoder), so on a site whose icons are all badges the undecodable legacy file would outrank the good SVG purely for being unexaminable. Absence of evidence is not evidence of a bare mark.

**The manifest records the losers.** A choice whose alternatives are invisible is indistinguishable from having had no choice — the failure mode that let a substituted icon go unnoticed in the first place.

### Manifest shape

`extracted/icons-manifest.json`, schema `hyperframes.capture.icons.v1`:

```jsonc
{
  "schema": "hyperframes.capture.icons.v1",
  "headline": {
    "file":   "assets/favicon.svg",              // the backwards-compatible stem
    "source": "assets/icon-icon-unsized.svg",    // the icons[] entry it was copied from
    "rank":   0,                                  // 0 = best by declared quality
    "shape":  "badge",
    "reason": "no bare-mark candidate among 3 icon(s); used the best-ranked badge"
  },
  "icons": [
    { "file": "...", "url": "...", "rel": "icon", "sizes": null, "type": "image/svg+xml",
      "rank": 0, "shape": "badge", "shapeReason": "first painted element is a full-bleed rounded rect (180x180 of 180x180)" }
  ]
}
```

`headline.source` lets a consumer avoid rendering the same image twice, since `headline.file` is a byte copy of it.

## A finding consumers should know

**Sites frequently declare no bare mark at all.** An `apple-touch-icon` is composited onto an opaque tile per Apple's spec, so it is a badge, not a mark — verified on a real site whose apple-touch icon is a fully opaque white square (alpha 255 at all four corners and edge midpoints). On that site all three declared icons are badges or undecodable, the headline falls back to the ranker, and the manifest says so rather than implying the preference was satisfied.

## Test plan

`iconClassifier.test.ts` (10 tests) covers both rules, including the disc-badge case corner sampling would miss and the `defs`-only case that reading elements in document order would misjudge.

`assetDownloader.test.ts` covers the pipeline: one file per declared icon named for its rel and sizes, the all-badges site recording the fallback reason, the bare-mark-only site, and the `assets/favicon.<ext>` stem surviving.

The ordering test is proven non-vacuous — flattening the shape preference back to pure ranker order fails with:
```
AssertionError: expected { file: 'assets/favicon.svg', …(4) } to match object { file: 'assets/favicon.png', …(3) }
```

Real output:
```
$ npx vitest run src/capture --poolOptions.forks.maxForks=4
 Test Files  3 failed | 19 passed (22)
      Tests  202 passed (202)
```
The 3 failures are collection errors (`Failed to resolve entry for package "@hyperframes/parsers"`) in a fresh worktree with that workspace package unbuilt; none imports a file in this diff and no assertion fails. `tsc --noEmit` is clean for `packages/cli`, and the pre-commit typecheck, lint, format and fallow gates all pass (no new complexity or dead-export findings).

Live capture of a real site:
```
assets/favicon.svg                        <- headline
assets/icon-icon-unsized.svg              badge    full-bleed rounded rect (180x180 of 180x180)
assets/icon-apple-touch-icon-180x180.png  badge    4 of 4 edge midpoints opaque
assets/icon-icon-48x48.ico                unknown  format could not be decoded
```

### DIFF
--- a/packages/cli/src/capture/assetDownloader.test.ts
+++ b/packages/cli/src/capture/assetDownloader.test.ts
@@ -12,6 +12,7 @@ import {
 import type { DesignTokens } from "./types.js";
 import type { IconCandidate } from "./faviconRanker.js";
 import { CAPTURE_USER_AGENT } from "./userAgent.js";
+import sharp from "sharp";
 
 describe("isPrivateUrl — SSRF denylist (security: F-003)", () => {
   it("blocks loopback, private, and metadata IPv4", () => {
@@ -199,6 +200,31 @@ function tokensWithNoSvgs(): DesignTokens {
   return { svgs: [], sections: [], ogImage: "" } as unknown as DesignTokens;
 }
 
+/**
+ * openai.com's declared icons, in DOM order, with the attributes the page really carries.
+ * `rankIconCandidates` puts `favicon.svg` first, so that is the file a capture must land.
+ */
+const OPENAI_ICONS: IconCandidate[] = [
+  {
+    rel: "icon",
+    href: "https://openai.example/favicon.svg",
+    sizes: null,
+    type: "image/svg+xml",
+  },
+  {
+    rel: "icon",
+    href: "https://openai.example/favicon.ico",
+    sizes: "48x48",
+    type: "image/x-icon",
+  },
+  {
+    rel: "apple-touch-icon",
+    href: "https://openai.example/apple-icon.png",
+    sizes: "180x180",
+    type: "image/png",
+  },
+];
+
 describe("drop counts — why a referenced asset is not in the capture", () => {
   afterEach(() => vi.unstubAllGlobals());
 
@@ -335,31 +361,6 @@ describe("asset fetches present the same identity as the page navigation", () =>
   afterEach(() => vi.unstubAllGlobals());
 
   /**
-   * openai.com's declared icons, in DOM order, with the attributes the page really carries.
-   * `rankIconCandidates` puts `favicon.svg` first, so that is the file a capture must land.
-   */
-  const OPENAI_ICONS: IconCandidate[] = [
-    {
-      rel: "icon",
-      href: "https://openai.example/favicon.svg",
-      sizes: null,
-      type: "image/svg+xml",
-    },
-    {
-      rel: "icon",
-      href: "https://openai.example/favicon.ico",
-      sizes: "48x48",
-      type: "image/x-icon",
-    },
-    {
-      rel: "apple-touch-icon",
-      href: "https://openai.example/apple-icon.png",
-      sizes: "180x180",
-      type: "image/png",
-    },
-  ];
-
-  /**
    * Measured against the real origin: `GET /favicon.svg` answers `403 text/html` to
    * `User-Agent: HyperFrames/1.0` and `200 image/svg+xml` to the browser UA the capture
    * already navigates with. The other two icons are served to either agent.
@@ -383,9 +384,9 @@ describe("asset fetches present the same identity as the page navigation", () =>
     // next candidate, so the file on disk is chosen by the CDN rather than by the ranker.
     await withTempDir(async (dir) => {
       vi.stubGlobal("fetch", serveLikeAnAntiBotEdge());
-      const { assets } = await downloadAssets(tokensWithNoSvgs(), dir, [], OPENAI_ICONS);
-      expect(assets.map((a) => a.localPath)).toEqual(["assets/favicon.svg"]);
-      expect(assets[0]?.url).toBe("https://openai.example/favicon.svg");
+      const { icons } = await downloadAssets(tokensWithNoSvgs(), dir, [], OPENAI_ICONS);
+      expect(icons.headline?.file).toBe("assets/favicon.svg");
+      expect(icons.icons[0]?.url).toBe("https://openai.example/favicon.svg");
     });
   });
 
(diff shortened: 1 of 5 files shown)


jerrai-bot-heygen (dismissed):
Approved at f60c805944, with one cross-repo follow-up that needs an EF-side change before the worker pin bump lands, and one hardening suggestion.

Verified by running: capture suite 202/202 at head (three files fail to *collect* for a missing `@hyperframes/parsers` build artifact — identical on main, environment not this PR); `iconClassifier` 10/10; `tsc` shows only the three pre-existing generated-artifact errors outside `cli`. Probe: flattening `SHAPE_PREFERENCE` to ranker order fails exactly the bare-mark-over-badge test. Verified by reading: rel/sizes slugified to `[a-z0-9-]` and `extname` cannot carry `/`, so no traversal; `isPrivateUrl` drops `data:`/non-http; collisions get `uniqueStem` `-N`; ordering is a stable sort with DOM-index tiebreak; zero icons / all fetches failing writes an honest empty manifest with no crash; `reason` is always populated; `assets/favicon.<ext>` keeps its stem (asserted).

**1. EF consumer leak (needs a follow-up in experiment-framework, not this PR).** `heygen/services/brand_context/executors/capture_kit_extract.py:23` filters `_IMAGE_NOISE = ^(…|favicon|…)` and `build_media_inventory` scans `assets/*` directly. The new `assets/icon-<rel>-<sizes>.<ext>` files match neither `favicon` nor `logo`, so up to eight opaque badge icons per capture would enter the brand media inventory as ordinary imagery — exactly the "solid blob" complaint, arriving through a different door. Either add `icon-` to `_IMAGE_NOISE` or have EF consume `extracted/icons-manifest.json`; I'll raise it on the EF side so it lands with the pin bump. Worth a line in this body so the OSS release notes carry it.

**2. No byte cap on icon downloads.** `fetchBuffer` has no Content-Length pre-check or byte cap; this PR raises exposure from one to up to eight full downloads held in `bytesByFile`. The 10 s timeout bounds time, not size. `readAssetBytes` already does a 25 MB pre-check — same guard here would do.

Low / nits, no action required: `parseFloat("100%") === 100`, so a full-bleed `width="100%"` rect on a viewBox'd SVG classifies as a bare mark; text wordmarks and `a`/`switch` containers classify `unknown` (conservative); `headlineReason`'s "bare mark was unavailable" branch is unreachable under the preference order; the assets array now carries up to nine `type: "favicon"` entries, so an external `find(type === "favicon")` gets the rank-0 icon rather than the headline (stem-matching consumers are safe).

Both bodies state what was verified and how; the openai.com evidence with hashes held up.

— Jerrai

jrusso1020 (commented):
**COMMENT at `f60c8059`, not a stamp — and the reason is state, not the code.**

## Read this first: this PR is not on its way to merging

As of this write, three consecutive polls apart:

- `mergeable=false`, `mergeable_state=dirty`. The retarget onto `main` left a conflict.
- The only review on it, `5123815153`, is **`DISMISSED`** — the base change dismissed it. `reviewDecision` is `REVIEW_REQUIRED`.
- `autoMergeRequest` is `null`. Nothing is armed.

So "approvals carried and auto-merge armed, lands on green" is not the state on GitHub. It needs
the conflict resolved, and then a fresh approval at whatever head that produces — a dismissed
review does not come back. I am not stamping a `DIRTY` head, so the findings below are a comment;
ping me at the new head and I will re-read the delta and stamp.

## The code

The design is right and the two hard parts are the ones you got right. Downloading every candidate
instead of stopping at the first success is the only way shape can participate at all. Sampling
edge midpoints rather than corners is the correct answer to the disc trap, and it is tested with a
real disc rather than argued. `SHAPE_PREFERENCE` tying `unknown` with `badge` and breaking on rank
is right for exactly the reason the docblock gives: an undecodable `.ico` must not win by being
unexaminable.

Checked and clean: `extracted/` is created at `index.ts:129` with `recursive: true`, well before the
manifest write at `:691`, so that write cannot throw on a missing directory. Stems are sanitised to
`[a-z0-9-]` and de-duplicated by `uniqueStem`, and `ext` comes from `extname()` of a URL pathname,
which cannot contain a separator — no traversal and no collision. `bytesByFile.get(chosen.file)!`
is safe; both maps are written in the same block.

### 1. A full-bleed rect written in percentages is classified as a bare mark

`num()` is `Number.parseFloat`, which parses `"100%"` as `100` and reports it finite:

```
parseFloat("100%") = 100 | isFinite = true
full-bleed test: 100 >= 171 -> false        // canvas 180 x FULL_BLEED_RATIO
verdict would be: bare-mark
```

So `<svg viewBox="0 0 180 180"><rect width="100%" height="100%" rx="90" .../>…` takes neither the
`rect` branch nor the `circle` branch and falls through to the final `bare-mark` return. A backdrop
rect written in percentages is one of the more common ways to write one, and this is the same
drawing as your openai fixture with two attributes spelled differently.

Impact is bounded and worth stating: it declines to improve rather than regressing. For the
mistake to change a headline, the misclassified badge has to out-rank a true bare mark, which is
what pure declared ranking did before this PR. Fix is a line in `num()` — reject a raw value with a
non-numeric suffix, or resolve `%` against the canvas.

### 2. The fallthrough asserts `bare-mark` where your own rule says `unknown`

`SHAPE_PREFERENCE`'s docblock is the sharpest sentence in the PR: *absence of evidence is not
evidence of a bare mark*. The classifier does not follow it. Only `rect` and `circle` are ever
measured; every other first painted element returns

```ts
{ shape: "bare-mark", reason: `first painted element is <${tag}> with margins` }
```

— "with margins" being an assertion nothing checked. `image` and `use` are both in `PAINTING` and
neither is measurable by this code, so `<svg><image width="180" height="180" href="data:…"/></svg>`,
which is a raster badge in an SVG wrapper, is positively identified as a bare mark and promoted
over a real one. Same for a `<use>` of a full-bleed `<symbol>` (`symbol` is not in `NON_PAINTING`,
so it is skipped, and the `<use>` is what gets judged).

Returning `unknown` for an unmeasured element makes the classifier agree with the ranking rule, and
`unknown` already sorts safely.

### 3. `headlineReason`'s "best-ranked bare mark was unavailable" branch cannot run

```ts
const bare = all.filter((i) => i.shape === "bare-mark").length;
if (bare > 0) return `best-ranked bare mark was unavailable; used a ${chosen.shape}`;
```

`all` is `manifest.icons`, which holds only icons that downloaded. If any of them is a bare mark,
`SHAPE_PREFERENCE` sorts it to index 0 and it *is* `chosen`, so this line is only reachable when
`chosen.shape !== "bare-mark"` and a bare mark exists — which cannot both hold. Dead, and its text
describes a state the code cannot produce, which is worse than dead in a field whose whole job is
to be trusted later.

### 4. No byte cap on an icon, and this PR multiplies that by eight

`fetchBuffer` bounds a fetch with a timeout and a content-type guard, but `await res.arrayBuffer()`
is unbounded. That was one icon before this PR. It is now up to `MAX_ICONS = 8`, and every buffer
is retained in `bytesByFile` until `promoteHeadline` runs, so they are live simultaneously rather
than one at a time. A `content-type: image/svg+xml` response of arbitrary size is the shape to
worry about, since the guard only rejects HTML and XML. A `content-length` pre-check plus a cap
matches what `streamToFile` already does for video.

### 5. Stray duplicated docblock

"Download every icon the page declared, classify each, and copy the best onto the historical
`assets/favicon.<ext>` stem" appears twice: once orphaned directly above `uniqueStem`, which it
does not describe, and once (reworded to "promote the best one") above `downloadDeclaredIcons`,
where it belongs. Delete the first.

## CI

Read at post time on `f60c8059`. I am deliberately not reporting a required-context reading here:
the base changed from `fix/capture-favicon-pick` to `main` after these runs, and the gate that
applies is `main`'s ruleset evaluated against whatever head the conflict resolution produces.
Nothing on the current head is failing, but that head is not the one that will merge.

-- Rames

jrusso1020 (approved):
**APPROVE at `d396dd89`.** This is the stamp my comment review at `f60c8059` withheld, and the only
thing that changed is the state it was withheld for: `mergeable` is now `MERGEABLE` where it was
`dirty`, and the earlier review is gone rather than stale — `5123815153` reads `DISMISSED`, which a
retarget does and which no later event undoes.

**The rebase carried the change and nothing else.** Same five files, same `+678/-54`, and all five
blobs are byte-identical to the head I reviewed: `assetDownloader.ts` `0f827be6`, `iconClassifier.ts`
`92ff5a7f`, `iconClassifier.test.ts` `21b54515`, `assetDownloader.test.ts` `857c898b`, `index.ts`
`848d8776`. The `+211/-26` on `assetDownloader.ts` is unchanged against the new base because #3726 is
in `main` now, which is the arithmetic you would want to see. So my review at `f60c8059` applies
verbatim and I am not re-deriving it — the five notes there stand exactly as written, all
non-blocking, and none of them changed status.

**On the red `Test`: it is a cancellation artifact of your own push, not a failing test.** I read
the job rather than inferring it. `Test` is an aggregator, and job `101414170720`'s entire body is:

```
Run echo "::error::Producer unit/integration tests did not succeed."
exit 1
```

Nothing ran and nothing asserted. It went red because the producer jobs it gates on were
`cancelled`, and they were cancelled because there are **two CI waves on this one SHA**: a first at
`02:23:02Z` that the concurrency group killed, and a second at `02:23:21Z` that is still in flight.
`Producer: unit tests` and `Producer: integration tests` are both `in_progress` again as of
`02:24:27Z`, and a fresh `Test` will run behind them. Same for `CLI: npx shim` on all three
platforms and `CLI smoke (required)` — every one of them is `cancelled` from the first wave with an
`in_progress` twin from the second. Branch protection honours the latest run per name, so these
resolve themselves; the thing not to do is read the first wave's corpses as failures. This PR
touches five files, all under `packages/cli/src/capture`, and none under `packages/producer`.

**CI as read at post time, latest run per name, and it is not green yet.** Of `main`'s eight
required contexts: `Semantic PR title` success `02:24:12Z` and `regression` success `02:24:32Z` are
through; `Build`, `Typecheck`, `Test: runtime contract` and `Render on windows-latest` are all
`in_progress`; `Tests on windows-latest` still shows only its `cancelled` first-wave result, with
no successor reported yet; and `Test` is the `failure` at `02:24:05Z` explained above. Worth one
more line on that timing, because it is what makes it self-resolving: that `Test` finished at
`02:24:05Z`, and the second wave's producer jobs did not start until `02:24:27Z` — so it was
reporting on the cancelled first wave and could not have seen the run that is going now.

I am claiming no green run. The verdict is on the code, which is byte-identical to a head I had
already read in full, and the second wave will say what it says.

Auto-merge is not armed as of this write and I am arming nothing.

-- Rames

jrusso1020:
One correction to the CI paragraph in my approval, in the same safe direction as the last one and
still worth fixing: I listed **`Test: runtime contract` as `in_progress`**. It had already finished
`success` — run `101414288377`, completed on the second wave. Five of the eight required contexts
now have a second-wave result or a first-wave pass; `Build`, `Typecheck` and
`Render on windows-latest` are genuinely still running, and `Tests on windows-latest` still has no
successor to its cancelled first-wave run.

I had the evidence at write time and did not read it: my own guard printed
`required-not-success='Build=CANCELLED,Render on windows-latest=CANCELLED,Test=FAILURE,Tests on
windows-latest=CANCELLED,Typecheck=CANCELLED'` — `Test: runtime contract` is absent from that list,
which is exactly the tell that it had gone green.

While correcting: **two tools will give you two different readings of this head and both are
right.** A guard that takes the last *completed* run per name sees `Build=CANCELLED`; a read that
takes the last *started* run sees `Build=in_progress`. The first wave was cancelled and the second
is still going, so both statements are true about different runs. Branch protection honours the
latest run per name, which is the one still in flight — so neither reading is a reason to act, and
the `CANCELLED`s in particular are not failures.

Verdict unchanged: APPROVED at `d396dd89`, on code byte-identical to `f60c8059`.

-- Rames

PR #3729, opened 2026-09-06
fix(render): probe runtime media source mutations

miguel-heygen:
## Summary

Known-duration compositions now launch the browser reconciliation probe when inline JavaScript synchronously mutates the source of an existing video, audio, or nested `<source>`. Extraction, video-frame injection, and audio mixing therefore receive the browser-live source instead of silently retaining the static fallback.

An AST-backed `media_runtime_src_mutation` warning identifies literal `.src = …` and `setAttribute("src", …)` writes tied to statically resolvable managed media. It recommends a final static source or declared `data-var-src`; image and script source mutations are excluded. The warning remains useful for delayed or external mutations that a pre-extraction probe cannot guarantee.

## Validation

- Probe-stage suite: 39/39 tests passed.
- Media lint suite: 67/67 tests passed.
- Parser, lint, and producer package typechecks passed.

The sanitized reconciliation fixture uses distinct SHA-256 identities:

- Video A: `986b0b8e67d317608d3edba09b7cdc49afb8237df7e4c7624b766f5d6e0951ac`
- Video B: `4f250966375ee944833a5633af41005817fd3dffd8975bf16d13f7e3bc3feb09`
- Audio A: `ec6b554699634f58f2783bfdf8345d3280c7217010b932564ea8c3c140f60935`
- Audio B: `28d1a173cb73315c777ac2970cb402267dae4661cb26863d02ad700bd3be469e`

The test starts with A in compiled metadata, discovers B from the browser, and proves B reaches media preflight for both video and audio. Original-project preview/output hash verification remains unavailable because the private assets and output were not supplied.

### DIFF
--- a/packages/lint/src/rules/media.test.ts
+++ b/packages/lint/src/rules/media.test.ts
@@ -379,6 +379,62 @@ describe("media rules", () => {
     expect(finding).toBeUndefined();
   });
 
+  it("warns on raw src mutations targeting existing video and audio", async () => {
+    const html = `
+<html><body>
+  <div data-composition-id="c1" data-width="1920" data-height="1080">
+    <video id="clip" src="video-a.mp4"></video>
+    <audio id="voice" src="audio-a.wav"></audio>
+  </div>
+  <script>
+    const clip = document.getElementById("clip");
+    clip.src = "video-b.mp4";
+    document.querySelector("#voice").setAttribute("src", "audio-b.wav");
+  </script>
+</body></html>`;
+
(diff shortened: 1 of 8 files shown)


miga-heygen (approved):
Reviewed: AST-based `extractMediaSrcMutations` correctly detects .src assignment and setAttribute on statically-resolvable DOM lookups, scoped to video/audio/source tags. Probe reconciliation and lint paths both wired. Tests thorough. — Miga

PR #3730, opened 2026-09-06
fix(cli): clarify publish visibility and claim links

miguel-heygen:
## Summary

- describe publish as private by default across CLI help, confirmation copy, README, and workflow references
- label the post-publish line `Requested visibility`, since the publish response carries no visibility field and the CLI can only report what it asked for
- keep an in-place re-publish from claiming it made the project private; it reports that the existing setting is kept unless `--public` is passed
- relabel anonymous auth-required links as Claim URL and explain that `--yes` only skips confirmation
- cover `--yes`/`--public` flag orthogonality, the in-place re-publish path, and anonymous claim-link output

## Testing

- `vitest run packages/cli/src/commands/publish.test.ts` (12 passing)
- `oxfmt --check .`, `bun run lint`, skills manifest `--check` — all clean
- the in-place test was confirmed non-vacuous by breaking the branch on purpose and watching it fail

## Scope note

CLI messaging only; no wire change and no change to how visibility is sent. Two behaviours were
confirmed against the publishing service before wording this:

- **A fresh publish is private by default.** A request without `--public` carries no visibility
  field, and the project is created private.
- **An in-place update preserves the project's existing visibility.** A re-publish without
  `--public` sends no visibility, and the existing value is kept — a plain re-publish never
  flips a public project private.

That second point is what the old copy got wrong. `--update` is only one of three routes into
the in-place path; the common one is the per-directory id this CLI already persists, so
`hyperframes publish --public` followed by a plain `hyperframes publish` in the same directory
updates in place and stays public. Printing "Private — authentication and access required"
there tells someone a public link is locked down. The copy now branches on whether the response
came back as an in-place update, so it covers all three routes rather than just the flag.


jrusso1020:
Post-merge note — I had this finished against `cf2c479a`, which is the exact head that merged as `672ea844`, so it all still applies to what is on `main`. Recording it here rather than as a review, since the stamp is moot on a merged PR.

**It would have been an approve.** The fix landed in better shape than what I proposed on Slack: I suggested hedging the label, and this hedges it *and* distinguishes the two cases the hedge was papering over.

What I checked:

- The three-way at `publish.ts:231-238` is correct against the server behaviour. `--public` earns its definitive `Public` because `is_public` really is transmitted on both paths (`publishProject.ts:575` direct, `:674` staged/complete); the in-place branch correctly says `Unchanged` because nothing about visibility is sent; the fresh branch keeps `Private`. The comment above it is accurate — `PublishedProjectResponse` (`publishProject.ts:29-37`) carries no visibility field and `parsePublishedProjectResponse` reads none, so the CLI can only state what it asked for.
- It keys on `updatedInPlace` (`publish.ts:225`, response id vs `requestedProjectId`) rather than on `--update`, so all three routes through `publish.ts:171` — `updateTarget ?? committedTeam?.projectId ?? ensureProjectId(dir)` — land on one condition. That includes the flag-less one, where `ensureProjectId` (`projectLink.ts:67-77`) returns the same per-directory id and makes a second plain `hyperframes publish` an update.
- The new test drives that route (real `ensureProjectId`, mock echoes the id) instead of the flag, and `not.toContain("Private — authentication and access required")` pins the exact string that was wrong.
- All eight required contexts were green at `cf2c479a`, checked name by name. The manifest hashes moved for exactly the three skills whose files changed, file counts unchanged.

### Shipped unfixed — three non-blocking items

**1. The label column jogs on the signed-out path.** `Project`, `Files`, `Proxies`, `Proxy note` (`publish.ts:210-218`) print *before* the `if (published.claimed)` split at `:223`, and were widened to column 21 to fit `Requested visibility`. The anonymous branch at `:274` still prints at column 11. Measured: lines 210/211/214/218/226/237/239/243 land at 21, lines 277/278 land at 11. A signed-out publish now renders as

```
  Project              my-video
  Files                12
  Claim URL  https://hyperframes.dev/p/...
  Access     Sign in required to claim
```

Everything sat at 11 before this change. Cosmetic, but signed-out is the first-run path.

**2. The new tests write into the real `~/.hyperframes/projects.json`.** `runPublish` reaches `publish.ts:171` -> `ensureProjectId(dir)`, and the in-place case calls it directly; `projectLink.ts:13-14` resolves that file under `homedir()`. Each run appends an entry keyed on a `/tmp/hf-publish-visibility-*` path that `rmSync` then removes, so the entries are dead on arrival and never pruned. The pre-existing tests in this file abort before `:171`, so this is new. Mocking `../utils/projectLink.js` the way `../utils/publishProject.js` is already mocked at the top of the file would keep it self-contained.

**3. The README didn't get the caveat the skill doc got.** It states "Published projects are private by default" with no in-place qualifier, while `preview-render.md` spells the distinction out. True for a fresh publish, incomplete for the update path — which is the path this change exists to describe correctly.

Minor, mentioning only because the copy is the point of the PR: the `Tip` at `:242-245` prints whenever `args.public !== true`, now including the in-place case, so "Re-publish with `--public` to allow anyone with the URL" sits one line under `Unchanged — keeps this project's current setting` and implies the project is not currently public.

Happy for 1-3 to ride the release PR or a follow-up — none of them justify holding v0.8.30.

-- Rames

PR #3732, opened 2026-09-06
fix(media): unify nested start coordinate mapping

miguel-heygen:
## Summary

- replace the runtime numeric-overlap heuristic with a shared local-by-default media start resolver used by runtime/snapshot and producer collection
- preserve legacy root-global media timing only through the explicit data-hf-media-start-basis="global" marker
- add nested H=2 / media=2 regressions for runtime visibility and producer video/audio collection
- warn on unmarked nonzero media starts in sub-compositions and document the coordinate contract

## Testing

- core runtime init suite: 97 passing
- producer renderMediaCollector suite: 2 passing
- lint media suite: 68 passing
- core, producer, and lint typechecks
- core package build
- package subpath contract
- changed-file oxfmt, oxlint, and git diff check

## Scope note

The source-level RED/GREEN fixture proves the exact +2s mapping defect and both local/global controls. The original private media and same-time snapshot/raw/decoded MP4 hashes were unavailable, so this PR does not claim pixel-level reproduction of that private output.


miga-heygen (approved):
Reviewed: replaces the fragile window-overlap heuristic with an explicit `data-hf-media-start-basis` attribute. The shared `resolveAbsoluteMediaStartSeconds` predicate is correctly wired into runtime, lint, and producer extraction paths. Tests cover local/global/legacy timing. — Miga

PR #851, opened 2026-05-14
test(producer): add mp4 H.265 SDR distributed fixture

jrusso1020:
## Description

Phase 4 of the distributed rendering plan: test fixtures (see DISTRIBUTED-RENDERING-PLAN.md §11 Phase 4 + §10 test strategy). This is PR 4.3 of the Phase 4 remainder, stacked on PR 4.3-pre (#850) which added the codec knob to `DistributedRenderConfig`.

This PR adds the mp4 H.265 SDR fixture (`tests/distributed/mp4-h265-sdr/`) plus the small harness plumbing that exposes the codec knob through `meta.json`. Composition mirrors the H.264 fixture: 2 seconds (60 frames) at 30fps with text, a crossfade transition straddling the frame-30 chunk seam, and a continuously rotating SVG icon. `renderConfig.format: "mp4"` + `renderConfig.codec: "h265"` + `chunkSize: 15` routes the distributed pipeline through libx265 with closed-GOP keyint params (`min-keyint=N:scenecut=0:open-gop=0:repeat-headers=1`) so concat-copy at assemble time round-trips losslessly.

**Cross-codec PSNR assertion**: the in-process renderer doesn't expose a codec hint (its RenderConfig only switches codec via HDR mode), so the in-process baseline for this fixture is rendered as h264. The harness's PSNR comparison therefore measures "libx265 chunked + concat" against "libx264 single-pass" on the same source frames. At "high" quality both encoders are near-lossless on simple vector+text content; observed PSNR is ~48dB across all 100 checkpoints — well above the 30dB threshold. This catches gross codec/encoder failures (e.g. libx265 emitting wrong bit depth or losing the IDR-at-chunk-seam contract) while accepting normal cross-codec PSNR drift. The in-process arm renders h264 vs the h264 baseline byte-identically.

Harness extensions:

1. **`TestMetadata.renderConfig.codec`** field accepted by `validateMetadata`. Rejected with format ∉ {mp4} for symmetry with the `DistributedRenderConfig` runtime check from 4.3-pre.
2. **`RunDistributedSimulatedInput.codec`** plumbed through to `plan()`. The non-mp4 plan-config branch keeps the field structurally absent (so byte-identical to pre-codec planDirs for mov/png-sequence) rather than passing `undefined`, which would surface in JSON.

## Testing

- `bun run --cwd packages/producer docker:test:update mp4-h265-sdr` — baseline rendered inside `Dockerfile.test` (h264 mp4 from in-process, used as the cross-codec reference)
[...]

PR #850, opened 2026-05-14
feat(producer): add codec knob to DistributedRenderConfig

jrusso1020:
## Description

Phase 4 prerequisite for PR 4.3 (the mp4 H.265 SDR distributed fixture). Splits the codec selection out of `DistributedRenderConfig.format` so callers can ask for libx265 without changing the format.

**Surface change** (`@hyperframes/producer/distributed`):

- `DistributedRenderConfig.codec?: "h264" | "h265"` — defaults to `"h264"`, ignored for non-mp4 formats. Passing `codec` with `format !== "mp4"` throws at plan time with a clear error so caller mistakes surface immediately rather than producing a silently-wrong planDir.
- `FORMAT_ENCODER_TABLE` is replaced by `resolveEncoderTriple(config)` — a small function that switches on `(format, codec)`. mp4 + h265 → `{encoder: "libx265-software", pixelFormat: "yuv420p"}`. mov and png-sequence are unchanged.

**Plumbing through `renderChunk`**:

The chunk worker reads `LockedRenderConfig.encoder` from `meta/encoder.json`. When that's `"libx265-software"`, the worker overrides `getEncoderPreset(quality, "mp4")`'s default `codec: "h264"` with `"h265"` so `runEncodeStage` invokes libx265 with the closed-GOP keyint params (`min-keyint=N:scenecut=0:open-gop=0:repeat-headers=1`) that survive concat-copy at assemble time. The engine layer (`packages/engine/src/services/chunkEncoder.ts`) already supports both codecs — this PR is purely the distributed config surface.

**Bit depth**: SDR-only, 8-bit yuv420p for both codecs. h265 + 10-bit yuv420p10le is HDR territory and lives in v1.5 (see plan §12).

## Testing

- `bun test packages/producer/src/services/distributed/plan.test.ts` — 14 tests pass including 3 new codec cases (`codec` defaults to h264, `codec: "h265"` maps to libx265-software, non-mp4 + codec throws)
- `bun test packages/producer/src/services/distributed/` — all 42 distributed unit tests pass
- `bunx oxlint` + `bunx oxfmt --check` clean
- `bunx tsc --noEmit` (producer package) clean

The H.265 fixture that exercises this end-to-end inside `Dockerfile.test` lands in the follow-up PR 4.3.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3735, opened 2026-09-06
fix(catalog): bind asset reads to checked project files

jrusso1020:
Catalog asset processing could check a source pathname and then read a different file after replacement. A shared reader now resolves references within the prepared project, opens read-only/nonblocking, checks and reads the same descriptor, and closes it on every path. It covers hosted/inlined references, mounted composition HTML, directory copies and download mirrors. This addresses CodeQL #850.

Directory publication reads at most the remaining 2 MB source budget plus one byte from each checked descriptor, in chunks of at most 64 KB. It charges the resulting buffer lengths before any output write, so oversized or concurrently growing media cannot cause an unbounded allocation. Both directory layouts and download mirrors reuse those buffers, so replacement between discovery and open cannot bypass the cap, and later changes cannot alter mirrored bytes. The existing source-counting policy, exact-limit acceptance, mirror collision precedence and internal downloads directory aliases are preserved.

External file links and sibling-prefix mounted-composition escapes remain unresolved instead of publishing files outside the prepared project. Internal links and symlinked project roots still work. Asset hashing, MIME handling, empty-file behavior, reference reporting and output write semantics are preserved. The prepared project tree remains trusted against hostile concurrent parent-directory renames. Cache-write alerts #851/#852 are reserved for the next narrow batch.

Validation: all 254 script/catalog tests (188 Node + 66 Vitest), script typecheck, dependency builds, lint/format and signed hooks pass. The 24 focused cases cover pathname replacement, external paths, directory budget replacement, all-or-nothing rejection, exact-limit mirrors, internal aliases, FIFO, short descriptor reads, sparse oversized inputs, growth after fstat and descriptor cleanup. Three bounded-read cases fail on the previous PR head and pass on this revision; the earlier four budget/mirror regression witnesses remain green. The suite is registered in the existing Windows lane; fresh CI and Windows results are pending.

Optional Fallow audit remains red locally: complexity and duplication findings in the small filesystem routines and unchanged cache-write blocks. This revision removes two duplicate directory walkers and simplifies repeated missing-file handling. Remaining new routines are small, covered filesystem checks/traversal; the descriptor cleanup pattern resembles the engine reader, and the two unchanged cache-write blocks are covered by the next batch. No audit suppressions were added. Required checks, CodeQL and approval of the current commit remain merge gates.

### DIFF
--- a/.github/workflows/windows-render.yml
+++ b/.github/workflows/windows-render.yml
@@ -484,6 +484,11 @@ jobs:
         shell: pwsh
         run: bunx vitest run packages/producer/src/services/hyperframeLint.file-race.test.ts --maxWorkers=2
 
+      - name: Verify catalog source reads on Windows
+        if: matrix.lane == 'studio-engine-cli'
+        shell: pwsh
+        run: bunx vitest run scripts/catalog/catalog-payload-assets.file-race.test.ts --maxWorkers=2
+
       - name: Run runtime contract test
         if: matrix.runtime_contract
         shell: pwsh
--- a/scripts/catalog-payload-assets.ts
+++ b/scripts/catalog-payload-assets.ts
@@ -13,6 +13,7 @@ import {
   mkdirSync,
   readdirSync,
   readFileSync,
+  readSync,
   statSync,
   writeFileSync,
   realpathSync,
@@ -152,8 +153,26 @@ function isMissingFile(error: unknown): boolean {
   );
 }
 
+/** Keep oversized or concurrently growing directory assets within the read budget. */
+function readWithinBudget(fd: number, maxBytes: number): Buffer<ArrayBuffer> {
+  const chunks: Buffer<ArrayBuffer>[] = [];
+  let remaining = maxBytes + 1;
+  while (remaining > 0) {
+    const chunk = Buffer.alloc(Math.min(64 * 1024, remaining));
+    const count = readSync(fd, chunk, 0, chunk.length, null);
+    if (count === 0) break;
+    chunks.push(chunk.subarray(0, count));
+    remaining -= count;
+  }
+  return Buffer.concat(chunks);
+}
+
 /** Read one checked file from the prepared project, including internal links. */
-function readProjectFile(root: string, filePath: string): Buffer<ArrayBuffer> | null {
+function readProjectFile(
+  root: string,
+  filePath: string,
+  maxBytes?: number,
+): Buffer<ArrayBuffer> | null {
   if (!isWithin(root, filePath)) return null;
   let source: string;
   try {
@@ -174,7 +193,7 @@ function readProjectFile(root: string, filePath: string): Buffer<ArrayBuffer> | 
   }
   try {
     if (!fstatSync(fd).isFile()) return null;
-    return readFileSync(fd);
+    return maxBytes === undefined ? readFileSync(fd) : readWithinBudget(fd, maxBytes);
   } finally {
     closeSync(fd);
   }
@@ -353,7 +372,11 @@ export function hostItemDirectory(projectDir: string, destDir: string, urlBase: 
   const files = new Map<string, Buffer<ArrayBuffer>>();
   let total = 0;
   for (const path of hostedPaths(projectDir)) {
-    const bytes = readProjectFile(projectDir, join(projectDir, path));
+    const bytes = readProjectFile(
+      projectDir,
+      join(projectDir, path),
+      MAX_HOSTED_DIRECTORY_BYTES - total,
+    );
     if (bytes === null) continue;
     total += bytes.length;
     if (total > MAX_HOSTED_DIRECTORY_BYTES) return "";
--- a/scripts/catalog-payload-assets.ts
+++ b/scripts/catalog-payload-assets.ts
@@ -8,8 +8,20 @@
  */
 
 import { createHash } from "node:crypto";
-import { existsSync, mkdirSync, readdirSync, readFileSync, statSync, writeFileSync } from "node:fs";
-import { extname, join, resolve } from "node:path";
+import {
+  existsSync,
+  mkdirSync,
+  readdirSync,
+  readFileSync,
+  statSync,
+  writeFileSync,
+  realpathSync,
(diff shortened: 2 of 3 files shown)


miguel-heygen (changes requested):
The individual source reads are substantially improved. `scripts/catalog-payload-assets.ts:142-183` rejects lexical escapes and resolved external links, opens read-only/nonblocking, checks and reads the descriptor, and closes on every path. The four callers now hash/inline/copy the bytes returned by that one owner; empty files, internal links, and symlinked project roots remain intact. The 15 new focused tests pass locally, as do changed-file lint/format. Exact-head Build and Typecheck are green; required Test/Windows and JavaScript CodeQL remain pending.

**Blocker — the directory byte cap is still a separate pathname check from the bytes published.** `directoryBytes()` at `scripts/catalog-payload-assets.ts:344-355` stats every candidate, then `hostItemDirectory()` authorizes the whole copy at `:368-369`; only afterward do `walk()`/`walkInto()` reopen the paths through `readProjectFile()` at `:372-406`. A file replaced after its budget stat but before the descriptor open is therefore charged at the old size and published at the new size.

I reproduced this at the exact head with an instrumented test: a one-byte internal `asset.png` was replaced with a 2,000,001-byte file immediately after `directoryBytes()` obtained its stat. `hostItemDirectory()` returned `/item/` and wrote the full 2,000,001-byte asset, bypassing `MAX_HOSTED_DIRECTORY_BYTES`. I removed the probe afterward. The current replacement witnesses swap only after `readProjectFile()` has opened its fd, so they cannot see the earlier budget→open window.

Please make the budget and copy consume the same checked bytes—e.g. collect `{ relativePath, bytes }` once, sum `bytes.length` before writing anything, then publish those exact buffers (including the `_downloads` mirror mapping). Add a replacement witness between budget discovery and source open, and preserve the current all-or-nothing behavior for over-budget directories.

**Non-blocking CI note:** optional `Fallow audit` is red at this head with 16 findings, including critical complexity on `readProjectFile` and a 61-line clone with the existing descriptor reader. It is not in the required set, but the PR should either simplify/share the helper or explicitly account for that result before calling CI clean.

**Verdict:** REQUEST CHANGES
**Reasoning:** Per-file identity is pinned correctly, but directory publication still authorizes bytes using stale pathname sizes, so the same replacement race bypasses the advertised 2 MB budget.

— Magi

miguel-heygen (changes requested):
The original budget/identity blocker is closed: `hostItemDirectory()` now collects each descriptor-checked buffer once, sums those exact lengths before any output write, and reuses the same map for original and `_downloads` mirror paths (`scripts/catalog-payload-assets.ts:351-378`). Internal aliases resolve to their real in-project prefix, external aliases avoid traversal, and the second publication loop preserves mirror-wins collision order. The new all-or-nothing, exact-limit, pre-open replacement, post-collection mutation, alias, and collision tests are well targeted. Focused suite is 20/20 and changed-file lint/format pass locally.

**Blocker — an over-budget file is now fully buffered before the cap can reject it.** `readProjectFile()` performs an unbounded `readFileSync(fd)` at `scripts/catalog-payload-assets.ts:155-180`; only after that allocation returns does `hostItemDirectory()` add `bytes.length` and compare against `MAX_HOSTED_DIRECTORY_BYTES` at `:355-360`. The old `directoryBytes()` stat path rejected a 2 MB+ file without reading it. Since `HOSTED_EXTENSIONS` includes `.mp4` and `.webm`, a large catalog asset—or the same replacement attacker swapping in a huge sparse file before descriptor open—can OOM the script before the 2 MB guard executes. The new cap+1 test itself necessarily reads all 2,000,001 bytes; scaling that fixture exposes the regression.

Please make the directory collector's read bounded by the remaining budget: read at most `remaining + 1` bytes from the checked descriptor (a fstat fast-reject is useful, but the bounded read must remain authoritative if the inode grows), signal over-budget without allocating the whole source, and keep reusing the accepted buffer for both output spellings. Add a test that proves an oversized/sparse source is never read beyond cap+1 while output remains all-or-nothing.

The optional Fallow tradeoff is adequately disclosed: the revision removes both duplicate walkers, reduces reader cyclomatic complexity 15→9, adds no suppression, and reserves unchanged cache-write clones for #851/#852. Its remaining red is not the blocker here.

Exact-head required checks are still running; the only completed failure is optional Fallow, already accounted for in the body.

**Verdict:** REQUEST CHANGES
**Reasoning:** The revised snapshot model fixes stale authorization and mirror drift, but enforcing the 2 MB output cap only after an unbounded read turns large or swapped-in media into a new memory-exhaustion path.

— Magi

miguel-heygen (approved):
Re-review at the bounded-read head: both directory-publication blockers are closed.

`readWithinBudget()` at `scripts/catalog-payload-assets.ts:155-168` reads the checked descriptor in ≤64 KiB chunks and stops after `maxBytes + 1`; `readProjectFile()` keeps that inside its existing descriptor `finally` at `:171-199`. `hostItemDirectory()` passes the remaining budget at `:372-384`, so accepted buffers total at most 2 MB, a sparse or growing source can consume at most cap+1 bytes, and the same accepted map remains authoritative for original and mirror writes. EOF, short reads, a zero-byte file at zero remaining budget, and inode growth all have the right result.

The expanded tests are strong: the 32 MB sparse file proves total actual reads stop at cap+1; the post-fstat growth case proves fstat is not the authority; the short-read case prevents truncation; the failure case pins descriptor cleanup; and the earlier all-or-nothing/mirror/alias/collision cases remain intact. I ran the focused suite (24/24) and changed-file lint/format; all pass.

The optional Fallow failure is adequately accounted for: no suppression, two duplicate walkers removed, and only one unchanged cache-write clone group remains alongside the small covered filesystem routines. It is not in the required set. Required Build/Typecheck/Windows are still running; exact-head required checks have no failures and remain landing gates with JavaScript CodeQL.

**Verdict:** APPROVE
**Reasoning:** Directory publication now enforces identity, output budget, and memory budget from the same bounded descriptor reads, then publishes only those captured buffers with preserved mirror semantics.

— Magi

PR #852, opened 2026-05-14
test(producer): add chunk-boundary fixtures per first-party adapter

jrusso1020:
## Description

Phase 4 of the distributed rendering plan: test fixtures (see DISTRIBUTED-RENDERING-PLAN.md §11 Phase 4 + §10 test strategy). This is PR 4.7 of the Phase 4 remainder — the final fixture PR.

This PR adds six per-adapter chunk-boundary fixtures under `tests/distributed/{gsap,anime,three,lottie,css,waapi}-boundary/` plus a single `bun:test` driver (`chunkBoundary.test.ts`) that exercises each fixture's seek-determinism contract. Each fixture is a 60-frame composition (2s @ 30fps, 320×180) that drives the named adapter through the HyperFrames runtime's seek hook. The test renders each at `chunkSize=60` (N=1 chunk, no seams) and `chunkSize=15` (N=4 chunks, three seams at frames 15/30/45), then asserts every PNG frame is byte-identical across the two runs.

**Why png-sequence**: mp4 bitstreams encode keyframe placement directly. At `chunkSize=60` libx264 emits 1 IDR; at `chunkSize=15` it emits 4 IDRs at frames 0/15/30/45. Those are legitimately different bytes even when the captured pixels are identical. png-sequence's assemble path merges chunk frame directories with no re-encode, so per-frame byte equality is exactly pixel equality — the strongest contract a distributed render can satisfy.

**Fixture design** (each is ~60 lines of HTML):
- `gsap-boundary` — single GSAP `tl.to(...)` driving translateX + rotation linearly across 2s.
- `anime-boundary` — anime.js v4 timeline registered via `window.__hfAnime`.
- `three-boundary` — minimal Three.js scene; cube rotation derived from `window.__hfThreeTime`.
- `lottie-boundary` — inline Lottie JSON (rectangle layer animating position+rotation) loaded via `lottie-web` and registered via `window.__hfLottie`.
- `css-boundary` — pure `@keyframes` animation; the HyperFrames CSS adapter seeks via `animation-delay`.
- `waapi-boundary` — `element.animate()` with linear keyframes; runtime sets `currentTime` per frame.

[...]

PR #3739, opened 2026-09-06
fix(catalog): publish cache assets without overwriting entries

jrusso1020:
Catalog asset references and large embedded data URIs checked for an existing hashed output and then wrote by pathname. A competing file or symlink could appear between those operations and be overwritten. Both callers now share cache publication that stages complete bytes in a private sibling directory, then atomically links them into place only if the destination is still absent. This addresses CodeQL #851 and #852.

Existing cache entries remain untouched, including dangling links, and cache hits require no staging or writable directory. Hashes, URLs, MIME handling, externalization thresholds and result counts are preserved. Cache misses reject a symlinked cache directory. Cleanup is best effort so it cannot mask a write/publication failure or turn a successful publication into an error. The output directory's ancestors remain trusted against hostile concurrent parent renames; cleanup failure may leave private staging residue. Directory-copy overwrite behavior is unchanged.

Validation: all 274 script/catalog tests (188 Node + 86 Vitest), script types, dependency builds, lint/format and signed hooks pass. Twenty new cases cover both callers: competing files/live or dangling links, existing dangling entries, external directory links, complete publication, cache reuse, partial-write cleanup and publication/cleanup failures. The competing-entry, dangling-link, directory-link and partial-write witnesses fail against main. The new suite joins the existing catalog source-read suite in the Windows lane.

Optional Fallow remains red locally: eight complexity findings, seven classified as inherited, and no duplication groups. The new cache helper has cyclomatic complexity 8; its branches enforce cache-hit preservation, directory type, competing-entry handling and cleanup. No suppressions were added. Fresh CI, Windows, CodeQL and Magi's review of the current commit remain merge gates.

### DIFF
--- a/.github/workflows/windows-render.yml
+++ b/.github/workflows/windows-render.yml
@@ -489,10 +489,10 @@ jobs:
         shell: pwsh
         run: bunx vitest run packages/producer/src/services/windowsFrameStaging.test.ts --maxWorkers=2
 
-      - name: Verify catalog source reads on Windows
+      - name: Verify catalog file operations on Windows
         if: matrix.lane == 'studio-engine-cli'
         shell: pwsh
-        run: bunx vitest run scripts/catalog/catalog-payload-assets.file-race.test.ts --maxWorkers=2
+        run: bunx vitest run scripts/catalog/catalog-payload-assets.file-race.test.ts scripts/catalog/catalog-payload-assets.cache-race.test.ts --maxWorkers=2
 
       - name: Run runtime contract test
         if: matrix.runtime_contract
--- a/.github/workflows/windows-render.yml
+++ b/.github/workflows/windows-render.yml
@@ -484,10 +484,10 @@ jobs:
         shell: pwsh
         run: bunx vitest run packages/producer/src/services/hyperframeLint.file-race.test.ts --maxWorkers=2
 
-      - name: Verify catalog source reads on Windows
+      - name: Verify catalog file operations on Windows
         if: matrix.lane == 'studio-engine-cli'
         shell: pwsh
-        run: bunx vitest run scripts/catalog/catalog-payload-assets.file-race.test.ts --maxWorkers=2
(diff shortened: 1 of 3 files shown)


miga-heygen:
Closing — this PR has merge conflicts and has been inactive. Thank you for the contribution! Please feel free to reopen with a rebased branch.

miguel-heygen (approved):
Exact-head security/concurrency review at `6cde0fab99d2dd3f81c227932cf7e922f44684a4`.

- `scripts/catalog-payload-assets.ts:217` preserves cache hits—including dangling leaves—without requiring `mkdir`, staging, or write access.
- `scripts/catalog-payload-assets.ts:221-229` stages complete bytes in an atomically created private sibling and makes `linkSync` the sole publication operation; only `EEXIST` is swallowed, so a competing regular file or live/dangling link is never replaced.
- `scripts/catalog-payload-assets.ts:230-234` keeps cleanup best-effort without masking the original write/link failure or turning a successful publication into an error.
- `scripts/catalog/catalog-payload-assets.cache-race.test.ts:95-162` covers both callers across cache reuse, competing entries, directory links, partial writes, publication failure, and cleanup failure. The two focused catalog race suites pass 44/44 locally.
- `.github/workflows/windows-render.yml:487-495` resolves the rebase conflict correctly: the newer frame-staging test remains, and both catalog file-operation suites run in the Windows lane.

No blocking findings. The disclosed remaining pathname trust boundary for hostile concurrent ancestor swaps is real but not widened by this patch; the optional Fallow failure is the disclosed cyclomatic-8 helper plus inherited findings, with no suppression. Required `Test` and Windows checks are still pending at review time, so branch protection remains the merge gate.

**Verdict:** APPROVE
**Reasoning:** The implementation gives both hashed-cache writers one atomic, no-replace publication boundary while preserving existing hit, hash, URL, MIME, threshold, and count behavior; the adversarial tests exercise the relevant state transitions and cleanup paths.

— Magi

PR #3750, opened 2026-09-07
fix(slideshow): retry init at DOMContentLoaded when the children were not parsed yet

BuddhiLW:
## What

`<hyperframes-slideshow>` retries `init()` at `DOMContentLoaded` when the deferred init ran before its children were parsed.

## Why

`connectedCallback` schedules `init()` on `setTimeout(0)` and expects the parser to have appended the children by then. That holds when the page arrives in one piece, and not otherwise. With `player.js` and `slideshow.js` loaded from `<head>`, which is the layout `hyperframes present` emits, the page bound when served by `present`'s own server, but the same HTML served by a plain static file server (`python3 -m http.server`) never bound in headless Chromium: `init()` ran on an empty subtree, returned, and the deck had no chrome and dead arrow keys. Calling `init()` by hand afterwards bound it.

I hit this exporting decks from Plato (a Clojure slide engine) to HyperFrames; it works around it by placing the scripts after the element.

## How

- `locateSlideshowParts(root)` classifies the subtree as `ready` or `incomplete` (`no-player`, `no-island`, `malformed-island`). `init()` uses it in place of the inline checks.
- On `incomplete`, `retryInitWhenParsed` arms one `DOMContentLoaded` listener, only while `document.readyState === "loading"`. Once parsing is over nothing more will arrive, so a missing or malformed island still fails quietly as before.
- The listener is dropped in `disconnectedCallback` and guarded by `initGeneration`, like the timer.

## Test plan

- [x] Unit tests: the classification and the readyState rule directly; the element binds at `DOMContentLoaded` once the children exist, retries after a truncated island, does not arm once parsing is over, drops the retry on disconnect. `vitest` in `packages/player`: 88 passed.
- [x] `typecheck`, `oxlint`, `oxfmt --check`, `fallow` via the pre-commit hook.
- [x] Manual, headless Chromium, the `present` page over a static file server: 0.8.30 bundle never binds; this branch's bundle binds and ArrowRight advances. Served by `hyperframes present` itself, both bind.

### DIFF
--- a/packages/player/src/slideshow/hyperframes-slideshow.test.ts
+++ b/packages/player/src/slideshow/hyperframes-slideshow.test.ts
@@ -1,7 +1,11 @@
 // fallow-ignore-file code-duplication
 import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
 import { handleRuntimeMessage } from "../runtime-message-handler.js";
-import { dropInvalidSlides } from "./hyperframes-slideshow.js";
+import {
+  childrenMayStillArrive,
+  dropInvalidSlides,
+  locateSlideshowParts,
+} from "./hyperframes-slideshow.js";
 import { slideshowChannelName } from "./slideshowPresenter.js";
 
 // Dynamic import defers custom-element registration until happy-dom is active.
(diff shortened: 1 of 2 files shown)


jrusso1020 (approved):
No blocking findings. The one-shot parser retry is bounded by readyState and guarded against stale connections (`packages/player/src/slideshow/hyperframes-slideshow.ts:532`); disconnect removes the listener (`:312`). Reconnect and parser-event ordering also check out.

Validation: all 88 slideshow tests passed locally. CI has no failures reported but several checks are still running; the manual Chromium repro was not independently repeated.

— Codex

**Verdict:** APPROVE
**Reasoning:** This narrowly fixes early initialization without changing post-parse failure behavior, and covers missing/truncated children and disconnect cleanup.

PR #3752, opened 2026-09-07
fix(studio): parse modern CSS colors in the inspector and gradient editor

dkryaklin:
## Problem

Select an element styled with `color: oklch(0.7 0.15 200)` in Studio. The inspector swatch shows black. Drag the picker and the element turns black.

`parseCssColor` in `packages/studio/src/components/editor/colorValue.ts` only read hex and comma-separated `rgb()`. That is not what reaches it:

- `getComputedStyle` keeps `oklch()`, `oklab()` and `lab()` as written, and turns `color-mix()` into `color(srgb ...)`.
- Gradient stops are read from the authored `background-image`, so `white`, `#0f172acc` or `rgb(255 0 0 / 50%)` are common there.

All of these returned `null`. The inspector fell back to black. In the gradient editor a stop inserted between two named stops copied its left neighbour instead of blending.

## Fix

- `parseCssColor` parses with `@colordx/core` plus its named-color plugin. It covers CSS Color 4, has no dependencies, is 8.4 KB gzipped, and is the fastest parser in its class. 
- `gradientValue.ts` drops its own parser and uses `parseCssColor`.
- `InlineTextToolbar.tsx` drops its canvas fallback, which only existed for named colors.

## Verification

- `colorValue.test.ts`: modern computed colors, named colors, 8-digit hex, gamut clipping, rejected input, picker hex conversion.
- `gradientValue.test.ts`: interpolation across named stops and across 8-digit hex stops with alpha.
- `bun run --cwd packages/studio test`: 428 files, 4786 passed.
- `oxlint`, `oxfmt --check`, `tsc --noEmit` clean. `bun install --frozen-lockfile` accepts the lockfile.


miguel-heygen (approved):

miguel-heygen:
@dkryaklin sign the commit to merge it pls

dkryaklin:
> @dkryaklin sign the commit to merge it pls

Done 👍

dkryaklin:
> @dkryaklin sign the commit to merge it pls

Friendly ping

jrusso1020:
Thanks for identifying this and adding regression tests. We'd prefer to avoid a new runtime dependency for color parsing. We prototyped a browser-native approach in #3931—could you adapt that into this PR while keeping the regression coverage?

dkryaklin:
> Thanks for identifying this and adding regression tests. We'd prefer to avoid a new runtime dependency for color parsing. We prototyped a browser-native approach in #3931—could you adapt that into this PR while keeping the regression coverage?

Sure, no problem. I've updated the PR with your approach from #3931 and kept my gradient test

Issue #3583, opened 2026-09-01
Frame-preset fonts (Google Fonts) are never staged as local files — 12/13 presets and media-use have no way to fetch them


julian-pani:
> **Note on provenance:** this report was written by an AI coding agent (Claude) investigating a finding from another agent instance (a "frame worker" sub-agent) during a video build, at the request of a human user who is filing it on the agent's behalf. The user has not independently verified the code-level claims below beyond a light read-through — treat the diagnosis as agent-generated analysis, not a human-confirmed bug report. If it turns out not to be a real bug, it may still be useful as a concrete example of where an agent's reasoning about this codebase went right or wrong — happy to have it closed either way with a note on what it got wrong.

## Summary

Building a video from the `blue-professional` frame-preset, a frame-worker sub-agent correctly followed the `font_family_without_font_face` rule in `skills/hyperframes-core/references/frame-worker-core.md` (*"Only use fonts that ship as files with the project... Never name a font that has no file"*) and substituted a system-font stack, because `frame.md`'s named fonts (Space Grotesk, Inter) had no matching `.woff2`/`.ttf`/`.otf` anywhere in the project, the preset's own directory, or the files staged by `build-frame.mjs`.

Tracing this back, it looks like a structural gap rather than a one-off: most shipped frame-presets name Google Fonts as their default identity typography, but nothing in the toolchain ever fetches those fonts as files unless the *calling project* happens to already have them from an unrelated brand-capture step.

## What I checked

**1. `build-frame.mjs`'s font-staging logic is real and working, but scoped narrowly.**
(`skills/product-launch-video/scripts/build-frame.mjs`, same logic duplicated in `skills/faceless-explainer/scripts/build-frame.mjs` and `skills/pr-to-video/scripts/build-frame.mjs`)

The "stage brand font files + emit @font-face" block only copies files it finds in the *project's* `capture/assets/fonts/` or `assets/fonts/`:

```js
const srcDirs = [
  join(hyperframesDir, "capture/assets/fonts"),
  join(hyperframesDir, "assets/fonts"),
].filter((d) => existsSync(d));
```

[...]

PR #3757, opened 2026-09-08
feat: offline asset ledger and vendor preflight CLI

mvanhorn:
## Summary

Agents and authors routinely pull CDN scripts and remote media into compositions, which silently breaks deterministic and offline renders. This PR adds an agent-readable asset graph and a one-command localization step:

- **`hyperframes ledger [dir] [--json] [--strict-offline]`** — static inventory of every declared asset reference (scripts, stylesheets, fonts, images, audio, video, iframes, text tracks) across all project HTML, each classified `remote | local | data | missing`. No browser, no network. Under `--strict-offline` it exits non-zero while any remote reference remains — the CI gate for deterministic renders.
- **`hyperframes vendor [dir] [--out assets/vendor] [--dry-run] [--strict-offline]`** — downloads each unique remote URL (scheme allowlist: `http`/`https` only; protocol-relative upgraded to `https`), saves it as `<basename>-<sha256-prefix><ext>`, rewrites every HTML reference to a path relative to the declaring file, and records provenance (`url`, `bytes`, `sha256`, `contentType`) in `vendor-manifest.json`. Remote iframes are deliberately never vendored — a live page cannot be made deterministic — so they still fail the strict gate.
- **Core module** `@hyperframes/core/asset-ledger` — the pure extraction/classification engine (linear tag scan + CSS `url()`/`@import`/`@font-face` scanning), reusing `@hyperframes/parsers/asset-resolution` for on-disk resolution. New subpath export synced via `package-subpaths.json`.
- **Lint rule** `remote_script_not_vendored` (info) — nudges toward `hyperframes vendor` when a remote `<script src>` is detected, without breaking the documented CDN quick-start path.
- **Docs** — `reference/cli-ledger` (flags, JSON schema, exit codes) and `guides/offline-deterministic-renders` (workflow guide), both wired into the Mintlify nav.

Scope note: this does **not** implement Google Fonts crawling (#3583) — only declared URLs are inventoried/vendored; font binaries nested inside remote CSS are documented as out of scope.

## Walkthrough video (rendered by HyperFrames itself)

The walkthrough is not a screen recording — it is a HyperFrames HTML composition, storyboarded first and rendered to MP4 by the CLI. Every terminal/JSON panel shows the real captured output of running the new commands on the demo project, and the demo is self-referential: the composition was authored with GSAP on jsDelivr, `hyperframes vendor` downloaded it to `assets/vendor/gsap.min-0274614511.js` and rewrote the `<script src>`, `hyperframes ledger --strict-offline` passed with zero remote references, and **the video was then rendered with the vendored file**.

▶ **Walkthrough MP4 (public):** https://raw.githubusercontent.com/mvanhorn/hyperframes/walkthrough-assets/offline-asset-ledger-walkthrough.mp4

(1920×1080, 30 s @ 30 fps, 2.2 MB — HyperFrames CLI render output)

Render command:

```bash
npx hyperframes render   # → renders/offline-ledger-demo_2026-09-08_01-32-43.mp4 (rendered in 32.9s)
```

Full demo pipeline (all real, exit codes verified):

```bash
npx hyperframes ledger --json            # counts: { total: 1, remote: 1, … } — flags the jsDelivr GSAP script
npx hyperframes vendor                   # ↓ …gsap.min.js → assets/vendor/gsap.min-0274614511.js (71.1 KB, sha256 c174bfce…)
npx hyperframes ledger --strict-offline  # 1 local · 0 remote — exit 0
npx hyperframes lint && npx hyperframes check   # 0 errors; 38/38 WCAG AA text checks pass
npx hyperframes render
```

## Test plan

- `packages/core/src/assets/ledger.test.ts` — 13 tests: URL classification (remote/data/local/skip, unknown schemes, templating placeholders), extraction (script tags, `link rel`/`as`, CSS `@import`/`@font-face`/backgrounds, `srcset`, `poster`, `<source>` parent-kind mapping, entity decoding, comment/script-string immunity), ledger assembly (counts, dedupe, resolver), and on-disk project scanning with file-relative resolution. Fixture HTML uses jsDelivr + local media; no browser required.
- `packages/cli/src/commands/ledger.test.ts` — 5 tests: `--json` output shape, `--strict-offline` exit codes (1 with remotes, 0 when fully local), human output, error paths.
- `packages/cli/src/commands/vendor.test.ts` — 10 tests: scheme allowlist, stable hashed filenames, content-type extension fallback, relative-path rewriting (incl. prefix-safe longest-first replacement), full command flow with mocked `fetch` (download + rewrite + manifest), `--dry-run` no-op guarantee, failed-download exit 1, `--strict-offline` failure on non-vendorable remotes.
- `packages/lint/src/rules/runtimeNetworkFetch.test.ts` — 3 tests: info finding with vendor fixHint, URL dedupe, silence on local/inline scripts.
- Full suites pass: `@hyperframes/cli` (vitest, exit 0), `@hyperframes/core` (exit 0), `@hyperframes/lint` (604 tests, exit 0). `bun run build`, root `bun run lint` (incl. `check:package-subpaths`), typecheck, oxlint/oxfmt, and the fallow audit all pass.
- End-to-end verification against the live jsDelivr CDN is shown in the walkthrough video above.

## AI disclosure

This PR was implemented by an AI agent (Claude Fable 5, running as a Cursor cloud agent), including the code, tests, docs, storyboard, and the HyperFrames composition used for the walkthrough video. A human (mvanhorn) directed scope and requirements.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -825,13 +825,15 @@
             "pages": [
               "reference/html-schema",
               "reference/color-grading",
-              "reference/audio-effects"
+              "reference/audio-effects",
+              "reference/cli-ledger"
             ]
           },
           {
             "group": "Rendering paths",
             "pages": [
               "guides/rendering",
+              "guides/offline-deterministic-renders",
               "deploy/overview",
               "deploy/cloud",
               "guides/deploy"
--- a/docs/guides/offline-deterministic-renders.mdx
+++ b/docs/guides/offline-deterministic-renders.mdx
@@ -1,0 +1,100 @@
+---
+title: "Offline & deterministic renders"
+sidebarTitle: "Offline renders"
+description: "Make a composition render byte-identically with the network unplugged: inventory assets, vendor remotes, and gate CI on zero network references."
+---
+
+A HyperFrames render is a function of the composition and its assets — nothing
+else. Two things quietly break that:
+
+- **Remote assets.** A CDN `<script>`, a hotlinked image, or a Google Fonts
+  stylesheet means render output depends on a third-party server being up,
+  fast, and serving the same bytes as yesterday.
+- **Runtime nondeterminism.** `Date.now()`, unseeded `Math.random()`, and
+  render-time `fetch()` calls produce different frames on different runs.
+
+The composition contract already bans the second category. This guide is about
+the first: getting to — and staying at — **zero network references**.
+
+## The workflow
+
+<Steps>
+  <Step title="Inventory: hyperframes ledger">
+    ```bash
+    npx hyperframes ledger --json
+    ```
+
+    The ledger scans every HTML file and classifies each declared asset as
+    `remote`, `local`, `data`, or `missing` — scripts, stylesheets, fonts,
+    images, audio, video, iframes, and text tracks, including URLs inside
+    `<style>` blocks and inline styles. It is purely static (no browser, no
+    network), so it is safe anywhere, CI included.
+  </Step>
+  <Step title="Localize: hyperframes vendor">
+    ```bash
+    npx hyperframes vendor
+    ```
+
+    Every remote `http(s)` asset is downloaded into `assets/vendor/`, every
+    reference is rewritten to a relative path, and `vendor-manifest.json`
+    records the source URL, size, and SHA-256 of each download. Your
+    composition now carries its own dependencies.
+  </Step>
+  <Step title="Enforce: --strict-offline">
+    ```bash
+    npx hyperframes ledger --strict-offline
+    ```
+
+    Exits non-zero while any remote reference remains. Put it in CI right
+    before `hyperframes render` so a stray CDN reference fails the pipeline
+    instead of the render.
+  </Step>
+</Steps>
+
+## Example
+
+A fresh composition that pulls GSAP from jsDelivr:
+
+```bash
+$ npx hyperframes ledger
+◆  Asset ledger for my-video
+   1 HTML file scanned, 6 asset references
+
+   4 local   1 remote   1 data   0 missing
+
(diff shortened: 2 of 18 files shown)


mvanhorn:
Pushed fixes for the CodeQL findings (0a2710a):

**Double-unescape in `decodeHtmlEntities` (`packages/core/src/assets/ledger.ts`)** — replaced the chained `.replace()` calls with a single-pass decoder driven by one regex and an entity map. Each entity now decodes exactly once, so a decoded `&` can never recombine with following text into a second entity: `&amp;lt;` decodes to `&lt;`, never `<`. Added unit tests pinning the single-decode contract and that unknown entities are left untouched.

**ReDoS risk in `stripCssComments`** — replaced the `/\/\*[\s\S]*?\*\//g` regex with a linear O(n) `indexOf`-based scanner (same shape as the existing `maskHtmlComments`), still length-preserving. Added a test that a 50k-character hostile comment run completes quickly and comments still mask correctly.

**Network→file in `vendor.ts`** — the download-to-disk flow is this command's entire purpose (localize remote assets for offline renders), so the flagged flow is intentional; it is now explicitly documented at the write site and hardened:
- http(s)-only scheme allowlist + URL validation retained (`toFetchableUrl`)
- new `resolveVendorTarget` traversal guard: every write is resolved and verified to stay under the vendor directory before `writeFileSync`
- `--out` itself is now verified to resolve inside the project root
- new per-download size cap (`--max-bytes`, default 100 MB), enforced against both `content-length` and the actual body

No repo-idiomatic CodeQL suppression exists, so no suppression comment was added — just real guards plus the intent comment.

Tests: `packages/core` ledger suite (16 passed) and `packages/cli` vendor + ledger suites (19 passed); changed files pass oxlint/oxfmt.

PR #3762, opened 2026-09-08
fix(engine): flag single-keyframe videos as sparse instead of skipping them

### LINKED ISSUE #3460: Sparse-keyframe check silently skips single-keyframe videos — the worst case goes unreported (0.8.10)

Abhishek-kumarsingh:
## Summary

Fixes #3460. `analyzeKeyframeIntervals` returned `isProblematic: false` whenever a video had fewer than two keyframes, treating a single-GOP video the same as a still image. But a video with exactly one keyframe is the *worst* case for the failure mode the check exists to catch: every seek past 0 lands inside a single GOP spanning the whole file. A 10s single-GOP video went unreported while a video with a 5s max interval triggered the warning.

## Fix

When exactly one keyframe is found, the effective interval is now the stream duration (fetched via the already-available `extractMediaMetadata`), not zero — matching the fix suggested in the issue. Still images and single-frame assets keep their current (non-problematic) behavior, since their duration is at or below the 2s threshold. Zero-keyframe handling is unchanged.

## Test plan

- [x] Added regression tests to `packages/engine/src/utils/ffprobe.test.ts` covering: a long-duration single keyframe (now flagged problematic), a short single keyframe (still not flagged), and zero keyframes (unchanged, not flagged) — confirmed the long-duration case failed before the fix
- [x] Full `@hyperframes/engine` suite passes (1641/1641)
- [x] `oxlint` and `oxfmt --check` clean on changed files

Note: the issue also raises a "side note" about whether the warning targets `render`'s decode path at all, based on the reporter's own frame-accuracy measurements — that's a separate scoping question I've left alone; this PR only fixes the single-keyframe detection gap.

### DIFF
--- a/packages/engine/src/utils/ffprobe.test.ts
+++ b/packages/engine/src/utils/ffprobe.test.ts
@@ -931,6 +931,91 @@ describe("ffprobe option separator", () => {
   });
 });
 
+describe("analyzeKeyframeIntervals — single-keyframe videos", () => {
+  afterEach(() => {
+    vi.resetModules();
+    vi.doUnmock("child_process");
+  });
+
+  it("treats a single keyframe spanning a long duration as problematic, not as a skip", async () => {
+    const { spawn } = createSpawnSpy([
+      // keyframe timestamp probe: one keyframe at t=0
+      { kind: "exit", code: 0, stdout: "0.000000\n" },
+      // duration probe (extractMediaMetadata), used to size the effective interval
+      {
+        kind: "exit",
+        code: 0,
+        stdout: JSON.stringify({
+          streams: [
+            {
+              codec_type: "video",
+              codec_name: "h264",
+              width: 640,
+              height: 360,
+              r_frame_rate: "30/1",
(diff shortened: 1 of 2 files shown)


jrusso1020:
Thanks for the fix. Closing alongside #3573 after current-main validation: the missing classification is real, but its only consumer is a render warning whose claimed failure was not reproduced; actual mid-GOP extraction matched FFmpeg ground truth. This implementation also uses container duration, which can falsely flag a 1s video with a 10s audio track. Keeping #3460 open for accurate diagnostic scope and video-stream duration handling.

PR #3797, opened 2026-09-09
fix(lint): flag seek-unsafe CSS transitions

### LINKED ISSUE #3493: lint: flag CSS transition (seek-unsafe; adapter only covers @keyframes)

Ftgn-dpA:
## What

Add a `css_transition_used` lint error for seek-unsafe CSS transition
declarations in `<style>` blocks and inline styles.

The rule covers standard and `-webkit-` transition properties, while allowing
explicit `transition: none` / `transition-property: none` declarations and
ignoring custom properties.

## Why

CSS transitions run on the browser clock rather than the seekable Hyperframes
timeline. Class or attribute changes can therefore restart a transition on
fresh render workers, causing parallel renders and snapshots to produce
different frames.

Fixes #3493.

## How

Use PostCSS to inspect actual declarations instead of matching raw CSS text.
Unsafe declarations report an error with guidance to keep state changes in
classes or attributes and move the visual animation onto the paused GSAP
timeline.

## Test plan

- [ ] Manual testing performed — not required for this parser-only rule
- [ ] Documentation updated — not applicable

- `bun run --filter @hyperframes/lint test` — 648 passed
- `bun run --filter @hyperframes/lint typecheck`
- `bunx oxlint packages/lint/src/rules/core.ts packages/lint/src/rules/core.test.ts`
- `bunx oxfmt --check packages/lint/src/rules/core.ts packages/lint/src/rules/core.test.ts`


miguel-heygen (approved):

miguel-heygen (commented):
Thanks for adding this — catching a wall-clock `transition` next to a paused GSAP timeline is a good rule to have.

One issue with the current exemption: it only allows the literal value `none`, so `transition-duration: 0s` (and `!important` variants) gets flagged as an error too. That's the pattern used for a reduced-motion / render-lock reset, and it's already used that way in about 20 files in this repo's own registry — with the current rule there's no way to keep using it without triggering an error. Could the rule treat a zero duration the same as `none`? That should unblock those existing files without weakening the check for anything that actually animates on the wall clock.

Issue #408, opened 2026-04-22
Codex install command fails with "invalid marketplace file"


yfrbl:
## Summary

The Codex install command shown in the README does not work with the current Codex CLI.

README currently suggests:

```bash
codex plugin marketplace add heygen-com/hyperframes --sparse .codex-plugin --sparse skills --sparse assets
```

Running that against `heygen-com/hyperframes` fails even though the repository contains `.codex-plugin/plugin.json`.

## Steps to reproduce

1. Install/update Codex CLI.
2. Run:
   ```bash
   codex plugin marketplace add heygen-com/hyperframes --sparse .codex-plugin --sparse skills --sparse assets
   ```

## Actual result

Codex exits with:

```text
Error: invalid marketplace file `/Users/.../.codex/.tmp/marketplaces/.staging/...`: marketplace root does not contain a supported manifest
```

## Expected result

Either:

- the plugin installs successfully using the documented command, or
- the README uses the correct installation flow for the current Codex CLI.

## Notes

- Tested with `codex-cli 0.122.0`.
- The repo does contain a plugin manifest at `.codex-plugin/plugin.json`:
  https://github.com/heygen-com/hyperframes/blob/main/.codex-plugin/plugin.json
- This looks like a mismatch between:
  - a repository that exposes a Codex plugin via `.codex-plugin/plugin.json`, and
  - `codex plugin marketplace add`, which appears to expect a marketplace root manifest rather than a single plugin repo.

If that command is no longer valid for current Codex versions, updating the README with the supported install path would likely fix the confusion.

jrusso1020:
@yfrbl you found that quick! We are working with the OpenAI team to get this into the marketplace it is not ready quite yet! 

Will update this issue once it’s in the marketplace , we needed to setup the plugin first though

jrusso1020:
this should be fixed now I believe @yfrbl

datasciencepocket-eng:
Nopes, this isn't fixed yet

PR #3798, opened 2026-09-09
fix(producer): isolate assembly scratch directories

jrusso1020:
Distributed assembly used a predictable `<output>.assemble-work` directory and recursively deleted anything already there. Create a private `mkdtemp` sibling for each invocation, then retain the existing finally cleanup for that invocation's directory. This addresses CodeQL #408's temporary concat-list path and prevents assembly from removing unrelated data at the old scratch path.

Final output paths, stream-copy/CFR options, audio normalization, provenance, and PNG assembly remain unchanged. Existing directories at the former scratch path are now preserved. The output parent remains a trusted caller-selected directory.

Validation: real FFmpeg assembly tests cover output format/frame count/rate/audio and successful/failing scratch cleanup; POSIX staging permissions are checked as 0700. All 13 assembly tests pass (66 assertions), the two security regressions fail against the baseline, and producer typecheck plus oxlint/oxfmt pass.

### DIFF
--- a/packages/producer/src/services/distributed/assemble.test.ts
+++ b/packages/producer/src/services/distributed/assemble.test.ts
@@ -16,7 +16,16 @@
 
 import { spawnSync } from "node:child_process";
 import { afterAll, beforeAll, describe, expect, it } from "bun:test";
-import { existsSync, mkdirSync, mkdtempSync, readdirSync, rmSync, writeFileSync } from "node:fs";
+import {
+  existsSync,
+  mkdirSync,
+  mkdtempSync,
+  readdirSync,
+  readFileSync,
+  rmSync,
+  statSync,
+  writeFileSync,
+} from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import {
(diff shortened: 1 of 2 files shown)


miguel-heygen (approved):
Exact-head security/lifecycle review at `973db1c5c48f403167f2b8b988aa0d982415eaa3`.

- `packages/producer/src/services/distributed/assemble.ts:153-161` creates the output parent first, then atomically allocates a private sibling with `mkdtempSync`. Production no longer names, probes, deletes, or recreates the legacy `<output>.assemble-work` path, so pre-existing data there is outside this invocation's ownership.
- The existing `try/finally` still encloses every container-format operation after acquisition: single remux, concat-list write/copy, optional CFR, audio normalization, mux, and faststart. Success, FFmpeg failure, typed CFR rejection, audio/mux failure, faststart failure, and abort all reach cleanup of exactly the minted directory. Cleanup failure remains warning-only and cannot replace the primary render result/error.
- PNG assembly returns before scratch acquisition and is byte-unchanged. Final `outputPath`, `-c copy`/CFR args, normalized-audio and provenance paths are unchanged; only scratch children receive the new parent. Concurrent invocations get distinct scratch directories while the disclosed final-output/trusted-parent overwrite boundary remains the same.
- `assemble.test.ts:254-280,326-356` observes the private directory while real FFmpeg is running, checks POSIX `0700`, proves the old sentinel survives, and proves both success and concat failure remove only the invocation directory. Both controls fail on the baseline mechanism. Exact-head Producer unit/integration, Build, Typecheck, runtime contract, lint/format, and JavaScript CodeQL are green; the PR scan has no alerts.

**Fallow reconciliation:** the audit reports 3 complexity findings and 4 clone groups, all `introduced:false`. The complexity is the pre-existing `assemble`/`mergePngFrameDirs` shape; this patch removes an existence/delete branch. All four clone groups are pre-existing FFmpeg test scaffolds, and neither new security regression is a clone source. No suppression or config change was added, so I have no concern with the seven inherited findings for this patch.

No blocking or follow-up findings. Required aggregate Test and Windows render are still running and remain merge gates.

**Verdict:** APPROVE
**Reasoning:** Private atomic scratch allocation removes ownership of the predictable path, existing cleanup covers every acquired-resource exit, and the media/final-output contracts are unchanged and regression-tested.

— Magi

PR #351, opened 2026-04-20
feat(cli): add --lang and auto-infer phonemizer locale from voice prefix

jrusso1020:
## What

Fixes multilingual TTS output in `hyperframes tts`. Adds:

- `--lang <code>` flag for explicit phonemizer locale override
- Auto-detection of the phonemizer locale from the Kokoro voice ID prefix when `--lang` is omitted (the actual bug fix)
- Graceful fallback when the installed `kokoro-onnx` version predates the `lang` kwarg

Closes #349.

## Why

`hyperframes tts` was calling Kokoro's `model.create(text, voice=, speed=)` with no language argument. Kokoro's text frontend defaults to `en-us` regardless of voice, so picking a non-English voice like `ef_dora` (Spanish) or `jf_alpha` (Japanese) and feeding it native-language text produced English-phonemized output — every non-English voice was effectively broken.

Kokoro's own voice ID convention encodes the language in the first letter (`a`=American, `b`=British, `e`=Spanish, `f`=French, `h`=Hindi, `i`=Italian, `j`=Japanese, `p`=Brazilian Portuguese, `z`=Mandarin), so the default can be derived mechanically from the voice. Explicit `--lang` is kept for cases where users intentionally want a mismatch (stylized accent, specific locale like `en-gb` vs `en-us`).

## How

**`packages/cli/src/tts/manager.ts`**
- New `SUPPORTED_LANGS` readonly tuple of the nine valid phonemizer codes
- New `inferLangFromVoiceId(voiceId)` helper mapping voice prefixes → locales
- New `isSupportedLang()` type guard
- Attach `defaultLang` to every `VoiceInfo`
- Expand `BUNDLED_VOICES` with `ef_dora`, `ff_siwis`, `jf_alpha`, `zf_xiaobei` so `--list` surfaces multilingual options (it was English-only before)

**`packages/cli/src/tts/synthesize.ts`**
- Accept optional `lang` in `SynthesizeOptions`; forward as `argv[7]` to Python
- Python worker introspects `Kokoro.create`'s signature and only passes `lang=` when the installed kokoro-onnx version supports it — older installs keep working with their default (English) phonemization
- Returned metadata adds `lang` and `langApplied` so the caller can detect silent no-ops
- Rename cached script to `synth-v2.py` so existing users automatically pick up the new script (the old cached copy doesn't know about `argv[7]`)

**`packages/cli/src/commands/tts.ts`**
- Add `--lang, -l` flag with validation against `SUPPORTED_LANGS`
[...]

PR #347, opened 2026-04-20
feat: add design system , theme.json, reusable components, docs

Hamzaa6296:
## Summary
Closes / contributes to #340

Adds a foundational Design System layer to Hyperframes so creators can maintain
consistent brand identity across multiple video compositions without repeating styles.

## Changes
- `templates/theme.json` , global design tokens (colors, typography, spacing)
- `templates/components/LowerThird.html` , reusable branded lower-third overlay
- `templates/components/Outro.html` , reusable branded end card
- `docs/design-system.md` , documentation for the design system

## How it works
- Theme values are declared once in `theme.json`
- CSS variables (`--hf-color-accent`, `--hf-font-family`, etc.) are injected at render
- Components in `templates/components/` can be copied into any composition
- Style inheritance: theme.json → composition → inline

## Testing
1. `npx hyperframes init my-video`
2. Copy `theme.json` and a component into the project
3. Run `npx hyperframes preview` , tokens should resolve correctly

PR #344, opened 2026-04-20
docs: add skillx quick-start to README and introduction

linisme:
## Summary
- Add a "Try it with skillx" block to `README.md` and `docs/introduction.mdx` so readers can try the `hyperframes` and `website-to-hyperframes` skills without installing anything.
- Block is wrapped in idempotency markers (`skillx:begin/end:setup-skillx`) so a later `setup-skillx` run can update it in place.
- Uses `--auto-approve` to match the task-based, one-shot flow these two skills drive end-to-end.

## Test plan
- [x] Render `README.md` on GitHub — confirm the shields.io badge, fenced code block, and italic disclaimer display correctly.
- [x] Build the docs site and confirm `introduction.mdx` renders cleanly (JSX comments hidden, `<Tip>` + code block layout intact).
- [x] Run `skillx run --auto-approve https://github.com/heygen-com/hyperframes/tree/main/skills/hyperframes "test"` in a scratch dir to verify the URL resolves and the skill is fetched.

Issue #343, opened 2026-04-20
build: use proper topological ordering instead of sequential cli-last hack


miguel-heygen:
## Context

In #342 the root build script was changed from:

```json
"build": "bun run --filter '*' build"
```

to:

```json
"build": "bun run --filter '!@hyperframes/cli' build && bun run --filter @hyperframes/cli build"
```

This was a targeted fix for two races the Windows CI surfaced:

1. `@hyperframes/cli`'s `build:runtime` reading `packages/core/dist/hyperframe.manifest.json` before `@hyperframes/core`'s build wrote it.
2. `@hyperframes/cli`'s `build:copy` reading `packages/studio/dist/` before `@hyperframes/studio`'s vite build populated it.

Sequencing CLI last fixes both, because CLI is the only package that depends on every other package's dist output.

## Why this is a hack

Bun's `--filter` does not enforce topological build-order for run scripts; adding `@hyperframes/studio` as a `workspace:*` dep of `@hyperframes/cli` doesn't help. So we sequence CLI by name. If another workspace-dep relationship is introduced between two currently-parallel packages that also needs build-output ordering, the same class of race returns silently.

## What we should do

- Either move to a build tool that respects workspace-dep topology (turbo, nx, or bun with proper `--filter` topology if/when that ships).
- Or declare the build graph explicitly (e.g., a root-level orchestration script that knows "build X, then Y, then Z in parallel, then W").

## Related

- #336 — the FFmpeg fix that motivated verifying Windows in the first place
- #342 — introduced the current sequential workaround

PR #3799, opened 2026-09-09
fix(producer): isolate regression harness temporary roots

jrusso1020:
The regression harness deleted and recreated a predictable system-temp directory for every suite. That shared root flowed into planning, chunk output, compiled HTML and capture diagnostics, producing twelve CodeQL alerts. Allocate a private root per suite invocation and preserve the existing finally cleanup, failure-report ordering and --keep-temp output/event paths.

Two mocked stage tests also supplied literal shared temporary paths. Give them private fixture roots with cleanup; the capture frames child still starts absent, preserving the fixture's disk-preflight behavior. No downstream production write, debug-directory, plan/chunk-output or media contract changes.

Alert/source map (independently traced by Magi):
- #727, #343, #344, #351–#354: harness root only.
- #339–#340: harness root plus captureStreamingStage fixture.
- #345–#347: harness root plus probeStage fixture.

Validation: 24 harness/routing tests, 12 capture tests and 41 probe tests pass. The new isolation regression fails against the old allocation mechanism. Producer typecheck and oxlint/oxfmt pass. Existing harness cleanup and keep-temp branches remain unchanged; full render regressions and CodeQL must pass before merge. No alert dismissals or query suppression.

### DIFF
--- a/packages/producer/src/regression-harness-parse.test.ts
+++ b/packages/producer/src/regression-harness-parse.test.ts
@@ -6,7 +6,10 @@
 // must move together.
 
 import { describe, expect, it } from "bun:test";
-import { parseArgs } from "./regression-harness.js";
+import { mkdtempSync, mkdirSync, readFileSync, rmSync, statSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { createRegressionTempRoot, parseArgs } from "./regression-harness.js";
 
 // parseArgs reads from index 2 onwards (node + script name are argv[0..1]).
 const withProgram = (rest: string[]): string[] => ["node", "regression-harness.ts", ...rest];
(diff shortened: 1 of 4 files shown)


miguel-heygen (approved):
Exact-head review at `fdd759b51eea89e2275a04c9debe9201cbe65510`, following the 12-alert source audit.

- `packages/producer/src/regression-harness.ts:966-979` makes `createRegressionTempRoot` the single harness allocator. Every suite ID comes from a direct fixture-directory entry, so it is a safe prefix component; `mkdtempSync` gives concurrent runs distinct private roots without touching `/tmp/hyperframes-tests/<id>`.
- The existing `runTestSuite` lifecycle is unchanged after acquisition: failure details still save first, non-`keepTemp` removes exactly the minted root, and `keepTemp` retains/logs that same random path and returns its rendered output path. Update, unsupported-mode early return, compile failure, render failure, and normal completion all still reach the same `finally`.
- `captureStreamingStage.test.ts:1-12,198-199,494-497,606-609` moves every #339/#340 secondary `framesDir` leg under one private test root with `afterAll` cleanup. The `frames` child intentionally remains absent until production code would create it; this preserves the minimal fixture's disk-preflight contract rather than preconditioning the test.
- `probeStage.test.ts:1-10,245-246` similarly moves the #345-#347 secondary `workDir` leg under a private root. Its mocked `writeCompiledArtifacts` boundary is unchanged. The other fixed test literals are not sources in the mapped alerts and remain mock-only values.
- This closes every mapped source leg: #727/#343/#344/#351-#354 through the common root; #339/#340 through common + capture fixture; #345-#347 through common + probe fixture. JavaScript CodeQL completed successfully with zero PR alerts, so no new source surfaced. No downstream plan/chunk scratch, persistent output, debug, media, or query configuration changed.
- Exact CI confirms 24 harness/routing, 12 capture, and 41 probe tests pass; Producer unit, Build, Typecheck, runtime contract, lint/format, Windows render, and CodeQL are green. The isolation control proves repeated same-suite roots differ, are POSIX `0700`, preserve the legacy sentinel, and deleting one cannot affect the other; it fails against the baseline allocator.

**Fallow reconciliation:** exactly four clone groups, all `introduced:false`, all pre-existing media-reconciliation scaffolds inside `probeStage.test.ts` (70 lines total). None includes the new private-root setup or isolation control; there are zero complexity/dead-code findings and no suppression/config change. No concern for this patch.

No blocking or follow-up findings. Required aggregate Test is still running and remains a merge gate.

**Verdict:** APPROVE
**Reasoning:** One private harness allocator plus two private mock-fixture roots cover all 12 audited source legs without changing persistent downstream contracts, and exact-head CodeQL confirms the source closure.

— Magi

PR #3804, opened 2026-09-09
fix(engine): reject malformed JPEG input before encoding

vanceingalls:
Malformed JPEG capture data currently reaches FFmpeg's image2pipe input and surfaces as `dqt: invalid precision` followed by secondary stream/codec errors. Validate JPEG marker lengths and quantization-table descriptors before writing each frame, and report the first rejected input frame index. Empty and truncated headers get distinct diagnostics. PNG and raw HDR inputs retain their existing paths.

Related: [PRINFRA-353](https://linear.app/heygen/issue/PRINFRA-353). Diagnosis confidence: 85% that the reported DQT signature requires malformed JPEG data; the original source of that corruption is still unconfirmed. This is a fail-fast diagnostic improvement, not a claim to repair Chrome's frame generation. It validates headers before the first scan, not entropy-coded pixels or later progressive scans.

Validation: 64 JPEG/streaming-encoder tests passed; engine typecheck, scoped oxlint/oxfmt, and fallow gate passed. A real FFmpeg smoke run encoded a valid 64x64 JPEG and rejected the reproduced precision-2 DQT frame at input index 1 before it entered the pipe; the remaining valid MP4 finalized successfully. Lifecycle/back-pressure tests now use a real JPEG fixture.


miguel-heygen (approved):
The validation boundary is correctly placed before `stdin.write`: the frame copy is stable, JPEG-only gating leaves PNG/raw paths untouched, and `inputFrameIndex` advances only after a successful write attempt (`packages/engine/src/services/streamingEncoder.ts:538-567`).

`jpegInputError` bounds every segment read, stops at the first SOS instead of scanning entropy-coded bytes as markers, handles fill bytes, and validates every DQT descriptor/table within a segment (`packages/engine/src/utils/jpegInput.ts:2-49`). The real FFmpeg JPEG fixture also repairs the older lifecycle tests so they continue exercising encoder behavior rather than newly invalid dummy bytes.

Scope is stated honestly: this identifies malformed headers and the first rejected input frame; FFmpeg still owns entropy/progressive-scan validation and the upstream corruption cause remains open. All applicable CI, CodeQL, Windows, and regression checks are green at `b22afd6b`.

**Verdict:** APPROVE
**Reasoning:** Malformed JPEGs now fail before pipe admission with bounded parsing and precise indexing, while all non-JPEG and streaming lifecycle contracts remain unchanged.

— Magi

Issue #3482, opened 2026-08-25
Fast-capture: fallback-ratio guard for #3423 x #3429 seam (100%-fallback renders pass artifact validation)


miga-heygen:
## Context

PR #3480 (branch `fix/draw-element-canvas-fallback`, closes #3423) added a
per-frame screenshot fallback for the drawElement fast-capture path: when
`drawElementImage` throws a recoverable error (canvas not initialized, or
the native "No cached paint record" InvalidStateError), the affected frame
falls back to `Page.captureScreenshot` instead of hard-failing the render.

Review feedback on #3480 (from tai) flagged a seam with PR #3429's artifact
validation (`ArtifactTransaction.validate()` /
`packages/producer/src/services/render/artifactTransaction.ts`):

> If a composition is 100% broken and every frame falls back to screenshot
> capture, the render passes #3429's artifact validation (correct frame
> count, correct duration) but ships N wrong-pixel frames as "complete."

To be precise about the actual risk: the screenshot fallback itself produces
*correct* pixels (it's the platform's normal, well-tested capture path) —
the real risk is a **silent, total loss of the fast-capture speedup** (and by
extension a masked persistent bug, e.g. a canvas-injection regression)
going completely unnoticed because the render still "succeeds" with the
right frame count/duration. Today nothing distinguishes "recovered a
handful of edge-case frames" from "drawElement never engaged for the whole
render" in either telemetry or artifact validation.

## What #3480 does today

`getCapturePerfSummary()` in `packages/engine/src/services/frameCapture.ts`
now emits a `console.warn` when `deNcprFallbacks / frames` exceeds 50%, so
the condition is at least visible in render logs / telemetry ingestion.
This is a diagnostic only — it does not fail or flag the render artifact.

## Follow-up needed (pick one, or combine)

1. **Fallback-ratio circuit breaker** (engine or producer-side): if the
   realized `deNcprFallbacks / frames` ratio exceeds some threshold (e.g.
   50%), fail the render explicitly instead of shipping it as "complete" —
   forcing a re-render with `forceScreenshot` (which the orchestrator already
   knows how to do for `DrawElementVerificationError`) rather than silently
   accepting a render that never got the fast-capture speedup.
2. **`deNcprFallbacks`/`frames` guard on the #3429 side**: thread the ratio
   (already available via `CapturePerfSummary.deNcprFallbacks` /
   `RenderPerfSummary`) into `ArtifactTransaction.validate()` or the
[...]

PR #3805, opened 2026-09-09
fix(render): retry drawElement failures on a fresh screenshot page

vanceingalls:
After drawElement injects its canvas, a same-page screenshot can capture the previous canvas bitmap instead of the newly sought composition frame. Recoverable paint-record/canvas errors, tiny-frame guards, and explicit boundary screenshot fallbacks currently use that unsafe surface.

Reject those captures with a structural `DrawElementCaptureError` and route the whole attempt through the existing fresh screenshot-session retry. Cover serial, pipelined, and batch paths; consume abandoned batch encode rejections during cleanup. Disk capture propagates the error even when frame files appear complete, clears the untrusted attempt, and forces screenshot capture. Keep cancellation/encoder-interruption exclusions and report `capture_error` separately from PSNR/blank verification failures.

Related: [PRINFRA-600](https://linear.app/heygen/issue/PRINFRA-600).

---

## Merge gate: met — and the defect is now confirmed, not inferred

The draft gate asked for real Chrome pixel comparisons across serial, worker and batch capture, including a hard scene cut and an induced missing paint record. Browser launch was sandbox-blocked on the investigation host; it works on macOS/hardware-GPU, so the comparisons ran there. **Full method, tables and caveats in [this comment](https://github.com/heygen-com/hyperframes/pull/3805#issuecomment-5614521179).**

Headline: with the NCPR error injected at the first frame after a hard cut, **`main` emits the previous frame's canvas bitmap** — its frame 60 scores 77.2 dB against baseline frame *59* and only 6.2 dB against its own correct frame 60 (the batch path corrupts all four frames of the batch, 13.2 dB). **This branch returns bit-identical pixels on all three paths.** Without injection, all three paths sit at 67.1 dB min with 0/180 frames below the 30 dB damage criterion.

So the original 75% diagnosis confidence understated one half and not the other: the stale-surface recovery defect is now **reproduced and pixel-confirmed**, while the original Windows compositor trigger is still **not** reproduced. This PR fixes the former and does not claim the latter.

### Costs and risks this accepts

- **A full retry costs ~+63% wall** (13.5s vs 8.3s on a 180-frame comp) versus the per-frame fallback it replaces. That is the deliberate trade: verified fresh-page recovery over a cheap read of an untrusted surface.
- Legitimately simple tiny JPEGs can still trip the tiny-frame guard and pay that retry.
- Evidence is macOS / hardware-GPU only.

## Validation

- New `frameCapture-freshFallback.test.ts`: 12 tests, covering that no same-page screenshot pixels are returned, pending batch prefixes reject safely, explicit boundary fallback is fresh-page-only, and complete disk files cannot mask an untrusted capture.
- Post-rebase re-run: 216 engine capture-suite tests, 221 producer orchestration/plan tests, engine + producer typechecks, scoped lint/format, tracked/large-file and fallow gates.
- Real-Chrome: zero unhandled promise rejections on the batch path, which exercises the abandoned-batch-encode drain outside the mocks.

Rebased onto `main` (the `renderOrchestrator.test.ts` conflict was an import union). `4c7fbde5` addresses @miguel-heygen's diagnostics finding: `captureFrameErrorDiagnostics` now runs **before** the recoverable-error wrapper in `captureFrameToBufferPipelined`, so NCPR/canvas failures emit the frame-error bundle like the serial path. It is bounded to one bundle per attempt and self-catches, so a dead page cannot mask the structural error the retry depends on. Covered by a test verified by mutation — restoring the old ordering fails it.

### DIFF
--- a/packages/engine/src/index.ts
+++ b/packages/engine/src/index.ts
@@ -416,3 +416,8 @@ export {
   PROVENANCE_VERSION,
   type RenderProvenance,
 } from "./utils/renderProvenance.js";
+
+export {
+  DrawElementCaptureError,
+  isDrawElementCaptureError,
+} from "./services/drawElementCaptureError.js";
--- a/packages/engine/src/services/drawElementCaptureError.ts
+++ b/packages/engine/src/services/drawElementCaptureError.ts
@@ -1,0 +1,25 @@
+/** A drawElement page cannot supply trusted pixels; retry on a fresh screenshot page. */
+export class DrawElementCaptureError extends Error {
+  readonly frameIndex: number;
+
+  constructor(frameIndex: number, reason: string, cause?: unknown) {
+    super(
+      `drawElement capture failed at frame ${frameIndex}: ${reason}; fresh screenshot capture required`,
+      { cause },
+    );
+    this.name = "DrawElementCaptureError";
+    // Consumed structurally across package/worker error boundaries.
+    Object.defineProperty(this, "deCaptureFailure", { value: true, enumerable: true });
+    this.frameIndex = frameIndex;
+  }
+}
+
+/** Structural matching survives engine/producer copies and worker error wrappers. */
+export function isDrawElementCaptureError(error: unknown): boolean {
+  let current = error;
+  for (let depth = 0; depth < 5 && typeof current === "object" && current !== null; depth++) {
+    if ("deCaptureFailure" in current && current.deCaptureFailure === true) return true;
+    current = "cause" in current ? current.cause : undefined;
+  }
+  return false;
+}
--- a/packages/engine/src/services/frameCapture-freshFallback.test.ts
+++ b/packages/engine/src/services/frameCapture-freshFallback.test.ts
@@ -1,0 +1,149 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import {
+  captureFrameToBuffer,
+  captureFrameToBufferPipelined,
+  captureFramesBatchPipelined,
+  createCaptureSession,
+} from "./frameCapture.js";
+import {
+  captureDrawElementFrame,
+  produceDrawElementFrame,
+  produceDrawElementFrameBatch,
+  DE_CANVAS_NOT_INITIALIZED_CODE,
+} from "./drawElementService.js";
+import { pageScreenshotCapture } from "./screenshotService.js";
+import { DrawElementCaptureError, isDrawElementCaptureError } from "./drawElementCaptureError.js";
+
+vi.mock("./browserManager.js", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("./browserManager.js")>()),
+  resolveBrowserGpuMode: async () => "hardware",
+  acquireBrowser: async () => ({
+    captureMode: "screenshot",
+    browser: {
+      version: async () => "Chrome/150.0.0.0",
+      newPage: async () => ({
+        evaluateOnNewDocument: async () => {},
+        setViewport: async () => {},
+        evaluate: async () => undefined,
+        screenshot: async () => {
+          throw new Error("diagnostic screenshot unavailable");
+        },
+      }),
+    },
+  }),
+}));
+vi.mock("./drawElementService.js", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("./drawElementService.js")>()),
+  captureDrawElementFrame: vi.fn(),
+  produceDrawElementFrame: vi.fn(),
+  produceDrawElementFrameBatch: vi.fn(),
+}));
+vi.mock("./screenshotService.js", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("./screenshotService.js")>()),
+  pageScreenshotCapture: vi.fn(),
+}));
+
+const dirs: string[] = [];
+async function session() {
+  const dir = mkdtempSync(join(tmpdir(), "hf-fresh-fallback-"));
+  dirs.push(dir);
+  const result = await createCaptureSession(
+    "http://localhost:0",
+    dir,
+    {
+      width: 1920,
+      height: 1080,
+      fps: { num: 30, den: 1 },
+      format: "jpeg",
+    },
+    null,
+    { forceScreenshot: true },
+  );
+  result.isInitialized = true;
+  result.captureMode = "drawelement";
+  return result;
+}
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  vi.stubEnv("HF_FORCE_DRAWELEMENT", "0");
+  vi.stubEnv("HF_FAST_CAPTURE_BOUNDARY_SS", "false");
+});
+afterEach(() => {
+  vi.unstubAllEnvs();
+  for (const dir of dirs.splice(0)) rmSync(dir, { recursive: true, force: true });
+});
+
+describe("fresh screenshot fallback", () => {
+  for (const [name, capture, produce] of [
+    ["serial", captureFrameToBuffer, captureDrawElementFrame],
+    ["pipelined", captureFrameToBufferPipelined, produceDrawElementFrame],
+  ] as const) {
+    it.each(["No cached paint record for element", DE_CANVAS_NOT_INITIALIZED_CODE])(
+      `${name} rejects untrusted capture instead of returning the injected canvas`,
+      async (message) => {
+        vi.mocked(produce).mockRejectedValueOnce(new Error(message));
+        const s = await session();
+        await expect(capture(s, 7, 7 / 30)).rejects.toBeInstanceOf(DrawElementCaptureError);
+        expect(pageScreenshotCapture).not.toHaveBeenCalled();
+        expect(s.deNcprFallbacks).toBe(1);
+      },
+    );
+    it(`${name} routes explicit boundary screenshots through a fresh page`, async () => {
+      vi.stubEnv("HF_FAST_CAPTURE_BOUNDARY_SS", "true");
+      const s = await session();
+      s.clipBoundaryFrames = new Set([7]);
+      await expect(capture(s, 7, 7 / 30)).rejects.toThrow("boundary screenshot requested");
+      expect(pageScreenshotCapture).not.toHaveBeenCalled();
+      expect(produce).not.toHaveBeenCalled();
+    });
+  }
+
+  it("rejects a suspect tiny frame without substituting a stale screenshot", async () => {
+    vi.mocked(captureDrawElementFrame).mockResolvedValueOnce(Buffer.alloc(100));
+    await expect(captureFrameToBuffer(await session(), 7, 7 / 30)).rejects.toThrow(
+      "suspect small frame",
+    );
+    expect(pageScreenshotCapture).not.toHaveBeenCalled();
+  });
+
+  it("returns a normal drawElement frame unchanged", async () => {
+    const buffer = Buffer.alloc(30_000);
+    vi.mocked(captureDrawElementFrame).mockResolvedValueOnce(buffer);
+    expect((await captureFrameToBuffer(await session(), 7, 7 / 30)).buffer).toBe(buffer);
+    expect(pageScreenshotCapture).not.toHaveBeenCalled();
+  });
+
+  it.each(["No cached paint record for element", DE_CANVAS_NOT_INITIALIZED_CODE])(
+    "rejects a failed batch and safely abandons pending prefix encodes",
+    async (error) => {
+      let rejectPrefix: (error: Error) => void = () => {};
+      const prefix = new Promise<Buffer>((_, reject) => {
+        rejectPrefix = reject;
+      });
+      vi.mocked(produceDrawElementFrameBatch).mockResolvedValueOnce({
+        encodeResults: [prefix],
+        failedAt: 1,
+        error,
+      });
+      await expect(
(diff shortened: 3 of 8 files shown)


miguel-heygen (commented):
The retry architecture reads coherently at `d71ad474`: all three drawElement entry points stop returning screenshot pixels from the injected-canvas page, the structural error survives the producer wrappers, cancellation and encoder interruption still short-circuit, and both streaming and disk recovery create a force-screenshot attempt. Disk capture also refuses to let a complete-looking prefix mask the untrusted frame set, and the batch prefix promises are observed before the page or worker is torn down.

I found one non-blocking diagnostics mismatch. In `captureFrameToBufferPipelined`, the new recoverable-error branch throws at `frameCapture.ts:3760-3768` before the Mirror captureFrameCore call to `captureFrameErrorDiagnostics` at `:3771-3781`. The serial path reaches its outer diagnostics catch, but the ordinary pipelined NCPR/canvas path does not emit the promised frame-error PNG/HTML/JSON/console bundle. That does not make the fresh-page retry unsafe, and it is not a regression from the old per-frame fallback, but either move the diagnostic call ahead of the wrapper throw or narrow the comment/body claim.

I am not stamping a draft over its own unmet merge gate. The current tests prove routing and rejection with mocks; they do not provide the real-Chrome pixel evidence the PR body requires for serial, worker, batch, hard-cut, and induced paint-record failure. Keep the draft until those comparisons exist, especially because a false-positive tiny-JPEG trip now converts one-frame fallback into a full recapture.

**Verdict:** COMMENT
**Reasoning:** The implementation has no code-level correctness blocker in the reviewed diff, but its explicit real-browser merge gate remains unmet; one pipelined diagnostics branch should also be reconciled before ready-for-review.

— Magi

vanceingalls:
## Real-Chrome pixel gate: run, and the defect reproduces

The merge gate this PR set for itself is now met. Browser launch was sandbox-blocked on the investigation host; it works on a macOS/hardware-GPU host, so I ran the comparisons there. Pushed alongside this: `4c7fbde5` moves the pipelined diagnostics call ahead of the recoverable wrapper (@miguel-heygen's non-blocking finding), and the branch is rebased onto `main` (the `renderOrchestrator.test.ts` conflict was an import union).

**Setup.** macOS arm64, M4 Pro, Chrome 152.0.7977.30, ANGLE/Metal hardware GPU. Purpose-built composition: 1920x1080 @30fps, 6s / 180 frames, three clips with hard cuts at t=2.0s and t=4.0s (11 clip-boundary frames), high-frequency texture so JPEGs clear the tiny-frame floor, no filter/blend/video/3D/webfont. Baseline is the same composition with `PRODUCER_EXPERIMENTAL_FAST_CAPTURE=false`, `--crf 0`. Damage criterion PSNR < 30 dB. Every run was verified to have actually engaged drawElement (`"captureMode":"drawelement"`) before being counted — an early batch of runs silently ran screenshot-vs-screenshot and was discarded.

### 1. Clean parity, all three capture paths

| path | selection | min PSNR | frames < 30 dB | wall |
|---|---|---|---|---|
| serial (`captureFrameToBuffer`) | `HF_DE_WORKER_ENCODE=false` | 67.1 dB | 0/180 | 10.9s |
| pipelined (`captureFrameToBufferPipelined`) | worker-encode + `HF_DE_BATCH=0` | 67.1 dB | 0/180 | 8.3s |
| batch (`captureFramesBatchPipelined`) | worker-encode + `HF_DE_BATCH=4` | 67.1 dB | 0/180 | 9.9s |

67 dB is JPEG-vs-screenshot encode noise. Screenshot baseline: 12.4s.

### 2. Induced missing paint record — branch vs clean `main`

Scratch env-gated hook (uncommitted, reverted) injecting the NCPR error at the first frame after the t=2.0s hard cut. Serial and pipelined inject a thrown `Error`; batch returns the `{failedAt, error}` shape the batch path actually classifies — injecting a throw there instead exercises a layer the production error never originates from, and that first attempt was discarded.

| run | fresh-page retry | frames < 30 dB | worst |
|---|---|---|---|
| serial, branch | yes | **0/180** | bit-identical |
| pipelined, branch | yes | **0/180** | bit-identical |
| pipelined, `main` | no | 1 (idx 60) | **6.2 dB** |
| batch, branch | yes | **0/180** | bit-identical |
| batch, `main` | no | 4 (idx 60-63) | **13.2 dB** |

**The stale-surface defect reproduces on `main`.** Its frame 60 scores 77.2 dB against baseline frame **59**, and only 6.2 dB against its own correct frame 60 — the same-page screenshot fallback returned the previous frame's canvas bitmap. Because the injection sits on a hard cut, that renders scene ALPHA (red) where scene BRAVO (blue) belongs. On the batch path the whole four-frame batch is corrupted. This branch recovers to bit-identical pixels on all three paths.

Main-side control was run for pipelined and batch, not serial; the serial mechanism is the same same-page fallback surface.

### 3. Native boundary-screenshot trigger

`HF_FAST_CAPTURE_BOUNDARY_SS=true` throws `DrawElementCaptureError` at clip boundaries without any patch: one fresh-page retry, output bit-identical to baseline, **zero unhandled promise rejections** — which exercises the abandoned-batch-encode drain outside the mocks.

### Retry cost

13.5s with one full retry vs 8.3s clean on this 180-frame composition, **+63%**. That is the cost trade this PR's description flags; it is real and worth stating in the merge decision rather than discovering in the fleet.

### What this does not show

macOS / hardware-GPU only. The original Windows compositor trigger is still not reproduced — this exercises the recovery path by injection, not that trigger. The 75% diagnosis confidence in the description stands.

Frame-60 PNG pair (baseline / `main` / branch) and the full harness are available; happy to attach or re-run any variant.

jrusso1020 (approved):
Reviewed the render half on its merits at `4c7fbde5`, with the engine and producer suites run and thirteen mutations applied at that head. The diagnosis is right, the fix is structural rather than a patch, and I found no correctness blocker. Two things are worth fixing, both about what is *measured* rather than what is written.

**Head reviewed:** `4c7fbde51e80f63a87d0e3be01ef3f7ec0c043f9`
**Prior review:** Magi's `5150984284` (COMMENTED) sat at `d71ad474`, a pre-rebase head — so the base moved under it. Both of its items are addressed: the pipelined diagnostics ordering by `4c7fbde5`, and the real-Chrome gate by the evidence comment. This review is additive to it.

**Audited:** `drawElementCaptureError.ts` (whole file), `frameCapture.ts` (every changed hunk plus the four capture entry points and their callers end-to-end), `capturePlan.ts` `replanAfterFailure` (whole function), `renderOrchestrator.ts` (the streaming and disk failure branches, `shouldRetryViaPinnedFallback`, the adaptive-retry guard), all three changed test files.
**Trusting:** the rest of `frameCapture.ts`'s 4435 lines, and the macOS pixel measurements (I cannot launch Chrome here — sandbox-blocked, same as your investigation host).

## Strengths

- **The fix is structural, not a per-site patch.** All four untrusted-surface reads become throws, and `isDrawElementCaptureError` (`drawElementCaptureError.ts:18-24`) matches on an own property through the cause chain rather than on a class identity, so it survives the package copies and worker wrappers that would defeat `instanceof`. The cycle guard at `:20` is the kind of thing that usually gets found in production instead.
- **The evidence comment is unusually honest.** It discloses the discarded runs, states plainly that the original Windows trigger is *not* reproduced, notes the serial `main`-side control was not run, and volunteers the +63% cost rather than burying it. That is the shape that makes a self-set merge gate worth setting.
- **Your own mutation claim on `4c7fbde5` holds.** Restoring the pre-fix ordering — recoverable wrapper before `captureFrameErrorDiagnostics` — fails exactly one test (`frameCapture-freshFallback.test.ts:107`) and nothing else. Verified, not taken on trust.
- **The engine half is genuinely pinned.** Eight mutations, eight kills: both boundary throws independently, the serial NCPR throw, the tiny-frame throw, the batch throw, the cause-chain depth walk (5→1), and the abandoned-encode drain — that last one only dies via an unhandled rejection, which is exactly the failure it exists to prevent, so the test is pinning the right thing.

## Important

**1. The disk-path recovery entry and the `capture_error` split are unpinned.** Two mutations survive the *full* 745-test producer suite, not just the changed files:

- `renderOrchestrator.ts:3884` — drop `isDrawElementCaptureError(err)` from the disk branch's throw-guard, so a capture error propagates and fails the render instead of triggering wipe-and-force-screenshot: **745 passed**.
- `renderOrchestrator.ts:3890-3892` — force `deSelfVerifyFallback = true`, so a capture error reports `fallbackReason: "psnr"` instead of `"capture_error"`: **745 passed**.

The streaming equivalents are all pinned — dropping `isDeCaptureError` from `shouldRetryViaPinnedFallback`, dropping `draw_element_capture` from `replanAfterFailure`, and dropping it from the adaptive-retry guard at `:1167` each kill a test. So this is an asymmetry between the two paths, not a general gap. Corroborating it: no file under `packages/producer/src` matching `*.test.ts` mentions `capture_error`, `deSelfVerifyFallback`, `isDrawElementCaptureError`, or `deCaptureFailure` — 103 test files swept.

That first mutation is the load-bearing one: it is the entry to the whole disk-path recovery this PR adds, and today nothing would notice if it regressed. One test that throws a `DrawElementCaptureError` out of `invokeDiskCapture` and asserts the replan carries `forceScreenshot`, `fallbackReason === "capture_error"` and `selfVerifyFallback === false` closes both.

**2. The tiny-frame abort is the one new trigger that can fire on a legitimately valid frame, and no measurement covers it.** At `frameCapture.ts:3667-3675`:

```ts
const sorted = sizes.length >= 12 ? [...sizes].sort((a, b) => a - b) : null;
const median = sorted ? (sorted[sorted.length >> 1] ?? 0) : 0;
const floor = Math.max(20000, median * 0.12);
```

Two properties compound. `sizes` is appended only in the *passing* branch, so the first 12 frames of every render are judged against a flat 20,000 B — `median` is 0, so the `Math.max` floor binds. And even once the median exists, `median * 0.12` only exceeds 20 kB when the median is above ~167 kB, so for most compositions the absolute floor is what is actually in force.

Encoding 1920×1080 at q80 with optimized Huffman tables, I measure solid black at **12,524 B** and a title card (white text bars on black) at **15,961 B** — both under the floor. Your own deleted comment puts a blank Chrome frame at 5-9 kB, the same order, which suggests the optimized column is the relevant one. The caveat is real and I will state it rather than overclaim: PIL's encoder is not Chrome's, so those two numbers bracket the question instead of settling it.

What changed is the price of being wrong. Before this PR a false positive cost one same-page screenshot and the render continued on drawElement; now it abandons the attempt and the whole render loses fast capture. And the composition behind the +63% figure was purpose-built with "high-frequency texture so JPEGs clear the tiny-frame floor", so this trigger is excluded from the evidence by construction. Worth noting for blast radius: the guard is reachable only through `captureFrameCore`, i.e. `captureFrame`/`captureFrameToBuffer` — `captureFrameToBufferPipelined` never calls it — and per the gate comment at `:3866` worker-encode is macOS-only, so the serial path carrying this guard is the BeginFrame path the Linux fleet runs.

Three ways out, cheapest first: require `sizes.length >= 12` before the abort can fire, making the floor relative-only; or lower the absolute floor toward the blank range (~10 kB) so it separates blank from simple-but-real content; or re-produce the suspect frame once before abandoning the attempt — a genuine paint-record drop will not come back identical, a legitimately small frame will, and the cost is one frame instead of a whole render. A byte-size histogram from one flat-design composition would settle which is needed.

## Nits

**3. `isRecoverableDrawElementError` is not idempotent.** `DrawElementCaptureError`'s message embeds the reason it was built from (`drawElement capture failed at frame N: No cached paint record; fresh screenshot capture required`), and the predicate is a substring match on exactly that text (`:3470-3472`). So an already-wrapped error still satisfies it. Nothing in-tree double-wraps today — I checked all three call sites — but any future outer check would, and each wrap costs a level against the depth-5 walk in `isDrawElementCaptureError`. An early `if (isDrawElementCaptureError(err)) return false;` makes the predicate self-limiting.

**4. `DE_FALLBACK_RATIO_WARN_THRESHOLD` is now effectively unreachable** (`:4378`). A recoverable error aborts the attempt, so `deNcprFallbacks` is at most 1 per session; `ncprFallbacks / frames > 0.5` therefore requires `frames <= 1`. It now means "the failure happened on the first frame", which is useful, but the new docstring ("Threshold for surfacing paint-record/canvas failures in partial-session diagnostics") still reads as a rate.

**5. Issue #3482 is still open** and this PR appears to resolve it. It tracks the "100%-fallback renders pass artifact validation" seam, and that seam is now unreachable: a paint-record failure aborts instead of silently screenshotting, so a render can no longer complete with a correct frame count on frames drawElement never captured. The PR deletes the comment that pointed at the issue; closing the issue with it would keep them from drifting apart.

## Note on my environment

My local producer run has four pre-existing test-file load failures (`mutedParity`, `planHash`, `probeFailures`, `probeStage`) and one unrelated media failure (`audioPadTrim.integration` true-peak ceiling). All five are identical with and without this change and none of the files are touched here, so I scoped every conclusion above to the suites that run clean; CI's producer jobs are green. Controls: engine capture 200 passed / 3 skipped, `frameCapture-freshFallback` 12 passed, producer changed-file suites 221 passed — the last matching your description exactly.

**Verdict:** APPROVE
**Reasoning:** The stale-surface defect is real and pixel-confirmed, and replacing every same-page screenshot fallback with a structural rejection is the right shape — correct-but-slower strictly beats shipping the previous frame's bitmap. Neither remaining item is a correctness fault: one is a test-coverage asymmetry on the disk path, the other a calibration question about a threshold whose false-positive cost this PR raises from one frame to one render.

— Rames Jusso

Issue #868, opened 2026-05-15
[CKS] Integration tests for url2video orchestrator step wiring (14-url-to-video-workflow)


cardinalconseils:
## Summary
Orchestrator (`packages/cli/src/url2video/orchestrator.ts`) wires scrape → generate → narrate → render and updates the run manifest between each step. Currently only individual step units are tested. Need integration tests around the orchestrator state machine: resume on partial failure, per-language fan-out, manifest transitions.

## Source
- **Phase:** 14 — URL-to-Video Workflow
- **Trigger:** backlog punt (4c) — Sprint Review verdict: ship as-is
- **Filed by:** CKS auto-filing (Dark Factory)
- **Date:** 2026-05-15
- **Intended label:** `cks:backlog` (label create currently blocked by token scope — using `enhancement`)

## Details
Tests to add:
- Resume a run mid-pipeline (manifest with `generate: ok`, `narrate: pending`) — orchestrator skips done steps.
- Per-language fan-out: `--langs fr,en` produces two `render-<lang>/` working dirs and two `output.mp4` paths.
- Failure surfaces correctly: a thrown step produces `failedStep` in manifest + BLOCKED.md, exit code non-zero.

## Evidence
`packages/cli/src/url2video/orchestrator.ts` (no `orchestrator.test.ts` present alongside `manifest.test.ts`, `blocked.test.ts`, `openrouter.test.ts`).

PR #759, opened 2026-05-12
perf(producer): hybrid layered/parallel path for SDR shader-transition renders (hf#732 PR 4/5)

vanceingalls:
## Summary

PR 4 of 5 in the hf#732 decomposition stack. **This is where the bulk of the shader-transition speedup lives** (`~2×` verified — see Empirical validation below).

Spreads per-frame DOM capture work across N DOM worker sessions and offloads the per-pixel shader-blend onto a `worker_threads` pool (the pool added in #758).

### Gating

The hybrid path is gated by `shouldUseHybridLayeredPath`:

- SDR content only — HDR raw-frame sources are fd-bound to one worker (per-worker `dup(fd)` is out of scope here).
- `workerCount >= 2`.
- Not every frame inside a transition window.

When the gate trips, the hybrid loop spawns `workerCount - 1` extra DOM sessions, allocates per-worker scratch buffers, and partitions the frame range into contiguous slices via `distributeLayeredHybridFrameRanges`. Each worker walks its slice; transitions dispatch through the shader-blend pool (with inline fallback). A frame-reorder buffer fences the encoder.

Pool teardown is guaranteed via try/finally on both the success and error paths.

### Structural change (heads-up to reviewers)

`captureHdrStage.ts` on main was already 921 lines (over the project's 500-line ceiling). Adding the hybrid path on top would push it past 1100 and the local pre-commit hook refuses to stage files past 500. **PR 4 splits `captureHdrStage.ts` into 5 files**:

- `captureHdrStage.ts` (orchestrator + cleanup invariants, 469 lines)
- `captureHdrResources.ts` (HDR video extraction + image decode + dim probing)
- `captureHdrFrameShared.ts` (gating predicates, partitioning, per-scene capture)
- `captureHdrSequentialLoop.ts` (legacy single-session loop)
- `captureHdrHybridLoop.ts` (new multi-worker path)

No behavior change in any pre-existing code path: the sequential loop is byte-equivalent to the previous inline implementation (both consume `captureSceneIntoBuffer` from the shared module, so behavior parity is enforced structurally rather than by comment-keeping).

`renderOrchestrator.ts` is intentionally unchanged — the stage computes its own worker budget via `calculateOptimalWorkers` rather than receiving it through the call signature.

## Stack

Stacked on top of #758 (PR 3: shaderTransition pool).

## Test plan

- [x] 14 new vitest tests in `captureHdrFrameShared.test.ts` pinning the hybrid gating predicate and the contiguous-chunking partitioner — all pass
[...]

PR #372, opened 2026-04-21
fix(producer): wire --crf and --video-bitrate CLI overrides into encoders

vanceingalls:
## Summary

Re-wire the `--crf` and `--video-bitrate` CLI flags through the three encoder spawn sites in `renderOrchestrator.ts`. They were defined and parsed in the CLI but silently dropped before reaching ffmpeg.

## Why

`Chunk 10` of `plans/hdr-followups.md`. PR #292 originally wired these through with a `baseEncoderOpts` object using `effectiveQuality`/`effectiveBitrate`; PR #268 rewrote the encode paths and reverted to `preset.quality` only, accidentally dropping the override. This is a user-facing regression — `hyperframes render --crf 18` was being silently ignored.

## What changed

- At the three encoder spawn sites (HDR streaming, SDR streaming, disk-based encode), `quality` defaults to `preset.quality` but is overridden by `job.config.crf` when set, and `bitrate` is set from `job.config.videoBitrate`. Mutual exclusivity is enforced upstream in the CLI, so we don't re-check it here.
- Fix the contradictory note in `docs/packages/cli.mdx` that claimed CRF/bitrate were now driven only by `--quality`. The flags table now lists `--crf` and `--video-bitrate` consistent with `docs/guides/rendering.mdx`.

## Test plan

- [x] `hyperframes render --crf 18 ...` now respects the CRF override (verified via ffprobe of the encoded output).
- [x] `hyperframes render --hdr ...` still works (no behavior change at the default path).
- [x] `hyperframes render --help` shows all flags consistent with the docs.

## Stack

Chunk 10 of `plans/hdr-followups.md`. Independent of all other chunks.

PR #3812, opened 2026-09-09
fix(studio): authenticate preview message senders

jrusso1020:
Studio's dimensions, caption activation, and caption overlay listeners accept any window's message when its payload claims `source: "hf-preview"`. A foreign frame can change the displayed composition dimensions or trigger caption activation/layout work. Require the browser-provided `MessageEvent.source` to match the current preview iframe's `contentWindow` before handling the payload.

- #759: pass the existing preview iframe ref from App and PreviewOverlays into useCompositionDimensions; reject unrelated and missing senders before parsing dimensions.
- #372: authenticate the current preview sender before caption detection can trigger DOM discovery and the caption file request.
- #868: authenticate the overlay's iframe sender before scheduling caption layout.

The check follows the current ref on each event, preserving iframe replacement and valid preview messages. It uses window identity rather than assuming an origin string, preserving existing srcdoc/sandbox behavior. Payload and runtime protocol checks remain in place; this change does not restrict authored code inside the trusted composition.

Validation: full workspace build, Studio typecheck, 4,799 Studio tests pass (18 existing todo). Three actual-handler regression tests reject foreign/null senders and retain current-preview behavior; each fails against the original source. Dimensions also covers replacing the preview ref and rejecting the old sender. Fallow passes with two test-fixture clone warnings, disclosed for independent reviewer assessment. No workflow changes or security suppressions.

Please independently review the sender boundary and compatibility, and verify #868/#759/#372 in the PR CodeQL scan. Alert closure will be counted only after merge and a successful main scan.

### DIFF
--- a/packages/studio/src/App.tsx
+++ b/packages/studio/src/App.tsx
@@ -333,7 +333,7 @@ export function StudioApp() {
         : null,
     effectiveTimelineDuration,
   });
-  const compositionDimensions = useCompositionDimensions();
+  const compositionDimensions = useCompositionDimensions(previewIframeRef);
   const { lintModal, linting, handleLint, closeLintModal, findingsByFile } = useLintModal(
     projectId,
     refreshKey,
--- a/packages/studio/src/captions/components/CaptionOverlay.sender.test.tsx
+++ b/packages/studio/src/captions/components/CaptionOverlay.sender.test.tsx
@@ -1,0 +1,47 @@
+// @vitest-environment jsdom
+import { act } from "react";
+import { createRoot } from "react-dom/client";
+import { expect, it, vi } from "vitest";
+import { CaptionOverlay } from "./CaptionOverlay";
+import { useCaptionStore } from "../store";
+
+Reflect.set(globalThis, "IS_REACT_ACT_ENVIRONMENT", true);
+
+it("only schedules caption layout for messages from its preview", () => {
+  const host = document.createElement("div");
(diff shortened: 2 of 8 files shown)


miguel-heygen (approved):
Reviewed exact head `db62e8f34296e2f51606218ebc3fbadc4459ec05`.

The sender boundary is correctly enforced at all three alerted handlers: `useCompositionDimensions.ts:39`, `useCaptionDetection.ts:134`, and `CaptionOverlay.tsx:119` compare the browser-controlled `MessageEvent.source` with the live preview iframe `contentWindow` before parsing or scheduling work. Reading the ref on every event preserves iframe replacement and rejects the old frame, foreign frames, nested frames, and null senders without breaking the existing srcdoc/sandbox origin model. Existing protocol and payload gates remain intact.

Verification: all three focused real-handler regressions pass locally; changed-file oxlint/oxfmt and `git diff --check` pass. All seven required checks, including Windows render and the full Test job, are green. Exact merge-ref CodeQL analysis `1748056217` reports zero results and there are zero open PR-ref alerts, so #868/#759/#372 are code-fixed rather than dismissed. The two overlapping test-fixture clone groups are small setup duplication and nonblocking.

— Magi

**Verdict:** APPROVE
**Reasoning:** The live iframe identity is the correct authentication boundary for these preview messages, every alerted handler applies it before side effects, compatibility is preserved, and native gates are fully green at the reviewed head.

PR #649, opened 2026-05-06
fix(player): replay from start when play is pressed after video ends

miguel-heygen:
## Summary

- When a non-looping composition reaches its end and the user presses play again, the player now seeks to frame 0 before resuming — previously it sent a `play` control with the playhead stuck at the final frame, which had no visible effect.
- The fix is a 3-line guard in `play()` that checks `_currentTime >= _duration` and calls `seek(0)` before dispatching the play command.
- Two new tests cover the replay-after-ended path and confirm mid-video play is unaffected.

## Test plan

- [x] `bun run --cwd packages/player test` — all 92 tests pass
- [x] `bun run --cwd packages/player build` — clean build
- [x] Lint + format + typecheck pass (lefthook pre-commit)
- [ ] Manual: open a composition in the player, let it play to the end, press play — should restart from 0:00

PR #634, opened 2026-05-05
feat: cache shader transition preview frames

vanceingalls:
## Summary
- Cache animated shader transition frame pairs in IndexedDB so browser previews can play outgoing and incoming scene motion through transitions instead of freezing snapshots.
- Include project/file signatures and normalized scene/style signatures in transition cache keys so edits invalidate only affected adjacent transitions and refreshes can reuse valid captures.
- Keep cached snapshots as blobs and upload only a bounded transition texture window, with stale-generation guards and cleanup so invalidated async uploads cannot overwrite newer state or leak textures.
- Gate playback on the active/upcoming transition texture window and repaint settled scene state outside transitions so scrub/playback cannot leave a hidden scene stack or black shader canvas.
- Move preview capture scale and transition-prep loading ownership to `<hyperframes-player>` via `shader-capture-scale` and `shader-loading`; HyperShader now reads player-provided preview options instead of composition config flags.
- Add the player-owned opaque transition loader with stable progress rows, shimmer title treatment, click/selection blocking, fade-out cleanup, and `shadertransitionstate` events.
- Coordinate Studio’s asset-loading overlay with the player-owned shader loader so the two overlays do not compete while transition captures are active.
- Default Studio shader previews to full-fidelity snapshots (`shader-capture-scale="1"`) while leaving direct non-player composition previews on the internal loading fallback.

## Verification
- `PATH=/Users/vanceingalls/.nvm/versions/node/v22.22.0/bin:$PATH bun run test src/hyperframes-player.test.ts` (packages/player)
- `bun run typecheck` (packages/player)
- `bun run typecheck` (packages/studio)
- `bun run typecheck` (packages/shader-transitions)
- `PATH=/Users/vanceingalls/.nvm/versions/node/v22.22.0/bin:$PATH bunx oxlint packages/player/src/hyperframes-player.ts packages/player/src/hyperframes-player.test.ts packages/studio/src/player/components/Player.tsx`
- `PATH=/Users/vanceingalls/.nvm/versions/node/v22.22.0/bin:$PATH bunx oxfmt --check packages/player/src/hyperframes-player.ts packages/player/src/hyperframes-player.test.ts packages/studio/src/player/components/Player.tsx`
- `git diff --check`
- `git diff --cached --check`
- `PATH=/Users/vanceingalls/.nvm/versions/node/v22.22.0/bin:$PATH bun run build` (packages/player)
[...]

PR #3813, opened 2026-09-09
fix(core): authenticate runtime control message senders

jrusso1020:
The runtime control bridge accepts any window's message when its payload claims hf-parent. A sibling or unrelated frame can therefore send playback, volume, seek, runtime-data, or other supported controls. Require the browser-provided sender to be the runtime's actual parent or its own window before inspecting or dispatching the control payload (#649).

Parent-to-composition controls and same-window controls remain supported, including embedded and standalone use. Existing payload/protocol validation is unchanged. Synthetic integration fixtures now include their legitimate parent sender instead of omitting MessageEvent.source.

Validation: full workspace build; all 2,703 core tests; core and runtime typechecks; lint/format; Fallow (no changed-file findings). The foreign/null sender regression fails against the original bridge. A real Chromium test with two sibling frames confirms parent and self controls are accepted while a foreign sibling is rejected. Unit coverage also distinguishes the embedding parent from the runtime window. No workflow changes or suppressions.

This follows the Studio sender batch but covers the opposite control direction. Other slideshow/playground handlers have different sender contracts and remain separate pending audit. Please independently review sender compatibility and #649's PR CodeQL result; closure is counted only after merge and main verification.

### DIFF
--- a/packages/core/src/runtime/bridge.test.ts
+++ b/packages/core/src/runtime/bridge.test.ts
@@ -27,11 +27,54 @@ function createMockDeps() {
 
 function makeControlMessage(action: string, extra?: Record<string, unknown>) {
   return new MessageEvent("message", {
+    source: window.parent,
     data: { source: "hf-parent", type: "control", action, ...extra },
   });
 }
 
 describe("installRuntimeControlBridge", () => {
+  it("rejects foreign and null senders before dispatching controls", () => {
+    const foreign = document.createElement("iframe");
+    document.body.append(foreign);
+    const deps = createMockDeps();
+    const handler = installRuntimeControlBridge(deps);
+    try {
+      for (const source of [foreign.contentWindow, null]) {
+        handler(
+          new MessageEvent("message", {
+            source,
+            data: { source: "hf-parent", type: "control", action: "play" },
+          }),
+        );
+      }
+      expect(deps.onPlay).not.toHaveBeenCalled();
+      handler(
+        new MessageEvent("message", {
+          source: window,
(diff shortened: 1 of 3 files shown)


miguel-heygen (approved):
Reviewed exact head `9efedf769f4739c4c9d30302288973cad48673f1`.

`packages/core/src/runtime/bridge.ts:126` puts browser-controlled sender identity ahead of every payload/protocol/action branch. The accepted set is correct for the runtime contract: the direct embedding parent covers Studio and `<hyperframes-player>`, while `window` preserves standalone/self control; foreign siblings, grandparents, nested frames, and null sources cannot dispatch. Cross-origin/sandboxed parents still work because WindowProxy identity comparison does not require origin access.

`bridge.test.ts:36-76` pins foreign/null rejection and distinct-parent/self acceptance, and the existing init fixtures now supply the legitimate parent source. The sibling slideshow navigation listener is a materially different bare-message embed contract with its own open alert #634, so keeping it for a separate audit is appropriate.

Verification: focused bridge + init suites pass 132/132 locally; changed-file oxlint/oxfmt and `git diff --check` pass. All six required checks, Windows render, and full Test are green. Exact merge-ref CodeQL analysis `1748384953` reports zero results and there are zero open PR-ref alerts, so #649 is code-fixed without dismissal.

— Magi

**Verdict:** APPROVE
**Reasoning:** The bridge now authenticates the only legitimate browser senders before any control side effect, preserves all production and standalone paths, and is fully green at the reviewed head.

PR #374, opened 2026-04-21
fix(engine,shader): handle matrix3d transforms and hide non-first scenes

vanceingalls:
## Summary

Two correctness fixes in the HDR transform & clipping pipeline: `parseTransformMatrix` now handles `matrix3d(...)` (GSAP's default `force3D: true`), and shader-transitions sets every non-first scene to `opacity: 0` at `t=0` so the engine doesn't over-composite at the start.

## Why

`Chunk 4` of `plans/hdr-followups.md`. Transform extraction and border-radius computation existed but were dead — an HDR video with `rotation: 45` rendered un-rotated, and 3-scene compositions ghosted at `t=0` because every scene defaulted to CSS `opacity: 1` and contributed to the first frame.

## What changed

**Matrix3d support in `parseTransformMatrix`.** `DOMMatrix.toString()` emits `matrix3d` whenever any ancestor in the chain has used a 3D transform — most importantly GSAP's default `force3D: true`, which converts `translate(...)` into `translate3d(..., 0)`. Without this, every GSAP-driven transform was silently dropped during HDR compositing because `videoFrameInjector.getViewportMatrix()` would return `matrix3d(...)` and the blit path would parse it as `null` and fall back to identity. The 16-value column-major form is converted to its 2D affine projection (indices 0, 1, 4, 5, 12, 13 → m11, m12, m21, m22, m41, m42); Z, perspective, and out-of-plane rotation components are dropped.

**Initial-state opacity in `initEngineMode`.** The browser preview branch uses a GL canvas overlay during transitions, so scene opacity at `t=0` doesn't matter visually. The engine branch reads scene opacity directly via `queryElementStacking()` to decide which layers to composite. Without an explicit initial-state tween, every scene defaulted to CSS `opacity: 1` and contributed to the very first frame, causing ghosting/overlap until the first transition fired. `tl.set()` at position 0 anchors the initial state in the timeline graph so reverse seeks from inside a later transition restore it correctly.

These two fixes together make `el.transform` and `el.borderRadius` (already wired in Chunk 7A's `compositeHdrFrame`) actually flow through the GSAP-animated case, and keep the engine's per-frame compositing aligned with what the user sees in browser preview.

## Test plan

- [x] 6 new `alphaBlit.test.ts` cases (identity matrix3d, translate3d, scale + translate3d, rotateZ, malformed arg count, non-finite values).
[...]

PR #367, opened 2026-04-21
refactor(shader-transitions): extract DEFAULT_DURATION and DEFAULT_EASE constants

vanceingalls:
## Summary

Extract `DEFAULT_DURATION = 0.7` and `DEFAULT_EASE = "power2.inOut"` as shared constants in `hyper-shader.ts` and apply them at all three fallback sites (metadata write, browser/render mode, engine mode).

## Why

`Chunk 2` of `plans/hdr-followups.md`. The three fallback sites had drifted apart: the metadata path used `1s` / `"none"` while the actual rendering used `0.7s` / `"power2.inOut"`. A transition that omitted `duration`/`ease` would render at 0.7 s but tell the producer it was 1 s, throwing off the producer's compositing window planning and producing a visible ~0.3 s brightness dropout.

This is a small, high-value correctness fix that runs before the larger Chunk 1 / Chunk 4 work.

## What changed

- New module-level `DEFAULT_DURATION` and `DEFAULT_EASE` constants in `packages/shader-transitions/src/hyper-shader.ts`.
- All three fallback call sites (metadata, browser, engine) now use the constants.
- Explicit `ease: "none"` on the timeline-length anchor tweens elsewhere in the file is intentional (those are linear interpolators driving the shader's progress uniform) and is left unchanged.

## Test plan

- [x] Render a composition with a transition that omits `duration` and `ease` — no brightness dip in the last ~0.3 s of the transition.
- [x] Preview (browser mode) and render (engine mode) produce matching blending curves.
- [x] Render with explicit `duration: 1.5` still works (constants are fallbacks only).

## Stack

Chunk 2 of `plans/hdr-followups.md`. Lands ahead of Chunk 1 (opacity) per the suggested merge order.

PR #366, opened 2026-04-21
refactor(types): tighten type safety, dedupe HfTransitionMeta, prune dead LUT export

vanceingalls:
## Summary

Four small, mechanical type-safety cleanups across `engine`, `producer`, and `shader-transitions`. Zero behavior change — pure pre-cleanup so the rest of the stack ships against a tighter baseline.

## Why

`Chunk 6` of `plans/hdr-followups.md`. Several non-null assertions and a duplicate interface had accumulated as rebase artifacts and leftover work-in-progress; lands first because it touches files later chunks edit and removes friction during review.

## What changed

- `renderOrchestrator.ts`: replace `layers[layerIdx]!` with a `for (const [layerIdx, layer] of layers.entries())` so both index and element come from the iterator.
- `engine/types.ts`: drop the duplicate `HfTransitionMeta` interface (rebase artifact); the original definition above it is the documented one. The orphaned doc comment now precedes `HfProtocol`.
- `shader-transitions/hyper-shader.ts`: keep the local `HfTransitionMeta` declaration (the package ships as a standalone CDN bundle and must not depend on `@hyperframes/engine`), but add a sync comment pointing at the source of truth in `engine/src/types.ts`.
- `alphaBlit.ts` + `engine/index.ts`: drop `export` from `getSrgbToHdrLut` and remove its re-export. It was only ever called by the internal `blitRgba8OverRgb48le`; the public surface was dead code.

## Test plan

- [x] `bun run --filter @hyperframes/engine typecheck`
- [x] `bun run --filter @hyperframes/producer typecheck`
- [x] `bun run --filter @hyperframes/shader-transitions typecheck`
- [x] `bun run --filter @hyperframes/engine test` — 308/308 pass (no test changes; assertions removed in code only).

## Stack

Chunk 6 of `plans/hdr-followups.md`. Mechanical cleanup landed early per the suggested merge order.

Issue #862, opened 2026-05-15
feat(cli): add --no-open flag to preview and play commands


na-navi:
## Problem

`hyperframes preview` and `hyperframes play` always open the system default browser on startup.

This is problematic for agent-driven workflows where the browser session is managed separately (e.g., Chrome DevTools Protocol, remote debugging, headless automation).

When an agent or script starts a preview server, the automatically opened browser window:
- Conflicts with an existing managed browser session
- Cannot be easily closed without human intervention
- Breaks automated test pipelines that expect no GUI

## Proposed Solution

Add a `--no-open` flag to both `preview` and `play` commands:

```bash
npx hyperframes preview --no-open
npx hyperframes play --no-open
```

This should start the server without calling `open()` (or equivalent) to launch the browser.

## Why This Matters

HyperFrames is positioned as agent-oriented. Agents often need to control the browser lifecycle explicitly. Automatically opening the default browser makes agent-driven workflows brittle.

A concrete scenario: an agent runs `hyperframes preview --no-open`, waits for HTTP 200, then connects via Chrome DevTools Protocol on a separate managed browser instance. The current behavior makes this impossible without manual cleanup.

## Scope

- `packages/cli/src/commands/preview.ts` — guard all 4 `import("open")` calls behind the flag
- `packages/cli/src/commands/play.ts` — guard the `import("open")` call behind the flag
- Both commands should accept `--no-open` as a boolean argument, defaulting to `false`

## Alternative

A `--browser <browser>` flag would also help, but `--no-open` is the smallest and safest primitive that covers the agent use case without changing existing behavior for human users.

na-navi:
Opened in wrong repo. Will reopen from fork.

PR #3814, opened 2026-09-09
fix(cli): bound and validate registry installation

jrusso1020:
Registry installation currently trusts network/cache manifest casts, buffers unbounded response bodies, and follows project symlinks when writing downloaded files and post-processing HTML. A crafted item name can escape the cache directory; a target directory symlink can send installation writes outside the project.

This change treats manifest parsing, download, transformation, and publication as one boundary:

- #366: validate network and cached manifests against the published schemas, constrain item names and requested name/type identity before cache lookup, and cap manifest/cache reads at 10 MiB. Valid stale fallback remains supported. Limits: 10,000 registry entries and 1,000 files per item.
- #367: cap downloads at 128 MiB per file and 512 MiB per install, share accounting across at most four concurrent files, and await every batch before returning errors. Missing or false Content-Length cannot bypass streamed byte accounting. The downloader returns bytes rather than writing a path.
- #858: validate item names, apply marker and variable transformations in memory, hash the final bytes, and publish once. Canonicalize the project root and every existing target component; reject escaping paths, Windows aliases/streams/device names, duplicate physical targets (including case and normalized Unicode aliases on Windows/macOS), and the reserved install record. File, cache, and lockfile publication use a private sibling stage and atomic replacement. Existing regular-file permission bits are preserved across replacement, including through internal leaf aliases. In-root symlinks remain supported by publishing to their resolved physical target; hard-linked external aliases are not overwritten through an inode write.
- #374/#862: structurally construct HTTPS URLs and validate every redirect, with a shared ten-second deadline and five-hop cap. Preserve private/custom HTTPS registries and explicit cross-origin signed CDN URLs. The independent preliminary audit identified these two as URL-selection flows, not local-file-payload transmission; independent review confirmed all five PR-ref results are removed; main CodeQL must verify closure after merge.

The filesystem guarantee assumes trusted ancestor directories are not concurrently replaced by another local process. It does not claim portable openat-style race immunity. Authored HTML/JS and arbitrary binary assets remain intentional registry capabilities; no MIME filtering or code stripping is added.

Validation: full workspace build, CLI typecheck, 3,160 CLI tests pass (3 skipped); 160 registry tests pass (1 skipped). The external-directory symlink regression fails the original installer. Coverage includes poisoned stale cache, name/identity rejection, stream cancellation and concurrent budgets, unsafe HTTPS redirects, external parent/leaf/lockfile links, internal parent/leaf links, hardlink isolation, Windows target spellings, and cleanup after failed publication. Fallow has no changed-file findings. Initial sandbox localhost failures passed when rerun with server access.

Please independently verify #366/#367/#858 remediation and classify any remaining intentional sinks plus #374/#862 individually. No alert dismissals or closure claims are made by this PR; verify main CodeQL after merge.

Review regression checks: 170 focused tests pass (1 skipped), including absent case/Unicode aliases on Windows/macOS and executable mode through regular files and internal symlinks. Alias and mode regressions fail the previous implementation. Rebuilt CLI telemetry E2E passes all checks using a test-only HTTPS-origin transport hook; production HTTPS enforcement remains intact. CLI types, lint, formatting and signed hooks pass. Prior Windows failures were timeouts in unchanged contactSheet/npxCommand tests; both pass locally and the new CI run must pass before merge.

### DIFF
--- a/packages/cli/src/registry/boundedFile.test.ts
+++ b/packages/cli/src/registry/boundedFile.test.ts
@@ -1,0 +1,17 @@
+import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+import { expect, it } from "vitest";
+import { readBoundedRegistryFile } from "./boundedFile.js";
+it("reads regular files up to the bound and rejects oversized or non-file cache entries", () => {
+  const dir = mkdtempSync(join(tmpdir(), "hf-cache-bound-"));
+  try {
+    const path = join(dir, "cache");
+    writeFileSync(path, "1234");
+    expect(readBoundedRegistryFile(path, 4).toString()).toBe("1234");
+    expect(() => readBoundedRegistryFile(path, 3)).toThrow(/limit/);
+    expect(() => readBoundedRegistryFile(dir, 100)).toThrow();
+  } finally {
+    rmSync(dir, { recursive: true, force: true });
+  }
+});
--- a/packages/cli/src/registry/boundedFile.ts
+++ b/packages/cli/src/registry/boundedFile.ts
@@ -1,0 +1,22 @@
+import { closeSync, fstatSync, openSync, readSync } from "node:fs";
+
+/** Read one regular file through a fixed descriptor without trusting a prior path stat. */
+export function readBoundedRegistryFile(path: string, maxBytes: number): Buffer {
+  const fd = openSync(path, "r");
+  try {
+    const stat = fstatSync(fd);
+    if (!stat.isFile() || stat.size > maxBytes) throw new Error("Registry file exceeds read limit");
+    const chunks: Buffer[] = [];
+    let total = 0;
+    while (total <= maxBytes) {
+      const chunk = Buffer.allocUnsafe(Math.min(64 * 1024, maxBytes + 1 - total));
+      const count = readSync(fd, chunk);
+      if (count === 0) return Buffer.concat(chunks, total);
+      total += count;
+      chunks.push(chunk.subarray(0, count));
+    }
+    throw new Error("Registry file exceeds read limit");
+  } finally {
+    closeSync(fd);
+  }
+}
--- a/packages/cli/src/registry/installer.test.ts
+++ b/packages/cli/src/registry/installer.test.ts
@@ -86,6 +86,34 @@ describe("installItem", () => {
       rmSync(dir, { recursive: true, force: true });
     }
   });
+  it.each(["darwin", "win32"] as const)(
+    "rejects absent case and Unicode aliases on %s before downloads",
+    async (platform) => {
+      vi.spyOn(process, "platform", "get").mockReturnValue(platform);
+      try {
+        for (const names of [
+          ["Foo.html", "foo.html"],
+          ["Café.html", "Cafe\u0301.html"],
+        ]) {
+          const dir = project();
+          try {
+            const conflicting = {
+              ...item,
+              files: names.map((name) => ({ ...item.files[0]!, target: `components/${name}` })),
+            };
+            await expect(installItem(conflicting, { destDir: dir, force: true })).rejects.toThrow(
+              /duplicate/,
+            );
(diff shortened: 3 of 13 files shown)


miguel-heygen (changes requested):
Reviewed exact head `f1c59ea2cbff9b181f9fa6ec761b543d4df84907`. The boundary shape is strong: fresh and stale manifests share one validator, the downloader returns bounded bytes, transforms/hash happen before one publish, and `registryTargetPath()` rechecks the physical parent while preserving in-root aliases. Exact merge-ref CodeQL analysis `1749046718` has zero results, so all five PR flows are removed without dismissal.

**Blockers**

1. `packages/cli/src/registry/publication.ts:62` — atomic replacement drops the existing file mode. The staged file is created with the process umask and renamed over the destination. On this exact head, replacing a `0755` target produced `0664`. Reinstalls therefore make executable registry scripts/assets non-executable even though the old path-based write preserved mode. Capture the physical destination mode when it is an existing regular file and apply it to the staged file before rename; add direct regular-file and internal-leaf-symlink mode tests.

2. `scripts/ci/cli-telemetry-e2e.mjs:112` / `packages/cli/src/registry/transport.ts:4` — the required `CLI smoke (required)` gate is red because its first-class custom-registry E2E still serves `http://127.0.0.1`, while the new transport rejects it before every install. Keep the HTTPS-only product boundary, but move this child-process fixture to an HTTPS-shaped URL intercepted by its existing fetch hook (or an equivalent TLS fixture), then rerun the required gate. Current failure: all install assertions fail with exit 1.

3. `packages/cli/src/registry/installer.ts:257` — duplicate physical-target protection only case-folds on Windows. On the default case-insensitive macOS filesystem, two absent targets such as `components/Foo.html` and `components/foo.html` pass preflight, enter the same four-file batch, and race atomic renames to one leaf while the lock records two logical hashes. Apply the portable alias key to Darwin too (including a consistent Unicode normalization) and pin the missing-leaf case; existing leaves already converge through `realpathSync`.

**Alert classification**

- #366, #367, #858: fixed by source changes; bounded/schema-validated ownership and single safe publication remove the original taint paths.
- #374, #862: the original CWE-200 classification was a URL-selection false positive (no local file payload), but the new structural HTTPS/redirect boundary removes the PR-ref results. Let main CodeQL close them as fixed; no dismissal is needed.

Local verification: 26 focused tests passed before two remaining suites hit a stale shared-dependency fixture; the exact mode probe above reproduces independently. Fallow is clear.

— Magi

**Verdict:** REQUEST CHANGES
**Reasoning:** The security boundary is coherent and CodeQL-clean, but it currently regresses file permissions, leaves a macOS duplicate-target race, and has a PR-owned required smoke failure.

miguel-heygen (approved):
Re-reviewed exact head `44121c3931f73412236b2b9b62f44bd94b642685`; this supersedes my prior request for changes.

All three blockers are resolved:

- `publication.ts:64-66` copies the existing physical regular-file mode onto the staged inode before atomic replacement. This also works through an internal leaf alias; my exact-head probe preserved `0755`, file contents, and the symlink.
- `installer.ts:254-263` uses NFC + case folding for Windows and Darwin before any download, so absent case/Unicode aliases and the reserved lock target converge deterministically.
- `cli-telemetry-e2e.mjs:112-135` presents a unique HTTPS fixture origin to production code and maps only that origin through the existing test hook. The HTTPS product boundary remains intact, and the fresh required CLI smoke passed.

Verification: focused publication + installer suites pass 20/20 locally. All seven required checks, full Test, Windows render, and CLI smoke are green. Exact merge-ref CodeQL analysis `1749328853` has zero results and there are zero open PR-ref alerts. #366/#367/#858 are fixed by the bounded validated ownership/publication changes; #374/#862 were URL-selection flows, and the structural HTTPS/redirect boundary removes them without dismissal.

— Magi

**Verdict:** APPROVE
**Reasoning:** The registry boundary now preserves existing file semantics while closing the manifest, download, redirect, and filesystem escape paths, and every native gate is green at the reviewed head.

PR #612, opened 2026-05-04
feat(cli): add remove-background command for transparent video

jrusso1020:
## What

Adds `hyperframes remove-background`, a new CLI subcommand that mattes a video or image locally with a small AI model and outputs a transparent WebM, ProRes 4444 `.mov`, or RGBA PNG. The output drops directly into any composition's `<video>` or `` element — no green screen, no API keys, no cloud upload.

```bash
npx hyperframes remove-background avatar.mp4 -o transparent.webm
npx hyperframes remove-background portrait.jpg -o cutout.png
npx hyperframes remove-background --info   # detected providers
```

## Why

Putting an avatar (or any subject) over an arbitrary background is one of the most common asks for a HyperFrames composition. Today the user has to wire up a third-party tool (rembg, Runway, unscreen.com), get the encoding flags right (`yuva420p` + `alpha_mode=1` — easy to miss; broken alpha in Chrome is silent), and integrate the output. Shipping it as a first-class CLI subcommand removes that friction and makes "transparent avatar over a scene" a one-liner.

The choice to run it locally (rather than punt to a SaaS) means: no API keys to configure, no upload of source media, runs fully offline once the model is cached.

## How

- **Model**: `u2net_human_seg` (MIT, ~168 MB ONNX) — the recommendation from a 12-cell matting eval that compared u²-net, MediaPipe SelfieSegmenter, BiRefNet, RVM, and BRIA RMBG-2.0 across CoreML, CUDA, and CPU. Single model on purpose: it's the only one that's MIT-clean, runs everywhere, and produces production-quality output for portraits at reasonable memory (~1.5 GB peak). Pre/postprocessing matches rembg's u²-net session exactly so output should be pixel-equivalent.
- **Runtime**: `onnxruntime-node` for inference, `sharp` for resize, `ffmpeg` for decode/encode (already required by other CLI commands). Provider auto-detect picks CoreML on Apple Silicon, CUDA when `HYPERFRAMES_CUDA=1`, CPU otherwise. Provider failure falls back to CPU rather than aborting.
- **Pipeline**: `ffmpeg` decode → per-frame ONNX → `ffmpeg` encode, with a streaming generator between the two and pre-allocated buffers for the hot path (Float32 input tensor, mask, RGBA output reused across every frame).
[...]

PR #611, opened 2026-05-03
feat(registry): add experimental HTML-in-canvas launch blocks

miguel-heygen:
## Problem

Chrome/Brave now expose the experimental CanvasDrawElement / HTML-in-canvas API, but HyperFrames did not have a launch-ready catalog story for it. The raw capability is easy to dismiss as a tech demo unless creators can install a block that turns one editable product UI into useful launch/tutorial shots.

`hyperframes validate` also needed to use the same browser args as the renderer so WebGL and CanvasDrawElement compositions are validated against the right environment.

## What this fixes

- Adds native `drawElementImage()` capture support for `@hyperframes/shader-transitions`, with the existing fallback path preserved.
- Enables `CanvasDrawElement` in renderer Chrome args and reuses renderer Chrome args from `hyperframes validate`.
- Adds registry `stability: "experimental"` metadata, schema/type/catalog/CLI surfacing, and generated docs warnings with agent setup instructions.
- Marks `canvas-dom-refractor` as experimental.
- Adds `experimental-screen-callout`, the simple workflow proof for one DOM screen -> zoom lens, spotlight, cursor trail, annotations, and crop strip.
- Adds `experimental-product-launch-portal`, the flagship creator block: one editable product UI becomes shadered portal warps, device panels, crop callouts, overload glitch beats, and a final launch CTA from the same HTML-in-canvas source.

## Root cause

The project had shader and catalog infrastructure, but not a native browser capture path or a creator-facing component that made the API useful. CanvasDrawElement needs a layout-subtree canvas, `requestPaint()`/paint lifecycle, and `drawElementImage()` working together; the catalog blocks now keep that native path gated and preserve DOM fallbacks for browsers without the experimental API.

### Local checks

- `bun run --filter @hyperframes/shader-transitions test`
- `bun run --filter @hyperframes/shader-transitions typecheck`
- `bun run --filter @hyperframes/shader-transitions build`
- `bun run --filter @hyperframes/engine test src/services/browserManager.test.ts`
- `bun run --filter @hyperframes/engine typecheck`
- `bun run --filter @hyperframes/core test src/registry/types.test.ts`
- `bun run --filter @hyperframes/core typecheck`
- `bun run --filter @hyperframes/cli typecheck`
- `bunx oxlint packages/cli/src/commands/catalog.ts packages/core/src/registry/types.ts packages/core/src/registry/index.ts packages/core/src/index.ts packages/core/src/registry/types.test.ts scripts/generate-catalog-pages.ts`
[...]

PR #3816, opened 2026-09-09
fix: restrict runtime and playground message dispatch

jrusso1020:
Runtime and SDK playground messages currently index ordinary objects with message-controlled action names. Names such as `__proto__` and `hasOwnProperty` can select inherited values and throw instead of being ignored. The playground also accepts messages from unrelated windows, allowing them to change preview selection, timing, and edits.

This change closes the related message-dispatch boundary:

- #648: use a Map for registered core control handlers, preserving the existing parent/self sender policy and every supported action.
- #611: require a string message type and an own entry in the playground handler table before dispatch.
- #612: accept playground messages only from the current preview iframe's non-null contentWindow. Replaced frames, unrelated windows and null senders are rejected before inspecting payloads.

Validation: 2,704 core tests and four playground tests pass; both new regression tests fail the prior implementation. The playground test exercises the real registered UI listener with current/replaced frames, inherited names and non-string coercion attempts. A real Chromium check confirms preview duration messages work while self/sibling and inherited/coerced types are ignored without exceptions. Core and playground typechecks, full workspace build, final core rebuild, lint/format and signed hooks pass.

No alert dismissals are requested. Verify the three PR-ref results and then main CodeQL closure after merge. Payload validation beyond the existing message contract is unchanged.

### DIFF
--- a/packages/core/src/runtime/bridge.test.ts
+++ b/packages/core/src/runtime/bridge.test.ts
@@ -33,6 +33,27 @@ function makeControlMessage(action: string, extra?: Record<string, unknown>) {
 }
 
 describe("installRuntimeControlBridge", () => {
+  it("ignores inherited and unknown action names from an authorized sender", () => {
+    const deps = createMockDeps();
+    const handler = installRuntimeControlBridge(deps);
+    try {
+      for (const action of [
+        "__proto__",
+        "constructor",
+        "hasOwnProperty",
+        "__defineGetter__",
+        "toString",
+        "unknown",
+      ]) {
+        expect(() => handler(makeControlMessage(action))).not.toThrow();
+      }
+      expect(deps.onPlay).not.toHaveBeenCalled();
+      handler(makeControlMessage("play"));
+      expect(deps.onPlay).toHaveBeenCalledOnce();
+    } finally {
+      window.removeEventListener("message", handler);
+    }
+  });
   it("rejects foreign and null senders before dispatching controls", () => {
     const foreign = document.createElement("iframe");
     document.body.append(foreign);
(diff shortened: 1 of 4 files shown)


miguel-heygen (approved):
Reviewed exact head `7c12a68285ecdb61dd424734992f3d1305adacbd`.

`packages/core/src/runtime/bridge.ts:57-90,135` follows the CodeQL recommendation directly: the fixed action registry is a `Map`, so inherited names cannot resolve, while every supported control remains registered. The existing parent/self sender gate still runs before payload/action inspection.

`packages/sdk-playground/src/main.ts:1359-1365` independently enforces the playground contract: it resolves the current iframe on every event, rejects null/replaced/foreign senders, requires a primitive string type, and checks an own registered handler before dispatch. The sandboxed srcdoc bridge remains the sole intended sender; supported click/deselect/drag/duration/time actions are unchanged.

Verification: core bridge tests pass 34/34 locally; changed-file oxlint/oxfmt and `git diff --check` pass. The playground test could not run in the isolated worktree because Vite denies the shared GSAP symlink outside its filesystem allowlist, so I relied on the passing hosted playground/full Test lanes for that surface. Both unrelated hosted timeouts passed on unchanged-head retry. All eight required checks, Windows, full Test, viewport, regression shards, and exact merge-ref CodeQL analysis `1749532267` are green with zero open PR-ref alerts. #648/#611/#612 are code-fixed; no dismissals.

— Magi

**Verdict:** APPROVE
**Reasoning:** Both dispatch tables now enforce explicit ownership and the playground authenticates the live iframe before any side effect, while all native gates are green at the reviewed head.

PR #637, opened 2026-05-05
feat(cli): add --background-output to remove-background

jrusso1020:
## Summary
- New `--background-output` / `-b` flag on `hyperframes remove-background` emits a second transparent video alongside the cutout — same source RGB, alpha is `255 − mask`. Single inference pass; dual ffmpeg encoders run in parallel and share the `--quality` preset, so the layers are pixel-aligned.
- Refactor: extracted pure `resolveRenderTargets` validator and a `spawnFfmpeg(args, label, stdio)` helper that consolidates the three near-identical decoder/encoder spawn blocks.
- Docs + skill explain that this produces a **hole-cut plate** (subject region transparent), **not** an inpainted clean plate, and call out which use cases each is right for so an agent doesn't reach for `--background-output` when a user actually needs an inpainter.

### What you get
```bash
npx hyperframes remove-background subject.mp4 \
  -o subject.webm \
  --background-output plate.webm
```

| Output | Alpha | Use as |
| --- | --- | --- |
| `subject.webm` | Mask — subject opaque | Foreground (top) |
| `plate.webm` | `255 − mask` — subject region transparent | Background (bottom); place text/graphics between this and `subject.webm` |

### Constraints (validated up front, before any I/O)
- Video input only (`--background-output` is rejected for image inputs — there's no temporal pairing to do).
- Both outputs must be `.webm` or `.mov`. `.png` is rejected for the plate.
- Hole-cut, not inpainted. The plate has a person-shaped transparent hole; composite something opaque under it.

### Files
- `packages/cli/src/background-removal/inference.ts` — `Session.process` returns `{ fg, bg }`, single pixel-loop pass writes both alpha tracks; pre-allocated `rgbaBgBuf` reused across frames.
- `packages/cli/src/background-removal/pipeline.ts` — added `RenderOptions.backgroundOutputPath`, exported `resolveRenderTargets`, extracted `spawnFfmpeg`. Frame loop issues both writes before any `await drain` so a slow encoder doesn't block the other.
- `packages/cli/src/commands/remove-background.ts` — flag plumbing + JSON/spinner output for the second path.
- `packages/cli/src/background-removal/pipeline.test.ts` — 7 new tests for `resolveRenderTargets` covering valid combos and the four rejection paths.
- `docs/guides/remove-background.mdx`, `docs/packages/cli.mdx`, `skills/hyperframes-media/SKILL.md` — updated.

## Test plan
- [x] `bun run --cwd packages/cli test src/background-removal` (33/33 pass)
[...]

PR #3819, opened 2026-09-09
fix(registry): authenticate deck sound effect messages

jrusso1020:
The Airbnb deck's sound-effect handler accepts messages from unrelated windows and indexes an ordinary object using the supplied clip name. A message naming `__proto__` can therefore assign `currentTime` to `Object.prototype`.

Require the sender to be the current deck player's non-null iframe window (#637), and require an own string clip name before lookup (#636). The four sound effects, gesture unlock, volume and slideshow mute behavior remain unchanged. The player is resolved for each message so a replaced iframe cannot keep controlling sound.

Verification: a regression executes the actual demo scripts in JSDOM with mocked Audio and covers all four clip names, unlock/mute/unmute, null/foreign/replaced senders, inherited names, coerced names and prototype integrity. It passes the fix and fails the original demo. A real Chromium check with the actual player iframe confirms valid sound messages work while self messages and prototype names are rejected. Full workspace build, core typechecks, lint/format and signed hooks pass; Fallow is clear.

Validation limitation: the repository CLI lint/check commands target the unchanged `index.html` composition, not the `demo.html` host page. They stop on its existing requestAnimationFrame lint finding (plus a file-size warning); no browser/layout pass is claimed from that command. The host-page behavior is covered by the separate real-browser check above.

Please verify both alert flows and compatibility. No dismissals; verify main CodeQL closure after merge. Test lives with the core JSDOM fixtures to use its existing dependency, with no new dependencies or production core changes.

### DIFF
--- a/packages/core/src/runtime/deckAudio.test.ts
+++ b/packages/core/src/runtime/deckAudio.test.ts
@@ -1,0 +1,74 @@
+import { JSDOM } from "jsdom";
+import { expect, it, vi } from "vitest";
+import { readFileSync } from "node:fs";
+import { resolve } from "node:path";
+const demoHtml = readFileSync(
+  resolve(__dirname, "../../../../registry/examples/airbnb-deck/demo.html"),
+  "utf8",
+);
+
+it("accepts only current-player sound names while preserving unlock and mute behavior", async () => {
+  const clips: Array<{
+    play: ReturnType<typeof vi.fn>;
+    pause: ReturnType<typeof vi.fn>;
+    volume: number;
+    currentTime: number;
+  }> = [];
+  const dom = new JSDOM(demoHtml, {
+    runScripts: "dangerously",
+    beforeParse(window) {
+      Object.defineProperty(window, "Audio", {
+        value: function () {
+          const clip = {
+            play: vi.fn(() => Promise.resolve()),
+            pause: vi.fn(),
+            volume: 1,
+            currentTime: 0,
+          };
+          clips.push(clip);
+          return clip;
+        },
+      });
+    },
+  });
+  try {
+    const w = dom.window;
+    const player = w.document.querySelector("hyperframes-player")!;
+    const frame = w.document.createElement("iframe");
+    player.append(frame);
+    Object.defineProperty(player, "iframeElement", { get: () => player.querySelector("iframe") });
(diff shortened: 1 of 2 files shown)


miguel-heygen (approved):
The two guards are correctly placed at the message boundary and preserve the demo’s existing sound contract.

- `registry/examples/airbnb-deck/demo.html:171-176` resolves the current player/iframe for every event and requires exact non-null `contentWindow` identity before reading the payload. That makes iframe replacement self-healing and rejects parent/self/sibling/null senders without relying on a fixed origin.
- `registry/examples/airbnb-deck/demo.html:176` requires a primitive string and an own key of the four-entry clip table before lookup. `__proto__`, `constructor`, coerced arrays, and unknown names cannot reach the `currentTime` assignment.
- `packages/core/src/runtime/deckAudio.test.ts:10-73` executes the shipped `demo.html` script rather than duplicating the handler. Its placement is reasonable: core already owns the JSDOM test dependency, no production core surface changes, and the witness covers valid sounds, gesture unlock, mute/unmute, current/foreign/null/replaced senders, inherited keys, and prototype integrity.

Independent verification at exact head `de65a8eae923a630c12ee4abd11dabcae2c0be34`:

- Focused JSDOM witness: 1/1 passed.
- Changed TypeScript: oxlint 0 warnings/errors; oxfmt check passed.
- Player build passed and verified the runtime pin.
- Local Chromium against the built real demo: the current iframe produced one valid play; same-frame `__proto__`, a foreign iframe, and a null-source event produced no additional plays; `Object.prototype` remained clean; no page errors.
- All hosted checks are now green after the Google apt mirror recovered, including the full Test, required CLI smoke, preview parity, perf parity/fps, Windows, regression matrix, and CodeQL. The PR-ref open-alert set is empty, and no review threads/comments are open.

Separate sequencing note: main alert #648 from merged #3816 is not part of this diff. I independently classified it as a scoped false positive; landing #3819 remains subject to that prior-batch disposition being recorded as James stated.

— Magi

**Verdict:** APPROVE

**Reasoning:** Both source authentication and clip selection are fail-closed at the boundary, the actual shipped script is exercised, independent browser behavior matches the unit witness, and every native gate is green at the unchanged head.

PR #797, opened 2026-05-13
fix(studio): align preview fonts with render

func25:
## What

Align Studio preview font handling with final render.

## Why

Preview and render use different font handling. I found this when making a video with a quite strict text's size and position.  This bug could change text width and make text layout look different between preview and final render.

## How

Add a preview HTML transform hook in the Studio API. Use it in preview adapters to inject the same deterministic fonts used by render.

_I only fixed the font handling. For other mismatches between preview and render, consider making another PR (or I can make it later if having more time)._

## Test plan

How was this tested?

- [x] Unit tests added/updated
- [x] Manual testing performed
- [ ] Documentation updated (if applicable)

PR #673, opened 2026-05-08
fix(player): drive src URL timelines without runtime

miguel-heygen:
## Problem

Fixes #672.

`<hyperframes-player src="...">` can load a same-origin standalone GSAP composition and parse its duration from `window.__timelines`, but playback controls do not drive that timeline when the iframe never installs the HyperFrames runtime bridge.

In the reported repro, `player.currentTime = 2` updated the player element's cached time, while `iframe.contentWindow.__timelines.main.time()` stayed at `0` and the visible title remained at its initial opacity.

## What this fixes

This PR adds a direct same-origin timeline adapter for standalone `window.__timelines` compositions when no runtime bridge is present.

- `seek()` now falls back from `window.__player.seek()` to direct `timeline.seek(seconds)` before using `postMessage`.
- `play()` / `pause()` directly call `timeline.play()` / `timeline.pause()` for standalone timeline-only embeds.
- The player maintains its parent-side `currentTime` while a direct timeline is playing by sampling `timeline.time()` on `requestAnimationFrame`.
- Runtime-backed compositions remain runtime-owned. If `window.__hf` or `window.__player` is installed, the player does not bypass the runtime to drive raw `__timelines`.

## Root cause

The iframe probe treated raw `window.__timelines` as a valid duration adapter and marked the player ready, but the control paths only talked to either:

1. `window.__player.seek()` for same-origin runtime installs, or
2. `postMessage({ source: "hf-parent", type: "control" })` for runtime bridge installs.

A standalone same-origin `src=URL` composition with only `window.__timelines` has neither a `window.__player` sync bridge nor a postMessage listener. The player could discover duration from the timeline but had no transport that actually drove it.

### Local checks

- `bun run --filter @hyperframes/player test` — 107/107 pass
- `bun run --filter @hyperframes/player typecheck`
- `bun run --filter @hyperframes/player build`
- `bunx oxlint packages/player/src/hyperframes-player.ts packages/player/src/hyperframes-player.test.ts`
- `bunx oxfmt --check packages/player/src/hyperframes-player.ts packages/player/src/hyperframes-player.test.ts`
- `git diff --check`
- Lefthook pre-commit passed lint, format, and typecheck before commit.

### Browser verification

Used `agent-browser` against a local same-origin harness matching issue #672:

- Parent page loads the locally built player bundle via `<script type="module" src="/player.js">`.
- Player embeds `<hyperframes-player src="/composition.html" shader-loading="none" controls>`.
[...]

PR #488, opened 2026-04-25
ci: guard release channel publishing

miguel-heygen:
## Summary
- add a release-channel validator that blocks stable tags from prerelease-only branches and prerelease tags from stable branches
- wire the publish workflow to fetch branch history and validate channel boundaries before publishing
- document the stable vs alpha branch policy and npm dist-tag behavior

## Testing
- node --test scripts/validate-release-channel.test.mjs
- bunx oxfmt --check scripts/validate-release-channel.mjs scripts/validate-release-channel.test.mjs package.json .github/workflows/publish.yml docs/docs.json docs/contributing/release-channels.mdx
- bunx oxlint scripts/validate-release-channel.mjs scripts/validate-release-channel.test.mjs

## Notes
- No browser flow exists for this CI/docs-only release guard change.

Issue #473, opened 2026-04-24
linter: flag missing tl.set() hard-kill on scene-boundary exits


ibrews:
## Background

[PR #364](https://github.com/heygen-com/hyperframes/pull/364) added Rule 5 to `step-6-build.md`:

> Every element whose visibility changes at a beat boundary needs a deterministic `tl.set()` kill after its exit tween — not just captions. Without it, non-linear seeking can land between the fade end and the next scene start, leaving the element partially visible or causing a black-frame gap.

## Problem

In a/b testing of PR #364 (see review [#4167523103](https://github.com/heygen-com/hyperframes/pull/364#pullrequestreview-4167523103)), the sub-agent produced 0 `tl.set()` hard-kills even with the rule present in the prompt — in both runs. The rule is documented but documentation alone isn't a reliable enforcement mechanism for sub-agents.

Meanwhile, scans of existing Stripe compositions found 86 exit animations with only 1 `tl.set()` hard-kill, and black-frame gaps at every beat boundary in rendered snapshots. The bug is real and widespread.

## Proposed linter rule

Flag any element that has an exit tween ending at or near a `data-start` boundary without a corresponding `tl.set()` clearing its opacity/visibility at that same position.

Heuristic:
- Find `tl.to(el, { opacity: 0, ... }, t)` where `t + duration` ≈ a scene boundary
- Check whether a `tl.set(el, { opacity: 0 })` (or equivalent) exists at the same position
- If not → warn: `missing hard-kill after exit tween on <selector> at t=<time>`

## Notes

- The existing caption hard-kill check is already a narrower version of this. This would generalize it to all scene-boundary exits.
- False positive risk: elements that are hidden by other means (e.g., parent clip, `display: none` on the scene root). May need to be scoped to elements inside `.clip` containers only.
- Related: linter heuristic inconsistency on `overlapping_gsap_tweens` (separate issue) may be masking some of these cases.

PR #3822, opened 2026-09-09
fix(security): verify CDN script integrity before inlining

jrusso1020:
CDN-backed compositions could execute changed third-party scripts, and producer compilation fetched and inlined scripts without checking their `integrity` attribute. Pin the existing GSAP 3.12.2 and Three.js r128 bytes in nine compositions and verify supported SRI digests before decoding or inlining downloaded scripts. Library versions and script contents stay unchanged. A verified integrity mismatch aborts compilation, producing no executable fallback HTML. Ordinary network/HTTP failures retain the existing fallback behavior. Nested and template composition hoisting carries integrity/CORS attributes into both preview and producer output. A child pin strengthens existing same-URL root tags; conflicting nonempty integrity strings fail explicitly rather than discarding a requirement. Protected local scripts remain external for browser verification.

Alert mapping:
- #797, #673, #488, #473, #393: five producer fixtures using GSAP 3.12.2.
- #405, #404, #403, #402: four liquid-glass registry blocks using Three.js r128.
- Shared compiler coverage verifies raw bytes (including BOM), strongest-algorithm selection, multiple digests, reserved options, URL-safe base64, case-insensitive algorithm tokens, unsupported metadata, and tampered-response rejection. Explicit hashing works under Node and Bun; Bun's fetch integrity option alone did not enforce the check in our probe.

Validation: CDN bytes independently match npm release artifacts. Full workspace build, producer typecheck, 135 compiler tests and 2,711 core tests, the full producer Bun unit lane, and Fallow pass. The new regression fails the original compiler. Real Chromium executes both exact library payloads and blocks modified responses; the actual compiler accepts/rejects the same payloads under Node and Bun. All five producer fixtures pass CLI lint/check. All four registry blocks pass lint, runtime, layout, and motion with their bundled assets and library dependencies preserved; the full check remains red only for pre-existing contrast findings, reproduced on both baseline and fixed HTML.

Local broader-suite limitations: the Vitest producer unit lane has 663 passing tests, one skip, and one unchanged audio loudness failure (3.9 LUFS difference versus a 3 LUFS limit), reproduced with the original compiler. Integration Vitest has 77 passing tests; the Bun integration runner reaches cross-worker tests that require beginframe capture, while this macOS browser selects screenshot capture. These are not claimed green; native CI must pass before merge. No workflow or threshold changes.

Nested validation exercises real bundleToSingleHtml and compileForRender paths, including an unpinned root duplicate, valid versus tampered CDN bytes, duplicate pin retention, conflicting pins, and protected local template scripts. The nested tampering regression fails the previous head. Full workspace build, core/producer types, and change-scoped Fallow pass.

Local script inline decisions are deferred until nested/template requirements are collected. Resolved-path matching covers aliases such as local.js and ./local.js; root, sibling, and template unpinned-first regressions all fail the earlier implementation and pass now. Emitted supported algorithm tokens are normalized to lowercase for browser enforcement. Unprotected script chunk order is retained.

### DIFF
--- a/packages/core/src/compiler/externalScripts.ts
+++ b/packages/core/src/compiler/externalScripts.ts
@@ -21,7 +21,14 @@ export function ensureExternalScriptTag(
   );
   const requirements = new Set(
     [attributes.integrity, ...existing.map((el) => el.getAttribute("integrity"))]
-      .map((value) => value?.trim())
+      .map((value) =>
+        value
+          ?.trim()
+          .replace(
+            /(^|[\t\n\f\r ])(sha256|sha384|sha512)-/gi,
+            (_match, space: string, algorithm: string) => `${space}${algorithm.toLowerCase()}-`,
+          ),
+      )
       .filter((value): value is string => Boolean(value)),
   );
   if (requirements.size > 1) {
--- a/packages/core/src/compiler/externalScripts.ts
+++ b/packages/core/src/compiler/externalScripts.ts
@@ -1,0 +1,42 @@
+export interface ExternalScriptAttributes {
+  integrity?: string;
+  crossorigin?: string;
+}
+
+export function readExternalScriptAttributes(el: Element): ExternalScriptAttributes {
+  const attributes: ExternalScriptAttributes = {};
+  if (el.hasAttribute("integrity")) attributes.integrity = el.getAttribute("integrity") || "";
+  if (el.hasAttribute("crossorigin")) attributes.crossorigin = el.getAttribute("crossorigin") || "";
+  return attributes;
+}
+
+/** Deduplicate scripts without discarding a nested composition's integrity requirement. */
+export function ensureExternalScriptTag(
+  doc: Document,
+  src: string,
+  attributes: ExternalScriptAttributes = {},
+): void {
+  const existing = [...doc.querySelectorAll("script[src]")].filter(
+    (el) => el.getAttribute("src")?.trim() === src,
+  );
+  const requirements = new Set(
+    [attributes.integrity, ...existing.map((el) => el.getAttribute("integrity"))]
+      .map((value) => value?.trim())
+      .filter((value): value is string => Boolean(value)),
+  );
+  if (requirements.size > 1) {
+    throw new Error(`Conflicting script integrity requirements for ${src}`);
+  }
+  const integrity = [...requirements][0];
+  const elements = existing.length ? existing : [doc.createElement("script")];
+  for (const el of elements) {
+    if (integrity) el.setAttribute("integrity", integrity);
+    if (attributes.crossorigin !== undefined)
+      el.setAttribute("crossorigin", attributes.crossorigin);
+  }
+  if (!existing.length) {
+    const el = elements[0]!;
+    el.setAttribute("src", src);
+    doc.body.appendChild(el);
+  }
+}
--- a/packages/core/src/compiler/htmlBundler.test.ts
+++ b/packages/core/src/compiler/htmlBundler.test.ts
@@ -1685,3 +1685,43 @@ it("keeps protected local scripts external when hoisting an inline template", as
     rmSync(dir, { recursive: true, force: true });
   }
 });
+
+it.each(["root", "sibling", "template"])(
+  "does not inline local bytes before a later %s integrity requirement",
+  async (placement) => {
+    const unpinned = '<script src="local.js"></script>';
+    const pinned =
+      '<script src="./local.js" integrity="sHa384-YQ==" crossorigin="anonymous"></script>';
+    const rootScript = placement === "root" ? unpinned : "";
+    const first =
+      placement === "sibling"
+        ? '<div data-composition-id="first" data-composition-src="first.html"></div>'
+        : "";
+    const templates =
+      placement === "template"
+        ? `<template id="first-template"><div data-composition-id="first">${unpinned}</div></template><template id="child-template"><div data-composition-id="child">${pinned}</div></template>`
+        : "";
+    const children =
+      placement === "template"
+        ? '<div data-composition-id="first" data-start="0" data-duration="1"></div><div data-composition-id="child" data-start="0" data-duration="1"></div>'
+        : `${first}<div data-composition-id="child" data-composition-src="child.html"></div>`;
+    const dir = makeTempProject({
+      "index.html": `<html><head>${rootScript}</head><body>${templates}<div data-composition-id="root" data-width="320" data-height="180" data-duration="1">${children}</div></body></html>`,
+      "first.html": `<html><head>${unpinned}</head><body><div data-composition-id="first" data-width="320" data-height="180" data-duration="1">First</div></body></html>`,
+      "child.html": `<html><head>${pinned}</head><body><div data-composition-id="child" data-width="320" data-height="180" data-duration="1">Child</div></body></html>`,
+      "local.js": "window.alteredLocalBytes = true;",
+    });
+    try {
+      const bundled = await bundleToSingleHtml(dir);
+      expect(bundled).not.toContain("window.alteredLocalBytes = true;");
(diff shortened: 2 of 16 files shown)


miguel-heygen (changes requested):
The raw-byte check is correctly placed before `TextDecoder`, and the fixture digests independently match the exact cdnjs/npm bytes. Strongest-algorithm selection, multiple same-strength hashes, unsupported-only metadata, reserved options, URL-safe base64, and mismatch fallback otherwise match the browser behavior I checked.

**Blocker — `packages/producer/src/services/htmlCompiler.ts:1169`: preserve case-insensitive SRI algorithm tokens.** The SRI contract defines a supported algorithm token by its ASCII-lowercased value, and Chromium accepts `SHA384-…`. This regex recognizes only lowercase names. Consequently a valid uppercase or mixed-case token produces no hashes, reaches `if (!strongest) return true`, and disables verification. I reproduced the concrete bypass with an uppercase SHA-384 digest for trusted bytes and a fetch returning `window.compromised=true;`: the compiler reported `compromisedInlined: true` and removed the external tag, while Chromium rejects the same mismatched response. Normalize the parsed algorithm (or make matching case-insensitive and lowercase the capture) before strongest selection, and add the mixed-case mismatch regression.

Focused evidence: `bun test packages/producer/src/services/htmlCompiler.test.ts` — 134 pass; exact CDN SHA-384 digests and npm byte equality both verified.

**Verdict:** REQUEST CHANGES
**Reasoning:** The nine pinned lowercase tags are correct, but the new shared verifier has a standards-valid case variant that bypasses the integrity check and inlines modified code.

— Magi

miguel-heygen (changes requested):
Correction to my prior review evidence: my first browser probe showed that a matching uppercase token executed, but did not exercise an uppercase mismatch. The author correctly called out that gap. I reran the exact case without request interception using `HeadlessChrome/152.0.7977.42`: lowercase mismatch → blocked; `SHA384` mismatch → altered script executed; `sHa384` mismatch → altered script executed.

The compiler-side casing fix at `packages/producer/src/services/htmlCompiler.ts:1169-1180` is correct and closes the original bypass: mixed-case supported algorithms are normalized for strongest-selection and comparison, and the new tests discriminate valid from tampered bytes.

**Blocker — integrity mismatch still falls through to an executable browser retry.** At `htmlCompiler.ts:1211-1213`, a mismatch becomes a rejected `Promise.allSettled` item; `htmlCompiler.ts:1234-1240` handles every rejection identically by warning and leaving the original external tag in the document. For lowercase metadata Chrome blocks that fallback. For the standards-valid mixed-case metadata the compiler now rejects, Chrome 152 ignores the token and executes the altered response. The exact attack path is therefore: compiler detects tampered bytes → preserves `src` + mixed-case integrity → render browser fetches the same tampered URL → browser executes it.

Treat a verified integrity mismatch differently from a download/HTTP failure: fail compilation, or remove/replace the executable tag so the result cannot run remote bytes. Failing compilation is the clearest contract. Keep the existing availability fallback for ordinary fetch failures if desired; a successful fetch whose bytes violate declared integrity is not an availability failure. Update the unit assertion at `htmlCompiler.test.ts:356-360`, which currently pins the unsafe retained-tag behavior for every mismatch, and add the mixed-case browser mismatch control if that browser behavior remains part of the rationale.

Fresh focused evidence: `bun test packages/producer/src/services/htmlCompiler.test.ts` — 134 pass; direct browser probe output: `lowercaseMismatch blocked`, `uppercaseMismatch altered`, `mixedCaseMismatch altered`. CI is still running on this head.

— Magi

**Verdict:** REQUEST CHANGES
**Reasoning:** Algorithm normalization is fixed, but the mismatch recovery path hands execution back to a browser that demonstrably does not enforce the same valid metadata spelling, so tampered CDN code can still run.

miguel-heygen (changes requested):
Re-review at exact `1728c44524749c9c8f605ac25bc3d41caf3a3ea8`. The remote/nested metadata model is coherent: `scriptItems` carries integrity/CORS, both producer and core use the same reconstruction helper, exact-URL root duplicates are strengthened, later unpinned references retain a pin, and distinct non-empty requirements fail explicitly. The prior mixed-case fallback blocker is also closed.

**Blocker — an unpinned local script can be inlined before a later nested pin is discovered.** `packages/core/src/compiler/htmlBundler.ts:865-899` inlines every relative root script before sub-compositions are parsed, without checking `integrity`. When the child pin is processed later at `htmlBundler.ts:941-963`, `ensureExternalScriptTag` adds a protected external tag but cannot retract the already-emitted inline chunk. I reproduced this with root `<script src="local.js">`, a child tag for the same `local.js` carrying the SHA-384 of trusted bytes, and altered bytes on disk. `bundleToSingleHtml` returned both `alteredInline: true` and `protectedExternal: true`: the browser executes the altered inline code before it ever verifies the protected duplicate.

The same ordering exists within collected children/templates: an unpinned relative item pushes bytes at `htmlBundler.ts:954-959` (or `hoistExternalScript` at `htmlBundler.ts:743-750`), while a later pinned sibling only creates/strengthens an external tag. The new protected-local-template test covers a pin seen first; it does not cover an unpinned-first/pinned-later requirement.

Collect and merge the integrity requirement for each resolved local script URL before inlining any bytes, then either verify those bytes before inlining or keep one protected external tag. Conflict detection also needs to happen before a chunk for that URL is committed. Please pin both root-unpinned + child-pinned local bytes and sibling/template unpinned-first + pinned-later; mismatched bytes must never appear in inline output.

Fresh focused suites are otherwise green: `bun test htmlBundler.test.ts inlineSubCompositions.test.ts htmlCompiler.test.ts` → **225 pass, 0 fail**. Native CI is still running on this head.

— Magi

**Verdict:** REQUEST CHANGES
**Reasoning:** CDN requirements now survive reconstruction, but the core bundler can still execute unverified local bytes when an integrity requirement arrives after an earlier inline decision.

miguel-heygen (approved):
Final re-review at exact `db10320fc04326b51a22cfb49daf59280106fccb`. All three prior security blockers are closed.

- Supported SRI tokens are parsed case-insensitively and canonicalized to lowercase before browser fallback; raw fetched bytes are checked before decoding. A verified mismatch raises through `Promise.allSettled` and aborts compilation, while ordinary network/HTTP failures retain their existing external-tag fallback.
- `scriptItems` carries integrity/CORS through nested and template extraction. Producer and core reconstruct through the same helper; an exact-URL child pin strengthens existing root duplicates, later unpinned references retain it, and distinct non-empty requirements fail before output.
- The core bundler now defers every local inline decision until sub-composition/template requirements are collected. Resolved-path comparison joins `local.js` / `./local.js`; root, sibling, and template unpinned-first cases remain external once any later pin exists. My exact altered-file probe now returns `alteredInline: false`, with every surviving alias tag carrying canonical `sha384-…`. Unprotected chunks retain their original document/chunk order.
- The nine alert mappings and CDN SHA-384 values were independently verified earlier against exact cdnjs/npm bytes. No PR-ref code-scanning alerts or unresolved review threads remain.

Fresh focused evidence: `bun test packages/core/src/compiler/htmlBundler.test.ts packages/core/src/compiler/inlineSubCompositions.test.ts packages/producer/src/services/htmlCompiler.test.ts` → **228 pass, 0 fail**. Every applicable native check is now green, including JavaScript CodeQL, full Test, build/type/lint/Fallow, producer unit/integration, catalog/preview, Windows, perf, and all regression shards.

Non-blocking compatibility note: resolved local aliases remain as separate protected external tags, preserving the pre-existing duplicate execution count rather than silently changing authored behavior.

— Magi

**Verdict:** APPROVE
**Reasoning:** Integrity requirements now survive every direct, nested, template, duplicate, alias, and local-inline boundary audited; mismatches cannot return executable fallback HTML, and all focused and native evidence is green.

PR #3825, opened 2026-09-09
fix(playground): render initialization errors as text

jrusso1020:
Playground initialization errors were interpolated into `innerHTML`, allowing error text containing markup to create active DOM elements. Build the same styled `<pre>` with `textContent` and replace the body with that element, preserving the error text, color, and padding.

Targets code-scanning alert #618. The regression exercises the real initialization failure handler with an HTML-bearing error and verifies literal text, no injected image, and the existing presentation. It fails the previous implementation.

Validation: five playground tests, playground typecheck, full workspace build, Fallow, and signed commit hooks pass.

### DIFF
--- a/packages/sdk-playground/src/main.error.test.ts
+++ b/packages/sdk-playground/src/main.error.test.ts
@@ -1,0 +1,34 @@
+// @vitest-environment jsdom
+import pageHtml from "../index.html?raw";
+import { expect, it, vi } from "vitest";
+
+vi.mock("@hyperframes/sdk", () => ({ openComposition: vi.fn() }));
+vi.mock("@hyperframes/core/gsap-parser-acorn", () => ({ parseGsapScriptAcorn: vi.fn() }));
+vi.mock("gsap/dist/gsap.min.js?raw", () => ({ default: "" }));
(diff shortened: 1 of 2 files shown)


miguel-heygen (approved):
Reviewed exact head `ee4703067dbc725745ac329fbc320c14bf8c0c5e` against merge base `8d80530ff4507a87a8689aa3f48e6062859540d2`.

The real initialization chain can surface SDK/parser/file-adapter exception text at the terminal `init().catch`; constructing a `<pre>`, assigning `textContent`, and replacing the body closes the HTML interpretation boundary while preserving the existing one-element error view and styling. The regression imports the actual entry module with only the failing adapter boundary mocked, and correctly asserts the attacker-shaped error remains literal text with no `` node.

Independent checks: 5/5 playground tests, package typecheck, production build, and Chrome 148 with `/api/composition` aborted (exactly one styled `<pre>`, literal `TypeError: Failed to fetch`). All applicable native checks are green, the PR-ref CodeQL open set is empty, and there are no unresolved threads/comments. I also verified main CodeQL 34417717468 fixed all nine prerequisite #3822 alerts.

No blocking findings.

PR #447, opened 2026-04-23
ci(windows-ffmpeg): pin BtbN release to specific autobuild tag

vanceingalls:
## What

Pin the BtbN/FFmpeg-Builds Windows download to a specific `autobuild-2026-04-23-13-16` release tag instead of the rolling `latest` nightly.

## Why

Follow-up to #436. The `release-url` input description claimed "pinned by default" but the actual default pointed at `releases/latest/download/…` — a nightly rolling build. A new upstream build could silently change encoder behavior or ABI between CI runs. The feature-inventory check catches codec removals but not within-codec behavioral shifts across nightlies.

## How

Replaced the `latest` URL with a specific dated autobuild tag + its git-hash-stamped asset filename. Updated the input description to note that both the tag and filename must be bumped together when upgrading (the asset filename embeds the git hash).

## Test plan

- [x] Verified pinned URL resolves (302 → 200) via `curl -sI -L`
- [x] Confirmed feature-inventory step uses `ffmpeg -encoders`/`-decoders` output, not the asset filename — no assumptions broken by the rename
- [ ] **Note:** Windows install/render/test jobs were skipped on this PR (YAML-only change didn't trigger Windows paths). The pinned download will be exercised by the next Windows-touching PR.

PR #3827, opened 2026-09-10
fix(producer): pass stub formatter path as an argument

jrusso1020:
A checkout path containing shell metacharacters could execute commands when the producer formats its generated early stub (CodeQL #447). Invoke `bun x oxfmt` with an argument array so the full output path stays data. `bunx` is the documented alias for `bun x`: https://bun.sh/docs/pm/bunx.

Generated stub contents, formatter selection, and the existing nonfatal formatter-failure behavior are preserved.

Validation:
- Executed the actual build script in a temporary checkout path containing spaces, semicolons, and a shell comment. The original code executed a controlled marker command; the fix passed the entire path as one argument and did not execute it.
- Compared generated bytes across baseline, fixed, and failing-formatter runs: identical. Formatter failure still exits successfully.
- Normal generator invocation, changed-file oxlint/oxfmt, and full workspace build pass.

Native CI and independent review are required before merge.

### DIFF
--- a/packages/producer/scripts/build-hf-early-stub.ts
+++ b/packages/producer/scripts/build-hf-early-stub.ts
@@ -14,7 +14,7 @@ import { mkdirSync, writeFileSync } from "node:fs";
 import { dirname, resolve } from "node:path";
 import { fileURLToPath } from "node:url";
 import { buildSync } from "esbuild";
-import { execSync } from "node:child_process";
+import { execFileSync } from "node:child_process";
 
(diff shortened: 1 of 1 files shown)


jrusso1020:
Superseded by merged PR #3834 (db51d5479): the producer now calls formatGeneratedSource, which invokes bun x oxfmt through execFileSync with an argument array. The vulnerable shell interpolation is gone. Code-scanning alert #447 is independently verified fixed on main at 2026-09-10T03:20:05Z. Closing this redundant PR without merging; preserving the branch/worktree.

